PR #27, opened 2014-05-17
Added new baro device

Advi42:
- support for Analog, I2C and SPI busses on IOIO
 - support for MPX5010, BMP85 and MS5611
 - added dynamic pressure for direct IAS calculation
 - common sensor setup dialog


Turbo87:

PR #30, opened 2014-09-18
Hm devel

hm17gh1:
Hello,

In this pull request are a few additions for XCSoar:

-- support for connecting to Open WiFi networks from KoboMenu
-- new layouts useful when using landscape orientation of the display (Right 16, Left 16, Left 24)
-- small, 1 line commit that moves the left parts of the menu, in landscape mode, to the right side, to allow right handed people use the menu without covering the screen with their hand, as it happens now. (this is a separate commit, so you can dismiss it if it is not desired)


Turbo87:

PR #36, opened 2014-12-10
CMPS10 Compass GUI patch

dadyeah:
This is to replace pull request #35


Turbo87:

PR #41, opened 2015-12-17
Add Loctome.com Live Tracking Server 

mrcheidel:
Add loctome.com live tracking server as option


Turbo87:

PR #42, opened 2016-11-08
Update URL of cup file format

pyrog:


Turbo87:
see https://github.com/XCSoar/XCSoar/blob/master/README#L35-L48

PR #43, opened 2017-03-06
Device\Driver\IMI\ IMIDevice: parse PGRMZ as altitude above 1013.25 hPa

petergasparovic:
The normal Garmin $PGRMZ line contains the "true" barometric altitude above MSL (corrected with QNH), but IMIDevice differs:
it emits the uncorrected barometric altitude (i.e. pressure altitude). That is the only reason why we catch this sentence in the driver instead of letting the generic class NMEAParser do it. (solution inspired by EWMicroRecorderDevice, and AltairProDevice. )

I have tested IMIDevice (i.e. IMI ERIXX logger) both with LK8000 and SeeYou Mobile and they work OK. Only XCSoar has the bug of incorrect interpretation of NMEA sequence "PGMRZ"

### DIFF
--- a/src/Device/Driver/IMI/Internal.cpp
+++ b/src/Device/Driver/IMI/Internal.cpp
@@ -40,3 +40,47 @@ IMIDevice::Disconnect(OperationEnvironment &env)
   // disconnect
   IMI::Disconnect(port, env);
 }
+
+static bool
+ReadAltitude(NMEAInputLine &line, fixed &value_r)
+{
+  fixed value;
+  bool available = line.ReadChecked(value);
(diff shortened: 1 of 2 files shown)


Turbo87:
Please send pull requests to the official mailing list (see https://github.com/XCSoar/XCSoar/blob/master/README#L35-L48)

This repository is just an unofficial mirror.

petergasparovic:
Thank you for help.
I will try to pull request to the mailing list, when I learn how to do it, 
and after my subscription to the list is confirmed.


Dňa 6. 3. 2017 o 9:43 Tobias Bieniek napísal(a):

PR #44, opened 2017-04-07
CAI 302 fix for downloading waypoint files; adds termination message …

dssherrill:
…after last waypoint

Using XCSoar to download waypoints to a Cambridge 302 leaves the 302 with no waypoints stored, or sometimes just one waypoint.  The problem arises from a requirement that is not mentioned in the 302 documentation.  You must write "C,-1" after downloading the last waypoint.  This is similar to the requirement to write "D,255" after the last waypoint when downloading a task.

The heart of the fix is this following method which must be invoked (indirectly) from **CAI302WaypointUploader::Run(...)** after downloading the last waypoint.

bool CAI302::CloseNavpoints(Port &port, OperationEnvironment &env)
{
  return DownloadCommand(port, "C,-1\r", env, 5000);
}

I was only able to identify this unwritten requirement by capturing the serial stream from Cambridge Aero's original 302 utility.

**This commit fixes this problem, but another problem remains: the waypoints get stored in a strange order.**  The 302's alphabetical waypoint listing is fine, but the sequential listing is a mess.  I will look into fixing this problem next.

### DIFF
--- a/src/Device/Driver/CAI302/Internal.hpp
+++ b/src/Device/Driver/CAI302/Internal.hpp
@@ -142,6 +142,8 @@ public:
 
   bool WriteNavpoint(unsigned id, const Waypoint &wp,
                      OperationEnvironment &env);
+
+  bool CloseNavpoints(OperationEnvironment &env);
 };
 
 #endif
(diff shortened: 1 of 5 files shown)


Turbo87:
Please send pull requests to the official mailing list (see https://github.com/XCSoar/XCSoar/blob/master/README#L35-L48)

This repository is just an unofficial mirror.

PR #45, opened 2017-09-05
correct function return for line with dx<0

ericcarden:
I stumbled upon what appears to be a problem with the FlatLine::IntersectOriginCircle function. The problem manifests itself as incorrect predicted entry points for a circular airspace area if the predicted track is westerly – with the error being largest on a track of 225 or 315 degrees. With a due west track or any easterly track, there is no error.

The following website describes how to calculate the locations of intersection points between a line and a circle (the stated purpose of this function): http://mathworld.wolfram.com/Circle-LineIntersection.html. The subject function does the calculation exactly as shown here except that it uses a call to the copysign function. This difference causes a sign error when dx (d.x) is negative (as on a westerly track). The result is that instead of returning the intersection points as (x1, y1) and (x2, y2), if on a westerly track, the locations (x1, y2) and (x2, y1) are returned. To clarify, if on a westerly track and the actual intersection point locations are (1, 4) and (5, 2), the function as it is now computes intersection point locations as (1, 2) and (5, 4). (The due west track case is exempt from this error, because the Y value is the same for both intersection points in this case.) You can easily see this by heading toward a circular airspace area on a NW or SW track and watching where the airspace intersection warning marker is placed. I’ve attached a screen shot of a chart I made to help me understand what was happening. It shows the actual and incorrect (as the function currently calculates them) intersection points for a particular scenario.

intersections

In case you’re interested in why using copysign causes a problem, here’s an explanation. The call copysign(x, y) returns the magnitude (absolute value) of x with the sign of y. So for x = -1 and y = 2, copysign(x, y) = 1. The Wolfram solution, however, (which I proved accurate using Excel and test scenarios approaching the circle from all quadrants) says that this part of the solution should be x multiplied by -1 if y is negative (otherwise by 1). So this would give, for x = -1 and y = 2, a result of -1. When x>=0, copysign coincidentally gives the correct result, but otherwise (westerly tracks), it reverses the sign for this part of the formula.

I’ve tested this trouble scenario in version 6.8.7 Android, and there’s no problem (approaching the airspace area from any direction). It looks like the file containing this function hasn’t been altered since version 6.8.7 was released (last change: 2016-03-04). Therefore, I think 6.8.7 uses this function just as it is now (including its use of the copysign function). I haven’t been able to answer why this problem doesn’t exist in 6.8.7. Any ideas?

This pull request is for my fix for this. I’ve compiled and tested it (for/on Android), and it corrects the airspace intersection marker issue.

### DIFF
--- a/src/Geo/Flat/FlatLine.cpp
+++ b/src/Geo/Flat/FlatLine.cpp
@@ -42,6 +42,7 @@ bool
 FlatLine::IntersectOriginCircle(const double r,
                                 FlatPoint &i1, FlatPoint &i2) const
 {
+  // http://mathworld.wolfram.com/Circle-LineIntersection.html
   const auto d = GetVector();
   const auto dr = GetSquaredDistance();
   const auto D = CrossProduct();
@@ -53,8 +54,9 @@ FlatLine::IntersectOriginCircle(const double r,
 
   det = sqrt(det);
   const auto inv_dr = 1. / dr;
-  i1.x = (D * d.y + copysign(d.x, d.y) * det) * inv_dr;
-  i2.x = (D * d.y - copysign(d.x, d.y) * det) * inv_dr;
+  const auto sign_dx = (d.y < 0) ? -d.x : d.x;
+  i1.x = (D * d.y + sign_dx * det) * inv_dr;
+  i2.x = (D * d.y - sign_dx * det) * inv_dr;
   i1.y = (-D * d.x + fabs(d.y) * det) * inv_dr;
   i2.y = (-D * d.x - fabs(d.y) * det) * inv_dr;
   return true;


felixhaedicke:
The explanation (at least the oart about what the patch does and why) should be in the commit message.

It would be great if you could provide at least one test case! See test/src.

ericcarden:
Ok. I'll add a descriptive commit comment and then submit a new pull request. I looked at test/src and don't quite understand how to provide a test case. If you can give me more guidance, I'll see what I can do. My testing was simply to fly all the way around a circular airspace area, occasionally turning toward it and seeing where the incursion warning is placed. Thanks.

felixhaedicke:
There are a lot example in test/src, e. g. test/src/TestGeoBounds.cpp

felixhaedicke:
and we already have test/src/TestFlatLine.cpp
Compile with
`make build-check`

ericcarden:
Thanks. That helps. I've reviewed TestFlatLine.cpp, see the need for an extra test case, and will add it. The only line-circle intersect test cases there now are for dx >= 0, cases which coincidentally are successful even as the FlatLine::IntersectOriginCircle function is now. I've installed the packages needed to compile with "make build-check". How do I run the test after compiling? I'm new to TAP. Thanks.

felixhaedicke:
It is a normal executable, see output/<your target>/bin/TestFlatLine
e. g.
`./output/UNIX/bin/TestFlatLine`

ericcarden:
Thanks. I figured out that "make check" runs the tests. I added two test cases that demonstrate the problem for dx<0 lines. As intended, these tests fail with the current code and pass with the fixed code. Once you like and accept the pull request I just sent for the test case additions, I'll redo the pull request for the fix. Thanks for your help.

MaxKellermann:
Agree with @felixhaedicke, the explanation should be in the commit message. A good pull request requires no further explanation at all, because each commit is already documented. Keep in mind that GitHub PR text is lost.

Turbo87:
> Keep in mind that GitHub PR text is lost.

I think if you use the "Merge" button in GitHub instead of merging manually it will include the PR description in the merge commit message :thin

MaxKellermann:
Maybe, but it really doesn't belong there.

felixhaedicke:
merged (edited the commit description)

ericcarden:
Thanks, @felixhaedicke.

PR #46, opened 2017-09-07
test/src: add test cases to show problem with dx<0 line

ericcarden:
These test cases demonstrate a problem with FlatLine::IntersectOriginCircle when dx<0 for the line. The calculated X values get reversed: x1, y2 and x2, y1 instead of x1, y1 and x2, y2. I'll send the fix for the problem separately, making it easier to test before and after the fix.


Turbo87:
could you add that commit to #45?

PR #47, opened 2017-09-09
Waypoint/WaypointReaderSeeYou: handle file with rwwidth field

ericcarden:
Some SeeYou WP files contain a field for runway width (called "rwwidth" in the file header, if present). For example, at least some versions of SeeYou Mobile create files like this (if saved due to user editing of a waypoint). This extra field is inserted just before the radio frequency field, thus shifting that field and the description field to the right. Without this change, when XCSoar uses such a file, the radio frequency isn't shown (in XCSoar's frequency field), the description/comments field in XCSoar contains only the radio frequency (from the WP file), and the description/comments from the WP file are nowhere in XCSoar.

### DIFF
--- a/src/Waypoint/WaypointReaderSeeYou.cpp
+++ b/src/Waypoint/WaypointReaderSeeYou.cpp
@@ -170,18 +170,11 @@ WaypointReaderSeeYou::ParseLine(const TCHAR* line, Waypoints &waypoints)
     iStyle = 6,
     iRWDir = 7,
     iRWLen = 8,
-    iFrequency = 9,
-    iDescription = 10,
   };
 
-  if (first) {
-    first = false;
-
-    /* skip first line if it doesn't begin with a quotation character
-       (usually the field order line) */
-    if (line[0] != _T('\"'))
-      return true;
-  }
+  // field positions for typical CUP file
+  static unsigned iFrequency = 9;
+  static unsigned iDescription = 10;
 
   // If (end-of-file or comment)
   if (StringIsEmpty(line) ||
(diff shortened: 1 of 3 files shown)


Turbo87:
@ericcarden this will need tests too including an example file

ericcarden:
I find no existing tests of this function in test/src, but I've attached two example files and give a test procedure here:

1.	Configure XCSoar to use a *.cup waypoint file wherein the 10th field in the header row is “rwwidth” (e.g., the attached runway_width.cup file).
2.	Start XCSoar .
3.	Examine a waypoint known to include a radio frequency and description (e.g., "01Moontown Apt").

The test is successful, if the frequency and description shown in XCSoar are correct. (With the current code, this test will fail. With the changes I made, it will pass.)

1.	Configure XCSoar to use a *.cup waypoint file that doesn't contain a runway width field (e.g., the attached no_runway_width.cup file).
2.	Start XCSoar .
3.	Examine a waypoint known to include a radio frequency and description (e.g., "01Moontown Apt").

The test is successful, if the frequency and description shown in XCSoar are correct. (This test will pass with both the current code and with my changes.)

[rwwidth test files.zip](https://github.com/XCSoar/XCSoar/files/1290476/rwwidth.test.files.zip)

Turbo87:
> I find no existing tests of this function in test/src

see https://github.com/XCSoar/XCSoar/blob/master/test/src/TestWaypointReader.cpp

ericcarden:
Thanks. I added testing for this (including an example waypoint file) and just created a pull request for only this added testing.

src/Waypoint/WaypointReaderSeeYou.cpp
```diff
@@ -170,18 +170,11 @@ WaypointReaderSeeYou::ParseLine(const TCHAR* line, Waypoints &waypoints)
...
-    first = false;
-
-    /* skip first line if it doesn't begin with a quotation character
-       (usually the field order line) */
-    if (line[0] != _T('\"'))
-      return true;
-  }
+  // field positions for typical CUP file
+  static unsigned iFrequency = 9;
+  static unsigned iDescription = 10;
```
MaxKellermann: Uh-oh, no `static` variables, please! This is completely broken code.

ericcarden:
Thanks. I'll redo this to not use static variables and will submit a new pull request.

felixhaedicke:
Btw. if you are going to revise the code for a pull request, there is no need to close the pull request and (and later create a new one). You can just force-push the new code to the branch you were using for your pull request.

PR #48, opened 2017-09-11
test/src: add testing for CUP file with runway width field

ericcarden:
Some CUP files (e.g., those saved/created by SeeYou Mobile) contain a runway width (rwwidth) field. This added testing demonstrates failures with such a file with the current code but demonstrates success with the pull request I submitted for a fix.


ericcarden:
@Turbo87 This is the testing related to our conversation in pull request #47. Let me know if it's what you had in mind. Thanks.

Turbo87:
since the commit is already part of #47 I think we should handle it over there

PR #50, opened 2017-09-13
Add basic TravisCI instructions

Turbo87:
/cc @felixhaedicke

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+sudo: required
+
+language: cpp
+
+env:
+  matrix:
+    - TARGET=UNIX
+
+addons:
+  apt:
+    packages:
+      - libjpeg-dev
+      - librsvg2-bin
+      - mesa-common-dev
+      - xsltproc
+
+before_script:
+  # Fix libjpeg-dev package which is missing pkg-config file
+  - sudo cp ./ide/ubuntu-trusty/libjpeg.pc /usr/lib/x86_64-linux-gnu/pkgconfig/libjpeg.pc
+
+script:
+  # Compile
+  - make -j4 TARGET=$TARGET everything;
+
+  # Run unit tests on platforms where possible (Unix)
+  - if [ "$TARGET" = "UNIX" ]; then
+      make TARGET=$TARGET check;
+    fi
--- a/ide/ubuntu-trusty/libjpeg.pc
+++ b/ide/ubuntu-trusty/libjpeg.pc
@@ -1,0 +1,10 @@
+prefix=/usr
+exec_prefix=${prefix}
+libdir=${prefix}/lib/x86_64-linux-gnu
+includedir=${prefix}/include
+
+Name: libjpeg
+Description: JPEG image codec
+Version: 
+Libs: -L${libdir} -ljpeg
+Cflags: -I${includedir}


felixhaedicke:
So Ubuntu Trusty is the default environment for Travis CI? I fear this cannot work, the missing libjpeg.pc, for which you have a workaround, is not the only problem. The compilers are too old. We need at least GCC 5.0 or Clang 3.8. I see that there is Clang 3.8 in trusty /but not as the default clang), so it might work if we compile Clang 3.8, but it still uses GCC 4.8's libstdc++ then.

felixhaedicke:
ok, I see it is for v6.8.x, so trusty should be fine. But CI would be much more helpful for master.

Turbo87:
> So Ubuntu Trusty is the default environment for Travis CI?

it used to be precise but they lately migrated to trusty

> The compilers are too old. We need at least GCC 5.0 or Clang 3.8.

I haven't had a detailed look yet, but there are probably backport PPAs for gcc and clang that we could use for `master`

Turbo87:
An alternative would be using CircleCI which seems to be based on Docker and then using a more recent Ubuntu container, but I don't have much experience with that so far.

felixhaedicke:
I have another solution in mind. Let's assume we want to build for Linux/X11, Android and Kobo.

For Linux/X11, we can create a (minimal) sbuild chroot with Ubuntu Xenial (or Debian stretch) inside the trusty environment, and build there. This the right thing to do anyway, and the host environment does not matter any more. I am going to provide documentation on how to create sbuild environments for different target platforms (Ubuntu, Debian, Raspbian, Armbian...).

For Android, we use the Android SDK / NDK, which should be fine on trusty.

For Kobo, we can use a Crosstool-NG based toolchain. I am already using one (on macOS) for this.

MaxKellermann:
This installs a GCC5 toolchain via PPA: https://github.com/MusicPlayerDaemon/MPD/blob/master/.travis.yml
Not a good solution, just a hack to make it work somehow.

What a pity Travis depends on this ancient distribution.

Turbo87:
> For Linux/X11, we can create a (minimal) sbuild chroot with Ubuntu Xenial (or Debian stretch) inside the trusty environment

that needs `sudo` access, right?

the issue with that is that the CI environments that require `sudo` on Travis are significantly slower than the ones based on containers.

I would recommend to either use what we currently have in this PR here, or instead use CircleCI which seems to be a little more flexible in what test environments are possible (all based on Docker containers though)

MaxKellermann:
I don't think we should merge something which breaks as soon as we merge into master, with nobody (yet) willing to adapt it to master.

Turbo87:
> with nobody (yet) willing to adapt it to master.

Not sure how you inferred that. I wanted to get CI going on the stable branch first and then work on `master`, but if you think doing it in a different order makes sense let me know.

MaxKellermann:
It's been known for more than a month that this PR is broken on master, but still nobody has repaired it. After that one month, you suggest to merge it anyway. Sounds wrong to me.
I'm always pessimistic about people willing to clean up after themselves. Without pressure, that rarely ever works. So my way of dealing with this (i.e. applying pressure) is: no merge until it's repaired.
Prove me wrong by submitting a universally working CI solution. Disproving my pessimism makes everybody happier.

PR #51, opened 2017-09-15
Add "Number of used Sats" infobox

teji24:
This (small) modification is to create a new InfoBox with de "Number Of Satellite" used by GPS module.  It can be used to monitor the gps reception quality during the flight, without going to status panel. very usefull for "light users", in paraglider, because we use very light gps modules, sometimes included INTO kobo itself. 
The infobox use existing gps status code, and just obtained infos into a conventional InfoBox.  Number of sats and status of connection (3d fix, ....).
As the used words for display are same as originals, the translating system of the app work on it too.
Hope it can be usefull for others.

### DIFF
--- a/src/InfoBoxes/Content/Other.cpp
+++ b/src/InfoBoxes/Content/Other.cpp
@@ -53,7 +53,7 @@ UpdateInfoBoxBattery(InfoBoxData &data)
   switch (Power::External::Status) {
     case Power::External::OFF:
       if (CommonInterface::Basic().battery_level_available)
-        data.UnsafeFormatComment(_T("%s; %d%%"),
+        data.UnsafeFormatComment(_T("%s; %u%%"),
                                  _("AC Off"),
                                  (int)CommonInterface::Basic().battery_level);
       else
@@ -188,13 +188,13 @@ UpdateInfoBoxNbrSat(InfoBoxData &data) {
     const NMEAInfo &basic = CommonInterface::Basic();
     const GPSState &gps = basic.gps;
 
-    data.SetComment(_(GetGPSStatus(basic)));
+    data.SetComment(gettext(GetGPSStatus(basic)));
     
     if (!basic.alive)
         data.SetComment(_("No GPS"));
     else if (gps.satellites_used_available) {
         // known number of sats
-        data.FormatValue(_T("%d"), gps.satellites_used);
+        data.FormatValue(_T("%u"), gps.satellites_used);
     } else {
         // valid but unknown number of sats
         data.SetValue(_("Unknown"));
--- a/src/InfoBoxes/Content/Other.cpp
+++ b/src/InfoBoxes/Content/Other.cpp
@@ -188,19 +188,13 @@ UpdateInfoBoxNbrSat(InfoBoxData &data) {
     const NMEAInfo &basic = CommonInterface::Basic();
     const GPSState &gps = basic.gps;
 
-    StaticString<80> Temp;
-    StaticString<20> GpsStatus;
-
-    GpsStatus = gettext(GetGPSStatus(basic));
-    data.SetComment(_(GpsStatus));
+    data.SetComment(_(GetGPSStatus(basic)));
     
     if (!basic.alive)
         data.SetComment(_("No GPS"));
-        //    ClearText(NumSat);
     else if (gps.satellites_used_available) {
         // known number of sats
-        Temp.Format(_T("%d"), gps.satellites_used);
-        data.SetValue(Temp);
+        data.FormatValue(_T("%d"), gps.satellites_used);
     } else {
         // valid but unknown number of sats
         data.SetValue(_("Unknown"));
--- a/src/InfoBoxes/Content/Other.cpp
+++ b/src/InfoBoxes/Content/Other.cpp
@@ -167,3 +167,44 @@ InfoBoxContentHorizon::Update(InfoBoxData &data)
 
   data.SetCustom();
 }
+
+
+gcc_pure
+static const TCHAR *
+GetGPSStatus(const NMEAInfo &basic)
+{
+  if (!basic.alive)
+    return N_("Disconnected");
+  else if (!basic.location_available)
+    return N_("Fix invalid");
+  else if (!basic.gps_altitude_available)
+    return N_("2D fix");
+  else
+    return N_("3D fix");
+}
+
+void
+UpdateInfoBoxNbrSat(InfoBoxData &data) {
+    const NMEAInfo &basic = CommonInterface::Basic();
+    const GPSState &gps = basic.gps;
+
+    StaticString<80> Temp;
+    StaticString<20> GpsStatus;
+
+    GpsStatus = gettext(GetGPSStatus(basic));
+    data.SetComment(_(GpsStatus));
+    
+    if (!basic.alive)
+        data.SetComment(_("No GPS"));
+        //    ClearText(NumSat);
+    else if (gps.satellites_used_available) {
+        // known number of sats
+        Temp.Format(_T("%d"), gps.satellites_used);
+        data.SetValue(Temp);
+    } else {
+        // valid but unknown number of sats
+        data.SetValue(_("Unknown"));
+    }
+}
+
+
--- a/src/Dialogs/StatusPanels/SystemStatusPanel.cpp
+++ b/src/Dialogs/StatusPanels/SystemStatusPanel.cpp
@@ -81,7 +81,7 @@ SystemStatusPanel::Refresh()
     ClearText(NumSat);
   else if (gps.satellites_used_available) {
     // known number of sats
-    Temp.Format(_T("%d"), gps.satellites_used);
+    Temp.Format(_T("%u"), gps.satellites_used);
     SetText(NumSat, Temp);
   } else
     // valid but unknown number of sats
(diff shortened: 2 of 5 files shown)


teji24:
screenshot from 2017-09-15 10-18-24

src/InfoBoxes/Content/Other.cpp
```diff
@@ -167,3 +167,44 @@ InfoBoxContentHorizon::Update(InfoBoxData &data)
...
+
+    StaticString<80> Temp;
+    StaticString<20> GpsStatus;
+
+    GpsStatus = gettext(GetGPSStatus(basic));
+    data.SetComment(_(GpsStatus));
+    
+    if (!basic.alive)
+        data.SetComment(_("No GPS"));
+        //    ClearText(NumSat);
```
MaxKellermann: What does this comment mean?

src/InfoBoxes/Content/Other.cpp
```diff
@@ -167,3 +167,44 @@ InfoBoxContentHorizon::Update(InfoBoxData &data)
...
+    else if (gps.satellites_used_available) {
+        // known number of sats
+        Temp.Format(_T("%d"), gps.satellites_used);
+        data.SetValue(Temp);
```
MaxKellermann: Try `data.FormatValue()` instead. That way, the value gets formatted right into the final buffer, without this temporary stack buffer, without copying it.

src/InfoBoxes/Content/Other.cpp
```diff
@@ -167,3 +167,44 @@ InfoBoxContentHorizon::Update(InfoBoxData &data)
...
+
+void
+UpdateInfoBoxNbrSat(InfoBoxData &data) {
+    const NMEAInfo &basic = CommonInterface::Basic();
+    const GPSState &gps = basic.gps;
+
+    StaticString<80> Temp;
+    StaticString<20> GpsStatus;
+
+    GpsStatus = gettext(GetGPSStatus(basic));
```
MaxKellermann: Why copy the string to a stack buffer?

src/InfoBoxes/Content/Other.cpp
```diff
@@ -167,3 +167,44 @@ InfoBoxContentHorizon::Update(InfoBoxData &data)
...
+    data.SetComment(_(GpsStatus));
```
MaxKellermann: Duplicate `gettext()` call.

src/InfoBoxes/Content/Other.cpp
```diff
@@ -167,3 +167,44 @@ InfoBoxContentHorizon::Update(InfoBoxData &data)
...
+GetGPSStatus(const NMEAInfo &basic)
+{
+  if (!basic.alive)
+    return N_("Disconnected");
+  else if (!basic.location_available)
+    return N_("Fix invalid");
+  else if (!basic.gps_altitude_available)
+    return N_("2D fix");
+  else
+    return N_("3D fix");
```
MaxKellermann: Returning an untranslated translatable string is an interesting micro-optimisation, but that is worth documenting. Please add API documentation to this function, mentioning that `gettext()` on the return value is necessary.

src/InfoBoxes/Content/Other.cpp
teji24: it mean i forget to remove it. it come from statusPanel

src/InfoBoxes/Content/Other.cpp
teji24: OK, i will do it.

src/InfoBoxes/Content/Other.cpp
teji24: i just take this code from existing StatusPanel code, then i presumed it was good.  I don't have experience on translating system in XCSoar...

teji24: i don't searched longer, cause (as you can see on screenshot), the translation work fine.

MaxKellermann: Documentation patches for old code are also welcome! Sometimes, the best code documentation comes from people who didn't write the code, because only they know that something is not obvious enough.

MaxKellermann: Now I see you copy'n'pasted somebody else's code. That's **bad**! It's bad because we now have two copies of the same code we have to maintain and keep up to date. You should insert one commit which makes the original code available as a library, and then call into that library.

teji24: the reason i don't do this is to not make big changes. as you probably already understand, i don't produce better code you has seen.  Then a little addition to help community, YES, big changes to (working fine) code of better programmers than me : NO ;-)

MaxKellermann: I appreciate your motivation to improve XCSoar. But if it takes more time for me to clean up after you than it would require me to write the same feature in the first place, then you're not helping. If you have no motivation to learn to produce better code (as I suggested), then we should close this pull request right now and go back to real work. It's up to you. Your post sounded like "but that's all I can do, and I won't learn it". Maybe I misunderstood, and I'll be happy to see my interpretation proven wrong when your new code arrives.

teji24: I'm actually testing modified code as you propose it to me.  My comment is just to explain that i come for the first time to open-source project.  I try to do by myself the little addition i would like, to not ask to others, to show that i can spend time too, not only profit about other's work. Then i do it, spend lot of time to test, to compile, to learn how to cross compile, and make it run effectively. the next step in my head is to share, not like an egoist... i know it's 1 gram in a ton, but i spend other big big time to understand git, push, pull, branch, fork, and other system. i just present thinks with modesty. i make 10 lines of code, more an arrangement of existing things than genius part of project.  
Sorry to make it not very well. My goal is really to help, give time to project in return of joy to use it, really conscient of lot of work it is from others.
of course i follow advices. i was testing new code when i receive your comment. I finish to test it, then try to understand how to submit again (just upload, in same fork probably ? ...)
If you prefer in future i don't participate cause too low programmer level, i can of course understand (and it is NOT ironic). i can also ask in forum "Please add nbr sat infobox", or keep my modifications.... i want first to respect others working in the project. really.  And one time again, i will thanks all participant for this fantastic work around xcsoar.

MaxKellermann: It's not about what you can do now, and you don't need to apologize or justify or explain imperfect code. It's about whether you're willing to learn and improve.
If you're already busy hacking the code, then welcome aboard, go ahead, don't let me distract you from that. Come back when you have something new to review, or when you need help.

src/InfoBoxes/Content/Other.cpp
```diff
@@ -188,19 +188,13 @@ UpdateInfoBoxNbrSat(InfoBoxData &data) {
     const NMEAInfo &basic = CommonInterface::Basic();
     const GPSState &gps = basic.gps;
 
-    StaticString<80> Temp;
-    StaticString<20> GpsStatus;
-
-    GpsStatus = gettext(GetGPSStatus(basic));
-    data.SetComment(_(GpsStatus));
+    data.SetComment(_(GetGPSStatus(basic)));
```
MaxKellermann: Using `_()` is wrong here, because this shortcut is only meant for string literals, to be collected into the PO template. `gettext()` is the function for non-literal strings.

src/InfoBoxes/Content/Other.cpp
teji24: i make a file update and new pull request. new code work fine too, but is cleaner. Thx for advices. ;-)

src/InfoBoxes/Content/Other.cpp
```diff
@@ -188,19 +188,13 @@ UpdateInfoBoxNbrSat(InfoBoxData &data) {
...
     
     if (!basic.alive)
         data.SetComment(_("No GPS"));
-        //    ClearText(NumSat);
     else if (gps.satellites_used_available) {
         // known number of sats
-        Temp.Format(_T("%d"), gps.satellites_used);
-        data.SetValue(Temp);
+        data.FormatValue(_T("%d"), gps.satellites_used);
```
MaxKellermann: `%d` is the wrong format. `satellites_used` is `unsigned`, and thus `%u` should be used. Yes, I know the old code you copied `SystemStatusPanel.cpp` is wrong as well. I'd welcome a fixup patch for that one.

src/InfoBoxes/Content/Other.cpp
teji24: ok, well understood.

src/InfoBoxes/Content/Other.cpp
teji24: ok, i modified my part, the status panel, and the battery infobox, also usinf %d in place of %d.
i compiled, tested. i re-submit.

MaxKellermann:
Author is wrong in the commits: teji24 <teji24@users.noreply.github.com>

What's your real name and email address?

MaxKellermann:
I've merged these changes with a bunch of fixups:
ee80749e54d1eaccca00cb49ef5bd2efb9844c91
44632028b29fda89434336b6eeb319b7a2bb0dff

teji24:
thx a lot for the time you spent to advice me, answer me and learn me good practices. It's a big pleasure (and honor) to participated.
And thanks to include new NbrSat InfoBox.

PR #53, opened 2017-09-16
musl libc (for Kobo)

felixhaedicke:
Add musl libc support, and use it instead of glibc for the KOBO target.

Advantages:
* very lightweight
* can be statically linked
* support for old Linux kernels

Uploaded my build of the toolchain here: http://s15356785.onlinehome-server.info/~felix/arm-unknown-linux-musleabihf-2017-09-17.tar.gz
sha256sum: ca2a4d104aacf364612f55353cfc16b1868722fa6eab2ecf11b2b35e917e430f
Built with ct-ng with the configuration "build/ct-ng/arm-unknown-linux-musleabihf.config" (see commit f7129dc).

### DIFF
--- a/build/ct-ng/armv7a-a8neon-linux-musleabihf.config
+++ b/build/ct-ng/armv7a-a8neon-linux-musleabihf.config
@@ -1,0 +1,544 @@
+# crosstool-ng 1.23.0
+#
+# Automatically generated file; DO NOT EDIT.
+# Crosstool-NG Configuration
+#
+CT_CONFIGURE_has_static_link=y
+CT_CONFIGURE_has_wget=y
+CT_CONFIGURE_has_curl=y
+CT_CONFIGURE_has_stat_flavor_GNU=y
+CT_CONFIGURE_has_make_3_81_or_newer=y
+CT_CONFIGURE_has_libtool_2_4_or_newer=y
+CT_CONFIGURE_has_libtoolize_2_4_or_newer=y
+CT_CONFIGURE_has_autoconf_2_63_or_newer=y
+CT_CONFIGURE_has_autoreconf_2_63_or_newer=y
+CT_CONFIGURE_has_automake_1_15_or_newer=y
+CT_CONFIGURE_has_gnu_m4_1_4_12_or_newer=y
+CT_CONFIGURE_has_svn=y
+CT_CONFIGURE_has_git=y
+CT_MODULES=y
+
+#
+# Paths and misc options
+#
+
+#
+# crosstool-NG behavior
+#
+# CT_OBSOLETE is not set
+CT_EXPERIMENTAL=y
+# CT_ALLOW_BUILD_AS_ROOT is not set
+# CT_DEBUG_CT is not set
+
+#
+# Paths
+#
+CT_LOCAL_TARBALLS_DIR="${HOME}/src"
+CT_SAVE_TARBALLS=y
+CT_WORK_DIR="${CT_TOP_DIR}/.build"
+CT_BUILD_TOP_DIR="${CT_WORK_DIR}/${CT_HOST:+HOST-${CT_HOST}/}${CT_TARGET}"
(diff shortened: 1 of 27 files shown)


MaxKellermann:
Sounds like a good idea.

Do we really need a custom toolchain? Is that custom toolchain more than just good defaults for Musl?

I'd like to be able to use the standard Debian cross-toolchain. And build Musl during the `libs` target, like we do with glibc currently.

Since we have problems with v6.8.x (Debian Stretch glibc doesn't work with ancient Kobo kernel, and some users have problems with our glibc 2.23 build), I'd also move that feature to v6.8.x.

felixhaedicke:
I fear that it is not so easy as just replacing glibc with musl in our thirdparty libs. It is not only musl, but we also need a c++ standard library.

libstdc++ from Debian's arm-linux-gnueabihf toolchain is not compatible with musl. At least not its headers, see https://www.musl-libc.org/faq.html:

> The existing libstdc++ is actually compatible with musl in most cases and could be used by copying it into the musl library path, but the C++ header files are usually not compatible. One option may be rebuilding just libstdc++ against musl; however, if C++ support is needed, it's recommended just to build a native toolchain targetting musl.
And as libstdc++ is tightly linked with GCC, it is not so easy to include in out thirdparty libs.

libc++ would be another alternative. Building it and its dependencies (libunwind, libcxxrt, libcxxabi) is described here: http://wiki.musl-libc.org/wiki/BuildingLLVM#Build_libcxx_against_libcxxrt
This seems to be what the ellcc people did - http://ellcc.org - and as a result, ellcc does not support C++ exceptions, at least not for ARM.

Another option could be just including libc++ and musl in out thirdparty libs, and let libc++ use libstdc++ and libsupc++ from Debian's arm-linux-gnueabihf toolchain. I have not yet tried if this works.

What do you think about including the build of ct-ng and the toolchain in our build process?

MaxKellermann:
ct-ng is just some glue which we don't need; our "thirdparty" approach replaces much of that. But if you find a simple and stable solution by building GCC in the "libs" stage, I'd be happy with that.

felixhaedicke:
> But if you find a simple and stable solution by building GCC in the "libs" stage, I'd be happy with that.

I'd rather try the libc++ approach first.

> ct-ng is just some glue which we don't need

ct-ng is a very simple, lightweight, elegant, and powerful tool. I like it very much!

MaxKellermann:
OK for me.

felixhaedicke:
Updated with the following changes:
* Can now be compiled with Debian's arm-linux-gnueabihf. For that, musl is installed as thirdparty library. And the required libstdc++ headers are installed from the GCC6 source tarball (these are compatible with g++-5 and g++-7). The arm-linux-gnueabihf toolchain's libstdc++ is used, but its headers are compatible with glibc only.
* The libc platform checks were rephrased, so that no explicit MUSL define is needed.
* A crosstool-ng toolchain can still be used, the configuration is included.  As before, the build variable USE_CROSSTOOL_NG=y can be used (but now for the musl-based toolchain).

There is no support for compiling with Clang, unless the crosstool-ng toolchain is used.

I dropped the libc++ approach. I had successfully included libc++, but libc++ has no exception_ptr support (which we use), unless libc++abi is used. And I could not get libc++abi working, it seems that it cannot be compiled with GCC's unwind library for ARM.

A workaround (using "-marm" to compile musl without Thumb instructions) was needed to circumvent a crash:
http://www.openwall.com/lists/musl/2017/10/08/2

MaxKellermann:
Excellent!

fraktal-1955:
since 6.8.13 I have problems to build the KOBO version; seems to relate to the 
switch to musl; see attached logs

[build_kobo.txt](https://github.com/XCSoar/XCSoar/files/4676312/build_kobo.txt)
[build_kobo-libs.txt](https://github.com/XCSoar/XCSoar/files/4676314/build_kobo-libs.txt)

fraktal-1955:
Compiling XCS 7.0 with gcc/g++ 8 works; this is OK for me; will test 7.0 in flight; still have my 6.8.12 version as backup

PR #54, opened 2017-09-17
XCTracer-Driver-backport from 7.0_preview11 to v6.8.x

commander-chris:
Copied the XCTracer driver files from the 7.0_preview11 sources to the v6.8.x branch. Edited build/driver.mk, src/Device/Register.cpp, and test/src/TestDriver.cpp to reflect the changes.

### DIFF
--- a/build/driver.mk
+++ b/build/driver.mk
@@ -82,6 +82,10 @@ BLUEFLY_SOURCES = \
 	$(DRIVER_SRC_DIR)/BlueFly/Settings.cpp \
 	$(DRIVER_SRC_DIR)/BlueFly/Register.cpp
 
+XCTRACER_SOURCES = \
+	$(DRIVER_SRC_DIR)/XCTracer/Parser.cpp \
+	$(DRIVER_SRC_DIR)/XCTracer/Register.cpp
+
 DRIVER_SOURCES = \
 	$(SRC)/Device/Driver.cpp \
 	$(SRC)/Device/Register.cpp \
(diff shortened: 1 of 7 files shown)


felixhaedicke:
Why did you not just cherry-pick commit b7fcd4e from the master branch to v6.8.x (and solve some trivial merge conflicts)? Looks like you just copy & pasted the code from the master branch. This is not a clean way, and has some disadvantages, e. g.:
* we cannot easily merge the 6.8 with your commit back master
* you (and not the real author, Thomas Ruf) of this driver is shown in the commit history

felixhaedicke:
And we should make a decision, if this is appropriate for the v6.8.x branch. There must be good reasons for introducing a new feature on the stable releases branch. Can you explain why you think this feature is important?

From my side, this would be ok to make an exception for this driver, if it is well-founded.
But it is Max's decision...

commander-chris:
Fair point, don't want to push (just needed the driver for my flying!)
And wasn't aware of the cherry-pick feature when I started the coding (quite new to git!)
Closing the pull-request and will redo with cherry-picking as soon as I find the time.

MaxKellermann:
Backporting drivers into a stable branch is "somewhat ok", because drivers are isolated, and the (potentially unstable) code is only ever used if a user explicitly chooses the driver. I'm not 100% happy with such a backport, but if a developer wants to push it, I won't reject it just for that one formal reason.

felixhaedicke:
Btw. if you are going to revise the code for a pull request, there is no need to close the pull request and (and later create a new one). You can just force-push the new code to the branch you were using for your pull request.

PR #55, opened 2017-09-17
doc/manual/dev: Add missing mesa packages

nallison:
On a new installation of Debian 9 both mesa-common-dev and libgl1-mesa-dev were needed as well as libegl1-mesa-dev to successfully compile to UNIX


MaxKellermann:
`mesa-common-dev` should be removed because it's an internal dependency of `libgl1-mesa-dev`, and depending on it would be fragile (and is unnecessary).
`libgl1-mesa-dev` however is truly needed. It should also be added to `debian/control`.

MaxKellermann:
Will you correct your patch, or shall we close this PR?

nallison:
Hi Max,

I plan to fix this according to your review but other work commitments have kept me from from completing it as I need to learn about using github to update the branch and PR correctly plus understand my mistakes (likely from misuse of apt-file search in locating the missing files).  I'm sorry for the delay.

-- 
Neil

MaxKellermann:
Cherry-picked to v6.8.x: d0b647f8af6a4c4673423f73f99c3d0cf37a5123 and bf9ca603654a4708bd8f579fdd61f77ea19981c8

PR #56, opened 2017-09-22
Waypoint/WaypointReaderSeeYou: handle file with rwwidth field

ericcarden:
Some SeeYou WP files contain a field for runway width (called "rwwidth" in the file header, if present). For example, at least some versions of SeeYou Mobile create files like this (if saved due to user editing of a WP). This extra field is inserted just before the radio frequency field, thus shifting that field and the description field to the right. Without this change, when XCSoar uses such a file, the radio frequency isn't shown (in XCSoar's frequency field), the description/comments field in XCSoar contains only the radio frequency (from the WP file), and the description/comments from the WP file are nowhere in XCSoar.

### DIFF
--- a/src/Waypoint/WaypointReaderSeeYou.cpp
+++ b/src/Waypoint/WaypointReaderSeeYou.cpp
@@ -169,19 +169,8 @@ WaypointReaderSeeYou::ParseLine(const TCHAR* line, Waypoints &waypoints)
     iElevation = 5,
     iStyle = 6,
     iRWDir = 7,
-    iRWLen = 8,
-    iFrequency = 9,
-    iDescription = 10,
+    iRWLen = 8
   };
-
-  if (first) {
-    first = false;
-
-    /* skip first line if it doesn't begin with a quotation character
-       (usually the field order line) */
-    if (line[0] != _T('\"'))
-      return true;
-  }
 
   // If (end-of-file or comment)
   if (StringIsEmpty(line) ||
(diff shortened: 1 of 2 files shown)


ericcarden:
This is a replacement of PR #47 (now closed), in which @MaxKellermann asked that I do this without using static variables. Done. In a few minutes, I'll create a pull request for the associated unit testing changes (the same as the now-closed PR #48).

Thanks,

Eric

src/Waypoint/WaypointReaderSeeYou.cpp
```diff
@@ -169,20 +169,9 @@ WaypointReaderSeeYou::ParseLine(const TCHAR* line, Waypoints &waypoints)
     iElevation = 5,
     iStyle = 6,
     iRWDir = 7,
-    iRWLen = 8,
-    iFrequency = 9,
-    iDescription = 10,
+    iRWLen = 8
```
MaxKellermann: Why remove the comma? This makes it hard to add more items, for no reason. There was a trailing comma before.

src/Waypoint/WaypointReaderSeeYou.hpp
```diff
@@ -44,6 +44,11 @@ class WaypointReaderSeeYou final : public WaypointReaderBase {
 protected:
   /* virtual methods from class WaypointReaderBase */
   bool ParseLine(const TCHAR* line, Waypoints &way_points) override;
+
+private:
+  /* field positions for typical SeeYou *.cup waypoint file */
```
MaxKellermann: Variables belong to the top, before the constructor, according to our coding style.

src/Waypoint/WaypointReaderSeeYou.hpp
```diff
@@ -44,6 +44,11 @@ class WaypointReaderSeeYou final : public WaypointReaderBase {
...
+  unsigned short iFrequency = 9;
```
MaxKellermann: Why is this a `short`? Are you aware that `short` is slower than a default-sized integer? Sure, your way, two `shorts` can be packed into one 32 bit word, saving 4 bytes of memory (theoretically, but only on a 32 bit CPU which is a fading technology). But the saving win must outweigh the performance loss, for example when you have thousands of instances; but this class has only temporary instances, and usually only one at a time, during startup only.

ericcarden:
Thanks for the review, @MaxKellermann. I just added a commit that restores the trailing comma, changes the variables to regular ints (instead of short ints), and puts the new variable declarations in the proper place. I hope this is a proper way of adding these changes. I didn't know about regular ints performing better. Thanks for that info.

MaxKellermann:
Squashed both commits into one: 9ddb0789624b3e44682a3ee348240e7d04b08a34

MaxKellermann:
Correct commit id: 3e3fa4fda125e88afdad60ae95244382493c776c

ericcarden:
Thanks, @MaxKellermann. I haven't squashed before, but I'll try to do that myself next time - to make things easier for you.

PR #57, opened 2017-09-22
test/src: add testing for CUP file with runway width field

ericcarden:
Some CUP files (e.g., those saved/created by SeeYou Mobile) contain a runway width (rwwidth) field. This added testing demonstrates failures with such a file with the current code but demonstrates success with the pull request I submitted for a fix (PR #56).


MaxKellermann:
I'd prefer to have the unit test change squashed in the #56 commit. If you add/change a feature, the same commit should also adjust the unit test. It belongs together naturally. (Not so important - just a hint, because I'm not merging #56 just yet.)

ericcarden:
Ok. I just guessed that you might want to merge the testing changes first/separately in order to see the result with the current code. No problem, though. I'll combine them in the future. Thanks.

PR #58, opened 2017-10-10
OpenVario remote Sentence

Parateam:
OpenVario Sentence (perhaps $POV,K) for remote xcsoar via BLE? Like Keyboard entrys? Perhaps it's simple to start an Keyboard event.

I think about the Keyboard values: 4 * Arrow, Enter ESC and F1

I've built a small Vario with a small joystick and the test with XCsoar in HID Keyboard mode was fantastic! To remote XCS with Handgloves during paragliding was fine! But the BLE module now is an HM-11 which i use for the Vario and GPS data. The HM-11 can not transmit HID-Keys, So i think a solution would be an "OpenVario Keyboard Event sentence".

Peter


MaxKellermann:
Everything in this pull request is wrong. I think you selected the wrong branch. Please resubmit with the correct branch.

PR #59, opened 2017-10-22
Add basic CircleCI instructions

Turbo87:
Addresses https://github.com/XCSoar/XCSoar/pull/50#issuecomment-337133766

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,35 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: buildpack-deps:stretch
+
+    steps:
+      - type: shell
+        name: Install System Dependencies
+        command: |
+          apt-get update -qqy
+          apt-get install -qqy --no-install-recommends \
+            ffmpeg \
+            gettext \
+            libasound2-dev \
+            libgeotiff-dev \
+            libgl1-mesa-dev \
+            liblua5.2-dev \
+            librsvg2-bin \
+            quilt \
+            xsltproc
+
+      - type: checkout
+
+      - type: run
+        name: Update git submodules
+        command: git submodule update --init --recursive
+
+      - type: run
+        name: Build
+        command: make -j2 TARGET=UNIX everything
+
+      - type: run
+        name: Tests
+        command: make TARGET=UNIX check


MaxKellermann:
Why not v6.8.x?

Turbo87:
because I interpreted

> It's been known for more than a month that this PR is broken on master, but still nobody has repaired it. [...] Sounds wrong to me.

as you wanting to have it on `master` first 🤷‍♂️

MaxKellermann:
No, I don't want to pull something into v6.8.x that's known to break once I merge it into master. Everything merged into v6.8.x will eventually go into master, but commits merged into master will (almost) never land in v6.8.x.

Turbo87:
so what is it that you would like me to do then?

MaxKellermann:
First of all, let me clarify that I don't think this CI is particularly useful. This builds just the UNIX target with exactly one configuration. Every time before I merge/push something, I build 15 different targets. Compared to that, the coverage provided by your PR is ridiculous, because most build failures are in other targets. But those targets may be added later to CI, which may turn this into something useful. Or not. I don't know, because I have yet to see something which convinces me.
But I agree to merge this if you believe it's useful, ridiculous coverage or not.
I asked why you did not want to have it in v6.8.x, and your decision was based on a misunderstanding.
So now that we cleared that out, does this change anything for you? Do you wish to have CI on both v6.8.x and master, or only on master?

Turbo87:
> So now that we cleared that out, does this change anything for you? Do you wish to have CI on both v6.8.x and master, or only on master?

I think it would make sense on both branches.

> Compared to that, the coverage provided by your PR is ridiculous

Calling PRs ridiculous is an awesome way to motivate contributors... 

> This builds just the UNIX target with exactly one configuration.

I wanted to get started with something that at least catches the most obvious errors. We can still extend this later with other configurations although that will obviously increase the build times.

Given the ridiculousness of the XCSoar test suite this might not actually provide too much benefit, but I think it's better than having nothing at all and only relying on manual reviews.

MaxKellermann:
> Calling PRs ridiculous is an awesome way to motivate contributors...

Irony is a very bad way of communication. Using it to criticize somebody is certainly an efficient way to demotivate him.

Now please turn on your language&logic circuits and find out that I never called this PR ridiculous - just in case your sentence attempted to imply that.

> Given the ridiculousness of the XCSoar test suite

You know very well that I authored almost all of them, and there were none at all before. I believe you call my work "ridiculous" as a kind of revenge for feeling hurt by me (when you really misunderstood that part). But I think we can have a global, always-true agreement: there are never enough unit tests, and existing unit tests are not good enough. So let me see your PRs to improve them in a way that you don't need to call them "ridiculous" anymore!

Be more constructive.

Turbo87:
> Be more constructive.

👇 

> so what is it that you would like me to do then?

MaxKellermann:
You wrote "it would make sense on both branches" so I thought it was obvious where to go from here (regarding CircleCI), after clearing up your misunderstandings.

I'm not interested in CircleCI integration, I only questioned whether merging to master was really what you wanted; and it turned out my doubts were right, and only master was **not** what you wanted. This PR contradicts with your stated intention, so I'm holding it back until this contradiction is resolved. But I don't "like you to do" something about it - I'm not convinced that this PR is useful. If you seriously don't know how to resolve this contradiction, I can explain it to you in more detail, but I think you're clever enough.

Now answering your question exactly (repeating myself): what I'd like you to do is improve the unit tests until you don't call them "ridiculous" anymore. Yes, really, that would be awesome!

Turbo87:
I rebased the branch on top of `v6.8.x` now and retargeted the PR accordingly.

PR #60, opened 2017-11-21
build/host.mk: Add CXX_FEATURES to HOST_CXXFLAGS

hor63:
The host compiler requires the C++ standards settings (C++14) like the target compiler does.
This is required to translate XCSoar on distries like Ubuntu Xenial LTS whose stock compiler
is still GCC 5.4. GCC 5.4 requires explicit -std=gnu++11 or gnu++14.


felixhaedicke:
Looks wrong to me. From my point of view, HOST_CXXFLAGS should not contain any target-specific flags, but CXX_FEATURES might do so.

Please check if a1bc377 fixes the problem for you.

minus5m:
Either one worked for me.
Thank you!

hor63:
Sure,

This will do it.

Thanks,
Kai

Sent from my iPad

On Dec 21, 2017, at 00:17, Felix Hädicke <notifications@github.com<mailto:notifications@github.com>> wrote:


Looks wrong to me. From my point of view, HOST_CXXFLAGS should not contain any target-specific flags, but CXX_FEATURES might do so.

Please check if a1bc377<https://github.com/XCSoar/XCSoar/commit/a1bc377a2eb85af3d3389d51a31bc0756f16046a> fixes the problem for you.

PR #61, opened 2018-01-15
Max communication payload size changed for IMI Erixx driver.

kicune:


MaxKellermann:
If your PR needs explanatory text, then your commit message is bad. Please fix it.

kicune:
I ammended the commit message & removed the explanatory text. Let me know if it's OK now.

lordfolken:
Please add a line to NEWS.txt so we can merge this.

kerel-fs:
Fixed the commit message, added a mention in NEWS.txt and rebased on `master` in this branch: [fix/IMI_PSize](https://github.com/kerel-fs/xcsoar/tree/fix/IMI_PSize)

Beside this housekeeping, I've no insight into the actual commit nor did I test it.

@kicune Thanks for your fix! If you have questions on how to perform these housekeeping steps yourself, just ask e.g. via Chat  (https://xcsoar.org/contact/).

MaxKellermann:
But which changes to the logger protocol? Where are these changes documented?

kicune:
I do not have the documentation available and after an e-mail discussion with Erixx producer I don't think it exists.

I came across the problem with logger; I reverse engineered the protocol and I identified and fixed the problem. Then I confronted the company that makes the logger and they confirmed that indeed my assumptions were corrrect and that the max communication payload size buffer was enlarged to 2kB in the new version of the logger's firmware. They also confirmed that this was the only significant change related to the comm protocol in this fw version.

I do not produce the HW, nor I have any special connection to them. I did not write the original driver. I am just the guy who own the thing and who's trying to help. That's about it.

lordfolken:
I have gotten some communication from imi. They will forward my question to the lead programmer. I will update this PR as soon as I have further information.

kicune:
Yes, Juraj Rojko. Let me know if you need his e-mail directly.
Here is the rough translation of our discussion from Jan 2018:

Me:
TY for your answer. Does it mean that you changed the value of COMM_MAX_PAYLOAD_SIZE and the change just did not make it to XCSoar source? That would be perfect and simple (and the only thing I need is the new magic number for the buffer size).

Juraj:
The driver for Erixx was written quite a long time ago, basicaly without our participation, I just provided them with data structures, some basic description and IIRC also with some simple demo code. Since then there were some minor changes in the Erixx FW. COMM_MAX_PAYLOAD_SIZE has now the value of 2048. I am not absolutely sure if that was the only change. If you can recompile the XCSoar, I would recommend to try the new size. If this does not help, then we could compare the definitions in XCSoar with ours. I am sure that the differences are not significant and it should be simple to make it work.

(then we discussed some details about the paging in the IMI protocol, which are not relevant to this topic)

Me: OK, thanks very much. With the changed buffer size it works perfectly on all available devices. I will fork XCSoar sources and create a pull request with the resized COMM_MAX_PAYLOAD_SIZE as the only change. Should you remember any further changes in the protocol, let me know so I can put it everything together.

(and I haven't heard from him since)

Hope this helps.

lordfolken:
The lk8000 guys seem to allocate dynamically:
https://github.com/LK8000/LK8000/search?q=tmsg&unscoped_q=tmsg

MaxKellermann:
@kicune please push the information you have about the problem and the solution into the commit message.

MaxKellermann:
@lordfolken this code doesn't look like dynamic allocation. It's ugly code involving CPP macros to calculate the size of a type definition, which is already known at compile time. Exactly the kind of code quality I wouldn't like to see in XCSoar.

MaxKellermann:
I don't understand the two questions.

The thing is that your commit message doesn't explain anything, but you posted more information in this PR as comments. Comments are lost eventually, so the commit message should contain everything one needs to understand the code change.
Improve the commit messsage!

MaxKellermann:
Is this going anywhere? Are you going to clean up and document this PR properly?

kicune:
No. I am honestly not able to follow your guidelines regarding the merge, nor I understand the problem you have with this trivial pull request. Please close it up and let somebody else who is more in sync with your process recreate the fix. The problem and the recommended resolution was described quit clearly, I believe.

PR #62, opened 2018-01-15
FEAT: fire 3 more GCE-events

raffiniert:
I've added these events to fire the following additional GCE-events:

- GCE_WIND_CHANGE
- GCE_AIRSPACE_INSIDE
- GCE_AIRSPACE_NEAR

Paragliding-pilots use them to fire XCI-events and send audible warnings to external devices like the Bluefly vario.

Example-code for XCI with bluefly (will output a 440Hz tone for 0.25s on "GCE_AIRSPACE_INSIDE"):

```
mode=default
type=gce
data=AIRSPACE_INSIDE
event=SendNMEAPort1 BSD 440 250
```


MaxKellermann:
As discussed on IRC, I object to this PR because:

- the stable branch won't get any new features (and contrary to what @raffiniert said, this does not implement missing documented features)
- this extends the deprecated "GCE" event system instead of adding Lua hooks

And:

- this PR has become a mess with those badly documented commits

PR #63, opened 2018-01-27
Show combined FLARM traffic from all available devices

kedder:
The motivation is to support traffic streams from both FLARM and
OGN (Open Glider Network) devices and display it on a single "FLARM Radar"
screen.

OGN devices can output position of nearby aircraft using FLARM protocol
(i.e. using $PFLAA sentences), so that XCSoar or other FLARM displays
could recognize and display the traffic data. If an aircraft is equipped
with both FLARM and OGN device and both transmit data to XCSoar, XCSoar
sees both devices as "FLARM", but displays traffic only from the first
device.

This patch fixes the problem by combining traffic lists from both devices.
If an aircraft with the same ID appears on more than one device, the data
from only the first device will be used.


Jarthianur:
Is I'm seeing this now, I think it could be interesting for You to have a look at my project [VFRB](https://github.com/Jarthianur/VirtualFlightRadar-Backend) here.

I don't want to advertise it at this point, but this project was exactly intended to merge all different kinds of input to, not only FLARM, but aircraft displays like XCSoar, in the first place.

An integration with XCSoar would be really awesome.

Yes, it's not perfect yet, but it's growing.

MaxKellermann:
Breaks the "everything" build target:
```src/FLARM/Data.hpp:61: undefined reference to `TrafficList::Complement(TrafficList const&)'```

kedder:
@MaxKellermann, I've fixed the build and rebased my branch on top of master. 

I wasn't initially comfortable with non-trivial implementation in header file, but probably that's the path of least resistance.

MaxKellermann:
I'm not comfortable with this either. You chose not to fix the problem, and found an ugly workaround instead.

kedder:
Ok, I can try to move the method to .cpp. 

There must be some reason why XCSoar has some methods implemented right in .hpp files though. @MaxKellermann, can you give some insights into this?

kedder:
I did not consider it a "problem". I'm not familiar with XCSoar code conventions very much and reasoning behind them, so I chose to follow the pattern instead - especially, since it produces smaller diff.

I'll try to move the method to cpp, but it seems that it will involve some refactoring. The method in question is being called in other header files - that code needs to be moved to respective .cpp too. It might create some domino effect.

MaxKellermann:
Don't worry about this right now - I merged your patch, I just expressed a bit of discomfort, nothing else. There are more important problems waiting to be solved. If you have some time, spend it on these, and send more PRs.

MaxKellermann:
This PR caused https://github.com/XCSoar/XCSoar/issues/925

PR #67, opened 2018-02-26
Add Python 3 compatibility

Turbo87:
see https://github.com/skylines-project/py-xcsoar/issues/15

/cc @roeles @TobiasLohner

### DIFF
--- a/python/src/Flight.cpp
+++ b/python/src/Flight.cpp
@@ -53,11 +53,24 @@ PyObject* xcsoar_Flight_new(PyTypeObject *type, PyObject *args, PyObject *kwargs
   self = (Pyxcsoar_Flight *)type->tp_alloc(type, 0);
   self->filename = nullptr;
 
+#if PY_MAJOR_VERSION >= 3
+  if (PyUnicode_Check(py_input_data)) {
+    Py_ssize_t length;
+    char *ptr = PyUnicode_AsUTF8AndSize(py_input_data, &length);
+    if (!ptr) {
+        return NULL;
+    }
+
+    // add one char for \0
+    self->filename = new char[length + 1];
+    strncpy(self->filename, ptr, length + 1);
+#else
   if (PyString_Check(py_input_data) || PyUnicode_Check(py_input_data)) {
     Py_ssize_t length = PyString_Size(py_input_data);
     // add one char for \0
     self->filename = new char[length + 1];
     strncpy(self->filename, PyString_AsString(py_input_data), length + 1);
+#endif
 
     Py_BEGIN_ALLOW_THREADS
     self->flight = new Flight(self->filename, keep);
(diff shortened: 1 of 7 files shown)


roeles:
Nice! Thanks for the mention.

Turbo87:
FYI The last commit was the thing that caused https://github.com/skylines-project/py-xcsoar/issues/15. Looking at `nm -D /usr/local/lib/python3.3/site-packages/xcsoar.so` revealed that `PyInit_xcsoar()` was indeed not visible, and in the end by turning random settings on and off I discovered that those options were responsible for it. Unfortunately I fail to understand why `__attribute__ ((visibility("default")))` does not seem to work properly.

roeles:
From a quick read I understand that it has something to do with function exporting. That is, making the fact that a function exists in a .so file visible to anyone that loads the .so file (I suspect python uses dlopen/dlsym for this).
I don't know what the `__attribute__ ((visibility("default")))` implies. Also, there exists `__attribute__ ((visibility("dllexport")))`.

python/src/Flight.cpp
```diff
@@ -53,11 +53,24 @@ PyObject* xcsoar_Flight_new(PyTypeObject *type, PyObject *args, PyObject *kwargs
   self = (Pyxcsoar_Flight *)type->tp_alloc(type, 0);
   self->filename = nullptr;
 
+#if PY_MAJOR_VERSION >= 3
+  if (PyUnicode_Check(py_input_data)) {
```
TobiasLohner: Wouldn't it be nice to allow byte strings in Python 3, too?

python/src/PythonConverters.cpp
```diff
@@ -394,6 +399,12 @@ bool Python::PyTupleToIGCFixEnhanced(PyObject *py_fix, IGCFixEnhanced &fix) {
 }
 
 bool Python::PyStringToString(PyObject *py_string, tstring &string) {
+#if PY_MAJOR_VERSION >= 3
+  if (PyUnicode_Check(py_string)) {
```
TobiasLohner: Wouldn't it be nice to allow byte strings in Python 3, too?

python/src/Util.cpp
```diff
@@ -64,12 +67,21 @@ PyObject* xcsoar_encode(PyObject *self, PyObject *args, PyObject *kwargs) {
 
   if (py_method == nullptr)
     method = UNSIGNED;
+#if PY_MAJOR_VERSION >= 3
```
TobiasLohner: Wouldn't it be nice to allow byte strings, too?

python/src/Flight.cpp
Turbo87: It would be nice, yes. My main goal was to make the py-xcsoar test suite work though and it seems to be happy enough with the changes in the PR.

roeles:
I suggest accepting the pull request and opening an issue for this.

Op za 24 mrt. 2018 17:55 schreef Tobias Bieniek <notifications@github.com>:

Turbo87:
> I suggest accepting the pull request and opening an issue for this.

Agreed. I've tested this PR locally inside of SkyLines and on SkyLines CI and it seems to work quite well. Everything that can be implemented on top of this can IMHO be added later if needed.

PR #71, opened 2018-03-18
Device Manager: Show if sensor data is available

fb:
This adds labels for Temperature and Humidity to the device dialog if a device provides that data.

### DIFF
--- a/src/Dialogs/Device/DeviceListDialog.cpp
+++ b/src/Dialogs/Device/DeviceListDialog.cpp
@@ -81,6 +81,8 @@ class DeviceListWidget final
     bool duplicate:1;
     bool open:1, error:1;
     bool alive:1, location:1, gps:1, baro:1, airspeed:1, vario:1, traffic:1;
+    bool temperature:1;
+    bool humidity:1;
     bool debug:1;
 
     void Set(const DeviceConfig &config, const DeviceDescriptor &device,
@@ -116,6 +118,8 @@ class DeviceListWidget final
       airspeed = basic.airspeed_available;
       vario = basic.total_energy_vario_available;
       traffic = basic.flarm.IsDetected();
+      temperature = basic.temperature_available;
+      humidity = basic.humidity_available;
       debug = device.IsDumpEnabled();
     }
   };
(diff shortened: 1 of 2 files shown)


src/InfoBoxes/Content/Weather.cpp
```diff
@@ -47,7 +47,7 @@ UpdateInfoBoxHumidity(InfoBoxData &data)
   }
 
   // Set Value
-  data.UnsafeFormatValue( _T("%d"), (int)basic.humidity);
+  data.UnsafeFormatValue( _T("%2.1f"), (float)basic.humidity);
```
Turbo87: this seems unrelated to the device manager changes 🤔

src/Dialogs/Device/DeviceListDialog.cpp
```diff
@@ -385,6 +389,12 @@ DeviceListWidget::OnPaintItem(Canvas &canvas, const PixelRect rc, unsigned idx)
     if (flags.traffic)
       buffer.append(_T("; FLARM"));
 
+    if (flags.temperature)
+      buffer.append(_T("; Temperature"));
+
+    if (flags.humidity)
+      buffer.append(_T("; Humidity"));
```
Turbo87: I'm not too sure about this. If we would add every datapoint we receive here this will result in a looong string... 🤔

src/InfoBoxes/Content/Weather.cpp
fb: true

src/Dialogs/Device/DeviceListDialog.cpp
fb: I actually had that same thought back when I implemented it. I tried `; T` and `; H`, which looked a bit cryptic.

fb:
Okay, it *results* in a long string. If you changed the language to German ("Luftgeschwindigkeit" instead of "Airspeed") or enabled debug mode the string would get too long even on a high DPI screen. Not sure if it's critical, though.

The flags on the device screen are very helpful to quickly find out if everything is working properly (without looking into the raw data stream, which I have to do at the moment), but abbreviations should be sufficient for that.

Turbo87:
can we maybe group it under a single flag called "Atmosphere" or "Environment" or something like that?

fb:
I rebased onto master and appended an entry in NEWS.txt

Issue #108, opened 2018-06-01
Downloadable options listed twice


jharting:
When selecting airspace/waypoint database to download sometimes I can see the options listed twice on one big list:

![screenshot_20180601_194514](https://user-images.githubusercontent.com/167352/40855379-b77c116e-65d4-11e8-85a8-61985bc457e1.png)

PR #109, opened 2018-06-01
prevent DownloadFilePicker options from being listed twice

jharting:
When the repository is cached locally  DownloadFilePickerWidget::RefreshList is called twice on initial load - once
for the local cache and then again when the repository download finishes. Since RefreshList() is not idempotent this results
in repository items to be added twice to the items vector. This change prevents that by clearing the items vector each time.

Fixes https://github.com/XCSoar/XCSoar/issues/108

### DIFF
--- a/src/Dialogs/DownloadFilePicker.cpp
+++ b/src/Dialogs/DownloadFilePicker.cpp
@@ -264,6 +264,7 @@ try {
 
   ParseFileRepository(repository, reader);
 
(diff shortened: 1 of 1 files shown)


Turbo87:
@jharting is that also an issue on v6.8.x?

jharting:
@Turbo87 yeah I think so. I'll send a separate PR for 6.8.x

PR #114, opened 2018-06-05
[6.8] prevent DownloadFilePicker options from being listed twice

jharting:
When the repository is cached locally  DownloadFilePickerWidget::RefreshList is called twice on initial load - once
for the local cache and then again when the repository download finishes. Since RefreshList() is not idempotent this results
in repository items to be added twice to the items vector. This change prevents that by clearing the items vector each time.

Fixes https://github.com/XCSoar/XCSoar/issues/108


jharting:
backport of https://github.com/XCSoar/XCSoar/pull/109

PR #122, opened 2018-06-23
Disabled set_idle call that caused crashes on Android 8.0

kicune:
Based on changes in the commit fc80aaf from v6.8.x branch.

### DIFF
--- a/src/Thread/Util.hpp
+++ b/src/Thread/Util.hpp
@@ -30,7 +30,7 @@
 #ifndef THREAD_UTIL_HXX
 #define THREAD_UTIL_HXX
 
-#ifdef __linux__
+#if defined(__linux__) && !defined(ANDROID)
 #include <sched.h>
 #include <sys/syscall.h>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


kerel-fs:
Thanks for trying to port the commit!
I can also confirm that this fixes the crash of the current master branch on LineageOS 15.1 / Android 8.1.0

Three problems thought:
- please don't edit the commit message when cherry-picking. It should still describe the commit, not how the commit ended somewhere (you could append the message at the end though)
- the original commit added a message to the changelog/NEWS.txt, this line is missing here (probably dropped accidentally when trying to resolve the merge conflict during cherry-picking).
- the new commit message refers to a commit 6076c50, which seems to be an accidental intermediate commit, not part of the public history.

I also did the cherry-pick, see:
https://github.com/kerel-fs/XCSoar/commits/fix/120
Note that the original commit message and author was preserved.

kicune:
Ok, understood - thanks for the explanation & example. I will try not to mess up next time. ;)
Would you create a push request from your fork? I am afraid that I would only add to the commotion if I tried to fix my push.

MaxKellermann:
The branch should be merged instead of cherry-picking commits.

MaxKellermann:
@lordfolken I objected merging this PR because it is a bad cherry-pick with wrong authorship information.

PR #126, opened 2018-07-14
Build system: Added glm lib.

Exadios:
This patch aims to address Issue 125.


MaxKellermann:
GLM is already available as a git submodule.

lordfolken:
You need to update the submodules before build. GLM is in there and then the path works.
```
git submodule init
git submodule update --recursive
```
Please read the build instructions: 
http://download.xcsoar.org/releases/6.8.10/XCSoar-developer-manual.pdf
Section 2.1

PR #127, opened 2018-07-15
Ticket 1687 on master

Exadios:
This branch adds AAT turn point type 'Variable Sector'. See ticket 1687 for context.


Turbo87:
This PR seems to contain a lot more than just a new turnpoint type 🤔

Exadios:
Are you referring to commits such as https://github.com/XCSoar/XCSoar/pull/127/commits/1f03cb00700bae50913573804d355cfde3c98df6? If so them these commits are the result of rebases from the upstream master.

Turbo87:
yes, commits like that. they shouldn't appear on this branch. the branch should only contain the relevant commits for this change.

Exadios:
OK. I was under the impression that you wanted the feature branches to be rebased. As I understand it these should drop out in the merge.

Turbo87:
> I was under the impression that you wanted the feature branches to be rebased.

that is correct, but you should only rebase your own commits, not any irrelevant commits. you can use `git log` or `gitk` to check how the branch looks like and `git rebase --interactive` to throw away irrelevant commits.

Exadios:
OK. I think I see what has happened. There's been a reset upstream. The actual rebase went OK but the repository Exadios/XCSoar then became out of sync. Instead of forcing the push to that repository I merged from it. This put my local copy and my remote repository in sync again but re introduced the commits that had been reset out upstream.

I'll fix it up.

MaxKellermann:
Hmmm.... "Ticket 1687 preparation for upstream push." this PR looks like a mess.

MaxKellermann:
Is this PR going anywhere? In the current state (28 confusing commits) it cannot be reviewed.

MaxKellermann:
Will be reopened after this branch has been sanitized.

PR #129, opened 2018-07-19
Revert "Disabled set_idle call that caused crashes on Android 8.0"

lordfolken:
Reverts XCSoar/XCSoar#122


MaxKellermann:
ac5d8999c2de96cf20cde003c118e53dd5b0b9ee

Issue #139, opened 2018-09-03
Possible null pointer dereference


Exadios:
In src/Form/DataField/Prefix.cpp, PrefixDataField::GetAsDisplayString() it is possible that 's' is null.
    const TCHAR *s = DataFieldString::GetAsDisplayString();
    if (StringIsEmpty(s))

MaxKellermann:
How can this happen?

MaxKellermann:
Closing due to lack of feedback.

Exadios:
I thought that this issue had been closed some time ago.

What feedback where you expecting?

MaxKellermann:
#140 was closed two weeks ago, because it was the wrong approach to fix this bug report.
I was expecting feedback as in "answering my question".

Exadios:
OK. But I will have to go over it again.

WRT XCSoar I'm presently working on a program termination and (what I am supposing to be) associated valgrind so I'll continue with that.

PR #140, opened 2018-09-03
Util/StringCompare: Handle the case when an arg is null.

Exadios:
It is possible that string is null in StringIsEmpty(const char *string)
Handle this case.

Issue #139


MaxKellermann:
You didn't explain how this "is possible".
Your commit adds global overhead to all callers, but it is illegal to pass `nullptr` here. If you believe somebody passes `nullptr`, fix the caller, not the function.

PR #141, opened 2018-09-09
Restarted working on manual topics by sorting an old  stash

hjr:


MaxKellermann:
I think the `.gitignore` patch is too broad. For example, `*.config` may match configuration files which we may want to add eventually, and your change will cost hours and hours of debugging why the file isn't there. Can you narrow it down so it really catches only QtCreator files?

hjr:
Agreed. The file names are <project name>.config. So it will be difficult to catch them much more precisely. I'll remove this part.

hjr:
Will create a new request.

PR #146, opened 2018-09-23
Identify 2D fixes

weshill:
Currently, XCSoar sets fix validity to A based on whether there is a valid position available and a valid GPS height available. However, both these conditions are true for a 2D fix which should have fix validity set to V.
See www.sacra.biz/2dfixes/DEBUG16082018skydrop.txt and www.sacra.biz/2dfixes/NMEAIN02092018oudie.txt for examples.

The correct way to identify 2D fixes is using the NMEA GSA sentence. This Pull Request will resolve the issue.

### DIFF
--- a/src/Device/Parser.cpp
+++ b/src/Device/Parser.cpp
@@ -307,9 +307,13 @@ NMEAParser::GSA(NMEAInputLine &line, NMEAInfo &info)
    */
 
   line.Skip();
-
-  if (line.Read(0) == 1)
-    info.location_available.Clear();
+  
+  // determine fix type
+  unsigned fix_mode;  
+  fix_mode = line.Read(0);  
+  if (fix_mode == 1)
+    info.location_available.Clear();  
+  info.fix_2d = (fix_mode == 2);
 
   // satellites are in items 4-15 of GSA string (4-15 is 1-indexed)
   for (unsigned i = 0; i < GPSState::MAXSATELLITES; i++)
--- a/src/Device/Parser.cpp
+++ b/src/Device/Parser.cpp
@@ -308,9 +308,7 @@ NMEAParser::GSA(NMEAInputLine &line, NMEAInfo &info)
 
   line.Skip();
   
-  // determine fix type
-  unsigned fix_mode;  
-  fix_mode = line.Read(0);  
+  unsigned fix_mode = line.Read(0);  
   if (fix_mode == 1)
     info.location_available.Clear();  
   info.fix_2d = (fix_mode == 2);
--- a/src/IGC/IGCFix.cpp
+++ b/src/IGC/IGCFix.cpp
@@ -36,7 +36,7 @@ IGCFix::Apply(const NMEAInfo &basic)
 
   /* "Use A for a 3D fix and V for a 2D fix (no GPS altitude) or for
      no GPS data" */
-  gps_valid = basic.location_available && basic.gps_altitude_available && !(basic.fix_2d);
+  gps_valid = basic.location_available && basic.gps_altitude_available && !basic.fix_2d;
 
   if (basic.location_available)
     location = basic.location;
--- a/src/IGC/IGCFix.cpp
+++ b/src/IGC/IGCFix.cpp
@@ -36,7 +36,7 @@ IGCFix::Apply(const NMEAInfo &basic)
 
   /* "Use A for a 3D fix and V for a 2D fix (no GPS altitude) or for
      no GPS data" */
-  gps_valid = basic.location_available && basic.gps_altitude_available && !basic.fix_2d;
+  gps_valid = basic.location_available && basic.gps_altitude_available && (!basic.fix_2d);
 
   if (basic.location_available)
     location = basic.location;
--- a/src/IGC/IGCFix.cpp
+++ b/src/IGC/IGCFix.cpp
@@ -36,16 +36,18 @@ IGCFix::Apply(const NMEAInfo &basic)
 
   /* "Use A for a 3D fix and V for a 2D fix (no GPS altitude) or for
      no GPS data" */
-  gps_valid = basic.location_available && basic.gps_altitude_available;
+  gps_valid = basic.location_available && basic.gps_altitude_available && !(basic.fix_2d);
 
   if (basic.location_available)
     location = basic.location;
 
   time = basic.date_time_utc;
-
-  gps_altitude = basic.gps_altitude_available
-    ? (int)basic.gps_altitude
-    : 0;
+  
+  gps_altitude = basic.fix_2d
+	  ? 0
+	  : (basic.gps_altitude_available
+         ? (int)basic.gps_altitude
+         : 0);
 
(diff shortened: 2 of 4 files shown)


hor63:
Treat 2D-fixes like no fix at all. They are trash, i.e. huge uncertainty, usually only 3 satellites fix.
Here's the quote from the authoritive document on ICG records:
FAI-IGC: TECHNICAL SPECIFICATION FOR GNSS FLIGHT RECORDERS
Second Edition with Amendment 4a - 10 April 2016

Page 29
Field "Fix validity" in the B-record:
"Use A for a 3D fix and V for a 2D fix (no GPS altitude) or for no GPS data
(pressure altitude data must continue to be recorded using UTC times from the
RTC)"
No fix and 2d-fixes are treated equal. The point of such B-records is continuous recording of the barometric height, and continuity of the recording. From IGC's point of view 2D fixes and no fixes are equivalent.

Regards,
Kai

weshill:
"Treat 2D-fixes like no fix at all. They are trash, i.e. huge uncertainty, usually only 3 satellites fix." 

Kai, I agree that 2D fixes are not accurate. XCSoar currently records 2D fixes as 3D fixes, so this change would rectify that. 

However, it needs the GSA sentence to do that and some instruments (eg LXNAV bluetooth) do not forward the GSA sentence - I am still looking at how to handle that issue.

You would be surprised how many 2D fixes use 4 satellites. I have tested PA6H and L80 modules and the chips are actually identifying 2D fixes with 4 satellites as 3D fixes. The NMEA log from my SkyDrop (L80) a few days ago shows many errors of this type. 
NMEA log - www.sacra.biz/2dfixes/DEBUG22092018skydrop.txt 
IGC file -  www.sacra.biz/2dfixes/igc22092108skydrop.IGC

src/IGC/IGCFix.cpp
```diff
@@ -36,7 +36,7 @@ IGCFix::Apply(const NMEAInfo &basic)
 
   /* "Use A for a 3D fix and V for a 2D fix (no GPS altitude) or for
      no GPS data" */
-  gps_valid = basic.location_available && basic.gps_altitude_available && !(basic.fix_2d);
+  gps_valid = basic.location_available && basic.gps_altitude_available && !basic.fix_2d;
```
MaxKellermann: Why?

src/IGC/IGCFix.cpp
```diff
@@ -36,7 +36,7 @@ IGCFix::Apply(const NMEAInfo &basic)
 
   /* "Use A for a 3D fix and V for a 2D fix (no GPS altitude) or for
      no GPS data" */
-  gps_valid = basic.location_available && basic.gps_altitude_available && !basic.fix_2d;
+  gps_valid = basic.location_available && basic.gps_altitude_available && (!basic.fix_2d);
```
MaxKellermann: Why?

src/Device/Parser.cpp
```diff
@@ -307,9 +307,13 @@ NMEAParser::GSA(NMEAInputLine &line, NMEAInfo &info)
    */
 
   line.Skip();
-
-  if (line.Read(0) == 1)
-    info.location_available.Clear();
+  
+  // determine fix type
+  unsigned fix_mode;  
+  fix_mode = line.Read(0);
```
MaxKellermann: Declare & assign in one line.

src/Device/Parser.cpp
```diff
@@ -307,9 +307,13 @@ NMEAParser::GSA(NMEAInputLine &line, NMEAInfo &info)
    */
 
   line.Skip();
-
-  if (line.Read(0) == 1)
-    info.location_available.Clear();
+  
+  // determine fix type
+  unsigned fix_mode;
```
MaxKellermann: The comment says "fix type", but the variable is named "fix_mode". Huh?
Why even the code comment, it only states the obvious (if its nomenclature wouldn't contradict with the actual code).

src/NMEA/Info.hpp
```diff
@@ -119,6 +119,12 @@ struct NMEAInfo {
    * @see AirDensityRatio
    */
   double indicated_airspeed;
+  
+  /**
+   * 2D fix information
+   * true for 2d fix
+   */  
+  bool fix_2D;
```
MaxKellermann: Who initializes this, and who expires this, and who merges this?

src/IGC/IGCFix.cpp
weshill: I have no experience with github and haven't done any C programming in 30 years. Eventually found the compiler output , identified the error and corrected it. Sorry about all the unnecessary commits.

MaxKellermann: Your commit contains no documentation on which error is corrected by this.
I don't think the parantheses are necessary, and they shouldn't be used; but your PR has one commit adding them, and one removing them. For no obvious reason.

hor63:
> However, it needs the GSA sentence to do that and some instruments (eg LXNAV bluetooth) do not forward the GSA sentence - I am still looking at how to handle that issue.
> 
> You would be surprised how many 2D fixes use 4 satellites. I have tested PA6H and L80 modules and the chips are actually identifying 2D fixes with 4 satellites as 3D fixes. The NMEA log from my SkyDrop (L80) a few days ago shows many errors of this type.
> NMEA log - [www.sacra.biz/2dfixes/DEBUG22092018skydrop.txt](http://www.sacra.biz/2dfixes/DEBUG22092018skydrop.txt)
> IGC file - [www.sacra.biz/2dfixes/igc22092108skydrop.IGC](http://www.sacra.biz/2dfixes/igc22092108skydrop.IGC)

I suggest that you assume a 2-D fix when the number of satellites is 3, or when the number of Satellites is 4, and the DOP is > 3 (or so). Both information is in the GPGGA record.
The threshold of the DOP is debatable, you may want to make that configurable to be adjustable based on experience and experiments, e.g. use examples from records with a GPGSA record to establich a baseline.

Kai

weshill:
> I suggest that you assume a 2-D fix when the number of satellites is 3, or when the number of Satellites is 4, and the DOP is > 3 (or so). Both information is in the GPGGA record.
> The threshold of the DOP is debatable, you may want to make that configurable to be adjustable based on experience and experiments, e.g. use examples from records with a GPGSA record to establich a baseline.
> 
> Kai

Kai,
2D fixes are calculated using an assumed altitude - when this altitude is wrong the errors can be significantly greater than indicated by the  DOP. Also, you can have 2D fixes with more than 3 satellites, so that is not a reliable way to detect them.
Here is an NMEA log which has up to 5 satellites in a 2D fix: www.sacra.biz/2dfixes/CAPTURE260718garmin.TXT

rawtaz:
@weshill There's a lot of good articles about writing good commit messages, but here's one that might be worth a read: https://chris.beams.io/posts/git-commit/

Please also note that the XCSoar developer manual specifices some rules to follow in this regard.

Let us know if you need any pointers or assistance!

MaxKellermann:
Closing due to lack of response

PR #151, opened 2018-10-31
Cache BLE device type and status in sharedpreferences

### LINKED ISSUE #149: Reconnect to BLE serial interface fails after Android Bluetooth restart

So two changes are needed:

clydebarrow:
Fixes #149


MaxKellermann:
Since this fixes a user-visible problem, we should close this one and continue work on #152, to be merged into master later.

PR #152, opened 2018-10-31
Backport BLE type/name fix to 6.8.x

### LINKED ISSUE #149: Reconnect to BLE serial interface fails after Android Bluetooth restart

clydebarrow:
Fixes #149 for V6.8.x


Turbo87:
just for understanding, this is the same as #151 but for the `v6.8.x` branch, correct?

clydebarrow:
Correct, same fix.

Turbo87:
okay, thanks for the clarification. I have enough permissions here to merge, but I'm not confident enough in the code at the moment to actually be comfortable doing so.

@MaxKellermann can you have a look?

MaxKellermann:
Ah, I see, so this is a backport of some unmerged code.
We shouldn't merge both PRs - if there is a bug that needs to be fixed in both branches, it should be merged to stable first, and later stable will be merged into master.

MaxKellermann:
I think your approach is a fragile solution, because it moves the decision which protocol to use to a cache, when it should really be configuration.
And I don't like that this commit tries to solve two problems at once: caching the name and fixing a connection problem.

clydebarrow:
I'm not sure that I understand the distinction here between cache and configuration. Are you suggesting the information should be stored elsewhere, i.e. in an existing datastore? It has to be stored somewhere.

MaxKellermann:
Huh, it was you who designed this distinction, and now you don't know it?
Configuration: XCSoar's profile.
Cache: your `class BluetoothCache`

clydebarrow:
The other approach I had considered which would address your concerns is to add a new type to Device/Config.hpp - currently RFCOMM is used for both legacy Bluetooth and Bluetooth LE. If a new type was added, e.g. BLE, to be used for BLE devices, and RFCOMM used only for legacy Bluetooth serial devices, this would solve the problem since there would be no ambiguity as to which type of Bluetooth device was associated with a port. This will be a much more extensive changeset though.

MaxKellermann:
That sounds like a good idea. BLE devices are not RFCOMM, after all. Note that XCSoar's usage of BLE is not "pure" BLE either, it's a protocol on top of BLE.

clydebarrow:
Quite true - with that in mind it might be better to choose a name that reflects the specific GATT characteristic that is used in the current implementation. It's a proprietary UUID, not part of an official GATT service, best as I can tell it originates from a device called an HM-10, so the device type could perhaps be BLE_HM10.

MaxKellermann:
5 months later, there is still no usable code, so closing this PR.

PR #160, opened 2018-12-31
Switch to PKG_CONFIG_PATH instead of PKG_CONFIG_LIBDIR in pkgconfig.mk

jkrautter:
See https://bugs.freedesktop.org/show_bug.cgi?id=88992
Causes build problems on systems where PKG_CONFIG_PATH is set, since PKG_CONFIG_PATH is used before falling back to PKG_CONFIG_LIBDIR

### DIFF
--- a/build/pkgconfig.mk
+++ b/build/pkgconfig.mk
@@ -1,27 +1,27 @@
 PKG_CONFIG = pkg-config
 
 ifeq ($(USE_THIRDPARTY_LIBS),y)
-  PKG_CONFIG := PKG_CONFIG_LIBDIR=$(THIRDPARTY_LIBS_ROOT)/lib/pkgconfig $(PKG_CONFIG) --static
+  PKG_CONFIG := PKG_CONFIG_PATH=$(THIRDPARTY_LIBS_ROOT)/lib/pkgconfig $(PKG_CONFIG) --static
 endif
 
 ifeq ($(TARGET_IS_DARWIN),y)
   ifeq ($(DARWIN_LIBS),)
     PKG_CONFIG := $(PKG_CONFIG) --static
   else
-    PKG_CONFIG := PKG_CONFIG_LIBDIR=$(DARWIN_LIBS)/lib/pkgconfig $(PKG_CONFIG) --static --define-variable=prefix=$(DARWIN_LIBS)
+    PKG_CONFIG := PKG_CONFIG_PATH=$(DARWIN_LIBS)/lib/pkgconfig $(PKG_CONFIG) --static --define-variable=prefix=$(DARWIN_LIBS)
   endif
 endif
 
 ifeq ($(HOST_IS_WIN32)$(HAVE_WIN32)$(HAVE_CE),nyn)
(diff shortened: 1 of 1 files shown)


MaxKellermann:
Your PR contains text which is not present in your commit message, which is bad.
But even the text in this PR doesn't explain what "build problems" you see.
Not setting PKG_CONFIG_LIBDIR is bad, because for cross builds, we don't want the default paths to be used.
If a user sets PKG_CONFIG_PATH and then sees a build failure because XCSoar uses that path ... well, duh! Then just don't set it to those incompatible values? Shit in, shit out? I mean, what's the real problem here? (This is not explained in your PR.)

jkrautter:
Sorry, I'm not too familiar with the github workflow yet. This is literally my first PR so I appreciate feedback.
On some distributions (Slackware in my case) PKG_CONFIG_PATH is set by default by the profile.d scripts. Unfortunately, the paths from PKG_CONFIG_PATH are used before PKG_CONFIG_LIBDIR (as shown in the link to the bug report). This leads to linker errors, as XCSoar uses the linker flags of the system libraries instead of the flags for the libraries built in the output directory. It took me a while to figure this out so I thought I'd push it back to you in case somebody else has this problem. What is the advantage of using PKG_CONFIG_LIBDIR over PKG_CONFIG_PATH for cross compiling? I had no problems cross compiling for kobo with my modified build script.

MaxKellermann:
If your distribution sets `PKG_CONFIG_PATH` for you, then complain to your distribution, or just undo the setting. I'm not sure if we should support broken configurations like that.
In any case, your reasoning about your change is not "this is the right thing to do" but only "this appears to fix my build problem". I'd like to argue about the former - what's the right thing to do?
Anecdotical evidence like "but this works for me" is not enough. We should know what is correct.
Setting `PKG_CONFIG_PATH` to invalid values and then seeing build failures should not be a surprise. Build failures due to invalid configurations are to be expected, this is correct behavior.

> What is the advantage of using PKG_CONFIG_LIBDIR over PKG_CONFIG_PATH for cross compiling?

It is the correct thing to do, and your change is the not correct thing to do. We don't want to have *additional* / *custom* search paths (which is the purpose of `PKG_CONFIG_PATH`), but we want to change the *default* search path. The original default must not be used, because it points to native libraries, not to cross-compiled ones.

jkrautter:
Well, whether setting PKG_CONFIG_PATH to the system library paths is 'invalid' is debatable. It doesn't say anywhere in the pkg-config manpage, that only PKG_CONFIG_LIBDIR should be used for system library paths:

> PKG_CONFIG_PATH
> A colon-separated (on Windows, semicolon-separated) list of directories to search for .pc files. The default directory will always be searched after searching the path; the default is libdir/pkgconfig:datadir/pkgconfig where libdir is the libdir where pkg-config and datadir is the datadir where pkg-config was installed. 

It doesn't say, that PKG_CONFIG_PATH is for _additional/custom_ search paths only.

In my experience, when working with other projects and teams in software development I would prioritize robustness over correctness. Sure, you may have implemented this specific networking protocol in your server as correctly as possible, but if that other popular client _doesn't_ implement it as correctly and your server refuses to talk with it, nobody is happy.

I doubt there's much sense in discussing this much more, if you'd prefer the current state I'm fine with that! In any case, this thread will be useful for other people having the same issue.

Issue #161, opened 2019-01-14
Driver for Air Control Display


Scumi:
The Air Control Display (Manufacturer Air Avionics) can be controlled via RS-232 interface. The device is sending NMEA sentences. Depending on the connected hardware and licence purchased, altimeter, COM and transponder interaction are possible.

An XCSoar driver for the altimeter and COM interaction would be desired.

The documentation seems pretty good in the Installation Manual of the Air Control Display, which can be found here: https://www.air-avionics.com?wpdmpro=air-control-display-57-installationshandbuch&lang=de&wpdmdl=982

PR #164, opened 2019-01-27
Driver for altimeter functionality of Air Control Display

Scumi:
Driver for the altimeter functionality of the Air Avionics - Air Control Display in accordance with the Installation Manual 2.0. Driver has been tested on my Openvario.

Ref.: #161

### DIFF
--- a/build/driver.mk
+++ b/build/driver.mk
@@ -124,6 +124,7 @@ DRIVER_SOURCES = \
 	$(DRIVER_SRC_DIR)/Zander.cpp \
 	$(DRIVER_SRC_DIR)/Vaulter.cpp \
 	$(DRIVER_SRC_DIR)/KRT2.cpp \
+	$(DRIVER_SRC_DIR)/AirControlDisplay.cpp \
 	$(DRIVER_SRC_DIR)/ATR833.cpp
 
 $(eval $(call link-library,driver,DRIVER))
--- a/src/Device/Driver/AirControlDisplay.cpp
+++ b/src/Device/Driver/AirControlDisplay.cpp
@@ -1,0 +1,113 @@
+/*
+Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2016 The XCSoar Project
+  A detailed list of copyright holders can be found in the file "AUTHORS".
+
+  This program is free software; you can redistribute it and/or
+  modify it under the terms of the GNU General Public License
+  as published by the Free Software Foundation; either version 2
+  of the License, or (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+
+#include "Device/Driver/AirControlDisplay.hpp"
+#include "Device/Driver.hpp"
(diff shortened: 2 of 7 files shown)


src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -0,0 +1,116 @@
...
+  /* virtual methods from class Device */
+  bool ParseNMEA(const char *line, struct NMEAInfo &info) override;
+  bool PutQNH(const AtmosphericPressure &pres,
+              OperationEnvironment &env) override;
+};
+
+bool
+ACDDevice::PutQNH(const AtmosphericPressure &pres, OperationEnvironment &env)
+{
+  if (!EnableNMEA(env))
```
MaxKellermann: Why this call?

build/driver.mk
```diff
@@ -124,6 +124,7 @@ DRIVER_SOURCES = \
 	$(DRIVER_SRC_DIR)/Zander.cpp \
 	$(DRIVER_SRC_DIR)/Vaulter.cpp \
 	$(DRIVER_SRC_DIR)/KRT2.cpp \
-	$(DRIVER_SRC_DIR)/ATR833.cpp
+	$(DRIVER_SRC_DIR)/ATR833.cpp \
+	$(DRIVER_SRC_DIR)/AirControlDisplay.cpp
```
MaxKellermann: Should be moved up to avoid modifying the ATR833 line in this patch

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include "NMEA/InputLine.hpp"
+#include "NMEA/Checksum.hpp"
+#include "Atmosphere/Pressure.hpp"
+#include "Units/System.hpp"
+#include "Math/Util.hpp"
+
+static bool
+ParsePAAVS(NMEAInputLine &line, NMEAInfo &info)
+{
+  double value;
```
MaxKellermann: According to your code comments, the values are integer. Why parse a `double` here?

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -0,0 +1,116 @@
...
+              OperationEnvironment &env) override;
+};
+
+bool
+ACDDevice::PutQNH(const AtmosphericPressure &pres, OperationEnvironment &env)
+{
+  if (!EnableNMEA(env))
+    return false;
+  char buffer[100];
+  unsigned QNHinPascal = uround(pres.GetPascal());
```
MaxKellermann: Coding style: bad variable name

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include "NMEA/Checksum.hpp"
+#include "Atmosphere/Pressure.hpp"
+#include "Units/System.hpp"
+#include "Math/Util.hpp"
+
+static bool
+ParsePAAVS(NMEAInputLine &line, NMEAInfo &info)
+{
+  int value;
+
```
MaxKellermann: There are many lines like this with trailing whitespace.

build/driver.mk
```diff
@@ -87,7 +87,7 @@ XCTRACER_SOURCES = \
 
 THERMALEXPRESS_SOURCES = \
 	$(DRIVER_SRC_DIR)/ThermalExpress/Driver.cpp
- 
+
```
MaxKellermann: This is an example of a line where we had a mistakenly added trailing space, and you removed it in this unrelated commit. This is confusing to read, and this is why there shouldn't be trailing whitespace. (See below for examples of trailing whitespace you added.)

Scumi:
Your request for the unit test was a good idea. I recognized a mismatch between the documentation and the transferred NMEA sentences. The manufacturer confirmed the mismatch. I reverted your requested change of the variable declaration, but this time updated the helptext. The altimeter sends the altitude with two decimal places.

PR #169, opened 2019-02-23
Moved the download directory out of the output directory.

wingunder:
Every time a 'make clean' or 'rm -rf output' gets run, the
output/download directory gets clobbered. This is pretty annoying,
if one does not have the luxury of a high speed Internet connection.

This patch fixes this problem, by moving the location of the download
directory to the XCSoar root directory.

In addition, the 'clean-download' make target was added, for the case
where one really wants to get rid of the download directory for some
reason.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 *.s
 *.so
 Doc/html/*
+download
 output
 perf.data
 perf.data.old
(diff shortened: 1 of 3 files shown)


MaxKellermann:
I don't agree with this change. You can always run `make DOWNLOAD_DIR=x/y/z` and have it in a permanent location outside of your git tree (not deleted by `git clean -dxf`, shared between different worktrees). This PR adds another generated top-level directory with little use.

wingunder:
Thanks for the explanation.
Maybe it would be worth documenting this in the developer manual.

PR #170, opened 2019-02-23
Acd provide gps

Scumi:
ACD driver modified to provide GPS data to Air Control Display unit. The ACD makes use of the GPS data for the NEAREST function of the radio.

GPSState was changed to store VDOP and HDOP and the Parser to parse those values from the GPGSA sentence.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -9,7 +9,7 @@ Version 7.0 - not yet released
   - parse wind from standard NMEA sentence WMV
   - driver for XC Tracer Vario
   - driver for KRT2 radio
-  - driver for Air Control Display altimeter
+  - driver for Air Control Display altimeter and radio
   - show detailed error message in device list
   - FLARM/OGN - make it possible to set/download registered device database
   - device manager: show flag if device provides data from
--- a/src/Device/Driver/AirControlDisplay.cpp
+++ b/src/Device/Driver/AirControlDisplay.cpp
@@ -27,6 +27,8 @@ Copyright_License {
 #include "NMEA/Info.hpp"
 #include "NMEA/InputLine.hpp"
 #include "NMEA/Checksum.hpp"
+#include "NMEA/MoreData.hpp"
+#include "Operation/Operation.hpp"
 #include "Atmosphere/Pressure.hpp"
 #include "RadioFrequency.hpp"
 #include "Units/System.hpp"
@@ -79,6 +81,8 @@ public:
   bool PutStandbyFrequency(RadioFrequency frequency,
                            const TCHAR *name,
                            OperationEnvironment &env) override;
+  void OnCalculatedUpdate(const MoreData &basic,
+                          const DerivedInfo &calculated) override;
 };
 
 bool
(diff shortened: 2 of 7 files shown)


Scumi:
> Those attributes need to be considered in `GPSState::Reset()`

done

MaxKellermann:
The commit "NMEA/GPSState.hpp: Add VDOP and PDOP" still doesn't contain proper support code. This belongs in the same commit.

Scumi:
Squashed the support code commits together so they can be cherry-picked.

src/NMEA/Info.cpp
```diff
@@ -38,6 +38,9 @@ GPSState::Reset()
   satellites_used_available.Clear();
   satellite_ids_available.Clear();
   replay = false;
+  pdop = 0;
+  hdop = 0;
```
MaxKellermann: This line is unrelated to this commit.

src/NMEA/Info.cpp
```diff
@@ -38,6 +38,9 @@ GPSState::Reset()
   satellites_used_available.Clear();
   satellite_ids_available.Clear();
   replay = false;
+  pdop = 0;
```
MaxKellermann: What does zero mean here? Is this some kind of "magic" value? Why isn't this documented?
(Don't be distracted from the fact that `hdop` isn't documented either - that's old and bad code. I just don't accept adding more equally bad code.)

src/NMEA/Info.cpp
```diff
@@ -38,6 +38,9 @@ GPSState::Reset()
...
+  hdop = 0;
+  vdop = 0;
```
MaxKellermann: Same here: what does zero mean?

src/NMEA/Info.cpp
hor63: Mmmh. It should be 1.0 as intial value. 0 is a bad initialization for that.
1.0 means that there is no dilution of precision, and is a much better default than 0 (no error at all, which is unrealistic)

HDOP, PDOP, and VDOP are well-defined terms in the NMEA specification. The only use of HDOP is in Logger/LoggerEPE where the meaning is epically described.
Here mentioning that the xDOP values are used for pass-through, and HDOP for error estimation should be sufficient.

Scumi: I agree, a value of 0 doesn't make sense and should not be used for initialization. 1.0 is not a good value either in my opinion, since that is the ideal state of this value (Ref.: https://en.wikipedia.org/wiki/Dilution_of_precision_(navigation)#Meaning_of_DOP_Values).

I would set it to 20 (or even better: 99). If you forward that value to other devices, they should discard it right away.

hor63: Better: -1 is an impossible value, and an indictor that no value is present. When sending it out leave the field empty.

MaxKellermann: @hor63 you mean like 2459e1e3378b7ca0a43e33ca0ce1ac759b97ae2f which I wrote 2 hours ago?

hor63: :D Exactly. @Scumi should build on that.

Scumi: Thanks for updating that code, Max. Updated mine to align with the new support code.

src/Logger/LoggerImpl.hpp
```diff
@@ -71,6 +71,10 @@ class LoggerImpl
     bool satellites_used_available;
     /** GPS Horizontal Dilution of precision */
     double hdop;
+    /** GPS Position (3D) Dilution of precision */
+    double pdop;
```
MaxKellermann: Why copy this in the logger? The logger uses `hdop`, but doesn't use any of the new attributes (nor does any of your other commits add usage of it).

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -123,6 +127,226 @@ ACDDevice::ParseNMEA(const char *_line, NMEAInfo &info)
     return false;
 }
 
+static void
```
MaxKellermann: Much of this is copy'n'pasted from somewhere else. Duplicate code is not acceptable.

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -123,6 +127,226 @@ ACDDevice::ParseNMEA(const char *_line, NMEAInfo &info)
...
+           info.gps.satellite_ids[11],
+           info.gps.pdop,
+           info.gps.hdop,
+           info.gps.vdop);
+
+  return true;
+}
+
+void
+ACDDevice::OnCalculatedUpdate(const MoreData &basic,
```
MaxKellermann: Why implement `OnCalculatedUpdate()` when this doesn't actually uses the calculated data? Better implement `OnSensorUpdate()` instead.

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -123,6 +127,226 @@ ACDDevice::ParseNMEA(const char *_line, NMEAInfo &info)
...
+                        const DerivedInfo &calculated)
+{
+  NullOperationEnvironment env;
+  char buffer[100];
+
+  if (FormatGPRMC(buffer, sizeof(buffer), basic))
```
MaxKellermann: Only send this if a new GPS location has been received; imagine one device has a 20 Hz pressure sensor, but doesn't provide location data; this would trigger sending those sentences 20 times a second.

NEWS.txt
```diff
@@ -9,7 +9,7 @@ Version 7.0 - not yet released
   - parse wind from standard NMEA sentence WMV
   - driver for XC Tracer Vario
   - driver for KRT2 radio
-  - driver for Air Control Display altimeter
+  - driver for Air Control Display altimeter and radio
```
MaxKellermann: This should be part of the commit that adds the feature.

src/Device/Driver/AirControlDisplay.cpp
Scumi: I agree, duplicate code makes everything a mess. Some of the code comes from another driver. As far as I understand this, I cannot use methods from another driver (which also rules out using NMEAOut to forward the data). Since each driver has to be maintained to what the device it is designed for can do, it might also be not practical to merge code and implement the methods. FLARM for example is an exception to that of course. 

Ideas for a way forward?

src/Device/Driver/AirControlDisplay.cpp
Scumi: Hm, `OnSensorUpdate()` made the app freeze every few seconds. Might have to do with sending the NMEA sentences to the device without checking if a new GPS location has been received (see your other comment below). `OnCalculatedUpdate()` works fine, even if I don't need the calculated data.

src/Device/Driver/AirControlDisplay.cpp
Scumi: I agree, makes sense.

Do you think I will still have a freezing app when would have a 20 Hz GPS receiver and use `OnSensorUpdate()`?

NEWS.txt
Scumi: Radio driver would be complete with the addition of this PR. Will squash this commit into the "main" one.

MaxKellermann:
There hasn't been any progress on fixing the problems for 2 months - let's close it until we have fixes.

Scumi:
Yes, I had a bunch of questions that were not answered for 2 months. I can try to improve my code, although I cannot guarantee that I solve all issues you had with the code.

PR #172, opened 2019-03-18
Added navigation through infoboxes with arrow keys.

glidingFisi:
This is necessary for Remote-Stick (e.g. openVario SteFly), because no
touchscreen input is available. Also the mouse over infobox focus is added.
To enter the infobox focus mode, a ScreenModes "infoboxfocus", is added. 
e.g. default.xci key left can set this screenmode, to enter in the infobox focus mode.
With this it is possible to set MC only with arrow key and enter.

### DIFF
--- a/src/InfoBoxes/InfoBoxManager.cpp
+++ b/src/InfoBoxes/InfoBoxManager.cpp
@@ -33,6 +33,7 @@ Copyright_License {
 #include "Profile/Current.hpp"
 #include "Interface.hpp"
 #include "UIState.hpp"
+#include "Event/KeyCode.hpp"
 
 namespace InfoBoxManager
 {
@@ -50,6 +51,7 @@ namespace InfoBoxManager
 
 static bool infoboxes_dirty = false;
 static bool infoboxes_hidden = false;
+static int selected_infobox = 0;
 
 static InfoBoxWindow *infoboxes[InfoBoxSettings::Panel::MAX_CONTENTS];
 
@@ -222,4 +224,37 @@ InfoBoxManager::ShowInfoBoxPicker(const int i)
   DisplayInfoBox();
 
   Profile::Save(Profile::map, panel, panel_index);
+}
+
+bool
+InfoBoxManager::OnKeyDown(InfoBoxWindow *ctx, unsigned key_code)
+{
+  int key = 0;
+
+  switch (key_code) {
+  case KEY_UP:
+    key = -1;
+    break;
+
+  case KEY_DOWN:
+    key = 1;
+    break;
+  }
+
+  for (unsigned i = 0; i < layout.count; i++) {
+    if (infoboxes[i] == ctx) {
+      selected_infobox = i + key;
+      if ((selected_infobox >= (int)layout.count))
+        selected_infobox = 0;
+      if (selected_infobox < 0)
+        selected_infobox = layout.count - 1;
+      infoboxes[selected_infobox]->SetFocus();
+    }
+  }
+  return true;
+}
+
+void
+InfoBoxManager::SetFocus() {
+  infoboxes[selected_infobox]->SetFocus();
 }
--- a/src/InfoBoxes/InfoBoxManager.hpp
+++ b/src/InfoBoxes/InfoBoxManager.hpp
@@ -24,6 +24,8 @@ Copyright_License {
 #ifndef XCSOAR_INFO_BOX_MANAGER_HPP
 #define XCSOAR_INFO_BOX_MANAGER_HPP
 
+#include "InfoBoxes/InfoBoxWindow.hpp"
+
 struct InfoBoxLook;
 class ContainerWindow;
 
@@ -51,6 +53,9 @@ namespace InfoBoxManager
    * then it configures the focused InfoBox if there is one.
    */
   void ShowInfoBoxPicker(const int id = -1);
+
+  bool OnKeyDown(InfoBoxWindow *ctx, unsigned key_code);
+  void SetFocus();
 };
 
 #endif
--- a/src/InfoBoxes/InfoBoxWindow.cpp
+++ b/src/InfoBoxes/InfoBoxWindow.cpp
@@ -22,6 +22,7 @@ Copyright_License {
 */
 
 #include "InfoBoxWindow.hpp"
+#include "InfoBoxManager.hpp"
 #include "InfoBoxSettings.hpp"
 #include "Border.hpp"
 #include "Look/InfoBoxLook.hpp"
@@ -346,6 +347,10 @@ bool
 InfoBoxWindow::OnKeyDown(unsigned key_code)
 {
   /* handle local hot key */
+  if (HasFocus()) {
+    /* Let the InfoBoxManager decide what is to do */
+    InfoBoxManager::OnKeyDown(this, key_code);
+  }
 
   switch (key_code) {
   case KEY_UP:
@@ -442,6 +447,8 @@ InfoBoxWindow::OnMouseDouble(PixelPoint p)
 bool
 InfoBoxWindow::OnMouseMove(PixelPoint p, unsigned keys)
 {
+  SetFocus();
+
   if (dragging) {
     SetPressed(IsInside(p));
     if (!pressed)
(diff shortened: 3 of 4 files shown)


src/InfoBoxes/InfoBoxWindow.cpp
```diff
@@ -346,6 +347,10 @@ bool
 InfoBoxWindow::OnKeyDown(unsigned key_code)
 {
   /* handle local hot key */
+  if (HasFocus()) {
+    /* Let the InfoBoxManager decide what is to do */
+    InfoBoxManager::OnKeyDown(this, key_code);
```
MaxKellermann: InfoBoxWindows should be agnostic to where they are integrated. This is a layering violation.

src/InfoBoxes/InfoBoxManager.hpp
```diff
@@ -51,6 +53,9 @@ namespace InfoBoxManager
    * then it configures the focused InfoBox if there is one.
    */
   void ShowInfoBoxPicker(const int id = -1);
+
+  bool OnKeyDown(InfoBoxWindow *ctx, unsigned key_code);
```
MaxKellermann: Never use a pointer when `nullptr` is an illegal value.

src/InfoBoxes/InfoBoxManager.hpp
```diff
@@ -24,6 +24,8 @@ Copyright_License {
 #ifndef XCSOAR_INFO_BOX_MANAGER_HPP
 #define XCSOAR_INFO_BOX_MANAGER_HPP
 
+#include "InfoBoxes/InfoBoxWindow.hpp"
```
MaxKellermann: Unnecessary include. Use forward declarations.

src/InfoBoxes/InfoBoxManager.cpp
```diff
@@ -223,3 +225,36 @@ InfoBoxManager::ShowInfoBoxPicker(const int i)
...
+  switch (key_code) {
+  case KEY_UP:
+    key = -1;
+    break;
+
+  case KEY_DOWN:
+    key = 1;
+    break;
+  }
+
```
MaxKellermann: Why does this function continue here if if `key==0`?

src/InfoBoxes/InfoBoxManager.cpp
```diff
@@ -50,6 +51,7 @@ namespace InfoBoxManager
 
 static bool infoboxes_dirty = false;
 static bool infoboxes_hidden = false;
+static int selected_infobox = 0;
```
MaxKellermann: Why is this `signed`?

src/InfoBoxes/InfoBoxWindow.cpp
```diff
@@ -442,6 +447,8 @@ InfoBoxWindow::OnMouseDouble(PixelPoint p)
 bool
 InfoBoxWindow::OnMouseMove(PixelPoint p, unsigned keys)
 {
+  SetFocus();
```
MaxKellermann: Uh-oh, forcing keyboard focus on mouse move? That sounds like a very bad idea, and will confuse lots of users.

src/InfoBoxes/InfoBoxWindow.cpp
glidingFisi: Sorry, I didn't understand the agnostic word in this context. Can you please explain the violation? Thx.

src/InfoBoxes/InfoBoxWindow.cpp
glidingFisi: The idea here is, with stefly you have a very small courser and to move this during flight is a little bit tricky. The hover would show the user directly that the courser is over the infobox. But I understand that this is a little bit confusing. 
Would it be better to have it configurable or ifdef openvario?

MaxKellermann: There is no compile-time switch for OpenVario.
Does the OpenVario even have a mouse? It has a touch screen, doesn't it?
`OnMouseMove()` triggers on touch screens only while the finger is down. Why would somebody want the keyboard focus to follow the finger? Tapping the finger, ok, but while moving the finger?
None of this makes any sense to me, and your text doesn't explain any of it.

src/InfoBoxes/InfoBoxWindow.cpp
MaxKellermann: `InfoBoxManager` manages `InfoBoxWindow`, but `InfoBoxWindow` doesn't know about `InfoBoxManager`. And it shouldn't.

hor63:
Usually no touch screen but arrow-keys and/or track-points emulating a mouse.
Operating a touch screen in the panel with streched-out arm in rough weather with some accuracy is close to impossible.

src/InfoBoxes/InfoBoxWindow.cpp
kerel-fs: The OpenVario doesn't have a touchscreen, but instead is often used with a remote control stick emulating a mouse (four arrow keys move the cursor + mouse click).
I think that this feature is useful only on non-touchscreen devices with such a cursor, so it should be guarded by a compile-time switch or configuration option?

MaxKellermann: Why would that be useful? If you want keyboard focus, click.

MaxKellermann: Oh, and emulating a pointing device with digital arrow keys sounds like the most horrible UI desaster I could imagine in a cockpit. This just doesn't make sense; it takes away too much of precious attention from the pilot, who should be watching outside instead of watching the mouse cursor reach a certain position on the screen.

kerel-fs: Some kind of highlighting if the mouse hovers over a clickable object is common practice and helps signalizing that clicking is a possible action thus I would consider it useful. If this should be done by SetFocus I can't answer (as I don't know the implications yet).
Or maybe I didn't understand the patch.

MaxKellermann: Agree with highlighting, but that's not what this code submission is about. It is about changing keyboard focus, which I consider a bad idea.

kerel-fs: I agree, thanks for the explanation (you already explained it in your first message but I didn't understand).

src/InfoBoxes/InfoBoxWindow.cpp
glidingFisi: Thank you, now I understand the violation.
But how to solve this, InfoBoxManager doesn't know about the keys. InfoBoxWindow doesn't know about the other InfoBoxWindow's.
In my opinion the InfoBoxWindow must signal the InfoBoxManager that he should select the next InfoBoxWindow.

src/InfoBoxes/InfoBoxWindow.cpp
glidingFisi: This was mentioned in the posts, it is possible with the steFly remote to move a mouse courser with the cross, which emulates the courser keys. My idea was in turbolent condition to give visual feedback, when the info box is reached with the mouse courser.
But I understand that this is not in the usability of XCSoar, and I will remove it in my next code clean up.
In addition with the "infoboxfocus" it is possible to set a key that focus to the last selected infobox is set.

src/InfoBoxes/InfoBoxWindow.cpp
MaxKellermann: There used to be code for this, but I removed it 4 years ago: https://github.com/XCSoar/XCSoar/commit/e8c8cde7b1f8566b1e361a4be9b07bba631af40c
Note that the arrow keys are already taken for other actions on InfoBoxes.

glidingFisi:
Closed because new PR with rework created: InfoBoxes navigation with courser key. #191

fb:
Instead of focusing infoboxes using a cursor keys or pointing device we could add actions like SelectInfobox(n), UnselectInfobox().

Using lua scripting, this action could be bound to something like TAB or Shift-Tab to cycle through infoboxes. A timer could be used to leave the *infobox selected* mode.

This approach would work without changing behaviour for other users.

PR #182, opened 2019-04-10
Initial alpha of Skysight weather integration

jfwells:
Hi, this is still work in progress and not yet ready to pull, but would appreciate some feedback and advice at this stage. Let me know if there's a better place for this.

In particular, but not limited to:
- I have had to make some kludges in order to get the additional libraries to build for Android (curl with SSL support, and netcdf4 and C++ bindings). For example:
   - build/python/build/openssl.py: Latest versions of the openssl configure script require various environment variables to be set, including "ANDROID_HOME" and various things in PATH. I've taken ndk location from toolchain.env['ANDROID_NDK'], but I'm not clear whether this will always be available. 
   - build/python/build/netcdfcxx.py is likewise a kludge -- I can only get this to build with these include files. I'm sure there's a better way to do his.

- I use the main "render" loop to trigger updates in Skysight (should the forecast data be updated, should the overlay be advanced). I've copied the approach used by the RASP implementation here. I'm not clear if this is the right approach, or if a timer would be better here. One particular problem is that on the ground, the initial overlay is not displayed until a render is triggered, which can be some time if GPS is not yet connected. On the flip side, when in flight, or if the map is being manipulated, the render loop gets called far more frequently than is necessary for my needs. Data downloads in background threads, but as these cannot access the map safely I need an asynchronous trigger from the main thread somewhere.

- The overlays in Android and Unix are pretty much identical, but they display quite differently. The images are fairly low-res, but when zoomed seem to be interpolated nicely on Unix. On Android they just show as larger, blocky pixels. Screenshot comparison on the forum. Any points as to why this might be would be much appreciated.

There are other stylistic issues to fix (not happy with the Metric/ActiveMetric/DisplayedMetric model, and I have an over-reliance on unneeded std::strings for a start). I also need to add documentaion, but I will fix those in the next round.

Note that this requires netcdf4 and netcdfcxx4 (development) libraries to compile. On Arch these are netcdf and netcdf-cxx.

Instructions for using the implementation are here: https://forum.xcsoar.org/viewtopic.php?f=7&t=3227&p=19689#p19689

Thanks in advance for any help/advice/review.


Plantain:
XCSoar developers are welcome to contact me (matthew .(at). skysight.io) if you need a SkySight account to work on or with this integration.

MaxKellermann:
Before we even begin: please note that linking XCSoar with OpenSSL is illegal, because the OpenSSL license is incompatible with the GPL. This means that you can be sued for copyright violation for distributing your build: http://jfwhome.com/XCSoar-debug.apk - please take it down quickly!

Now if you really enable CURL's https features, please make that a separate commit. (And of course, choose a TLS library which is compatible with the GPL.)

The OpenSSL/GPL situation is very unfortunate, but that's what it is, and we need to obey copyright law and licenses.

Keep your branch rebased on the official master branch. Do not *merge*! No "Merge remote-tracking branch 'upstream/master'" commits!

src/Weather/Skysight/Request.hpp
```diff
@@ -0,0 +1,118 @@
...
+  SkysightRequest(const SkysightRequestArgs _args) : args(_args) {};
+  bool Process();  
+  bool ProcessToString(tstring &response);
+  
+  SkysightCallType GetType();
+  void SetCredentials(const TCHAR *_key, const TCHAR *_username = nullptr,
+                      const TCHAR *_password = nullptr);
+  
+  class FileHandler final : public Net::ResponseHandler {
+    FILE *file;
```
MaxKellermann: Please use `class FileOutputStream` instead.

src/Weather/Skysight/Request.cpp
```diff
@@ -0,0 +1,249 @@
...
+void SkysightAsyncRequest::Process() {
+  ScopeLock protect(mutex);
+  if (IsBusy()) return;
+  Trigger();
+}
+
+void SkysightAsyncRequest::Tick() {
+  status = Status::Busy;
+
+  mutex.Unlock();
```
MaxKellermann: Unmanaged mutex operation is not exception-safe!

src/Weather/Skysight/Request.cpp
```diff
@@ -0,0 +1,249 @@
...
+  FileHandler handler(file);
+  Net::Request request(session, handler, args.url.c_str());
+  request.AddHeader("X-API-Key", key.c_str());
+
+  tstring pBody;
+  if(username.length() && password.length()) {
+    request.AddHeader("Content-Type", "application/json");
+    NarrowString<1024> creds;
+    creds.Format("{\"username\":\"%s\",\"password\":\"%s\"}", username.c_str(), password.c_str());
+    pBody = creds.c_str();
```
MaxKellermann: `char`/`wchar_t` conflict

src/Weather/Skysight/Request.cpp
```diff
@@ -0,0 +1,249 @@
...
+    request.SetRequestBody(pBody.c_str(), pBody.length());
+    request.SetNoFailOnError();
+  }
+  
+  request.SetVerifyPeer(false);
+  
+  try {
+    request.Send(10000);
+  } catch (const std::exception &exc) {
+    success = false;
```
MaxKellermann: Preserve error information.

src/Weather/Skysight/Request.cpp
```diff
@@ -0,0 +1,249 @@
...
+  if (!success) File::Delete(temp_path);
+
+  file = NULL;
+
+  if(success) {
+    if (!File::Delete(final_path) && File::ExistsAny(final_path)) {
+      File::Delete(temp_path);
+      return false;
+    }
+    std::rename(temp_path.c_str(), args.path.c_str());
```
MaxKellermann: With `FileOutputStream`, you don't need a `temp_path`!

src/Weather/Skysight/Request.hpp
```diff
@@ -0,0 +1,118 @@
...
+#include "Net/HTTP/Session.hpp"
+#include "Net/HTTP/Request.hpp"
+#include "Net/HTTP/Handler.hpp"
+
+#include "IO/FileLineReader.hpp"
+
+#include "LogFile.hpp"
+
+
+class SkysightRequestError {};
```
MaxKellermann: Needs documentation. Since this is not derived from `std::exception`, the usual `catch` blocks won't catch it.

src/Weather/Skysight/Request.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+struct SkysightRequest {
+  
+public:
+  enum class Status {Idle, Busy, Complete, Error};
```
MaxKellermann: Why is this `public`?

src/Weather/Skysight/Request.hpp
```diff
@@ -0,0 +1,118 @@
...
+    size_t received = 0;
+
+  public:
+    FileHandler(FILE *_file) :file(_file) {};
+    void DataReceived(const void *data, size_t length) override;
+    void ResponseReceived(int64_t content_length) override;
+  };
+
+  class BufferHandler final : public Net::ResponseHandler {
+    uint8_t *buffer;
```
MaxKellermann: Who owns this buffer?

src/Weather/Skysight/Settings.hpp
```diff
@@ -0,0 +1,47 @@
...
+}
+*/
+
+#ifndef XCSOAR_WEATHER_SKYSIGHT_SETTINGS_HPP
+#define XCSOAR_WEATHER_SKYSIGHT_SETTINGS_HPP
+
+#include "Util/StaticString.hxx"
+
+struct SkysightSettings {
+	StaticString<64> email;
```
MaxKellermann: Wrong indent

src/Weather/Skysight/Skysight.cpp
```diff
@@ -0,0 +1,623 @@
...
+  if(p == tstring::npos)
+    return;
+    
+  tstring reg = file_base.substr(0, p);  
+  tstring rem = file_base.substr(p+1);  
+  
+  p = rem.find(_("-"));
+  if(p == tstring::npos)
+    return;    
+  tstring met = rem.substr(0, p);
```
MaxKellermann: Please don't create so many `std::string` temporaries. This is heavy bloat.

src/Weather/Skysight/Skysight.cpp
```diff
@@ -0,0 +1,623 @@
...
+
+  GetActiveMetricState(id, m);
+    
+  active_metrics.push_back(m);
+  SaveActiveMetrics();
+  return active_metrics.size() - 1;
+}
+
+void Skysight::RefreshActiveMetric(tstring id) {
+  std::vector<SkysightActiveMetric>::iterator i;
```
MaxKellermann: Use `auto`

src/Weather/Skysight/Skysight.cpp
```diff
@@ -0,0 +1,623 @@
...
+  for(i = active_metrics.begin(); i<active_metrics.end();++i) {
```
MaxKellermann: Don't use `<` with iterator objects.

src/Weather/Skysight/Skysight.cpp
```diff
@@ -0,0 +1,623 @@
...
+  
+  if(self->api.metrics.size()) {
+    self->LoadActiveMetrics();
+    self->Render(true);
+  }
+}
+
+bool Skysight::GetActiveMetricState(tstring metric_name, SkysightActiveMetric &m) {
+  
+  tstring search_pattern = region + "-" + metric_name + "*";
```
MaxKellermann: Again so many `std::string` temporaries.

src/Weather/Skysight/Skysight.cpp
```diff
@@ -0,0 +1,623 @@
...
+  std::vector<SkysightImageFile> img_files = ScanFolder(search_pattern);
+  
+  if(img_files.size() > 0) {
+    uint64_t min_date = (uint64_t)std::numeric_limits<uint64_t>::max;
```
MaxKellermann: This cast is unnecessary.

src/Weather/Skysight/Skysight.cpp
```diff
@@ -0,0 +1,623 @@
...
+    self->Render(true);
+  }
+}
+
+bool Skysight::GetActiveMetricState(tstring metric_name, SkysightActiveMetric &m) {
+  
+  tstring search_pattern = region + "-" + metric_name + "*";
+  std::vector<SkysightImageFile> img_files = ScanFolder(search_pattern);
+  
+  if(img_files.size() > 0) {
```
MaxKellermann: Use `empty()`

src/Weather/Skysight/Skysight.hpp
```diff
@@ -0,0 +1,153 @@
...
+  Path fullpath;
+  Path filename;
+  tstring metric;
+  tstring region;
+  uint64_t datetime;
+  bool is_valid;
+  uint64_t mtime;
+};
+
+class Skysight final : private NullBlackboardListener { //: public Timer {
```
MaxKellermann: What *is* this class? It is a Skysight. But what is a Skysight?

src/Weather/Skysight/Skysight.hpp
```diff
@@ -0,0 +1,153 @@
...
+    static void DownloadComplete(const tstring &&details,  const bool success,  
+                const tstring &&layer_id,  const uint64_t time_index);
+    
+    static void APIInited(const tstring &&details,  const bool success,  
+                const tstring &&layer_id,  const uint64_t time_index);
+
+    std::map<tstring, tstring> GetRegions() {
+      return api.regions;
+    }
+    tstring GetRegion() {
```
MaxKellermann: This allocates a `std::string` copy for no reason. Please keep the bloat in your code smaller!

src/Weather/Skysight/Skysight.hpp
```diff
@@ -0,0 +1,153 @@
...
+    SkysightMetric GetMetric(int index) {
+      return api.GetMetric(index);
+    }
+    SkysightMetric GetMetric(const tstring id) {
+      return api.GetMetric(id);
+    }
+    bool MetricExists(const tstring id) {
+      return api.MetricExists(id);
+    }
+    int NumMetrics() {
```
MaxKellermann: Why is this a signed integer?

src/Weather/Skysight/SkysightAPI.cpp
```diff
@@ -0,0 +1,653 @@
...
+#include "Operation/Operation.hpp"
+
+
+SkysightAPI *SkysightAPI::self;
+
+SkysightAPI::~SkysightAPI() {
+  Timer::Cancel();
+}
+
+SkysightMetric SkysightAPI::GetMetric(int index) {
```
MaxKellermann: Why is this a signed integer?

src/Weather/Skysight/SkysightAPI.cpp
```diff
@@ -0,0 +1,653 @@
...
+  for(i = metrics.begin(); i<metrics.end();++i)
+    if(!i->id.compare(id)) {
+      return true;
+    }
+  return false;
+}
+
+
+int SkysightAPI::NumMetrics() {
+  return (int)metrics.size();
```
MaxKellermann: Why cast to signed?

src/Weather/Skysight/SkysightAPI.hpp
```diff
@@ -0,0 +1,143 @@
...
+#include <boost/property_tree/json_parser.hpp>
+
+//maintain two-hour local data cache
+#define SKYSIGHTAPI_LOCAL_CACHE 7200 
+
+#define SKYSIGHTAPI_BASE_URL "https://skysight.io/api"
+
+struct BrokenDateTime;
+
+class SkysightAPI final : public Timer {
```
MaxKellermann: What *is* this class? What is a SkysightAPI?

jfwells:
> Before we even begin: please note that linking XCSoar with OpenSSL is illegal, because the OpenSSL license is incompatible with the GPL. This means that you can be sued for copyright violation for distributing your build: http://jfwhome.com/XCSoar-debug.apk - please take it down quickly!
> 
> Now if you really enable CURL's https features, please make that a separate commit. (And of course, choose a TLS library which is compatible with the GPL.)
> 
> The OpenSSL/GPL situation is very unfortunate, but that's what it is, and we need to obey copyright law and licenses.
> 
> Keep your branch rebased on the official master branch. Do not _merge_! No "Merge remote-tracking branch 'upstream/master'" commits!

Thanks. I'll hive off the Android changes into a separate commit for now, assuming that most people can build/test on machines that have curl libraries with SSL support.

Shame about OpenSSL (it looks like https support was removed from the codebase and I had wondered why). I must admit I had tried LibreSSL as getitng openSSL to compile was a bit of a pain. No luck. It looks like there are a few options: OpenSSL 3.0 (unstable), LibreSSL, BoringSSL and GnuTLS. I'll probably try them in reverse order, unless anyone has any experience or preferences that suggest otherwise.

Thanks very much for the additional comments and pointers.

MaxKellermann:
Since both LibreSSL and BoringSSL are OpenSSL forks (IIRC), I think they share the licensing problem.

hor63:
Lets hope for the next version of OpenSSL which is hopefully released under the Apache 2.0 license in May.
See here: https://www.openssl.org/blog/blog/2018/03/01/last-license/

jfwells:
> The next major version of OpenSSL 3.0 is supposed to be released under the Apache 2.0 license which is compatible with the GPL. However I did not find any hint when this version is going to be released. I would not hold my breath while waiting on it.

Thanks. It does seem a bit early to use v3. I'm trying to get GnuTLS to compile. It depends on Nettle which in turn needs GMP.

I can get GnuTLS and Nettle (without GMP) to compile using Automake, but I can't (yet) get GMP to pass the "configure" step as it fails on various compiler tests. Looks like it might be messy.

hor63:
> but I can't (yet) get GMP to pass the "configure" step as it fails on various compiler tests. Looks like it might be messy.

What are the error messages? GMP is a prerequisite for compiling GCC. I don't care to remember how often I configured and compiled GMP over that course. You may send me a message directly.

jfwells:
I've got a working compile with WolfSSL replacing openSSL. Sorry for the force-push mess. 

The code is not ready for review yet -- I have more changes to make, and to clean up the commits. Hopefully within the next couple of weeks.

lordfolken:
So I got this to compile on Debian stable. 

The following changes where necessary for the UNIX target:
in:

diff --git a/src/Weather/Skysight/CDFDecoder.cpp b/src/Weather/Skysight/CDFDecoder.cpp
index b30e101063..22dd125cf1 100644
--- a/src/Weather/Skysight/CDFDecoder.cpp
+++ b/src/Weather/Skysight/CDFDecoder.cpp
@@ -29,8 +29,8 @@ Copyright_License {
 #include <netcdfcpp.h>
 #endif
 
-#include <geotiffio.h>
-#include <xtiffio.h>
+#include <geotiff/geotiffio.h>
+#include <geotiff/xtiffio.h>
 #include "SkysightAPI.hpp"
 #include "OS/FileUtil.hpp"
 #include "LogFile.hpp"

The files where not found in the search path as they are indeed in the subdirectory on debian. I'm not sure how one does this correctly. 

diff --git a/build/netcdf.mk b/build/netcdf.mk
index 5439bb28cc..c0f50dd4d4 100644
--- a/build/netcdf.mk
+++ b/build/netcdf.mk
@@ -3,7 +3,7 @@ NETCDF = y
 ifneq ($(TARGET),ANDROID)
 $(eval $(call pkg-config-library,NETCDF,netcdf-cxx4))
 $(eval $(call link-library,netcdfcpp,NETCDF))
-NETCDF_LDLIBS = -lnetcdf_c++4 -lnetcdf -lhdf5_hl -lhdf5 -lsz -ldl
+NETCDF_LDLIBS = -v -lnetcdf_c++4 -lnetcdf -lhdf5_hl_cpp -lhdf5_cpp -lsz -ldl
 else
 NETCDF_LDLIBS += -l:libnetcdf_c++.a -l:libnetcdf.a
 endif

For some reason the hdf5_hl library ends with _cpp in debian.

jfwells:
> So I got this to compile on Debian stable.


I'm not sure about the correct approach for the first issue, but for the second, I know that my NETCDF_LDLIBS lines that manually specifies the packages are temporary kludges. The pkg_config line above it should be doing the trick (I'm not sure what else needs to be done to make it so!)

lordfolken:
So I'm still working on compiling the android version. Since the netcdf library is not in the toolchain, I think it needs to be added to the thirdparty.mk in the build directory. But I haven't yet managed to compile it.

lordfolken:
Any progress on the implementation? :)

MaxKellermann:
No progress, closing for now

PR #183, opened 2019-04-12
Update METAR & TAF url to https

Developer66:
tgftp.nws.noaa.gov switched from http to https... METARs and TAFs must be downloaded from https.

METARs and TAFs are not loaded in current version 6.8x and 7.0.x


MaxKellermann:
fb75a9eec04d8ed4c86afe3ca3fe4cb07e450dc2

lordfolken:
This will break on the kobo.

MaxKellermann:
If I understand this correctly, the Kobo was already broken because HTTPS was made mandatory. So this commit fixes those ports which are capable of HTTPS, and leaves the others just as broken as before.

PR #186, opened 2019-04-17
Update SkyLines server IP

Turbo87:
We have migrated SkyLines to a new server today, and apparently the v6.8.x branch still has the hardcoded IP in it. This commit changes the code to use the new IP. The old IP still works for now as we have set up UDP forwarding, but it will be turned off in the near future, so it would be good to get this released soon.


Turbo87:
@MaxKellermann since you're the only one that can release XCSoar: do you have time to put out another v6.8 release in the next days?

lordfolken:
I would second this @MaxKellermann

PR #187, opened 2019-04-19
Update NOAADownloader.cpp

Developer66:
Update METAR & TAF url to https like #183... Changes were not accepted?


MaxKellermann:
https://github.com/XCSoar/XCSoar/pull/183 was merged.

Issue #155, opened 2018-12-03
Building XCSoar with CMake


Tymolc:
Hi!

I recently decided that I might want to look further into contributing to XCSoar.
I feel like the build system is kind of difficult to grasp, especially for non-Linux devs, so I was thinking about providing a CMake alternative for building with conan as a dependency manager.
My desired outcome would be that it is possible to build XCSoar natively on all desktop operating systems (disclaimer: I unfortunately have very little experience with Android and iOS development).

Did anyone look into this before and would this be a desired change?

Feedback is very welcome.

MaxKellermann:
9 years ago, [I suggested moving to autotools](http://xcsoar.1045713.n5.nabble.com/A-few-general-questions-on-development-td2650276.html), but it was vetoed.

As much as I've tried to optimize our Makefile over the last 9 years, it's still horrible. So yes, I'd look forward to switch to something else. Meanwhile, I've moved past autotools, now this had to wait until [Meson](https://mesonbuild.com/) emerged (I've [switched MPD to Meson](https://github.com/MusicPlayerDaemon/MPD/commit/94592c14062d5afc9482d11baa401648082022c0) which is as complex was XCSoar). I've always hated CMake, and never considered it superior to autotools, which is why I was stuck with autotools for so long.

In any case:
- a submission which omits any currently supported target cannot be merged, so any volunteer needs to figure out all targets (iOS is not a supported target, but Android certainly is)
- there will not be two build systems; the submission must delete the old build scripts

As much as I hate CMake, and as much as I'd prefer you to use Meson instead, the guy who does the job (and is willing to keep supporting it) may decide. I will not veto a well-done and complete CMake port.

Tymolc:
Thanks for your quick reply.
I agree that CMake can be very painful. I guess it might be Stockholm syndrome, but I've gotten kind of used to it. I am happy to look into Meson though. New projects are always a way to learn new things, after all!

Exadios:
I have a branch on Exadios/XCSoar and Exadios/xcsoar-exp called 'master-cmake-build' which may be of interest to you. Presently it only builds the "UNIX" target for the master branch. I use it for debugging since it generally does a faster and better job of building code changes than the supplied Makefile (I use ninja as the back end build agent). This branch is currently up to date with the current master branch.

sbrodehl:
What is the current state here?

Tymolc:
[...]

PR #188, opened 2019-04-26
Android: Add Android legacy to support Honeycomb (Android 3.1, API 12) onward

hor63:
A commit series to support older Android versions from Honeycomb onwards.

Per Max' suggestion legacy mode implements an own native library libxcsoar_legacy.so.
The legacy library excludes all functions which are not available in older Android version before Lollipop.

Legacy mode can be built for X86 and Arm V7. 64-bit architectures are not supported because they were not existing for old Android versions.

Build system supports pure full mode builds, pure legacy builds, and combined builds (legacy and full mode libraries in one APK)
Target ANDROIDFAT with DEBUG=n includes all targets with full and legacy modes.

All builds are executed with the same, and most recent SDK and NDK.

As by-product all MIPS targets are removed because NDK support ceased.

Tested devices:
- Emulator X86:
  - System image API 16 in legacy mode
  - System image API 24 in full mode
  External data sources for both: NMEA data stream from UBlox GPS mouse via TCP client port from remote Linux computer works.
  Map display, config dialogs, file download functionality are working correctly.

- HTC Flyer tablet with Honeycomb (Android 3.1, API 12) in legacy mode.
  Data sources: NMEA data stream from UBlox GPS mouse via Bluetooth port works.
  Builtin sensors with GPS works
  Map display, config dialogs, file download functionality are working correctly (except https due to outdated certificates).

- Cubieboard 2 (Allwinner A20) with on-chip Android 4.2 in legacy mode
  Data sources: NMEA data stream from UBlox GPS mouse via TCP client port works.
  Map display, config dialogs, file download functionality are working correctly (except https due to outdated certificates).

- My beloved Samsung S7 with Oreo (Android 8.0) in full mode.
  Data sources: NMEA data stream from UBlox GPS mouse via Bluetooth port works. Flarm via K6 Team BT2 with Bluetooth works.
  Builtin sensors with GPS and Baro works.
  Map display, config dialogs, file download functionality are working correctly incl. HTTPS.


build/android.mk
```diff
@@ -27,6 +27,11 @@ DX = $(ANDROID_BUILD_TOOLS_DIR)/dx
 
 ANDROID_LIB_NAMES = xcsoar
 
+ifeq ($(TARGET)$(ANDROID_LEGACY),ANDROIDy)
```
MaxKellermann: Why check `$(TARGET)` here? Setting `ANDROID_LEGACY=y` on a target other than Android is illegal, isn't it?

build/android.mk
```diff
@@ -71,6 +76,8 @@ ifeq ($(TESTING),y)
 	JAVA_SOURCES += $(wildcard android/src/testing/*.java)
 endif
 
+
+
```
MaxKellermann: Why add two blank lines here?

build/android.mk
```diff
@@ -209,29 +216,42 @@ ifeq ($(FAT_BINARY),y)
 ANDROID_LIB_BUILD =
 ANDROID_THIRDPARTY_STAMPS =
 
-# Example: $(eval $(call generate-abi,xcsoar,armeabi-v7a,ANDROID7))
+# Example: $(eval $(call generate-abi,xcsoar,armeabi-v7a,ANDROID7),n) ("n" sets ANDROID_LEGACY)
```
MaxKellermann: Line too long

hor63:
Supporting Gingerbread is not possible without jumping through some weird hoops.
Mostly because configure for 3rd party libs cannot detect missing functions with the headers, and libraries not supporting older versions than API 16. This was not an issue down to API 12.
Sorry to say good-bye to Dell Streak.

build/android.mk
```diff
@@ -209,29 +216,42 @@ ifeq ($(FAT_BINARY),y)
...
 define generate-abi
 
 ANDROID_LIB_BUILD += $$(ANDROID_BUILD)/lib/$(2)/lib$(1).so
 
-$$(ANDROID_BUILD)/lib/$(2)/lib$(1).so: $$(OUT)/$(3)/$$(XCSOAR_ABI)/bin/lib$(1).so | $$(ANDROID_BUILD)/lib/$(2)/dirstamp
+OUT_BUILD_DIR := $(3)
+ifeq ($(4),y)
+OUT_BUILD_DIR := $$(OUT_BUILD_DIR)_LEGACY
```
MaxKellermann: This creates a whole new output directory - this is rather surprising, because it can be expected that output is found in `output/$(TARGET)`. Below this, there are directories `dbg` and `opt` for different build options which cause an ABI change. Let's integrate it at this level.

build/android.mk
```diff
@@ -248,7 +268,9 @@ $(call SRC_TO_OBJ,$(SRC)/Android/InternalSensors.cpp): $(NATIVE_HEADERS)
 $(call SRC_TO_OBJ,$(SRC)/Android/Battery.cpp): $(NATIVE_HEADERS)
 $(call SRC_TO_OBJ,$(SRC)/Android/GliderLink.cpp): $(NATIVE_HEADERS)
 $(call SRC_TO_OBJ,$(SRC)/Android/NativePortListener.cpp): $(NATIVE_HEADERS)
+ifneq ($(ANDROID_LEGACY),y)
```
MaxKellermann: Is this necessary?

build/android.mk
```diff
@@ -260,13 +282,30 @@ ANDROID_LIB_BUILD = $(patsubst %,$(ANDROID_ABI_DIR)/lib%.so,$(ANDROID_LIB_NAMES)
...
+  ANDROID_LEGACY_LIB_DIR = $(TARGET_OUTPUT_DIR)_LEGACY/$(XCSOAR_ABI)/bin
+  ANDROID_LEGACY_LIB_BUILD = $(patsubst %,$(ANDROID_ABI_DIR)/lib%.so,$(ANDROID_LIB_NAMES)_legacy)
+
+  $(ANDROID_LEGACY_LIB_BUILD): $(ANDROID_LEGACY_LIB_DIR)/lib$(ANDROID_LIB_NAMES)_legacy.so
+	$(Q)cp $< $@
+
+  $(ANDROID_LEGACY_LIB_DIR)/lib$(ANDROID_LIB_NAMES)_legacy.so : ANDROID_LEGACY_FORCE
+	$(Q)$(MAKE) TARGET=$(ANDROID_ORG_TARGET) ANDROID_LEGACY=y ANDROID_INCLUDE_LEGACY_LIB=n DEBUG=$(DEBUG) USE_CCACHE=$(USE_CCACHE) V=$(V) $@
+
+  # Dummy target to force the sub-make run each time.
```
MaxKellermann: Horrible kludge. Please not.

build/main.mk
```diff
@@ -1,7 +1,7 @@
 ifeq ($(HAVE_POSIX),y)
-PROGRAM_NAME = xcsoar
+PROGRAM_NAME = xcsoar$(PROGRAM_NAME_SUFFIX)
```
MaxKellermann: What's that?

build/targets.mk
```diff
@@ -240,7 +275,7 @@ ifeq ($(TARGET),OSX64)
   TARGET_IS_OSX = y
   OSX_MIN_SUPPORTED_VERSION = 10.7
   HOST_TRIPLET = x86_64-apple-darwin
-  LLVM_TARGET = $(HOST_TRIPLET)
+  LLVM_TARGET := $(HOST_TRIPLET)
```
MaxKellermann: Huh?

build/targets.mk
```diff
@@ -410,15 +472,15 @@ ifeq ($(TARGET),ANDROID)
   LLVM_PREFIX = $(ANDROID_TOOLCHAIN)/bin/
 
   ifeq ($(X86),y)
-    LLVM_TARGET = i686-none-linux-android
+    LLVM_TARGET := i686-linux-android
```
MaxKellermann: Huh?

build/targets.mk
```diff
@@ -448,6 +512,7 @@ ifeq ($(TARGET),ANDROID)
   HAVE_WIN32 := n
   HAVE_MSVCRT := n
   HAVE_VASPRINTF := y
+
```
MaxKellermann: Why blank line here?

build/targets.mk
```diff
@@ -529,7 +594,7 @@ ifeq ($(TARGET_IS_KOBO),y)
 
   ifeq ($(USE_CROSSTOOL_NG),y)
     HOST_TRIPLET = $(ACTUAL_HOST_TRIPLET)
-    LLVM_TARGET = $(ACTUAL_HOST_TRIPLET)
+    LLVM_TARGET := $(ACTUAL_HOST_TRIPLET)
```
MaxKellermann: Huh?

build/targets.mk
```diff
@@ -551,7 +616,7 @@ ifeq ($(TARGET),ANDROID)
   TARGET_CPPFLAGS += --sysroot=$(ANDROID_SYSROOT)
   TARGET_CPPFLAGS += -isystem $(ANDROID_SYSROOT)/usr/include/$(HOST_TRIPLET)
   TARGET_CPPFLAGS += -DANDROID
-  TARGET_CPPFLAGS += -D__ANDROID_API__=21
+  TARGET_CPPFLAGS += -D__ANDROID_API__=$(NDK_API)
```
MaxKellermann: Making this a variable should be a separate commit

src/Android/posix_memalign.cpp
```diff
@@ -0,0 +1,49 @@
...
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+
+#include <stdlib.h>
+
+
+#if defined __ANDROID_API__ && __ANDROID_API__ < 17
+extern "C" {
```
MaxKellermann: Is this necessary?

MaxKellermann:
I cherry-picked d8b69db65382d3965eade3f34db1310340478aec

Makefile
```diff
@@ -32,6 +32,7 @@
 #   ICF         "y" enables Identical Code Folding (gold --icf=all)
 #
 #   DEBUG       If set to "y", the debugging version of XCSoar is built
+#               To disable debugging it must be set to "n"!
```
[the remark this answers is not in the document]
hor63: Removed.

Makefile
```diff
@@ -53,6 +54,16 @@
 #
 #   IWYU        "y" to run "include-what-you-use" on all sources
 #
+#   ANDROID_LEGACY "y" to build an Android APK which is compatible down to Android 3.1 (Honeycomb). Bluetooth LE is not supported.
+#                Some system calls are replaced with earlier, simpler ones, or direct Linux Kernel calls when not supported by older Bionic C libraries.
+#                This option is not supported for X64 and AARCH64.
+#                Default is "n"
+#
+#   ANDROID_INCLUDE_LEGACY_LIB "y" to build an Android APK which includes the current library with all functionality for Android 5.0 and later,
+#                but also the legacy library which is automatically used for older Andoid versions. Incompatible with ANDROID_LEGACY
```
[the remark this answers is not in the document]
hor63: (Hopefully) clarified.

build/android.mk
hor63: Removed.

build/android.mk
hor63: Removed.

build/android.mk
hor63: Split lines.

build/targets.mk
hor63: Reverted.

build/targets.mk
hor63: See distinct commit describing adjustments to the LLVM_TARGET for Android builds.

build/targets.mk
hor63: Deleted.

build/targets.mk
hor63: Reverted.

build/targets.mk
hor63: Moved to separate commit within this series.

src/Android/posix_memalign.cpp
hor63: No. It is redundant. Compiling and linking is already governed by the build system.

build/android.mk
hor63: Changed to one output directory per architecture.
However it is necessary not only to distinguish the ABI directory but also the third-party directory because the configure runs must run once for API 21, the other time with android-16 platform and API 12.
That does not make the build system easier nor prettier.

build/android.mk
hor63: No. Removed.

build/android.mk
hor63: Worked around it. Now the legacy lib depends on the standard lib, and forces a legacy make when the standard lib is newer, i.e. re-compiled.

build/main.mk
hor63: PROGRAM_NAME_SUFFIX is necessary for the linker command. For Android Legacy it is set to "_legacy".

hor63:
I reverted the -Ofast change.
Uploaded the entire series edited and rebased according to your remarks.

hor63:
> I cherry-picked [d8b69db](https://github.com/XCSoar/XCSoar/commit/d8b69db65382d3965eade3f34db1310340478aec)

hor63:
Reopen pull request. I closed it bz accident.

build/targets.mk
MaxKellermann: "The template can be taken from the prefixed compiler commands in toolchains/llvm/prebuilt/linux-x86_64/bin."  .... but these are commands, not LLVM targets. The LLVM_TARGET variable is not used to build a command, it is only a parameter. This needs further explanation!

build/targets.mk
MaxKellermann: I could have cherry-picked this one, but unfortunately, you made this depend on the other commits. Meanwhile, I had to implement a similar variable for v6.8.x to address a build problem. I suggest you rebase your PR on current master, fix the conflicts, and force-push.

build/targets.mk
```diff
@@ -668,11 +668,9 @@ ifeq ($(TARGET),ANDROID)
...
   endif
 
-  # clang as linker driver adds the option '-pie' to the linker command for the X64 platform.
-  # This option which is incompatible with the option '-shared'.
-  ifeq ($(X64),y)
-    TARGET_LDFLAGS += -no-pie
-  endif
+  # clang as linker driver adds the option '-pie' to the linker command.
+  # This option is incompatible with the option '-shared'.
+  TARGET_LDFLAGS += -no-pie
```
MaxKellermann: There's no explanation why this is suddenly necessary.

android/src/XCSoar.java
```diff
@@ -103,14 +103,15 @@
...
 
-    /* Workaround for layout problems in Android KitKat with immersive full
-       screen mode: Sometimes the content view was not initialized with the
-       correct size, which caused graphics artifacts. */
-    getWindow().addFlags(WindowManager.LayoutParams.FLAG_LAYOUT_IN_SCREEN|
-                         WindowManager.LayoutParams.FLAG_LAYOUT_NO_LIMITS|
-                         WindowManager.LayoutParams.FLAG_LAYOUT_INSET_DECOR|
-                         WindowManager.LayoutParams.FLAG_TRANSLUCENT_NAVIGATION);
-
+    if (!Loader.isLegacy) {
```
MaxKellermann: Explanation missing in your commit message

android/src/Loader.java
```diff
@@ -29,12 +29,39 @@
...
+        loaded = true;
+      }
+    } catch (UnsatisfiedLinkError e) {
+      ;
+    }
+
+    /* Library xcsoar.so was not be loaded because
+     * - It is running on an older Android version < API 21
+     * - It is a legacy-only build of the APK
+     * - Something with loading of the library xcsoar simply went wrong.
```
MaxKellermann: Why is that a good reason to load the legacy library? If this solves anything, we have serious problems that must be addressed.

android/src/Loader.java
```diff
@@ -29,12 +29,39 @@
...
+       * in the Bionic C library
+       */
+
+      if (android.os.Build.VERSION.SDK_INT >= 21) {
+        System.loadLibrary("xcsoar");
+        Log.i(TAG, "Loaded native library xcsoar");
+        loaded = true;
+      }
+    } catch (UnsatisfiedLinkError e) {
+      ;
```
MaxKellermann: Don't just ignore the exception!

src/VALI-XCS.cpp
```diff
@@ -105,3 +105,26 @@ try {
...
 }
+
+#if defined __ANDROID_API__ && __ANDROID_API__ < 17
+extern "C" {
+
+  /** \brief Mimic posix_memalign as a thin wrapper around memalign.
+   *
+   * posix_memalign is not present in the API version 16 bionic library and before, but is required by the builtin "new" operator of clang in NDK19 or later.
+   */
+  int posix_memalign(void **memptr, size_t alignment, size_t size) {
```
MaxKellermann: No duplicate code!

build/targets.mk
hor63: With the most recent NDK V19.2.5435600 all targets including the 32 bit targets ANDROID7 and ANDROID86 report the error that the parameters -shared and -pie are incompatible.

Here the output from make TARGET=ANDROID7:
  LINK    output/ANDROID7/dbg/bin/libxcsoar-ns.so
/home/hor/Android/ndk-bundle/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: fatal error: -shared and -pie are incompatible
clang++: error: linker command failed with exit code 1 (use -v to see invocation)

android/src/Loader.java
hor63: Remove that line as a legitimate reason. See also below dealing with the exception handling for library xcsoar.

android/src/Loader.java
hor63: Will add an appropriate warning log message.

src/VALI-XCS.cpp
hor63: Looking at it again, absolutely agree. Just being lazy.
Will remove that commit, and instead add the source to the program.

android/src/XCSoar.java
hor63: Will add source code comment.

build/targets.mk
hor63: Sure will do. No problem

build/targets.mk
hor63: The mentioned commands a simple shell scripts which invode clang with the correct -target option. I meant the triplet which is the -target parameter in these command shell scripts.
I will amend the commit message.

hor63:
Rebased my commit series on the latest HEAD.
Fixed compatibility issues with Notification.Builder too.

build/targets.mk
MaxKellermann: But this commit does three things:
1. change `=` to `:=` (why?)
2. drop `-none` from the LLVM target (why?)
3. append $(ANDROID_NDK_API) to the LLVM target

You explained only 3.
About 2 (dropping `-none`): that may or may not be correct, but you didn't explain it. Look at https://developer.android.com/ndk/guides/standalone_toolchain and you'll see that Google's official NDK documentation tells you to use `-none`, while the shell wrappers don't use it. In any case, if you want to drop `-none`, explain why, and make it a separate commit.

android/src/XCSoar.java
```diff
@@ -220,7 +222,10 @@ private void getHapticFeedbackSettings() {
 
   private void enableImmersiveModeIfSupported() {
     // Set / Reset the System UI visibility flags for Immersive Full Screen Mode, if supported
-    ImmersiveFullScreenMode.enable(getWindow().getDecorView());
+    // In legacy mode immersive full screen mode is not supported.
+	if (!Loader.isLegacy) {
```
MaxKellermann: Wrong indent

android/src/XCSoar.java
MaxKellermann: But this is the wrong check. `isLegacy` checks whether the "legacy" library was loaded; it does not check the Android version. Of course, the Android version determiners whether the "legacy" library should be loaded, but I don't like this indirection; it is fragile and non-obvious.
If you really believe this code should be disabled, check `android.os.Build.VERSION.SDK_INT`.
Look at https://github.com/XCSoar/XCSoar/commit/b1f8d0ccf1dc6cdc3d2c5f21191667237237f556#diff-58a404defd924a09562ddea6a3f1cc86L112

src/Android/BluetoothHelper.cpp
```diff
@@ -67,10 +67,16 @@ BluetoothHelper::Initialise(JNIEnv *env)
...
                                                "()Lorg/xcsoar/AndroidPort;");
 
+#if __ANDROID_API__ >= 21
+
   startLeScan_method = env->GetStaticMethodID(cls, "startLeScan",
                                               "(Landroid/bluetooth/BluetoothAdapter$LeScanCallback;)Z");
   stopLeScan_method = env->GetStaticMethodID(cls, "stopLeScan",
                                              "(Landroid/bluetooth/BluetoothAdapter$LeScanCallback;)V");
+#else /* __ANDROID_API__ >= 21 */
+  startLeScan_method = nullptr;
```
MaxKellermann: If you really want to make this a compile-time check (why?), why does the variable `startLeScan_method` even exist?
Second question: why do you want to make this a compile-time check? It never was, and it has always worked with even older Android versions.

src/OS/FileDescriptor.cxx
```diff
@@ -240,7 +241,14 @@ FileDescriptor::CreateEventFD(unsigned initval) noexcept
 bool
 FileDescriptor::CreateSignalFD(const sigset_t *mask) noexcept
 {
-	int new_fd = ::signalfd(fd, mask, SFD_NONBLOCK|SFD_CLOEXEC);
+
+#if !defined (__ANDROID_API__) || __ANDROID_API__ >= 21
+  int new_fd = ::signalfd(fd, mask, SFD_NONBLOCK|SFD_CLOEXEC);
```
MaxKellermann: Wrong indent

MaxKellermann:
Much of commit "Android: Add Android legacy to support Honeycomb (Android 3.1, API 12) onward" reimplements code which I_ removed a while ago when I dropped support for old Android versions. Instead of redoing this code and repeating the mistakes, why not revert those commits?

build/targets.mk
MaxKellermann: I don't see this message with NDK r19c (the latest release). What is this version number you posted?

hor63:
> Much of commit "Android: Add Android legacy to support Honeycomb (Android 3.1, API 12) onward" reimplements code which I_ removed a while ago when I dropped support for old Android versions. Instead of redoing this code and repeating the mistakes, why not revert those commits?

This commit is more focused, and re-enables only things which are really necessary for Honeycomb onward. The solutions are the ones from the V6.x series.

build/android.mk
```diff
@@ -256,17 +277,35 @@ $(call SRC_TO_OBJ,$(SRC)/Android/NativeI2CbaroListener.cpp): $(NATIVE_HEADERS)
...
 	$(Q)cp $< $@
 
+ifeq ($(ANDROID_INCLUDE_LEGACY_LIB),y)
+  ANDROID_LEGACY_LIB_DIR = $(TARGET_OUTPUT_DIR)/$(XCSOAR_ABI)_legacy/bin
+  ANDROID_LEGACY_LIB_BUILD = $(patsubst %,$(ANDROID_ABI_DIR)/lib%.so,$(ANDROID_LIB_NAMES)_legacy)
+
+  $(ANDROID_LEGACY_LIB_BUILD): $(ANDROID_LEGACY_LIB_DIR)/lib$(ANDROID_LIB_NAMES)_legacy.so
+	$(Q)cp $< $@
+
+# When the standard library was built or re-build build the legacy lib too.
```
MaxKellermann: Confusing typo

MaxKellermann:
> This commit is more focused, and re-enables only things which are really necessary for Honeycomb onward. The solutions are the ones from the V6.x series.

Then do a partial revert, but the commit message should reflect that this reverts a commit (and which one). your commit message reads like you're the author of that code, which made me think you reimplemented it from scratch; but you're not the author after all.

build/android.mk
```diff
@@ -256,17 +277,35 @@ $(call SRC_TO_OBJ,$(SRC)/Android/NativeI2CbaroListener.cpp): $(NATIVE_HEADERS)
...
+  $(ANDROID_LEGACY_LIB_DIR)/lib$(ANDROID_LIB_NAMES)_legacy.so : $(ANDROID_LIB_BUILD)
```
MaxKellermann: Why does the legacy library require building the non-legacy library first?

build/android.mk
```diff
@@ -256,17 +277,35 @@ $(call SRC_TO_OBJ,$(SRC)/Android/NativeI2CbaroListener.cpp): $(NATIVE_HEADERS)
...
+	$(Q)$(MAKE) TARGET=$(ANDROID_ORG_TARGET) ANDROID_LEGACY=y ANDROID_INCLUDE_LEGACY_LIB=n DEBUG=$(DEBUG) USE_CCACHE=$(USE_CCACHE) V=$(V) $@
+
+endif # ANDROID_INCLUDE_LEGACY
```
MaxKellermann: Typo in comment

build/android.mk
```diff
@@ -256,17 +277,35 @@ $(call SRC_TO_OBJ,$(SRC)/Android/NativeI2CbaroListener.cpp): $(NATIVE_HEADERS)
 $(call SRC_TO_OBJ,$(SRC)/Android/NativeNunchuckListener.cpp): $(NATIVE_HEADERS)
 $(call SRC_TO_OBJ,$(SRC)/Android/NativeVoltageListener.cpp): $(NATIVE_HEADERS)
 
-ANDROID_LIB_BUILD = $(patsubst %,$(ANDROID_ABI_DIR)/lib%.so,$(ANDROID_LIB_NAMES))
+ifeq ($(ANDROID_LEGACY),y)
+  ANDROID_LIB_BUILD = $(patsubst %,$(ANDROID_ABI_DIR)/lib%.so,$(ANDROID_LIB_NAMES)_legacy)
```
MaxKellermann: Code duplication; you could easily do `ANDROID_LIB_BUILD += _legacy`

build/android.mk
```diff
@@ -256,17 +277,35 @@ $(call SRC_TO_OBJ,$(SRC)/Android/NativeI2CbaroListener.cpp): $(NATIVE_HEADERS)
...
+else
+  ANDROID_LIB_BUILD = $(patsubst %,$(ANDROID_ABI_DIR)/lib%.so,$(ANDROID_LIB_NAMES))
+endif
 $(ANDROID_LIB_BUILD): $(ANDROID_ABI_DIR)/lib%.so: $(ABI_BIN_DIR)/lib%.so | $(ANDROID_ABI_DIR)/dirstamp
 	$(Q)cp $< $@
 
+ifeq ($(ANDROID_INCLUDE_LEGACY_LIB),y)
+  ANDROID_LEGACY_LIB_DIR = $(TARGET_OUTPUT_DIR)/$(XCSOAR_ABI)_legacy/bin
```
MaxKellermann: I don't like that the formula for this variable has been implemented already elsewhere; this kind of code duplicate is fragile.

build/android.mk
```diff
@@ -209,29 +213,46 @@ ifeq ($(FAT_BINARY),y)
 ANDROID_LIB_BUILD =
 ANDROID_THIRDPARTY_STAMPS =
 
-# Example: $(eval $(call generate-abi,xcsoar,armeabi-v7a,ANDROID7))
+# Example: $(eval $(call generate-abi,xcsoar,armeabi-v7a,ANDROID7,n))
+# ("n" sets ANDROID_LEGACY)
```
MaxKellermann: Why is this "legacy" flag here even necessary? Will these rules ever be used?

build/android.mk
```diff
@@ -209,29 +213,46 @@ ifeq ($(FAT_BINARY),y)
...
 ANDROID_LIB_BUILD += $$(ANDROID_BUILD)/lib/$(2)/lib$(1).so
 
-$$(ANDROID_BUILD)/lib/$(2)/lib$(1).so: $$(OUT)/$(3)/$$(XCSOAR_ABI)/bin/lib$(1).so | $$(ANDROID_BUILD)/lib/$(2)/dirstamp
+# Legacy lib or not?
+ifeq ($(4),n)
+  ANDROID_THIRDPARTY_STAMP := $$(OUT)/$(3)/lib/$(2)/stamp
+  ANDROID_ABI_LIB := $$(OUT)/$(3)/$$(XCSOAR_ABI)/bin/lib$(1).so
+else
+  ANDROID_THIRDPARTY_STAMP := $$(OUT)/$(3)/lib_legacy/$(2)/stamp
+  ANDROID_ABI_LIB := $$(OUT)/$(3)/$$(XCSOAR_ABI)_legacy/bin/lib$(1).so
```
MaxKellermann: For the legacy build, `XCSOAR_ABI` already contains the `_legacy` suffix. So what this does is add it again!

MaxKellermann:
The whole "Android: Extend the build system for Andoid Legacy library build" commit still looks overly complex, confusing and wrong (because it contains redundant rules which appear to be useful, but are actually never used). Maybe I'm wrong about this - but that would just be one more reason to refactor it ...
Our Makefile is really ugly and confusing already, and we should avoid making it more ugly. This might be a good time to reimplement the "AndroidFat" build properly, and make it easier to add new ABIs, without invoking `make` again.
I wish we weren't using `make` at all, see https://github.com/XCSoar/XCSoar/issues/155 .... sigh

hor63:
> This might be a good time to reimplement the "AndroidFat" build properly, and make it easier to add new ABIs, without invoking `make` again.
> I wish we weren't using `make` at all, see #155 .... sigh

Yeah, agree. I am willing to do it next.
However at this time the build is working, and I suggest we get the system out as is as preview for testing for people with older Androids.
I will do a re-work of the Android.mk next. 
I will address the other issues today, and force-push again.

MaxKellermann:
> However at this time the build is working

Even if it works - if there are unused (and confusing) rules, this does more harm than good. It is important that other people can understand what you've been doing (and this is too hard already). Let's get this right.

If you have smaller fixup/improvement patches which are eligible for merging, move them to the top; that allows easier cherry-picking, because this way, they don't have dependencies on other (more complex and harder-to-merge) commits.

hor63:
OK. Got more work to do at *work* right now.
Will pick it up next week.

hor63:
> The whole "Android: Extend the build system for Andoid Legacy library build" commit still looks overly complex, confusing and wrong (because it contains redundant rules which appear to be useful, but are actually never used). Maybe I'm wrong about this - but that would just be one more reason to refactor it ...
> Our Makefile is really ugly and confusing already, and we should avoid making it more ugly. This might be a good time to reimplement the "AndroidFat" build properly, and make it easier to add new ABIs, without invoking `make` again.

I actually have no good idea how to implement multiple compiler runs with for the different targets within the Android FAT build with different compile options as well as the two runs for standard and legacy lib combined for one architecture without running make again.
Make is not procedural with local variables. It expands variables once and constructs the dependency tree with the recipies.
To circumvent that I would have to re-write the compile.mk and link.mk completely, thus creating more ugly kludges to avoid a more straight-forward solution invoking make again.
The current way android.mk handles it deserves a re-write for sure. But I do not see a way without invoking make again. But I will try to make it straight-forward and consistent.
This will probably include targets.mk too.

hor63:
At this point I am putting this pull request on the back-burner and first start with another commit-set wich will clean up, and harmonize the existing Android build.
This will be the foundation of this legacy-enabling pull request.

Besides more descriptive naming (ANDROID_ABI2-5 anyone?) the make process will be in two parts:
The outer part which you invoke will compose and build the Android APK.
Within it will call make to build one or more flavors of the actual XCSoar binary as shared lib. This will be a straight-forward binary build (as far as there is anything straight-forward with the XCSoar make).
All necessary information will be collected in targets.mk.
android.mk will be independent of the architecture, which and how many .so binaries are being built. All that needed information is handed down from targets.mk.
A one-architecture build, a fat build, a combined standard and legacy build... are all handled in one consistent way in android.mk.
Thus adding a new architecture or removing one will become a lot easier and less error prone.

On this foundation I will re-start this pull request.
Thoughts?

MaxKellermann:
> At this point I am putting this pull request on the back-burner and first start with another commit-set wich will clean up, and harmonize the existing Android build.

This is always a good idea. First get cleanups merged, then do more work on cleaner code base.

> The outer part which you invoke will compose and build the Android APK.
> Within it will call make to build one or more flavors of the actual XCSoar binary as shared lib. This will be a straight-forward binary build (as far as there is anything straight-forward with the XCSoar make).

Good idea.

> The Java and other processor independent stuff will be build once only independent from the architecture.

That's tricky, because Java (javac+javah) needs to be invoked to generate JNI C headers, which are needed by the C++ part ...

MaxKellermann:
Depends on https://github.com/XCSoar/XCSoar/pull/209 - reopen after that one is merged

PR #189, opened 2019-05-03
Added driver for Becker AR62xx radios

eku60:
The drivers have been positive tested on UNIX and Windows with a Becker AR6201 attached to ttys1 and com1. After establishing connection the actual settings were read from the AR6201 and then several times the active or standby frequencies have been changed with these from waypoint data using the gui-buttons. Switching the radio on/off and cutting of the connection was showing no problem during testing...


MaxKellermann:
This PR is confusing because it contains several commits which add code, delete it, add different code etc. - please clean this up so it can be reviewed.

eku60:
Hello,

sorry, on the first commit I forgot useful comments. So I deleted the first commits (ar62xx.cpp and ar62xx.hpp) and committed them again with useful comments.
In addition the make file  and Register.cpp are committed.

I don't know how to delete the first three steps out of the pull request. Need help.

MaxKellermann:
I recommend using [stgit](http://www.procode.org/stgit/) to properly author patches. There are other tools and ways for doing it, but that's how I do it.
What you can do now is use `git rebase -i`, see https://git-scm.com/docs/git-rebase#_interactive_mode
To edit the top-most commit, use `git commit --amend`
That's the short story.

antonka2000:
how does the AR62xx driver perform. I would need it in my glider. LK8000 has it inside. But I am more used to XCsoar.

MaxKellermann:
@antonka2000, Becker products will not be supported by XCSoar, because of Becker acts hostile towards this project. Sorry. Buy a radio from a different manufacturer. Or get used to LK8000. In any case, discussing this is off-topic here.

eku60:
Hi Antonka2000

when Max writes, that Becker acts hostile towards an implementation of a driver for their radios in XCSoar, it would be bad news.

Otherwise I used some methods from the LK8000 driver and the XCSoar driver for the KRT2 to create the driver. This worked with the images 21xxx of my own. You also will have to use a RS232<->RS422 converter, as Becker uses a RS422 interface.

To get the driver into XCSoar there is some work necessary, but if Becker forbids it, there would hardly be any chance.
Our club is hard working on a new hangar, so I have no time to prepare it in a way that it would be acceptable by XCSoar.

I assume, Max has more infos on Becker's suggestions, than me. Maybe someone asked Becker for protocol-information...

eku60

PR #191, opened 2019-05-04
InfoBoxes navigation with courser key.

glidingFisi:
I created a new PR, because I did a mistake in the rebase of my fork.
The navigation with the courser key is now configurable, in the Look
dialog, where the other InfoBoxes settings are placed.
The default behavior is like it is now.
To have all features from an InfoBox, a new mode was added. If the
InfoBoxes courser mode is activ and on a InfoBox is clicked, the InfoBox
background turns red. And with the courser can be moved through the
infoboxes. If the mouse is pressed again or the enter key is pressed, it
gets in the focus mode "light red" background. With pressing the enter
again the InfoBox dialog will be open. With escape it can be exit from
each mode.
The mentioned layer problem that the InfoBoxWindow call the
InfoBoxManager is still available.
The Altair approach is not useful for Stefly, because Altair have a lot
of buttons, this isn't available.
For e.g. a custom xci file can created, with editing the default keys:
mode=default
type=key
data=LEFT
event=ScreenModes infoboxfocus
So the InfoBox courser mode can reached by pressing the left courser.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -51,6 +51,7 @@ Version 7.0 - not yet released
       climbing cruise (green), circling non-climb (orange)
   - improved auto-scaling of vario-like graphical infoboxes
   - new infobox: "Number Of Satellites"
+  - infobox courser mode: navigate with courser keys through the InfoBoxes
 * Windows
   - drop support for Windows CE
   - require Windows Vista or later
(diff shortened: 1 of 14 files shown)


MaxKellermann:
I don't like that you closed the old PR and created a new one; and I don't like that your new PR doesn't address all of my complaints. I suggest we continue talking about your code as soon as everything is fixed.

PR #193, opened 2019-05-06
WIP: Driver/KRT2: fix initializer of CMD_TIMEOUT

kerel-fs:
Note: Can't compile locally atm, thus using CircleCI to test the patch. I'm not sure this patch is correct.


MaxKellermann:
Explain

kerel-fs:
Nvm, this was bs. :)

PR #194, opened 2019-05-06
Driver/KRT2: Add workaround for possible GCC bug

kerel-fs:
~~WORK IN PROGRESS~~
This commit adds the redeclaration of CMD_TIMEOUT at namespace
scope as required before C++17. This commit is introduced in order
to fix a potential gcc bug, where gcc 6.3 and 8.3 doesn't inline
the variable (despite the constexpr modifier).

### DIFF
--- a/src/Device/Driver/KRT2.cpp
+++ b/src/Device/Driver/KRT2.cpp
@@ -150,6 +150,13 @@ public:
                             struct NMEAInfo &info) override;
 };
 
+/* Workaround for some GCC versions which don't inline the constexpr
+   despite being defined so in C++17, see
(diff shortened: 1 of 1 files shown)


Turbo87:
I guess the other option would be to look into updating gcc on CI. Since it's using Docker underneath it should be somewhat straight-forward and reproducible.

kerel-fs:
In this case a note should be added in the dev manual or the requirements have to be updated to gcc 9.1 (https://gitlab.com/kerel-fs/xcsoar-docker-build/pipelines/59926311 confirms that this also happens with gcc 8.3), as @lordfolken pointed out via IRC.

PR #198, opened 2019-05-14
Logger/FRecord: Add redeclaration of constexpr variables

kerel-fs:
This commit adds the redeclaration of those variables at namespace
scope as required before C++17. This commit is introduced as a
workaround for gcc (6.3 and 8.3) which don't inline the variables
(despite its constexpr modifier).

Fixes XCSoar#195.

### DIFF
--- a/src/Computer/RouteComputer.cpp
+++ b/src/Computer/RouteComputer.cpp
@@ -31,6 +31,13 @@ Copyright_License {
 
 #include <algorithm>
 
+/* Workaround for some GCC versions which don't inline the constexpr
+   despite being defined so in C++17, see
+   http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0386r2.pdf */
+#if GCC_OLDER_THAN(9,0)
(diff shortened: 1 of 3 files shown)


kerel-fs:
Do you know a better way to fix the build for debian stable / GCC 6.3.0 & GCC 8.3.0? I don't really like to sprinkle this workaround across more and more places in the code base.

MaxKellermann:
I don't think there is an easier way to work around this GCC bug. It's an annoying bug, and it's always annoying to support known-buggy toolchains. That's why I like dropping support for those, but on the other hand, I want XCSoar to be buildable with Debian Stable (Stretch atm). Buster will be released soon.

kerel-fs:
ok.
If buster will come with GCC 8.3.0 (which is the gcc version in debian testing currently), we have to stay with those workarounds even longer, as GCC 8.3.0 does also show this behavior.

MaxKellermann:
My computer runs Buster with GCC 8.3.0, and I never saw this problem.

PR #209, opened 2019-05-28
Android: Re-structure the Android build system, and third-party build

hor63:
The build is restructured in following ways:
- The native build specific stuff was moved to the new
  android_native.mk.
- The build process always consists of two steps:
  - The make run invoked by the user builds the Java stuff and
    architecture independent resources like sounds and images.
    The output is always ./output/ANDROID
  - The inner make run or runs to build one or multiple native shared
    lib. The output is ./output/ANDROID/<ANDROID_APK_LIB_ABI>, i.e. the
    subdirectory depends on the native architecture.
- The build process is for a single architecture and a Fat build
    identical. The native architecture or architectures are set in
    targets.mk in the variable ANDROID_TARGETS which can contain a
    single or multiple targets.
- android.mk and android_native.mk are target and architecture
    agnostic. All necessary information is defined in targets.mk.
- All architecture specific variables are listed in a single block in
    targets.mk. Adding a new architecture requires adding a new
    architecture block here.


build/targets.mk
```diff
@@ -314,6 +314,12 @@ endif
 ifeq ($(TARGET),ANDROID)
   ANDROID_NDK ?= $(HOME)/opt/android-ndk-r20-beta3
 
+  # Control if the Java and other machine independent parts are being built
+  # or if the native libxcsoar.so is being built.
+  # The caller must never set this 'n'. This is being done by make itself
+  # which calls itself with 'ANDROID_BUILD_APK = n' to build the native lib.
+  ANDROID_BUILD_APK ?= y
```
MaxKellermann: `make` has a built-in syntax to declare which targets shall be built. This sounds like it attempts to imitate this `make` feature, but badly?

build/android.mk
```diff
@@ -202,75 +189,48 @@ $(ANDROID_BUILD)/classes.dex: $(JAVA_SOURCES) $(ANDROID_BUILD)/gen/org/xcsoar/R.
...
-$(NATIVE_HEADERS): $(ANDROID_BUILD)/classes.dex
-
 .DELETE_ON_ERROR: $(ANDROID_BUILD)/unsigned.apk
-$(ANDROID_BUILD)/unsigned.apk: $(ANDROID_BUILD)/classes.dex $(ANDROID_BUILD)/resources.apk $(ANDROID_LIB_BUILD)
+$(ANDROID_BUILD)/unsigned.apk: cleanNativeLibs $(ANDROID_BUILD)/classes.dex $(ANDROID_BUILD)/resources.apk $(ANDROID_LIB_BUILD)
 	@$(NQ)echo "  APK     $@"
 	$(Q)cp $(ANDROID_BUILD)/resources.apk $@
 	$(Q)cd $(dir $@) && zip -q -r $(notdir $@) classes.dex lib
 
+# Remove potentially stale native shared libs before building and copying new ones to the lib directory
```
MaxKellermann: What does this mean?

build/targets.mk
hor63: Removed the use this variable to (re) define the make targets. It is still used internally to distinguish between the outer and inner make.

build/android.mk
```diff
@@ -202,75 +189,48 @@ $(ANDROID_BUILD)/classes.dex: $(JAVA_SOURCES) $(ANDROID_BUILD)/gen/org/xcsoar/R.
...
-	$(Q)cp $< $@
-
-endif # !FAT_BINARY
-
-
-$(NATIVE_HEADERS): $(ANDROID_BUILD)/classes.dex
-
 .DELETE_ON_ERROR: $(ANDROID_BUILD)/unsigned.apk
-$(ANDROID_BUILD)/unsigned.apk: $(ANDROID_BUILD)/classes.dex $(ANDROID_BUILD)/resources.apk $(ANDROID_LIB_BUILD)
+$(ANDROID_BUILD)/unsigned.apk: cleanNativeLibs $(ANDROID_BUILD)/classes.dex $(ANDROID_BUILD)/resources.apk $(ANDROID_LIB_BUILD)
```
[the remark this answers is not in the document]
hor63: Moved the cleanNativeLibs target to the appropriate place: Before copying the native libs.

build/android.mk
hor63: Better word is outdated. I re-wrote the comment to make the purpose clear. Please see the updated source.

hor63:
There is another issue with parallel makes: The sources of the third-party libraries are going into the directory output/src.
thirdparty.py checks the dependency of the target lib, and the originating tarball. When the target lib is outdated or not present the source (unpack) directory is purged, and the sources are un-packed again into the output/src/... directory. This leads to errors in a parallel build when parallel deletes, and parallel un-tars clash (in the best case) in the worst case it leads to an incomplete source directory causing strange build errors.

I suggest, not to mix that fix with the current Android build improvements.
I am going to fix that in a different pull request in the next days.

hor63:
The seemingly simple task of downloading and unpacking third-party libs in the outer make for Android to avoid concurrent purging, downloading, and unpacking in the sub-makes for the native libraries ended with pretty substantial modifications to thirdparty.py and friends, hopefully for the better.

hor63:
Any chance for a review to move this and the Android Legacy option forward?
Thanks

MaxKellermann:
You make it extremely difficult to review your PR; for example, the first commit moves "make" rules around **and** changes them, both in one commit. And contains lots of completely unrelated changes (adding useless `echo` statements) which further obscure your work.
The practical problem is that whenever I take some time to look at this PR, I close the tab quickly because I figure I don't have enough time for this. Even if I took enough time, it would feel somewhat frustrating because the complexity or this commit is unnecessary.

MaxKellermann:
Needs serious cleanup. Will reopen when something new can be reviewed.

PR #210, opened 2019-05-31
Upates to allow extra key codes for android and to hardcode BUTTON_L1…

be-adairing:
… & R1 to control the modal menus.


src/Event/Windows/KeyCode.hpp
```diff
@@ -69,6 +69,19 @@ enum {
...
+  KEYCODE_BUTTON_L2 = '',
+  KEYCODE_BUTTON_A = '',
+  KEYCODE_BUTTON_B = '',
+  KEYCODE_BUTTON_C = '',
+  KEYCODE_BUTTON_X = '',
+  KEYCODE_BUTTON_Y = '',
+  KEYCODE_BUTTON_Z = '',
+  KEYCODE_MEDIA_NEXT = '',
+  KEYCODE_MEDIA_PREVIOUS = '',
+  KEYCODE_MEDIA_PLAY_PAUSE = ''
```
MaxKellermann: What do these declarations mean?

src/Event/Android/KeyCode.hpp
```diff
@@ -38,6 +38,20 @@ enum {
...
+  KEYCODE_BUTTON_L2 = 0x68,
+  KEYCODE_BUTTON_A = 0x60,
+  KEYCODE_BUTTON_B = 0x61,
+  KEYCODE_BUTTON_C = 0x62,
+  KEYCODE_BUTTON_X = 0x63,
+  KEYCODE_BUTTON_Y = 0x64,
+  KEYCODE_BUTTON_Z = 0x65,
+  KEYCODE_MEDIA_NEXT = 0x57,
+  KEYCODE_MEDIA_PREVIOUS = 0x58,
+  KEYCODE_MEDIA_PLAY_PAUSE = 0x55,
```
MaxKellermann: These key codes should be in a separate commit, with a link to the official documentation where you found their definitions.

src/Form/Button.cpp
```diff
@@ -150,6 +150,9 @@ Button::OnKeyCheck(unsigned key_code) const
 {
   switch (key_code) {
   case KEY_RETURN:
+  #ifdef ANDROID
```
MaxKellermann: Bad indent

src/Form/Button.cpp
```diff
@@ -150,6 +150,9 @@ Button::OnKeyCheck(unsigned key_code) const
...
+  case KEYCODE_BUTTON_L1:
```
MaxKellermann: This change is not explained. What is `BUTTON_L1` and why do you want it to act as replacement for `RETURN`?

src/Form/Button.cpp
```diff
@@ -163,9 +166,13 @@ Button::OnKeyDown(unsigned key_code)
   switch (key_code) {
   case KEY_RETURN:
   case KEY_SPACE:
+  #ifdef ANDROID
+  case KEYCODE_BUTTON_L1:
+  #endif
     Click();
     return true;
 
+
```
MaxKellermann: Why add a second empty line here?

src/Form/Form.cpp
```diff
@@ -391,6 +392,13 @@ WndForm::ShowModal()
       }
 #endif
 
+#ifdef ANDROID
+      if (event.GetKeyCode() == KEY_ESCAPE || event.GetKeyCode() == KEYCODE_BUTTON_R1) {
```
MaxKellermann: The `ESCAPE` check is redundant. The `BUTTON_R1` check is not explained.

src/Form/Form.cpp
```diff
@@ -44,6 +44,7 @@ Copyright_License {
 #ifdef ANDROID
 #include "Event/Shared/Event.hpp"
 #include "Event/Android/Loop.hpp"
+#include "Event/Android/KeyCode.hpp"
```
MaxKellermann: Redundant.

src/InfoBoxes/InfoBoxWindow.cpp
```diff
@@ -381,7 +384,7 @@ InfoBoxWindow::OnKeyDown(unsigned key_code)
 
   if (InputEvents::ProcessKey(InputEvents::MODE_INFOBOX, key_code))
     return true;
-
+
```
MaxKellermann: Why add whitespace here?

src/Input/InputKeys.cpp
```diff
@@ -61,6 +79,9 @@ static constexpr struct string_to_key string_to_key[] = {
   { _T("F18"), KEY_F18 },
   { _T("F19"), KEY_F19 },
   { _T("F20"), KEY_F20 },
+
+
+
```
MaxKellermann: Why do you suggest to add three empty lines here, and what does this have to do with this overall change?

src/Event/Windows/KeyCode.hpp
be-adairing: They are redundant now so i have removed them.

src/Event/Android/KeyCode.hpp
be-adairing: They are from the same source as the others in the class.

src/Form/Button.cpp
be-adairing: have moved this and all the others all the way to the left.

src/Form/Button.cpp
be-adairing: I have removed this extra line.

src/Form/Form.cpp
be-adairing: I will remove it.

src/InfoBoxes/InfoBoxWindow.cpp
be-adairing: Will remove it.

src/Event/Windows/KeyCode.hpp
MaxKellermann: No, they're not redundant. They're wrong.

src/Input/InputKeys.cpp
be-adairing: will remove it.

src/Event/Windows/KeyCode.hpp
be-adairing: I limited the change to android after checking the keycodes for windows on the following page: 

https://docs.microsoft.com/en-us/windows/desktop/inputdev/virtual-key-codes

The MS VK codes don't seem to have the controller codes - if there is one available then they can be added.

src/Form/Button.cpp
be-adairing: Gamepads/Controllers by convention use the primary buttons (BUTTON_L1 & BUTTON_R1) to control 'click/select' and 'escape/back' UI behaviour. Modal dialogs in XCS currently assume these behaviours are linked to KEY_ENTER and KEY_ESCAPE; BUTTON_L1 & BUTTON_R1 are their controller analogues.

Is there a better place for this explanation?

src/Form/Form.cpp
be-adairing: This is another place where the BUTTON_L1/BUTTON_R1 change needs to be made so that the controller and 'escape' from the menu.

be-adairing: Would it be a good idea for these to be commented in each location?

src/Form/Button.cpp
MaxKellermann: Yes, explanations belong in code comments (or commit messages). The code must be comprehensible without any network connection, only with the data in your local git repository.
After your explanation, I think it's a bad idea to hook `BUTTON_L1`/`R1` into each and every code location where we currently check `RETURN`/`ESCAPE`. **If** we decide that `BUTTON_L1` acts as a full-featured `RETURN` replacement *everywhere*, then we should translate this key code earlier. For example here: https://github.com/XCSoar/XCSoar/blob/master/src/Android/EventBridge.cpp#L47-L63

src/Form/Form.cpp
MaxKellermann: My brain is unable to parse your replies.

src/Form/Button.cpp
be-adairing: I totally agree that is far better than my approach - I will rework the code to use this method.

src/Form/Form.cpp
be-adairing: i'll use the Bridge method to map all inputs so this can be removed.

PR #211, opened 2019-05-31
Add Android controller keycodes and default behaviours for controllers.

be-adairing:
With Android it is possible to have extra buttons / external controllers that send unknown to XCS keycodes.

This patch adds these missing keycodes to XCS and enables the controller to use it's L1 & R1 buttons act as Return and Escape so XCS modal dialogs can be controlled. It uses the Android EventBridge to map the inbound keycodes to the correct XCS keycodes. 

The attached init.lua file could be a reasonable starting point for a controller config.
[init.lua.zip](https://github.com/XCSoar/XCSoar/files/3242017/init.lua.zip)


MaxKellermann:
Look at your commit list, this is a mess! I can't review that.

fb:
In general, I think adding support on Android for those keycodes is a good idea.
However I don't agree with your decision about the default behavior for L1 and R1.

The lua example can just be posted like this instead of a zip file containing a text file with 25 lines.

```lua
xcsoar.input_event.new("key_BUTTON_Y",
	function(e)
		xcsoar.fire_legacy_event("QuickMenu","");
	end
);
```

And please clean up your git history before creating a PR.

be-adairing:
Sorry but i've never contributed to an open source project or GitHub before this, I thought that the PR would just be looked at as the files that had changed. I'll get in into one commit. 

I attached the lua as a zip because github doesn't allow attaching lua files and I thought it might be of some help.

Most importantly - ow do you think the buttons should behave? They do have to be mapped because of the way the modal menus work.

fb:
I'd keep things simple at first and just add support for the keycodes to allow to use them with a custom configuration.

Maybe _R1 = Back_ makes sense for some input devices, but maybe it's not intuitive for all.

MaxKellermann:
> I thought that the PR would just be looked at as the files that had changed.

I look at individual commits, because I expect each commit to represent a separate and documented change.

> I'll get in into one commit.

Sometimes, folding several commits into one can be a good idea, but not always. Keep separate changes in separate commits. For example:

> I'd keep things simple at first and just add support for the keycodes to allow to use them with a custom configuration

Agree with that. Adding general support for those keycodes is a good first step, and should not be mixed with your other proposed changes. If you make it a separate commit, I can cherry-pick that commit, and we can continue discussing the other commits.

But the PR as it is now cannot be reviewed nor can I cherry-pick the good parts.

be-adairing:
I think you are correct that we should let the user choose what buttons do.

I will separate the changes.

Perhaps the best way to enable controllers is to allow the LUA scripts to send keystrokes with LUA? Rather than me dive in and start writing it, is there a place to discus and agree what should be done?

Paul

Sent from my iPhone

On 31 May 2019, at 18:28, Fabian Bartschke <notifications@github.com<mailto:notifications@github.com>> wrote:


I'd keep things simple at first and just add support for the keycodes to allow to use them with a custom configuration.

Maybe R1 = Back makes sense for some input devices, but maybe it's not intuitive for all.

PR #213, opened 2019-06-03
Add keypad keys

be-adairing:
Android can have extra keypad keys, typically used on BlueTooth controllers. This patch enables these extra KEYCODES to be used within XCS, ie with LUA scripts.

For example:
xcsoar.input_event.new("key_BUTTON_X",
	function(e)
    		if (xcsoar.map.is_panning) then
    			xcsoar.map.leavepan();
    		else
    			xcsoar.map.enterpan();
    		end
	end
)


MaxKellermann:
Cherry-picked the good commit to 862a178dfd9d2ed1c8deb4ca4a41b276888da84e and ignored the misplaced merge commit.

PR #219, opened 2019-06-20
Extend SeeYouReader to parse the ICAO Field in CUP files

lordfolken:
The ICAO field aka short name is currently unused. This parses the entry and displays it in the waypoint details dialog. It also makes the short name searchable. 

Waypoint Details
2020-08-12-104438_640x480_scrot

Waypoint Search List Dialog
2020-08-12-104219_640x480_scrot

- [x] fix the merge and rebase code
- [x] make shortname field searchable in waypoint search
- [x] actually display field in Waypoint dialog
- [x] display shortname field in Waypoint List
- [x] update NEWS.txt

### DIFF
--- a/src/Dialogs/Waypoint/WaypointInfoWidget.cpp
+++ b/src/Dialogs/Waypoint/WaypointInfoWidget.cpp
@@ -130,6 +130,9 @@ WaypointInfoWidget::Prepare(ContainerWindow &parent, const PixelRect &rc)
   if (!buffer.empty())
     AddReadOnly(_("Runway"), nullptr, buffer);
 
+  if (!waypoint->shortname.empty())
+    AddReadOnly(_("Short Name"), nullptr, waypoint->shortname.c_str());
+
   if (FormatGeoPoint(waypoint->location,
                      buffer.buffer(), buffer.capacity()) != nullptr)
     AddReadOnly(_("Location"), nullptr, buffer);
--- a/src/Engine/Waypoint/Waypoints.cpp
+++ b/src/Engine/Waypoint/Waypoints.cpp
@@ -104,16 +104,22 @@ void
 Waypoints::WaypointNameTree::Add(WaypointPtr wp)
 {
   TCHAR normalized_name[wp->name.length() + 1];
+  TCHAR normalized_shortname[wp->shortname.length() +1];
   NormalizeSearchString(normalized_name, wp->name.c_str());
+  NormalizeSearchString(normalized_shortname, wp->shortname.c_str());
   RadixTree<WaypointPtr>::Add(normalized_name, std::move(wp));
+  RadixTree<WaypointPtr>::Add(normalized_shortname, std::move(wp));
 }
 
 void
 Waypoints::WaypointNameTree::Remove(const WaypointPtr &wp)
 {
   TCHAR normalized_name[wp->name.length() + 1];
+  TCHAR normalized_shortname[wp->shortname.length() + 1];
   NormalizeSearchString(normalized_name, wp->name.c_str());
+  NormalizeSearchString(normalized_shortname, wp->shortname.c_str());
   RadixTree<WaypointPtr>::Remove(normalized_name, wp);
+  RadixTree<WaypointPtr>::Remove(normalized_shortname, wp);
 }
 
 Waypoints::Waypoints()
--- a/src/Renderer/WaypointListRenderer.cpp
+++ b/src/Renderer/WaypointListRenderer.cpp
@@ -89,7 +89,13 @@ Draw(Canvas &canvas, PixelRect rc,
   row_renderer.DrawSecondRow(canvas, rc, buffer);
 
   // Draw waypoint name
-  row_renderer.DrawFirstRow(canvas, rc, waypoint.name.c_str());
+  if (!waypoint.shortname.empty()) {
+    std::string waypoint_title = waypoint.name+" ("+waypoint.shortname+")";
+    row_renderer.DrawFirstRow(canvas, rc, waypoint_title.c_str());
+  }
+  else {
+    row_renderer.DrawFirstRow(canvas, rc, waypoint.name.c_str());
+  }
 }
 
 void
@@ -153,5 +159,11 @@ WaypointListRenderer::Draw(Canvas &canvas, PixelRect rc,
   row_renderer.DrawSecondRow(canvas, rc, buffer);
 
   // Draw waypoint name
-  row_renderer.DrawFirstRow(canvas, rc, waypoint.name.c_str());
+  if (!waypoint.shortname.empty()) {
+    std::string waypoint_title = waypoint.name+" ("+waypoint.shortname+")";
+    row_renderer.DrawFirstRow(canvas, rc, waypoint_title.c_str());
+  }
+  else {
+    row_renderer.DrawFirstRow(canvas, rc, waypoint.name.c_str());
+  }
 }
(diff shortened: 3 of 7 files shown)


src/Dialogs/Waypoint/WaypointInfoWidget.cpp
```diff
@@ -130,10 +130,13 @@ WaypointInfoWidget::Prepare(ContainerWindow &parent, const PixelRect &rc)
...
     AddReadOnly(_("Runway"), nullptr, buffer);
 
+  if (!waypoint->icao.empty())
+    AddReadOnly(_("ICAO"), nullptr, waypoint->icao.c_str());
+
   if (FormatGeoPoint(waypoint->location,
                      buffer.buffer(), buffer.capacity()) != nullptr)
     AddReadOnly(_("Location"), nullptr, buffer);
-
+
```
rawtaz: I'm guessing this shouldn't be here :)

lordfolken: done

src/Renderer/WaypointListRenderer.cpp
```diff
@@ -89,7 +89,13 @@ Draw(Canvas &canvas, PixelRect rc,
   row_renderer.DrawSecondRow(canvas, rc, buffer);
 
   // Draw waypoint name
-  row_renderer.DrawFirstRow(canvas, rc, waypoint.name.c_str());
+  if (!waypoint.shortname.empty()) {
+    std::string waypoint_title = waypoint.name+" ("+waypoint.shortname+")";
```
MaxKellermann, line 93: This is very inefficient, because it will:
1. copy `waypoint.name`
2. do another copy with `waypoint.name + " ("
3. do another copy of everything so far, appending `waypoint.shortname`
4. do yet another copy of everything so far, appending `")"`
It would be OK to be very inefficient now and optimize later, but this is so bad, it deserves a TODO comment.
But... does this compile on Windows?

src/Engine/Waypoint/Waypoints.cpp
```diff
@@ -104,16 +104,22 @@ void
 Waypoints::WaypointNameTree::Add(WaypointPtr wp)
 {
   TCHAR normalized_name[wp->name.length() + 1];
+  TCHAR normalized_shortname[wp->shortname.length() +1];
```
MaxKellermann, line 107: This is a variable-length array which is illegal in C++.
The line above it does the same, but it's also bad. We should fix all VLAs in our code, and not write new code using this feature.

src/Engine/Waypoint/Waypoints.cpp
```diff
@@ -104,16 +104,22 @@ void
...
   NormalizeSearchString(normalized_name, wp->name.c_str());
+  NormalizeSearchString(normalized_shortname, wp->shortname.c_str());
   RadixTree<WaypointPtr>::Add(normalized_name, std::move(wp));
+  RadixTree<WaypointPtr>::Add(normalized_shortname, std::move(wp));
```
MaxKellermann, line 111: What is the value of `wp` before executing this line?
By answering this question, you'll find out that this is broken and cannot work.

src/Renderer/WaypointListRenderer.cpp
lordfolken: I understand. The new implementation is now using string append();

src/Engine/Waypoint/Waypoints.cpp
lordfolken: I replaced this vla with a dynamic array. I left the original code, as this cleanup isn't part of this feature.

lordfolken:
@MaxKellermann could you review again?

src/Engine/Waypoint/Waypoints.cpp
```diff
@@ -106,6 +106,11 @@ Waypoints::WaypointNameTree::Add(WaypointPtr wp)
   TCHAR normalized_name[wp->name.length() + 1];
   NormalizeSearchString(normalized_name, wp->name.c_str());
   RadixTree<WaypointPtr>::Add(normalized_name, std::move(wp));
+  if (!wp->shortname.empty()) {
```
MaxKellermann, line 109: This will crash. See also https://github.com/XCSoar/XCSoar/pull/219#discussion_r475554551 (which you did not resolve).

src/Renderer/WaypointListRenderer.cpp
```diff
@@ -89,7 +89,15 @@ Draw(Canvas &canvas, PixelRect rc,
   row_renderer.DrawSecondRow(canvas, rc, buffer);
 
   // Draw waypoint name
-  row_renderer.DrawFirstRow(canvas, rc, waypoint.name.c_str());
+  if (!waypoint.shortname.empty()) {
+    std::string waypoint_title = waypoint.name.c_str();
+    waypoint_title.append(" (",2);
+    waypoint_title.append(waypoint.shortname);
+    waypoint_title.append(")",1);
```
MaxKellermann, line 96: This is just as inefficient as before (see https://github.com/XCSoar/XCSoar/pull/219#discussion_r475553548).

src/Renderer/WaypointListRenderer.cpp
MaxKellermann: append() isn't any more efficient than operator+. It's just another name for the same thing.

src/Engine/Waypoint/Waypoints.cpp
```diff
@@ -114,6 +119,11 @@ Waypoints::WaypointNameTree::Remove(const WaypointPtr &wp)
   TCHAR normalized_name[wp->name.length() + 1];
   NormalizeSearchString(normalized_name, wp->name.c_str());
   RadixTree<WaypointPtr>::Remove(normalized_name, wp);
+  if (!wp->shortname.empty()) {
+    TCHAR *normalized_shortname = new TCHAR;
+    NormalizeSearchString(normalized_shortname, wp->shortname.c_str());
```
MaxKellermann, line 124: This will also crash, because `normalized_shortname` has room for exactly one character, i.e. barely enough for the null terminator.

lordfolken:
@MaxKellermann could you have a look again please?

src/Dialogs/Settings/Panels/WaypointDisplayConfigPanel.cpp
```diff
@@ -95,7 +95,10 @@ WaypointDisplayConfigPanel::Prepare(ContainerWindow &parent,
     { (unsigned)WaypointRendererSettings::DisplayTextType::FIRST_FIVE,
       N_("First 5 letters"),
       N_("The first 5 letters of the waypoint name are displayed.") },
-    { (unsigned)WaypointRendererSettings::DisplayTextType::NONE,
+    { (unsigned)WaypointRendererSettings::DisplayTextType::SHORT_FIVE,
+      N_("Shortname then first 5 letters"),
+      N_("Displays the shortname. If unavailable, the first 5 letters of the waypoint name are displayed.") },
+     { (unsigned)WaypointRendererSettings::DisplayTextType::NONE,
```
MaxKellermann, line 101: Whitespace change

src/Renderer/WaypointRendererSettings.hpp
```diff
@@ -33,6 +33,7 @@ struct WaypointRendererSettings {
   enum class DisplayTextType : uint8_t {
     NAME = 0,
     OBSOLETE_DONT_USE_NUMBER,
+    SHORT_FIVE,
```
MaxKellermann, line 36: Inserting a new enum value in the middle will break all existing configurations

PR #220, opened 2019-06-20
LXNAV Vario firmware 7.0.5 (WIP)

lordfolken:
The current firmware does not allow to declare or download flights in 6.8.12 or xcsoar-test. 

These are some fixes that restore this communication.

This needs testing before merge.


MaxKellermann:
The first commit is a big confusing mess. This should not be merged.

lordfolken:
I agree.
I will have a look at it and rework it. Separate the fixes from the renaming.

lordfolken:
Closing for now.

PR #224, opened 2019-06-25
Add missing packages to developer manual.

sbrodehl:
Adds `quilt` and `glm` package to compile instructions (_2.3.1 Compiling for Linux/UNIX_).

### DIFF
--- a/doc/manual/en/compiling.tex
+++ b/doc/manual/en/compiling.tex
@@ -79,6 +79,8 @@ similar operating systems:
 \item \href{http://www.libpng.org/}{libpng}
 \item \href{http://libjpeg.sourceforge.net/}{libjpeg}
 \item OpenGL (Mesa)
+\item \href{https://glm.g-truc.net}{OpenGL Mathematics (GLM)} 
+\item \href{https://savannah.nongnu.org/projects/quilt/}{Quilt} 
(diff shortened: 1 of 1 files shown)


MaxKellermann:
glm is not necessary; it's included as a git submodule.

lordfolken:
Have a look at https://github.com/XCSoar/XCSoar/blob/master/ide/provisioning/install-debian-packages.sh
and please add any missing packages to the manual

MaxKellermann:
Fold those two patches. The first one is wrong, and I don't want to merge a commit when we already know it's wrong.
Rule of thumb: if your PR needs a text, then your commit message is bad. This PR has text.

sbrodehl:
I checked the source files again, and it looks like everything is fine.
Section _2.2 Requirements_, together with section _2.3.1 Compiling for Linux/UNIX_ lists all needed packages for Linux

But, reading the [v6.8.11 XCSoar-developer-manual.pdf](http://download.xcsoar.org/releases/6.8.11/XCSoar-developer-manual.pdf) it turns out that the PDF was build on August 17, 2018!

So, the only thing to do is to build an up-to-date version of the developer manual and publish it, I guess?

Sorry for the confusion! :-/

PR #225, opened 2019-06-25
Detect which lua version syntax to use.

sbrodehl:
Detect which lua version syntax to use (with or without dot).

On Arch Linux the lua libraries are loaded as lua or lua52, but lua5.2 produces the error below (pkg-config):
```
Package lua5.2 was not found in the pkg-config search path.
Perhaps you should add the directory containing `lua5.2.pc'
to the PKG_CONFIG_PATH environment variable
Package 'lua5.2', required by 'virtual:world', not found
build/compile.mk:107: *** library not found: lua5.2.  Stop.
```

### DIFF
--- a/build/lua.mk
+++ b/build/lua.mk
@@ -6,7 +6,9 @@ ifeq ($(USE_THIRDPARTY_LIBS),y)
 LIBLUA_LDLIBS = -llua
 LIBLUA_CPPFLAGS =
 else
-$(eval $(call pkg-config-library,LIBLUA,lua5.2))
+TARGET_LUAVER=5.2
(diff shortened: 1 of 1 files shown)


MaxKellermann:
You can't just call the host's `pkg-config`. This breaks the cross build.
Also this causes redundant `pkg-config` calls, which slows down the build further.

sbrodehl:
Yes, I agree.

kerel-fs:
@sbrodehl Did you find another solution to compile XCSoar on archlinux without applying this patch?

sbrodehl:
I hacked a few things together, so the answer is „yes“.
But I don’t have a clean solution at the moment.

kerel-fs:
Ok, I see.
After applying your patch I finally managed to compile XCSoar on/for archlinux, so this simplifies testing for me. Thanks! :)

As it is possible to build pre-/releases for our supported targets using virtualbox or docker this issue is of rather low priority I guess.

PR #226, opened 2019-06-30
Added CoPilot to logger settings

Luap-Paul:
Added a field for an copilot to the logger settings, and expanded the LX Driver, so the declaration for the nano includes the copilot.

### DIFF
--- a/NanoDeclare.cpp
+++ b/NanoDeclare.cpp
@@ -1,0 +1,165 @@
+/*
+Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2016 The XCSoar Project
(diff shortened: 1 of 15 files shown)


lordfolken:
Thanks for the effort! Unfortunately like this its not possible to merge this change. We would need only the differences that you made to the original source. These are entire files. This way its very hard to identify future bugs and what caused them. 

Also we require an entry into the NEWS.txt file. 

Have a look at rebasing and try to clean up the patches. 

https://www.geeksforgeeks.org/ultimate-guide-git-github/

is a guide. 

If you need help, we are on freenode channel #xcsoar

Luap-Paul:
Thank you for your effort, and the help! Sorry for my late reply. Im working on it, but can take a while because recently i dont have that much time. :)

PR #236, opened 2019-08-10
Update fr.po

scls19fr:
Minor translation fixes
First letter upper cased
Typography fixes (space before ? or !)
Plural

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -52,7 +52,7 @@ msgstr "Aller-retour FAI"
 #: src/Task/TypeStrings.cpp:34
 #, no-c-format
 msgid "FAI goal"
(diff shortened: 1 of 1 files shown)


scls19fr:
I wonder if services such as 

Transifex https://www.transifex.com
POEditor https://poeditor.com/
Crowdin https://crowdin.com/

shouldn't be used

it could help to have translation synchronised and to reduce the technical knowledges to provide more translations (more languages).

So users with not Git/Github experience could contribute more easily.

lordfolken:
We had a translation service in the past. If you can set up something that integrates nicely into github, it would be more than welcome.

PR #237, opened 2019-08-10
Update XCSoar-in-a-flash.tex

### LINKED ISSUE #235: Sim mode - define initial position
Hello,

Reading sim mode it's unclear to me how to define initial position (using GPS position or using a waypoint).
Is there a config file to modify? (to set initial position, but also maybe to set some others constants such as max cruise speed in sim mode ...)

Kind regards

scls19fr:
Closes #235


lordfolken:
I like this improvement with the reference. Please also add yourself to the AUTHORS file in the documentation section.

PR #239, opened 2019-08-12
wolfSSL

### LINKED ISSUE #222: Implement WolfSSL Library
The libcurl we currently ship is not built with ssl support. This causes several issues for downloading files and communication in general. 

https://www.wolfssl.com/
The wolfSSL embedded SSL library is a lightweight, portable, C-language-based SSL/TLS library targeted at IoT, embedded, and RTOS environments primarily because of its size, speed, and feature set. It works seamlessly in desktop, enterprise, and cloud environments as well. wolfSSL supports industry standards up to the current TLS 1.3 and DTLS 1.2, is up to 20 times smaller than OpenSSL, offers a simple API, an OpenSSL compatibility layer, OCSP and CRL support, is backed by the robust wolfCrypt cryptography 

The library is GPL 2.0 Licensed so including it would not be a problem.
https://www.wolfssl.com/license/

The version of libcurl we currently use has wolfssl support:
https://www.wolfssl.com/wolfssl-build-option-in-curl/

Issues that would be solved with this implementation:
- https://github.com/XCSoar/xcsoar-data-repository/issues/73
- #119 
- https://github.com/XCSoar/XCSoar/pull/182

dspreitz:
This is my first attempt to fix #222 
- Updated libs.py as suggested by @lordfolken 
- Updated compiling instructions in docu

The patch builds on my ubuntu 18.04 LTS system for the default UNIX target.
However, I am a bit clueless how to test if SSL is now supported in XCS. Pls advice.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -56,6 +56,7 @@ The following people and organizations have contributed code to XCSoar:
  Bruno de Lacheisserie <bruno.de.lacheisserie@gmail.com>
  Peter F Bradshaw <pfb@exadios.com>
  Stefan Schumann <stefan.schumann@posteo.de>
+ Dominic Spreitz <dominic@spreitz.de>
 
 Documentation:
 
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -62,6 +62,8 @@ Version 7.0 - not yet released
   - drop support for ARMv6 and MIPS CPUs
 * Kobo
   - support Kobo Glo HD
+* Added SSL support through wolfSSL
+  - added support for SSL secured connections e.g. for download manager
 
 Version 6.8.13 - not yet released
 
--- a/build/python/build/libs.py
+++ b/build/python/build/libs.py
@@ -127,7 +127,7 @@ curl = CurlProject(
         '--disable-manual',
         '--disable-threaded-resolver', '--disable-verbose', '--disable-sspi',
         '--disable-crypto-auth', '--disable-ntlm-wb', '--disable-tls-srp', '--disable-cookies',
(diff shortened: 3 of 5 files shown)


dspreitz:
I am a bit surprised thiw compiles in the CI. How did the CI know about the updated build dependencies for wolfSSL?

lordfolken:
Really cool. :) Could you add yourself to AUTHORS and add an entry to NEWS.txt under the KOBO section?

dspreitz:
I just checked (using wireshark) if the new wolfSSL libs are actually being used by XCS when downloading the German airspace from the DAEC server. Indeed the airspace file is being downloaded using TLS encryption. So it seems to work as expected.

lordfolken:
+1 awesome. Could you please add the new build dependencies also to the dev manual?

dspreitz:
The dev manual is already updated in commit fc9d5ca. I should have been more explicit in the commit message, sorry.

MaxKellermann:
I don't understand this PR. It adds an option to the CURL cross-build configuration to enable its wolfSSL support, but the documentation says you need a native wolfSSL build. But a cross-built CURL cannot possibly use a native wolfSSL binary. How is this supposed to work?

MaxKellermann:
No reply. I'll close this PR, hoping that this will induce some action to either explain the code or fix it.

kerel-fs:
Implicitly mentioned in the previous comments, a summary of the TODOs required to get libcurl with SSL/TLS support:
- Add Makefile/python scripts to cross-compile wolfSSL for all targets for which libcurl gets cross-compiled already (all/most of darwin, pc, android, kobo)
- Modify libcurl cross-compilation flags to build against wolfSSL (done by this PR)

Correct?

PR #240, opened 2019-08-13
Fix Android build environment

kerel-fs:

PR #241, opened 2019-08-14
Add gitlab-CI

kerel-fs:
This PR is depends on #240.

todo:
- [x] Fix KOBO target (build log: https://gitlab.com/kerel-fs/xcsoar/-/jobs/271564405 )
- [x] Add upload of build artifacts to download.xcsoar.org

edit:
20190815: fixed ToDos.

### DIFF
--- a/ide/provisioning/install-debian-packages.sh
+++ b/ide/provisioning/install-debian-packages.sh
@@ -62,7 +62,7 @@ apt-get install $APTOPTS g++-mingw-w64
 echo
 
 echo Installing dependencies for the Android target, not including SDK / NDK...
-apt-get install $APTOPTS openjdk-8-jdk-headless vorbis-tools adb
+apt-get install $APTOPTS openjdk-8-jdk-headless vorbis-tools adb unzip
 echo
 
 echo Clean up downloaded resources in order to free space
--- a/ide/provisioning/install-debian-packages.sh
+++ b/ide/provisioning/install-debian-packages.sh
@@ -3,9 +3,10 @@
 set -e
 
 export DEBIAN_FRONTEND=noninteractive
+APTOPTS="--assume-yes --no-install-recommends"
 
 echo Installing base dependencies...
-apt-get --assume-yes install make \
+apt-get install $APTOPTS make \
   librsvg2-bin xsltproc \
   imagemagick gettext ffmpeg \
   git quilt zip \
@@ -14,7 +15,7 @@ apt-get --assume-yes install make \
 echo
 
 echo Installing Manual dependencies...
-apt-get --assume-yes install texlive \
+apt-get install $APTOPTS texlive \
   texlive-latex-extra \
   texlive-luatex \
   texlive-lang-french \
@@ -25,7 +26,7 @@ apt-get --assume-yes install texlive \
 echo
 
 echo Installing dependencies for the Linux target...
-apt-get --assume-yes install make g++ \
+apt-get install $APTOPTS make g++ \
   zlib1g-dev \
   libfreetype6-dev \
   libpng-dev libjpeg-dev \
@@ -41,21 +42,21 @@ apt-get --assume-yes install make g++ \
 echo
 
 echo Installing dependencies for compiling with LLVM / Clang...
-apt-get --assume-yes install llvm clang libc++-dev
+apt-get install $APTOPTS llvm clang libc++-dev
 echo
 
 echo Installing dependencies for compiling targets which need libinput or GBM...
-apt-get --assume-yes install libinput-dev libgbm-dev
+apt-get install $APTOPTS libinput-dev libgbm-dev
 echo
 
 echo Installing dependencies for ARM Linux targets...
-apt-get --assume-yes install g++-arm-linux-gnueabihf
+apt-get install $APTOPTS g++-arm-linux-gnueabihf
 echo
 
 echo Installing PC/WIN64 dependencies...
-apt-get --assume-yes install g++-mingw-w64
+apt-get install $APTOPTS g++-mingw-w64
 echo
 
 echo Installing dependencies for the Android target, not including SDK / NDK...
-apt-get --assume-yes install default-jdk-headless vorbis-tools adb
+apt-get install $APTOPTS default-jdk-headless vorbis-tools adb
 echo
--- a/ide/provisioning/install-debian-packages.sh
+++ b/ide/provisioning/install-debian-packages.sh
@@ -41,7 +41,8 @@ apt-get install $APTOPTS make g++ \
   librsvg2-bin xsltproc \
   imagemagick gettext \
   mesa-common-dev libgl1-mesa-dev libegl1-mesa-dev \
-  fonts-dejavu
+  fonts-dejavu \
+  xz-utils
 echo
 
 echo Installing dependencies for compiling with LLVM / Clang...
@@ -61,7 +62,7 @@ apt-get install $APTOPTS g++-mingw-w64
 echo
 
 echo Installing dependencies for the Android target, not including SDK / NDK...
-apt-get install $APTOPTS default-jdk-headless vorbis-tools adb
+apt-get install $APTOPTS openjdk-8-jdk-headless vorbis-tools adb
 echo
 
 echo Clean up downloaded resources in order to free space
--- a/ide/provisioning/install-debian-packages.sh
+++ b/ide/provisioning/install-debian-packages.sh
@@ -2,6 +2,8 @@
 
 set -e
 
+export DEBIAN_FRONTEND=noninteractive
+
 echo Installing base dependencies...
 apt-get --assume-yes install make \
   librsvg2-bin xsltproc \
--- a/ide/provisioning/install-debian-packages.sh
+++ b/ide/provisioning/install-debian-packages.sh
@@ -62,7 +62,7 @@ apt-get install $APTOPTS g++-mingw-w64
 echo
 
 echo Installing dependencies for the Android target, not including SDK / NDK...
-apt-get install $APTOPTS openjdk-8-jdk-headless vorbis-tools adb unzip
+apt-get install $APTOPTS openjdk-8-jdk-headless vorbis-tools adb libtool unzip
 echo
 
 echo Clean up downloaded resources in order to free space
--- a/ide/provisioning/install-debian-packages.sh
+++ b/ide/provisioning/install-debian-packages.sh
@@ -5,6 +5,9 @@ set -e
 export DEBIAN_FRONTEND=noninteractive
 APTOPTS="--assume-yes --no-install-recommends"
 
+echo Updating Repositories
+apt-get update
+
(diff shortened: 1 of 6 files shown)


kerel-fs:
All CI builds finish successfully, see https://gitlab.com/kerel-fs/xcsoar/pipelines/76549132

The only untested part of these patches is the deployment of nightlies using gitlab-CI to download.xcsoar.org. This will require a gitlab repository which mirrors xcsoar/xcsoar and correctly configured CI variables.

MaxKellermann:
This PR contains unrelated commits from your other PR

kerel-fs:
Correct. Fortunately you merged the branch of my other PR using fast-forward, so it's not a problem (otherwise I would've rebased this branch).

ide/provisioning/install-debian-packages.sh
```diff
@@ -2,8 +2,12 @@
 
 set -e
 
+# To prevent apt from displaying dialogs
+DEBIAN_FRONTEND=noninteractive
```
MaxKellermann: Unrelated change, and one without effect.

ide/provisioning/install-debian-packages.sh
```diff
@@ -2,8 +2,12 @@
...
+APTOPTS="--no-install-recommends --assume-yes"
+
 echo Installing base dependencies...
-apt-get --assume-yes install make \
+apt-get install $APTOPTS make \
```
MaxKellermann: Your commit not only adds `--no-install-recommends`, but also refactors the existing `--assume-yes` into a global variable. This makes the commit obscure.

ide/provisioning/install-debian-packages.sh
```diff
@@ -13,9 +13,10 @@ echo Installing base dependencies...
 apt-get install $APTOPTS make \
   librsvg2-bin xsltproc \
   imagemagick gettext ffmpeg \
-  git quilt zip \
+  git quilt unzip zip \
```
MaxKellermann: What for?

ide/provisioning/install-debian-packages.sh
```diff
@@ -13,9 +13,10 @@ echo Installing base dependencies...
...
   m4 automake wget \
-  ttf-bitstream-vera fakeroot
+  ttf-bitstream-vera fakeroot \
+  ccache
```
MaxKellermann: What for?

ide/provisioning/install-debian-packages.sh
```diff
@@ -42,11 +43,12 @@ apt-get install $APTOPTS make g++ \
   librsvg2-bin xsltproc \
   imagemagick gettext \
   mesa-common-dev libgl1-mesa-dev libegl1-mesa-dev \
-  fonts-dejavu
+  fonts-dejavu \
+  libnetcdf-dev
```
MaxKellermann: What for?

ide/provisioning/install-debian-packages.sh
```diff
@@ -42,11 +43,12 @@ apt-get install $APTOPTS make g++ \
...
 echo
 
 echo Installing dependencies for compiling with LLVM / Clang...
-apt-get install $APTOPTS llvm clang libc++-dev
+apt-get install $APTOPTS llvm clang-4.0 libc++-dev
```
MaxKellermann: Why hard-code the clang version? This is fragile, and 4.0 is an ancient version.

MaxKellermann:
Sigh, this PR adds yet another copy of dependency and build commands (to `ide/docker`). Note that I will not take care for keeping those up to date. If I ever push commits, I will not check if they break anything here. This is cumbersome and fragile.

ide/provisioning/install-debian-packages.sh
```diff
@@ -44,11 +44,12 @@ apt-get install $APTOPTS make g++ \
...
   mesa-common-dev libgl1-mesa-dev libegl1-mesa-dev \
   fonts-dejavu \
-  libnetcdf-dev
+  libnetcdf-dev \
+  xz-utils
 echo
 
 echo Installing dependencies for compiling with LLVM / Clang...
-apt-get install $APTOPTS llvm clang-4.0 libc++-dev
+apt-get install $APTOPTS llvm clang libc++-dev
```
MaxKellermann: Oh, now we're back to plain unversioned `clang`. Why submit a PR changing it forth and back?

ide/provisioning/install-debian-packages.sh
```diff
@@ -64,7 +65,7 @@ apt-get install $APTOPTS g++-mingw-w64
 echo
 
 echo Installing dependencies for the Android target, not including SDK / NDK...
-apt-get install $APTOPTS default-jdk-headless vorbis-tools adb libtool
+apt-get install $APTOPTS openjdk-8-jdk-headless vorbis-tools adb libtool
```
MaxKellermann: Hard-coding OpenJDK 8 will break building on Buster, which doesn't have 8.

ide/provisioning/install-debian-packages.sh
```diff
@@ -68,10 +68,6 @@ echo Installing dependencies for the Android target, not including SDK / NDK...
 apt-get install $APTOPTS openjdk-8-jdk-headless vorbis-tools adb libtool
 echo
 
-echo Installing dependencies for KOBO
-apt-get --assume-yes install libav-tools
```
MaxKellermann: This line was added in this very PR, and now remove it again? Why add it in the first place?

MaxKellermann:
This PR doesn't make me happy. It is obscure, underdocumented, changes things forth and back, creeps in unrelated changes hidden in other commits.

kerel-fs:
Thanks for your review!
I don't want to make you unhappy. :(
My intention is to automatically build as many targets as possible using a unmodified xcsoar source tree. I'll try to address all your annotations.

ide/provisioning/install-debian-packages.sh
kerel-fs: Fixed and moved into a separate commit.

ide/provisioning/install-debian-packages.sh
kerel-fs: Moved into a separate commit.

ide/provisioning/install-debian-packages.sh
kerel-fs: Removed, a left-over I didn't tidy up (introduced without explanation in https://github.com/lordfolken/xcsoar-docker-build/pull/1).

ide/provisioning/install-debian-packages.sh
kerel-fs: The upcoming Skysight integration will add this requirement, but as long as it's not merged we don't need it. Removed.

ide/provisioning/install-debian-packages.sh
kerel-fs: Sorry, yet another left-over from xcsoar-docker-build (using debian stretch). Removed.

ide/provisioning/install-debian-packages.sh
kerel-fs: > This commit incorporates all changes to install-debian-packages.sh
> available in the current master branch (bb41583) of
> xcsoar-docker-build [1].

As explained in the commit message c019eecc, I based these patches on lordfolken/xcsoar-docker-build@bb41583. Thus I first incorporated all changes in this repository and then added my patches.

As this clutters the xcsoar git history (and is harder to review, sry) I edited my patchset accordingly now.

ide/provisioning/install-debian-packages.sh
kerel-fs: See https://github.com/XCSoar/XCSoar/pull/241#discussion_r314697580. Removed.

ide/provisioning/install-debian-packages.sh
kerel-fs: `ide/provisioning/install-android-tools.sh` uses `unzip` to extract the Android NDK files. Moved into a separate commit.

MaxKellermann:
You did not fix my complaint about the OpenJDK 8 switch;

> The currently used Android SDK Plaform Android 8.0 / API level 26 requires OpenJDK 8 [1].

Does it? I don't have OpenJDK 8, only 11, and I can build XCSoar. How come?

kerel-fs:
> You did not fix my complaint about the OpenJDK 8 switch;

Correct, since I'm not finished updating this branch. :)

kerel-fs:
> Does it? I don't have OpenJDK 8, only 11, and I can build XCSoar. How come?

Interesting, this is good news! I based my assumption on https://source.android.com/setup/build/older-versions#jdk which explicitly states that Android 8.0 (which is our target using android platform level 26) requires OpenJDK 8. I'll give OpenJDK another try (previously it failed).

kerel-fs:
Using OpenJDK 11 the installation of the Android SDK using the sdkmanager fails.
https://github.com/XCSoar/XCSoar/blob/3cf15face691dc27985abaf555a16c2f099d7ed8/ide/provisioning/install-android-tools.sh#L36-L38

Error log:
```
Installing Android SDK packages...
Exception in thread "main" java.lang.NoClassDefFoundError: javax/xml/bind/annotation/XmlSchema
	at com.android.repository.api.SchemaModule$SchemaModuleVersion.<init>(SchemaModule.java:156)
	at com.android.repository.api.SchemaModule.<init>(SchemaModule.java:75)
	at com.android.sdklib.repository.AndroidSdkHandler.<clinit>(AndroidSdkHandler.java:81)
	at com.android.sdklib.tool.SdkManagerCli.main(SdkManagerCli.java:117)
	at com.android.sdklib.tool.SdkManagerCli.main(SdkManagerCli.java:93)
Caused by: java.lang.ClassNotFoundException: javax.xml.bind.annotation.XmlSchema
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:583)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
	... 5 more
```

Afaik this is caused by using an incompatible OpenJDK version.
Maybe you can explain to me how to install the Android SDK package using OpenJDK 11?
There is an ugly [workaround](https://stackoverflow.com/a/55982976) available, but I expect there are better ways to do it.

If we do not find a fix or workaround for this, I'd like to recommend to install OpenJDK 8 from Debian unstable as done in this PR to provide setup scripts for a working build environment.

kerel-fs:
>  Maybe you can explain to me how to install the Android SDK package using OpenJDK 11?

Answering the question myself:
The recommended way to install the Android SDK tools & platform is using the Android Studio (using a GUI).

Unfortunately for an unattended installation using the old `sdkmanager` CLI is still the recommended way, see [Update your tools with the SDK Manager
](https://developer.android.com/studio/intro/update#sdk-manager), so I guess we have to accept using OpenJDK 8 here.

PR #245, opened 2019-08-29
Add InfoBoxes for Radio settings

roeles:
This pull request adds two infoboxes:
- One to show the currently configured Active Frequency, with the name as comment
- One to show the currently configured Standby Frequency, with the name as comment


roeles:
I realized that this doesn't take into account when settings are changing on the remote device. I'll investigate that first, before someone else spends time on this.

PR #247, opened 2019-09-01
Radio frequency infoboxes

roeles:


src/Computer/Settings.hpp
```diff
@@ -122,6 +123,23 @@ struct PlacesOfInterestSettings {
...
+  RadioFrequency active_frequency;
+  RadioFrequency standby_frequency;
+
+  StaticString<32> active_name;
+  StaticString<32> standby_name;
+
+  void SetDefaults() {
+    active_frequency.Clear();
+    standby_frequency.Clear();
+    active_name = "";
```
MaxKellermann: Use `clear()`, which sets one byte; this line calls `strcpy()` which is unnecessary overhead.

src/ActionInterface.cpp
```diff
@@ -291,3 +291,46 @@ ActionInterface::SendUIState()
...
   main_window->SetUIState(GetUIState());
 }
+
+void
+ActionInterface::SetActiveFrequency(const RadioFrequency & freq, const TCHAR * freq_name, bool to_devices)
+{
+  /* update interface settings */
+
+  SetComputerSettings().radio.active_frequency = freq;
+  if(freq_name != nullptr) {
```
MaxKellermann: If `freq_name==nullptr`, then the old name will persist - is that what you want?

src/InfoBoxes/Content/Radio.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include <tchar.h>
+
+void
+UpdateInfoBoxFrequency(InfoBoxData & data, const RadioFrequency & freq, const TCHAR * freq_name)
+{
+  if(freq.IsDefined()) {
+    TCHAR buffer[32];
+    freq.Format(buffer, 32);
+    data.SetValue(buffer);
```
MaxKellermann: Use `InfoBoxData::FormatValue()`

src/InfoBoxes/Content/Radio.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "InfoBoxes/Content/Radio.hpp"
+#include "InfoBoxes/Data.hpp"
+#include "Interface.hpp"
+#include "Units/Units.hpp"
+#include "Formatter/UserUnits.hpp"
+#include "Language/Language.hpp"
+
+#include <tchar.h>
+
+void
```
MaxKellermann: Why is this a public function?

src/InfoBoxes/Content/Radio.cpp
```diff
@@ -0,0 +1,63 @@
...
+UpdateInfoBoxFrequency(InfoBoxData & data, const RadioFrequency & freq, const TCHAR * freq_name)
+{
+  if(freq.IsDefined()) {
+    TCHAR buffer[32];
+    freq.Format(buffer, 32);
+    data.SetValue(buffer);
+    data.SetComment(freq_name);
```
MaxKellermann: This will crash if `freq_name==nullptr`. The parameter is not documented, and is not documented that `nullptr` is an invalid value. In another part of your code, `freq_name==nullptr` is a legal parameter. So this really needs clarification.

src/Computer/Settings.hpp
roeles: Used clear()

src/ActionInterface.cpp
roeles: Implemented freq_name.clear() in else block

src/InfoBoxes/Content/Radio.cpp
roeles: Implemented FormatValue and removed TCHAR buffer[32]

src/InfoBoxes/Content/Radio.cpp
roeles: Changed parameter to `const StaticString<32> &`. This does not allow nullptr, and matches the type that is read from ComputerSettings.
With the changes indicated above, nullptr is not a valid option in this part of the code.

src/InfoBoxes/Content/Radio.cpp
roeles: Removed from header and made static.

MaxKellermann:
Please amend the commits instead of adding a new commit.

roeles:
@MaxKellermann I'll do that, but it will be the first time I amend commits using git. If I screw things up, I'll open a new pull request.

Am I correct to assume that code-wise everything is okay, or is amending the commits a prerequisite for reviewing further?

roeles:
@MaxKellermann please let me know if this is okay, commit-wise.

src/InfoBoxes/Content/Radio.cpp
MaxKellermann: That's a bad API decision because it is extremely narrow.

roeles: Thanks for the response. I have expanded the length to 64, but if that too isn't satisfactory I kindly ask for some insights.
My reasoning is this:
1. I don't want to use a char *, because I can't guarantee that the name is still in memory when I change task/waypoints after setting the frequency. So I need a copy.
2. The CUP file format (for example) does not define a maximum length for waypoint names.
3. I think using a tstring is not a good option, since I am not sure that ComputerSettings would remain trivial. It would also imply copying from this tstring to the StaticString<32> defined in InfoBoxData.

Thus, since the only usage so far is the InfoBox, I decided that StaticString<32> would be best for now.

If any of the points above are invalid, please let me know.

MaxKellermann: About 1.: Your `StaticString<32>&` just as well doesn't guarantee that the name is still in memory.
2. and 3. are valid, but I have no idea how any of this (including 1.) is relevant for this API design.

roeles: Thanks for taking the time to respond.

I will look into 1, even though it might not be relevant.

I think I don't understand what you mean by "API decision" and "narrow" in this context. The function I think we are discussing is a static function, which in my view exists only because I don't want to repeat it's contents two times. I don't think it is part of any API.

I thought with "narrow" you meant the maximum size of 32, but I'm starting to doubt that now. Can you please clarify?

MaxKellermann: "Narrow" means the choice of types for this parameter is very narrow. Only `const StaticString<32>&` is allowed, even though there is no technical reason for this limitation. Previously, you had `const TCHAR *` which was just fine, and I don't know why you changed it. Any `StaticString` is implicitly convertibly to `const TCHAR *` and that without any overhead; but you can convert `const TCHAR *` to a `StaticString` only by allocating memory and copying data.

roeles: Okay, thanks for clarifying!

roeles: I've changed the parameter back to `const TCHAR *`. I believe there are still copies made during assignment in `ComputerSettings` and assignment of the `StaticString` inside `InfoBoxData`, but I currently don't see how to avoid those while keeping `ComputerSettings` trivial.

PR #251, opened 2019-09-18
LX Driver: Fixed possible buffer overflow

Exadios:
Fixes a problem where data that originates external to XCSoar could cause
a buffer overflow.


MaxKellermann:
80f9b0d345d9b7796e7391b0f01a216e400d0dca

PR #252, opened 2019-09-18
doc/manual/dev: Add libinput-dev dependency

kerel-fs:


kerel-fs:
Actually this dependency might be required for all UNIX targets except the Kobo or targets using wayland, see 87d501e and
https://github.com/XCSoar/XCSoar/blob/2a5719536f488a86ef49d87a113c71bdf5ed26a2/build/libevent_options.mk#L10-L17
I'll update this PR.

kerel-fs:
Updated the patch according to my previous comment here.

MaxKellermann:
985b8f06a4b75db0798aa85a38e34d2449682612

PR #253, opened 2019-09-18
WIP: Fix upload of nightlies to download.xcsoar.org

kerel-fs:
This would enable the upload of nightlies for every commit in the master branch to download.xcsoar.org.
Since this are ~6GB per commit currently, I propose to upload artifacts only for preview releases, but using DEBUG=n and reduced into a directory structure which resembles the structure known from the existing releases.


MaxKellermann:
@lordfolken I have no idea why `scp` should be changed to `rsync`, and I'd wish there was an explanation in the commit message - but I trust your judgement. If you think a PR is good, merge it. Don't depend on me.

PR #264, opened 2019-09-29
Update Active and Standby Radio Frequency setting from KRT2

roeles:
When receiving an STX frame (identical to the one we send), update the Computer Setting for either active or standby frequency.

### DIFF
--- a/src/Device/Driver/KRT2.cpp
+++ b/src/Device/Driver/KRT2.cpp
@@ -55,6 +55,15 @@ class KRT2Device final : public AbstractDevice {
 
   static constexpr size_t MAX_NAME_LENGTH = 8; //!< Max. radio station name length.
 
+  struct stx_msg {
+    uint8_t start = STX;
+    uint8_t command;
+    uint8_t mhz;
+    uint8_t khz;
+    char station[MAX_NAME_LENGTH];
+    uint8_t checksum;
+  };
+
   //! Port the radio is connected to.
   Port &port;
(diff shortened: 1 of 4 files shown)


roeles:
Crap, forgot to fix something. Will correct it.

src/Device/Driver/KRT2.cpp
```diff
@@ -374,21 +396,39 @@ KRT2Device::GetStationName(char *station_name, const TCHAR *name)
...
+    return;
+  }
+
+  if(msg->checksum != (msg->mhz ^ msg->khz)) {
+    return;
+  }
+
+  RadioFrequency freq;
+  freq.SetKiloHertz((msg->mhz * 1000) + (msg->khz * 5));
+  StaticString<MAX_NAME_LENGTH> freq_name(msg->station);
```
MaxKellermann: Buffer overflow bug here. You don't know if `msg->station` is null-terminated.

src/NMEA/ExternalSettings.hpp
```diff
@@ -72,6 +74,12 @@ struct ExternalSettings {
   /** the volume of the device [0-100%] */
   unsigned volume;
 
+  /** the radio frequencies of the device */
+  RadioFrequency active_frequency;
+  RadioFrequency standby_frequency;
+  StaticString<32> active_name;
```
MaxKellermann: Undocumented variable with a very generic name. Active name of what?

PR #267, opened 2019-09-30
Documents: Sans serif font are back

mlep:
For some time I was wondering why the manual in version 6.8.11 was in sans serif fonts (https://download.xcsoar.org/releases/6.8.11/XCSoar-manual.pdf) while the more recent 7.0 manual was in serif (https://download.xcsoar.org/nightly_builds/7.0_preview14~git%23923cf10d/manual/XCSoar-manual.pdf).

Removing the helvet package from the xcsoar style file solved the issue.
Moreover, the difference in font style when \emph{} is used is reestablished.

Note: the serif font in the table of contents is another issue (coming PR)

Before:
Manual-Before

After:
Manual-after

### DIFF
--- a/doc/manual/xcsoar.sty
+++ b/doc/manual/xcsoar.sty
@@ -12,7 +12,6 @@
 \makeatletter
 
 % Font selection
-%\usepackage{helvet}
 \renewcommand{\familydefault}{\sfdefault}
 \fontfamily{phv}\selectfont
 
--- a/doc/manual/xcsoar.sty
+++ b/doc/manual/xcsoar.sty
@@ -12,7 +12,7 @@
 \makeatletter
 
 % Font selection
-\usepackage{helvet}
+%\usepackage{helvet}
 \renewcommand{\familydefault}{\sfdefault}
 \fontfamily{phv}\selectfont
 


doc/manual/xcsoar.sty
```diff
@@ -12,7 +12,7 @@
 \makeatletter
 
 % Font selection
-\usepackage{helvet}
+%\usepackage{helvet}
```
MaxKellermann: What's the use of commenting this out instead of just removing it? If we want this back, we can look at the git history.

mlep: I hesitated between the two possibilities.
Now, commented line is deleted.

MaxKellermann:
Your commit message implies that the change is an "issue" that needs to be "solved", but doesn't explain it.

mlep:
The ("official") manual of version 6.8.11 was in Sans Serif font.
With the current compilation, there are 2 issues related to fonts:
- the manual is coming in Serif font
- the \emph command ("emph" for "emphasis") did not create an emphasis (typically text in bold or italic) See in the "before" example, the words "as well as".

This PR solves both issues (see example "after")

mlep:
Superseeded by PR #307.

PR #268, opened 2019-09-30
Doc: Table of contents: Better font and layout

mlep:
Even when the documents using the xcsoar.sty file where producing text in sans serif font (see PR #267), the titles in the table of content was in sans serif. For example: https://download.xcsoar.org/releases/6.8.11/XCSoar-manual.pdf .
Moreover, the section numbers were sometimes touching the section titles. 

With this PR, the titles in the table of contents uses the same font style (i.e. sans serif) as the rest of the document, and there is always a space between the section number and title.

Before:
TOC-before

After:
TOC-after

### DIFF
--- a/doc/manual/toc.tex
+++ b/doc/manual/toc.tex
@@ -1,7 +1,10 @@
 \begingroup
-\fontfamily{ptm}
-\normalsize
-\fontseries{c}\selectfont
+%\fontfamily{phv}
+%\normalsize
+%\fontseries{c}\selectfont
 \setlength{\parskip}{0.05\baselineskip}
+\makeatletter
+\renewcommand*\l@section{\@dottedtocline{2}{1.3em}{2.6em}}
+\makeatother
 \tableofcontents
 \endgroup
--- a/doc/manual/toc.tex
+++ b/doc/manual/toc.tex
@@ -1,7 +1,4 @@
 \begingroup
-%\fontfamily{phv}
-%\normalsize
-%\fontseries{c}\selectfont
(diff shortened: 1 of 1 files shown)


doc/manual/toc.tex
```diff
@@ -1,7 +1,10 @@
 \begingroup
-\fontfamily{ptm}
-\normalsize
-\fontseries{c}\selectfont
+%\fontfamily{phv}
+%\normalsize
+%\fontseries{c}\selectfont
```
MaxKellermann: Why comment out these lines, and why a modified version? This is misleading.

doc/manual/toc.tex
```diff
@@ -1,7 +1,10 @@
...
 \setlength{\parskip}{0.05\baselineskip}
+\makeatletter
+\renewcommand*\l@section{\@dottedtocline{2}{1.3em}{2.6em}}
```
MaxKellermann: This is obscure and looks like a kludge. Is there no better way to accomplish this?

mlep: A better alternative would be to use the package tocloft.
Can I introduce it?

doc/manual/toc.tex
mlep: Was due to preliminary testing. Commented lines are now removed.

doc/manual/toc.tex
lordfolken: Tocloft is part of the texlive-latex-extra package which we install in the provisioning scripts. Go for it.

mlep:
Closing because superseded by PR #308.

PR #274, opened 2019-10-04
Make cursor on framebuffer devices resizeable

hvennekate:
This is an attempt to implement a requested user feature, that is to have the cursor resizable on embedded devices (using framebuffer). Please review thoroughly -- I am doing this for the first time and am not sure whether I should also update/extend tests & docs.

### DIFF
--- a/src/Screen/FB/TopWindow.cpp
+++ b/src/Screen/FB/TopWindow.cpp
@@ -27,6 +27,7 @@ Copyright_License {
 #include "Event/Poll/Loop.hpp"
 #include "Event/Queue.hpp"
 #include "Event/Globals.hpp"
+#include "Interface.hpp"
 
 #ifdef DRAW_MOUSE_CURSOR
 #include "Util/Macros.hpp"
@@ -91,10 +92,14 @@ TopWindow::OnPaint(Canvas &canvas)
   /* draw the mouse cursor */
 
   const auto m = event_queue->GetMousePosition();
+  uint8_t cursorSize = CommonInterface::GetUISettings().display.cursor_size;
+  unsigned shortDistance = Layout::Scale(cursorSize * 4);
+  unsigned longDistance = Layout::Scale(cursorSize * 6);
+
   const BulkPixelPoint p[] = {
     { m.x, m.y },
-    { m.x + Layout::Scale(4), m.y + Layout::Scale(4) },
-    { m.x, m.y + Layout::Scale(6) },
+    { m.x + shortDistance, m.y + shortDistance },
+    { m.x, m.y + longDistance },
   };
 
   canvas.SelectBlackPen();
--- a/src/Screen/FB/TopWindow.cpp
+++ b/src/Screen/FB/TopWindow.cpp
@@ -27,7 +27,6 @@ Copyright_License {
 #include "Event/Poll/Loop.hpp"
 #include "Event/Queue.hpp"
 #include "Event/Globals.hpp"
-#include "Interface.hpp"
 
 #ifdef DRAW_MOUSE_CURSOR
 #include "Util/Macros.hpp"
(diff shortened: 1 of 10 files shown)


src/Screen/FB/TopWindow.cpp
```diff
@@ -27,6 +27,7 @@ Copyright_License {
 #include "Event/Poll/Loop.hpp"
 #include "Event/Queue.hpp"
 #include "Event/Globals.hpp"
+#include "Interface.hpp"
```
MaxKellermann: The "Screen" library should not include this header, because it's a higher-level collection of global variables (aka "shit code needs to live somewhere, and this header it is") - layering violation.
Maybe the cursor size setting should be copied to a `TopWindow` variable?

test/README
```diff
@@ -10,7 +10,7 @@ References:
 	- http://en.wikipedia.org/wiki/Test_Anything_Protocol
 
 What is in a number?
-	0*	Test the test system itself - psudo tests
+	0*	Test the test system itself - pseudo tests
```
MaxKellermann: Sneaking in a typo fix in an unrelated commit?

hvennekate: Well, sort of, yes, my apologies. This was originally not intended for public release and I just cherry-picked the relevant commit and did not check it again. Is this ok, though?

src/Screen/FB/TopWindow.cpp
hvennekate: Sorry, my mistake. Guess I'll have to check out the architecture. Also: the include should probably be tied to the pre-compiler conditional as well, i suppose.

hvennekate: At that: Is it considered better style to have both includes and code in the same pre-compiler conditional or separate? (I'm looking at `USE_FB` in the same file, where both seem to be kept in the same conditional bracket)

MaxKellermann: I don't understand the question.

The "architecture" is not documented, no need to be sorry.
The `Screen` library (the low-level UI toolkit which higher levels like `Form` build on top of) should be generic and not depend on XCSoar's specific libraries and global variables. If some global configuration needs to be known by the `Screen` library, it should be passed during initialization.
There are still some historic layering violations; after many years of code cleanup, it's far from perfect. But I'd like new code to be as good as possible, while still trying to improve the old code.

test/README
MaxKellermann: Typo fixes are always welcome - a patch is worth merging if the project is better after that. It is just confusing to review a patch if it contains several unrelated things.

src/Screen/FB/TopWindow.cpp
hvennekate: Ok, I'll try my best (well, maybe I'm just a little confused here). Added a commit where the cursor size is an attribute of `TopWindow`. Since I would like to instantly apply the settings change (i.e. not require the user to restart), I updated the value in the `LayoutConfigPanel`. Also the update encompasses the more generic `DisplaySettings` as a whole. Let me know if this is not acceptable.

hvennekate:
> Before you push, fold the corrections into first commit.

Sorry, I'm new to this: What precisely do you mean? squash/fixup and force-push? Or cherry-pick to a single commit on a different branch? (I'm not entirely sure how github handles force-pushes -- Atlassian Crucible usually gets confused by those which is why I usually avoid them)

src/Screen/TopWindow.hpp
```diff
@@ -47,6 +47,7 @@ union SDL_Event;
 struct SDL_Window;
 #endif
 
+#include "DisplaySettings.hpp"
```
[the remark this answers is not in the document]
hvennekate: Well, the approach I chose seems a bit overblown in retrospect: changed to only update the cursor size directly. (The problem here is that the function's definition is also in the header file...)

MaxKellermann:
I meant fold and force-push. Keep separate things in separate commits, but if one commit is just a fixup for another unmerged commit, they should be folded. GitHub can deal with force-pushes just fine - and in your private repository, you can force-push as much as you like. Of course, official branches should never be force-pushed.

I've done the folding for you: ff3367a400f08b68cf1658e06b0c2b687729ea2b
Thanks and looking forward to more contributions.

PR #285, opened 2019-11-11
Fixed trailing space after \xc command

CptFrikadel:
When using the `\xc` shortcut latex skips the typed white-space that comes after it. So I added a control space `\␣`.


CptFrikadel:
I should probably have clarified: It's chapter 8 of the dev manual.

Issue #234, opened 2019-08-09
Scale is not displayed in panorama mode


scls19fr:
Hello,

I'm facing issue with panorama mode... scale is not displayed (see screenshot).
I'm using a Huawei P30 Lite (6.15" - 2312 x 1080 resolution - 415 DPI).

![IMG_20190809_113708](https://user-images.githubusercontent.com/109167/62770829-23cfdb80-ba9c-11e9-854e-e24c83f0525d.jpg)

On a much bigger screen (Acer Chromebook CB5-132T-C8VM with 11" screen) scale is displayed correctly.

![IMG_20190809_114827](https://user-images.githubusercontent.com/109167/62770868-36e2ab80-ba9c-11e9-8218-1d686de85613.jpg)

Scale should never be hidden by buttons.

Kind regards

lordfolken:
Yes, please write a bugfix.

scls19fr:
I currently haven't dive into XCSoar code previously but I have some Java / Kotlin skills... 
So could you point me where is the code which manage scale feature?

lordfolken:
have a look into `src/Renderer/MapScaleRenderer.cpp`

scls19fr:
I'm wondering what are the reasons of this `3` factor in

https://github.com/XCSoar/XCSoar/blob/b2ae619d507419cd92f0fb629dc078a5b36be9fe/src/Renderer/MapScaleRenderer.cpp#L51

I have been able to reproduce this issue on my Macbook, so I wanted to use it as a testing environment... but wasn't able to compile XCSoar.

```
$ make
```

raises

```
configure: error: You will need to substantially rewrite libxtiff to
build libgeotiff without libtiff
Traceback (most recent call last):
  File "./build/thirdparty.py", line 140, in <module>
    x.build(toolchain)
  File "/Users/scls/github/scls19fr/XCSoar/build/python/build/autotools.py", line 99, in build
    build = self.configure(toolchain)
  File "/Users/scls/github/scls19fr/XCSoar/build/python/build/autotools.py", line 86, in configure
    subprocess.check_call(configure, cwd=build, env=toolchain.env)
  File "/Users/scls/anaconda3/lib/python3.7/subprocess.py", line 347, in check_call
    raise CalledProcessError(retcode, cmd)
[...]

PR #289, opened 2019-11-12
Fix rendering of map scale in Pan mode (#234)

CptFrikadel:
Fixed issue #234. Map scale and mode indicator now render above the buttons when panning in portrait mode.

### DIFF
--- a/src/MapWindow/GlueMapWindowOverlays.cpp
+++ b/src/MapWindow/GlueMapWindowOverlays.cpp
@@ -271,10 +271,20 @@ void
 GlueMapWindow::DrawMapScale(Canvas &canvas, const PixelRect &rc,
                             const MapWindowProjection &projection) const
 {
-  RenderMapScale(canvas, projection, rc, look.overlay);
 
-  if (!projection.IsValid())
-    return;
+  PixelRect scaleRef(rc.left, rc.top, rc.right, rc.bottom);
+
+  /*
+    If buttons are drawn on bottom (Pan mode and portrait);
+    Change reference for rendering so scale is drawn above.
+  */
+  if (IsPanning() && !Layout::landscape)
+    scaleRef.bottom -= rc.GetHeight() / 6; // Scale factor as in MenuBar.cpp
+
+  RenderMapScale(canvas, projection, scaleRef, look.overlay);
+
+    if (!projection.IsValid())
+      return;
 
   StaticString<80> buffer;
 
@@ -327,6 +337,9 @@ GlueMapWindow::DrawMapScale(Canvas &canvas, const PixelRect &rc,
     TextInBoxMode mode;
     mode.vertical_position = TextInBoxMode::VerticalPosition::ABOVE;
     mode.shape = LabelShape::OUTLINED;
+
+    if (IsPanning() && !Layout::landscape)
+      y -= rc.GetHeight() / 6; // Scale factor as in MenuBar.cpp
 
     TextInBox(canvas, buffer, 0, y, mode, rc, nullptr);
   }
--- a/src/MapWindow/GlueMapWindowOverlays.cpp
+++ b/src/MapWindow/GlueMapWindowOverlays.cpp
@@ -42,6 +42,7 @@ Copyright_License {
 #include "Look/GestureLook.hpp"
 #include "Input/InputEvents.hpp"
 #include "Renderer/MapScaleRenderer.hpp"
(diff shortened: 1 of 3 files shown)


src/Renderer/MapScaleRenderer.cpp
```diff
@@ -28,6 +28,7 @@
 #include "Look/OverlayLook.hpp"
 #include "Util/StaticString.hxx"
 #include "Formatter/UserUnits.hpp"
+#include "Pan.hpp"
```
MaxKellermann: This is a layering violation. The MapScaleRenderer is generic code and shouldn't depend on this library.

MaxKellermann:
This PR contains a known-bad commit which a following commit reverts. Please remove this known-bad commit from the PR instead of reverting it.

src/Renderer/MapScaleRenderer.cpp
CptFrikadel: I changed it so that now it offsets the reference PixelRect from GlueMapwindow's DrawMapScale. The MapScaleRenderer is now the same is it was before.

CptFrikadel:
This branch has indeed become a quite a mess. Should I make a new less messy PR from a new and clean branch?

MaxKellermann:
This has nothing to do with this PR. You can clean up your branch and force-push it to your public GitHub fork. And yes, that would be necessary for a proper code review.

CptFrikadel:
Like this? Or should I also remove the first other two commits?

MaxKellermann:
Without looking at the code: this PR now contains 3 commits with nearly identical commit messages. That doesn't look right. A PR should only contain known-good commits. It's good to have separate things in separate commits, but your commits don't look like they do separate things.

CptFrikadel:
Ok, I think it's correct now..

src/MapWindow/GlueMapWindowOverlays.cpp
```diff
@@ -271,10 +271,20 @@ void
...
-  if (!projection.IsValid())
-    return;
+  PixelRect scaleRef(rc.left, rc.top, rc.right, rc.bottom);
+
+  /*
+    If buttons are drawn on bottom (Pan mode and portrait);
+    Change reference for rendering so scale is drawn above.
+  */
+  if (IsPanning() && !Layout::landscape)
+    scaleRef.bottom -= rc.GetHeight() / 6; // Scale factor as in MenuBar.cpp
```
MaxKellermann: Don't duplicate this formula. If somebody ever changes the button layout in `MenuBar.cpp`, this will break. This is very fragile code!

src/MapWindow/GlueMapWindowOverlays.cpp
```diff
@@ -271,10 +271,20 @@ void
...
+
+  RenderMapScale(canvas, projection, scaleRef, look.overlay);
+
+    if (!projection.IsValid())
```
MaxKellermann: Bad whitespace change.

src/MapWindow/GlueMapWindowOverlays.cpp
CptFrikadel: I changed that in the new commit. It does however introduce ```MenuBar.hpp``` as a new dependency and makes the ```GetButtonPosition(..)``` no longer static.

MaxKellermann:
Your approach is still fragile, because `GlueMapWindow` doesn't know when panning has been enabled and will not trigger a redraw. This just appears to work, because there are enough other triggers for redraws, but it isn't correct.
Instead of calling into `MenuBar`, I'd prefer to have, for example, a method like `GlueMapWindow::SetBottomMargin(unsigned)` which tells `GlueMapWindow` to not put any HUD features in that area.
That way, `GlueMapWindow` can decide whether a redraw must be triggered, but is still decoupled from the `MenuBar`.

CptFrikadel:
Apologies for the long silence. I've looked at it again, but I don't think I understand what you mean. The only times a button might be drawn on top of HUD features is when pan mode has been entered or when the screen is resized. On both occasions a redraw is already triggered and the scale is drawn correctly because the decision whether or not to offset happens inside `DrawMapScale`. 
The call into `MenuBar` is unavoidable as that is the only way to obtain its size. I don't understand how having `GlueMapWindow::SetBottomMargin()` would fix this as now the routine creating the buttons (`InputEvents::drawButtons()`) has to have a reference to the `GlueMapWindow`.

PR #294, opened 2019-11-26
Cursor improvements

hvennekate:
Some more improvements after I discussed the cursor with the actual users ;)

### DIFF
--- a/doc/manual/en/ch11_configuration.tex
+++ b/doc/manual/en/ch11_configuration.tex
@@ -663,7 +663,7 @@ This page once more details the appearance of the graphical user interface of XC
   {\bf Box}: Draws boxes around each InfoBox. \\
   {\bf Tab}: Draws a tab at the top of the InfoBox across the 
   title. \\
-\item[Cursor size*]  Control the size of the cursor. Default is 2.
+\item[Cursor size]  Control the size of the cursor. Default is 1.
   Only available if cursor is drawn by XCSoar.
 \item[Black cursor] Use black instead of white cursor. Default is {\bf Off} (white cursor).
   Only available if cursor is drawn by XCSoar.
--- a/doc/manual/en/ch11_configuration.tex
+++ b/doc/manual/en/ch11_configuration.tex
@@ -663,6 +663,10 @@ This page once more details the appearance of the graphical user interface of XC
   {\bf Box}: Draws boxes around each InfoBox. \\
   {\bf Tab}: Draws a tab at the top of the InfoBox across the 
   title. \\
+\item[Cursor size*]  Control the size of the cursor. Default is 2.
+  Only available if cursor is drawn by XCSoar.
+\item[Black cursor] Use black instead of white cursor. Default is {\bf Off} (white cursor).
+  Only available if cursor is drawn by XCSoar.
 \end{description}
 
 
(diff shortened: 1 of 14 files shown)


doc/manual/en/ch11_configuration.tex
```diff
@@ -663,6 +663,10 @@ \section{Look / Screen Layout}\label{sec:interface-appearance}
   {\bf Box}: Draws boxes around each InfoBox. \\
   {\bf Tab}: Draws a tab at the top of the InfoBox across the 
   title. \\
+\item[Cursor size*]  Control the size of the cursor. Default is 2.
```
hvennekate: Question: do we document features that depend on the build target?

po/ca.po
```diff
@@ -5722,7 +5722,7 @@ msgstr ""
 
 #: src/Dialogs/Settings/Panels/AirspaceConfigPanel.cpp:69
 #: src/Dialogs/Settings/Panels/InterfaceConfigPanel.cpp:169
-#: src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp:60
```
hvennekate: Forgot to update those the last time around...

src/Screen/TopWindow.hpp
```diff
@@ -144,6 +144,7 @@ class TopWindow : public ContainerWindow {
 #endif
 #ifdef DRAW_MOUSE_CURSOR
   uint8_t cursor_size = 1;
```
hvennekate: This should really be "2"

src/Screen/FB/TopWindow.cpp
```diff
@@ -91,17 +91,23 @@ TopWindow::OnPaint(Canvas &canvas)
...
 
   const BulkPixelPoint p[] = {
     { m.x, m.y },
     { m.x + shortDistance, m.y + shortDistance },
     { m.x, m.y + longDistance },
   };
 
-  canvas.SelectBlackPen();
-  canvas.SelectWhiteBrush();
+  unsigned penSize = cursor_size / 2 + cursor_size % 2;
```
hvennekate: The poor man's version of rounding up...

MaxKellermann:
Updates to po/* should not be part of a PR, because this can break easily and cause conflicts.
New Canvas methods should be a separate commit.
Cursor size change and inverted colors should be separate commits. It is hard to review a commit when you do two different things mixed in one commit.

hvennekate:
> Updates to po/* should not be part of a PR, because this can break easily and cause conflicts.
> New Canvas methods should be a separate commit.
> Cursor size change and inverted colors should be separate commits. It is hard to review a commit when you do two different things mixed in one commit.

HI Max! thanks for the review -- will adapt accordingly.
Just one question: where do I put the po/* file changes? This is basically to update the line numbers and add the translations for the cursor stuff.

hvennekate:
Changes have been split into multiple commits and po/* file changes removed.

doc/manual/en/ch11_configuration.tex
MaxKellermann: Yes, please - maybe with a footnote?

MaxKellermann:
> Just one question: where do I put the po/* file changes?

Not at all. Let me update those.

hvennekate:
> > Just one question: where do I put the po/* file changes?
> 
> Not at all. Let me update those.

:joy:  If only I had known that sooner! You sure you do Chinese as well?

PR #296, opened 2019-11-29
Vagrant debian10

### LINKED ISSUE #292: Vagrant build fails
When compiling with `make TARGET=ANDROID -j` on the Vagrant box the build fails with:

> /bin/sh: 1: pkg-config not found

lordfolken:
This updates the vagrant box to debian-10. We are also switching to the generic box provider from vagrantup. These boxes are built by lavabit and the entire [process](https://github.com/lavabit/robox/blob/master/http/generic.debian10.vagrant.cfg) is here on github.
The boxes support all possible providers for vagrant, such as: virtualbox,hyperv,libvirt and even parallels on OSX.

The clone process also now initializes submodules without which a build would fail.

Fixes #292 

I'm not entierly happy about adding the repos in the vagrantfile. This should be a seperate script in ide/provisioning/setuprepos.sh or similar and should be references by both docker and vagrant.

### DIFF
--- a/ide/vagrant/Vagrantfile
+++ b/ide/vagrant/Vagrantfile
@@ -9,6 +9,7 @@ EOF
 
 sudo apt-get update
 
+sudo /xcsoar-host-src/ide/provisioning/add-debian-unstable.sh
 sudo /xcsoar-host-src/ide/provisioning/install-debian-packages.sh
 /xcsoar-host-src/ide/provisioning/install-android-tools.sh
 
--- a/ide/vagrant/Vagrantfile
+++ b/ide/vagrant/Vagrantfile
@@ -22,7 +22,7 @@ git remote add master https://github.com/XCSoar/XCSoar.git
 SCRIPT
 
 Vagrant.configure("2") do |config|
-    config.vm.box = "koalephant/debian9-amd64"
+    config.vm.box = "generic/debian10"
 
     config.vm.network "private_network", type: "dhcp"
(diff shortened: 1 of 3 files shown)


ide/vagrant/Vagrantfile
```diff
@@ -15,14 +18,14 @@ sudo /xcsoar-host-src/ide/provisioning/install-debian-packages.sh
 sudo apt-get clean
 
 cd
-git clone -o host /xcsoar-host-src xcsoar-src
+git clone --recurse-submodules -o host /xcsoar-host-src xcsoar-src
```
kerel-fs: The git checkout on the host system should already be a full checkout with submodules, so why should we git-clone it in the provisioning script instead of using it directly?
Since the host checkout is cloned only once at provisioning time, it might get outdated when the devloper modified it after the first provisioning of the box.

lordfolken: Even if the submodules are checked out locally, git clone alone will not clone the submodules. What bothers me is that it actually checks them out from the defined upstream sources, and doesn't copy it from the host source tree. So a change in a submodule could be lost, or the state could differ from other reason. 

Options that we could have a look at:
- use rsync instead of git clone
- use host export mount

For the moment this works, so it can be applied.

lordfolken:
@kerel-fs could you review so we could merge this?

lordfolken:
This fixes #374 for the Vagrant environment

PR #298, opened 2019-12-07
Skysight

CazYokoyama:
Re-factor the code to clarify what each commit does.  The previous code was divided by who wrote. Its drawback is that one of the commits is large and not clear what it does.

The first couple of commits have a lot of files involved because of re-factoring. I thought it is better to keep the same file structure between the first and the last commit because it reduces difference between them.

I have no option other than clicking "Create pull request". Sorry if I create another PR. I hope not.

### DIFF
--- a/build/python/build/autotools.py
+++ b/build/python/build/autotools.py
@@ -44,6 +44,11 @@ class AutotoolsProject(MakeProject):
             build = self.make_build_path(toolchain)
 
         if self.autogen:
+            # run autogen.sh in wolfssl
+            autogen_sh = os.path.join(src, "autogen.sh")
+            if os.path.exists(autogen_sh):
+                subprocess.check_call([autogen_sh], cwd=src)
+
             if sys.platform == 'darwin':
                 subprocess.check_call(['glibtoolize', '--force'], cwd=src)
             else:
--- a/build/python/build/libs.py
+++ b/build/python/build/libs.py
@@ -281,3 +281,38 @@ libsalsa = AutotoolsProject(
     ],
     patches=abspath('lib/salsa-lib/patches')
 )
+    
+netcdf = AutotoolsProject(
+    'ftp://ftp.unidata.ucar.edu/pub/netcdf/netcdf-c-4.6.2.tar.gz',
+    'https://www.gfd-dennou.org/library/netcdf/unidata-mirror/netcdf-c-4.6.2.tar.gz',
+    'c37525981167b3cd82d32e1afa3022afb94e59287db5f116c57f5ed4d9c6a638',
+    'lib/libnetcdf.a',
+    [
+        '--disable-netcdf-4',
+        '--disable-dap',
+        '--disable-largefile',
+        '--disable-testsets',
+        '--disable-utilities',
+        '--disable-examples',
+        '--disable-doxygen',
+        '--disable-maintainer-mode',
+        '--disable-examples',
+        '--disable-shared', '--enable-static'
+    ],
+    patches=abspath('lib/netcdf/patches'),
+    ldflags='-Wl,--gc-sections'
+)
+
+netcdfcxx = AutotoolsProject(
+    'ftp://ftp.unidata.ucar.edu/pub/netcdf/netcdf-cxx-4.2.tar.gz',
+    'https://www.gfd-dennou.org/arch/netcdf/unidata-mirror/netcdf-cxx-4.2.tar.gz',
+    '95ed6ab49a0ee001255eac4e44aacb5ca4ea96ba850c08337a3e4c9a0872ccd1',
+    'lib/libnetcdf_c++.a',
+    [
+        '--disable-shared', '--enable-static',
+        '--disable-large-file-tests',
+        '--disable-extra-tests',
+        '--disable-valgrind-tests'
+    ],
+    autogen=True,
+)
--- a/build/python/build/libs.py
+++ b/build/python/build/libs.py
@@ -108,6 +108,26 @@ freetype = FreeTypeProject(
     ],
 )
 
+wolfssl = AutotoolsProject(
+    # fossies.org no longer keeps wolfssl-4.0.0.tar.gz, but keep it because
+    # AutotoolsProject does not allow wolfssl-4.0.0-stable as directory name,
+    # i.e. untaring v4.0.0-stable.tar.gz creates wolfssl-4.0.0-stable.
+    'https://fossies.org/linux/misc/wolfssl-4.0.0.tar.gz',
+    'https://github.com/wolfSSL/wolfssl/archive/v4.0.0-stable.tar.gz',
+    '6cf678c72b485d1904047c40c20f85104c96b5f39778822783a2c407ccb23657',
+    'lib/libwolfssl.a',
+    [
+      '--disable-all',
+      '--disable-shared',
+      '--enable-static',
+      '--enable-sslv3',
+      '--enable-tls13',
+      '--enable-alpn',
+      '--enable-opensslall',
+    ],
+    autogen=True,
+)
+
 curl = CurlProject(
     'http://curl.haxx.se/download/curl-7.64.1.tar.xz',
     'https://github.com/curl/curl/releases/download/curl-7_64_1/curl-7.64.1.tar.xz',
@@ -127,7 +147,7 @@ curl = CurlProject(
         '--disable-manual',
         '--disable-threaded-resolver', '--disable-verbose', '--disable-sspi',
         '--disable-crypto-auth', '--disable-ntlm-wb', '--disable-tls-srp', '--disable-cookies',
-        '--without-ssl', '--without-gnutls', '--without-nss', '--without-libssh2',
+        '--without-ssl', '--with-wolfssl', '--without-gnutls', '--without-nss', '--without-libssh2',
     ],
     patches=abspath('lib/curl/patches'),
 )
--- a/build/python/build/project.py
+++ b/build/python/build/project.py
@@ -1,5 +1,5 @@
 import os, shutil
-import re
+import re, glob
 
 from build.download import download_and_verify
 from build.tar import untar
@@ -52,6 +52,10 @@ class Project:
         path = untar(self.download(toolchain), parent_path, self.base)
         if self.patches is not None:
             push_all(toolchain, path, self.patches)
+        if os.path.isdir(path) is not True:
+            # rename wolfssl-4.0.0-stable to wolfssl-4.0.0
+            filename = glob.glob(path+"*")
+            os.rename(filename[0], path)
         return path
 
     def make_build_path(self, toolchain):
--- a/build/thirdparty.py
+++ b/build/thirdparty.py
@@ -97,6 +97,7 @@ elif 'apple-darwin' in actual_host_triplet:
     ]
 elif target == 'ANDROID':
     thirdparty_libs = [
+        wolfssl,
         curl,
         lua,
         proj,
--- a/build/thirdparty.py
+++ b/build/thirdparty.py
@@ -98,6 +98,8 @@ elif 'apple-darwin' in actual_host_triplet:
 elif target == 'ANDROID':
     thirdparty_libs = [
         wolfssl,
+        netcdf,
+        netcdfcxx,
         curl,
         lua,
         proj,
--- a/src/Dialogs/Settings/Panels/WeatherConfigPanel.cpp
+++ b/src/Dialogs/Settings/Panels/WeatherConfigPanel.cpp
@@ -29,6 +29,10 @@ Copyright_License {
 #include "Interface.hpp"
 #include "UIGlobals.hpp"
 #include "Util/NumberParser.hpp"
+#include "Form/DataField/Enum.hpp"
+#include "Form/DataField/Listener.hpp"
+#include "DataGlobals.hpp"
+#include "Weather/Skysight/Skysight.hpp"
 
 enum ControlIndex {
 #ifdef HAVE_PCMET
@@ -37,10 +41,15 @@ enum ControlIndex {
   PCMET_FTP_USER,
   PCMET_FTP_PASSWORD,
 #endif
+#ifdef HAVE_SKYSIGHT
+  SKYSIGHT_EMAIL,
+  SKYSIGHT_PASSWORD,
+  SKYSIGHT_REGION
+#endif
 };
 
 class WeatherConfigPanel final
-  : public RowFormWidget {
+  : public RowFormWidget, DataFieldListener {
 public:
   WeatherConfigPanel()
     :RowFormWidget(UIGlobals::GetDialogLook()) {}
@@ -49,13 +58,30 @@ public:
   /* methods from Widget */
   virtual void Prepare(ContainerWindow &parent, const PixelRect &rc) override;
   virtual bool Save(bool &changed) override;
+
+private:
+  /* methods from DataFieldListener */
+  virtual void OnModified(DataField &df) override {};
 };
+
+static void
+FillRegionControl(WndProperty &wp, const TCHAR *setting)
+{
+  DataFieldEnum *df = (DataFieldEnum *)wp.GetDataField();
+  auto skysight = DataGlobals::GetSkysight();
+
+  for (auto &i: skysight->GetRegions())
+    df->addEnumText(i.first.c_str(), i.second.c_str());
+
+  // if old region doesn't exist any more this will fall back to first element
+  df->Set(setting);
+  wp.RefreshDisplay();
+}
 
 void
 WeatherConfigPanel::Prepare(ContainerWindow &parent, const PixelRect &rc)
 {
   const auto &settings = CommonInterface::GetComputerSettings().weather;
-
   RowFormWidget::Prepare(parent, rc);
 
   AddText(_T("pc_met Username"), _T(""),
@@ -67,6 +93,14 @@ WeatherConfigPanel::Prepare(ContainerWindow &parent, const PixelRect &rc)
           settings.pcmet.ftp_credentials.username);
   AddPassword(_T("pc_met FTP Password"), _T(""),
               settings.pcmet.ftp_credentials.password);
+
+  AddText(_T("Skysight Email"), _T("The e-mail you use to log in to the skysight.io site."),
+          settings.skysight.email);
+  AddPassword(_T("Skysight Password"), _T("Your Skysight password."),
+              settings.skysight.password);  
+  WndProperty *wp = AddEnum(_T("Skysight Region"), _T("The Skysight region to load data for."), this);
+  FillRegionControl(*wp, settings.skysight.region);
+
 }
 
 bool
@@ -88,6 +122,18 @@ WeatherConfigPanel::Save(bool &_changed)
 
   changed |= SaveValue(PCMET_FTP_PASSWORD, ProfileKeys::PCMetFtpPassword,
                        settings.pcmet.ftp_credentials.password);
+#endif
+
+#ifdef HAVE_SKYSIGHT
+  changed |= SaveValue(SKYSIGHT_EMAIL, ProfileKeys::SkysightEmail,
+                       settings.skysight.email);
+
+  changed |= SaveValue(SKYSIGHT_PASSWORD, ProfileKeys::SkysightPassword,
+                       settings.skysight.password);
+
+  changed |= SaveValue(SKYSIGHT_REGION, ProfileKeys::SkysightRegion,
+                    settings.skysight.region);        
+  DataGlobals::GetSkysight()->Init();         
 #endif
 
   _changed |= changed;
--- a/src/Dialogs/Weather/SkysightDialog.cpp
+++ b/src/Dialogs/Weather/SkysightDialog.cpp
@@ -82,6 +82,37 @@ private:
   std::shared_ptr<Skysight> skysight;
 };
 
+//TODO: Could extend this to extract Skysight data at point and display on list (see NOAAListRenderer::Draw(2)
+void
+SkysightListItemRenderer::Draw(Canvas &canvas, const PixelRect rc, unsigned i) {
+  const ComputerSettings &settings = CommonInterface::GetComputerSettings();
+  SkysightActiveMetric m = SkysightActiveMetric(skysight->GetActiveMetric(i));
+
+  tstring first_row = tstring(m.metric->name);
+  if (skysight->displayed_metric == m.metric->id.c_str())
+    first_row += " [ACTIVE]";
+
+  StaticString<256> second_row;
+
+  if (m.updating) {
+    second_row.Format("%s", _("Updating..."));
+  } else {
+    if (!m.from || !m.to || !m.mtime) {
+      second_row.Format("%s", _("No data. Press \"Update\" to update."));
+    } else {
+      uint64_t elapsed = BrokenDateTime::NowUTC().ToUnixTimeUTC() - m.mtime;
+
+      second_row.Format(_("Data from %s to %s. Updated %s ago"), 
+                        FormatLocalTimeHHMM((int)m.from, settings.utc_offset).c_str(),
+                        FormatLocalTimeHHMM((int)m.to, settings.utc_offset).c_str(),
+                        FormatTimespanSmart(elapsed).c_str());  
+    }
+  }
+
+  row_renderer.DrawFirstRow(canvas, rc, first_row.c_str());
+  row_renderer.DrawSecondRow(canvas, rc, second_row.c_str());
+}
+
 /**
  * RENDERER FOR METRICS POPUP LIST
  */
@@ -119,13 +150,18 @@ class SkysightWidget final
   : public ListWidget, private ActionListener, private Timer
 {
   enum Buttons {
+    ACTIVATE,
+    DEACTIVATE,
+    ADD,
+    REMOVE,
     UPDATE,
     UPDATE_ALL
   };
 
   ButtonPanelWidget *buttons_widget;
 
-  Button *update_button, *updateall_button;
+  Button *activate_button, *deactivate_button, *add_button, *remove_button,
+    *update_button, *updateall_button;
 
   struct ListItem {
     StaticString<255> name;
@@ -151,8 +187,13 @@ public:
 
 private:
   void UpdateList();
+
+  void ActivateClicked();
+  void DeactivateClicked();
+  void AddClicked();
   void UpdateClicked();
   void UpdateAllClicked();
+  void RemoveClicked();
 
 public:
   /* virtual methods from class Widget */
@@ -183,6 +224,10 @@ private:
 void
 SkysightWidget::CreateButtons(ButtonPanel &buttons)
 {
+  activate_button = buttons.Add(_("Activate"), *this, ACTIVATE);
+  deactivate_button = buttons.Add(_("Deactivate"), *this, DEACTIVATE);
+  add_button = buttons.Add(_("Add"), *this, ADD);
+  remove_button = buttons.Add(_("Remove"), *this, REMOVE);
   update_button = buttons.Add(_("Update"), *this, UPDATE);
   updateall_button = buttons.Add(_("Update All"), *this, UPDATE_ALL);
 }
@@ -215,10 +260,12 @@ SkysightWidget::UpdateList()
 {
   unsigned index = GetList().GetCursorIndex();
   bool item_updating = false;
+  bool item_active = false;
 
   if ((int)index < skysight->NumActiveMetrics()) {
     SkysightActiveMetric a = skysight->GetActiveMetric(index);
     item_updating = a.updating;
+    item_active = (skysight->displayed_metric == a.metric->id.c_str());
   }
 
   bool any_updating = skysight->ActiveMetricsUpdating();
@@ -228,13 +275,49 @@ SkysightWidget::UpdateList()
   list.SetLength(skysight->NumActiveMetrics());
   list.Invalidate();
 
+  add_button->SetEnabled(!skysight->ActiveMetricsFull());
+  remove_button->SetEnabled(!empty && !item_updating);
   update_button->SetEnabled(!empty && !item_updating);
   updateall_button->SetEnabled(!empty && !any_updating);
+  activate_button->SetEnabled(!empty && !item_updating); 
+  deactivate_button->SetEnabled(!empty && item_active);
 }
 
 void
 SkysightWidget::OnPaintItem(Canvas &canvas, const PixelRect rc, unsigned index)
 {
+  row_renderer.Draw(canvas, rc, index);
+}
+
+void SkysightWidget::AddClicked()
+{
+  if (!skysight->IsReady()) {
+    ShowMessageBox(
+      _("Please check your Skysight settings and internet connection."),
+      _("Couldn't connect to Skysight"),
+      MB_OK
+    );
+    return;
+  }
+
+  SkysightMetricsListItemRenderer item_renderer;
+ 
+  int i = ListPicker(_("Choose a parameter"),
+                     skysight->NumMetrics(), 0,
+                     item_renderer.CalculateLayout(UIGlobals::GetDialogLook()),
+                     item_renderer,
+                     false, /*timer */
+                     nullptr, /*TCHAR help text */
+                     &SkysightMetricsListItemRenderer::HelpCallback,
+                     nullptr /*Extra caption*/);
+
+  if (i < 0)
+    return;
+
+  assert((int)i < skysight->NumMetrics());
+  skysight->AddActiveMetric(skysight->GetMetric(i).id.c_str());
+  
+  UpdateList();
 }
 
 void SkysightWidget::UpdateClicked()
@@ -255,10 +338,63 @@ void SkysightWidget::UpdateAllClicked()
   UpdateList();
 }
 
+void SkysightWidget::RemoveClicked()
+{
+  unsigned index = GetList().GetCursorIndex();
+  assert(index < (unsigned)skysight->NumActiveMetrics());
+  
+  SkysightActiveMetric a = skysight->GetActiveMetric(index);
+  StaticString<256> tmp;
+  tmp.Format(_("Do you want to remove \"%s\"?"),
+             a.metric->name.c_str());
+
+  if (ShowMessageBox(tmp, _("Remove"), MB_YESNO) == IDNO)
+    return;
+
+  skysight->RemoveActiveMetric(a.metric->id);
+
+  UpdateList();
+}
+
+inline void
+SkysightWidget::ActivateClicked()
+{
+  unsigned index = GetList().GetCursorIndex();
+  assert(index < (unsigned)skysight->NumActiveMetrics());
+  
+  SkysightActiveMetric a = skysight->GetActiveMetric(index);  
+  if (!skysight->DisplayActiveMetric(a.metric->id.c_str()))
+    ShowMessageBox(_("Couldn't display data. There is no forecast data available for this time."),
+		   _("Display Error"), MB_OK);
+  UpdateList();
+}
+
+inline void
+SkysightWidget::DeactivateClicked()
+{
+  unsigned index = GetList().GetCursorIndex();
+  assert(index < (unsigned)skysight->NumActiveMetrics());
+
+  skysight->DisplayActiveMetric();
+  UpdateList();
+}
+
 void
 SkysightWidget::OnAction(int id) noexcept
 {
   switch ((Buttons)id) {
+  case ACTIVATE:
+    ActivateClicked();
+    break;
+
+  case DEACTIVATE:
+    DeactivateClicked();
+    break;
+
+  case ADD:
+    AddClicked();
+    break;
+
   case UPDATE:
     UpdateClicked();
     break;
@@ -266,7 +402,9 @@ SkysightWidget::OnAction(int id) noexcept
   case UPDATE_ALL:
     UpdateAllClicked();
     break;    
-  default:
+
+  case REMOVE:
+    RemoveClicked();
     break;
   }
 }
--- a/src/Dialogs/Weather/SkysightDialog.cpp
+++ b/src/Dialogs/Weather/SkysightDialog.cpp
@@ -1,0 +1,284 @@
+/*
+Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2016 The XCSoar Project
+  A detailed list of copyright holders can be found in the file "AUTHORS".
+
+  This program is free software; you can redistribute it and/or
+  modify it under the terms of the GNU General Public License
+  as published by the Free Software Foundation; either version 2
+  of the License, or (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+
+#include "SkysightDialog.hpp"
+#include "Dialogs/Message.hpp"
+#include "Dialogs/ListPicker.hpp"
+#include "Dialogs/JobDialog.hpp"
+#include "Language/Language.hpp"
+#include "Weather/Features.hpp"
+
+#ifdef HAVE_SKYSIGHT
+#include "UIGlobals.hpp"
+#include "Look/DialogLook.hpp"
+#include "Dialogs/TextEntry.hpp"
+#include "Form/Button.hpp"
+#include "Form/ButtonPanel.hpp"
+#include "Form/ActionListener.hpp"
+#include "Widget/ListWidget.hpp"
+#include "Widget/ButtonPanelWidget.hpp"
+#include "Weather/Skysight/Skysight.hpp"
+#include "Operation/VerboseOperationEnvironment.hpp"
+#include "Util/TrivialArray.hxx"
+#include "Util/StringAPI.hxx"
+#include "Renderer/TextRowRenderer.hpp"
+#include "Renderer/TwoTextRowsRenderer.hpp"
+#include "Event/Timer.hpp"
+
+#include "Protection.hpp"
+#include "DataGlobals.hpp"
+#include "Interface.hpp"
+
+#include "Language/Language.hpp"
+#include "LocalPath.hpp"
+#include "Net/HTTP/Session.hpp"
+
+#include "MapWindow/OverlayBitmap.hpp"
+#include "MapWindow/GlueMapWindow.hpp"
+#include "Dialogs/Error.hpp"
+
+#include "Util/StaticString.hxx"
+#include "Language/Language.hpp"
+#include "Time/BrokenDateTime.hpp"
+#include "Formatter/TimeFormatter.hpp"
+#include "Formatter/LocalTimeFormatter.hpp"
+
+#include "Weather/Skysight/Skysight.hpp"
+
+class SkysightListItemRenderer
+{
+  TwoTextRowsRenderer row_renderer; 
+
+public:
+  SkysightListItemRenderer(): skysight(DataGlobals::GetSkysight()) {}
+
+  void Draw(Canvas &canvas, const PixelRect rc, unsigned i); 
+
+  unsigned CalculateLayout(const DialogLook &look) {
+    return row_renderer.CalculateLayout(*look.list.font_bold, look.small_font);
+  }
+
+private:
+  std::shared_ptr<Skysight> skysight;
+};
+
+/**
+ * RENDERER FOR METRICS POPUP LIST
+ */
+class SkysightMetricsListItemRenderer: public ListItemRenderer
+{
+  TextRowRenderer row_renderer;
+  std::shared_ptr<Skysight> skysight;
+
+public:
+  SkysightMetricsListItemRenderer(): skysight(DataGlobals::GetSkysight()) {}
+  
+  unsigned CalculateLayout(const DialogLook &look) {
+    return row_renderer.CalculateLayout(*look.list.font);
+  }
+
+  void OnPaintItem(Canvas &canvas, const PixelRect rc, unsigned i) override {
+    row_renderer.DrawTextRow(canvas, rc, skysight->GetMetric(i).name.c_str());
+  }
+  
+  static const TCHAR* HelpCallback(unsigned i)
+  {
+    std::shared_ptr<Skysight> skysight = DataGlobals::GetSkysight();
+
+    if (!skysight)
+      return _("No description available.");
+
+    tstring helptext = skysight->GetMetric(i).desc;
+    TCHAR *help = new TCHAR[helptext.length() + 1];
+    std::strcpy(help, helptext.c_str());
+    return help;
+  }  
+};
+
+class SkysightWidget final
+  : public ListWidget, private ActionListener, private Timer
+{
+  enum Buttons {
+    UPDATE,
+    UPDATE_ALL
+  };
+
+  ButtonPanelWidget *buttons_widget;
+
+  Button *update_button, *updateall_button;
+
+  struct ListItem {
+    StaticString<255> name;
+
+    gcc_pure bool operator<(const ListItem &i2) const
+    {
+      return StringCollate(name, i2.name) < 0;
+    }
+  };
+
+  SkysightListItemRenderer row_renderer;
+  std::shared_ptr<Skysight> skysight;
+
+public:
+  explicit SkysightWidget(std::shared_ptr<Skysight> &&_skysight)
+    : skysight(std::move(_skysight)) {}
+
+  void SetButtonPanel(ButtonPanelWidget &_buttons) {
+    buttons_widget = &_buttons;
+  }
+
+  void CreateButtons(ButtonPanel &buttons);
+
+private:
+  void UpdateList();
+  void UpdateClicked();
+  void UpdateAllClicked();
+
+public:
+  /* virtual methods from class Widget */
+  void Prepare(ContainerWindow &parent,
+                       const PixelRect &rc) override;
+  void Unprepare() override;
+
+protected:
+  /* virtual methods from ListItemRenderer */
+  void OnPaintItem(Canvas &canvas, const PixelRect rc,
+                           unsigned idx) override;
+
+  /* virtual methods from ListCursorHandler */
+  bool CanActivateItem(unsigned index) const override {
+    return true;
+  }
+
+  void OnActivateItem(unsigned index) override {};
+
+private:
+  /* virtual methods from class ActionListener */
+  void OnAction(int id) noexcept override;
+
+  /* virtual methods from Timer */
+  void OnTimer() override;
+};
+
+void
+SkysightWidget::CreateButtons(ButtonPanel &buttons)
+{
+  update_button = buttons.Add(_("Update"), *this, UPDATE);
+  updateall_button = buttons.Add(_("Update All"), *this, UPDATE_ALL);
+}
+
+void
+SkysightWidget::Prepare(ContainerWindow &parent, const PixelRect &rc)
+{
+  CreateButtons(buttons_widget->GetButtonPanel());
+  const DialogLook &look = UIGlobals::GetDialogLook();
+  CreateList(parent, look, rc, row_renderer.CalculateLayout(look));
+  UpdateList();
+  Timer::Schedule(std::chrono::milliseconds(500));
+}
+
+void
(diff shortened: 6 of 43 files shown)


src/Weather/Skysight/Skysight.cpp
```diff
@@ -427,6 +428,29 @@ Skysight::FromUnixTime(uint64_t t)
   return api.FromUnixTime(t);
 }
 
+void
+Skysight::Render(bool force_update)
+{
+  if (displayed_metric.metric) {
+    LogFormat("In Render!");
```
MaxKellermann: Useless log message

src/Weather/Skysight/Skysight.cpp
```diff
@@ -577,5 +606,21 @@ Skysight::DisplayActiveMetric(const TCHAR *const id)
...
+  auto *map = UIGlobals::GetMap();
+  if (map == nullptr)
+    return false;
+
+  LogFormat("Setting %s", path.c_str());
+  std::unique_ptr<MapOverlayBitmap> bmp;
+  try {
+    bmp.reset(new MapOverlayBitmap(path));
+  } catch (const std::exception &e) {
+    return false;
```
MaxKellermann: Why ignore the exception details?

src/Weather/Skysight/Skysight.cpp
```diff
@@ -577,5 +606,21 @@ Skysight::DisplayActiveMetric(const TCHAR *const id)
...
+  }
+
+  bmp->SetAlpha(0.6);
+  bmp->SetLabel(label);
+  map->SetOverlay(std::move(bmp));
+  LogFormat("%s", "Success");
```
MaxKellermann: Useless log message

src/Weather/Skysight/SkysightAPI.cpp
```diff
@@ -531,6 +531,18 @@ SkysightAPI::GetData(SkysightCallType t, const TCHAR *const layer,
...
+bool
+SkysightAPI::LoadFromFile(Path path, tstring &content)
+{
+  FileLineReader reader(path, Charset::AUTO);
+  content = "";
+  char *line;
+  while ((line = reader.ReadLine()) != NULL)
+    content += line;
+
+  return true;
```
MaxKellermann: What use is this return value?

src/Weather/Skysight/SkysightAPI.cpp
```diff
@@ -531,6 +531,18 @@ SkysightAPI::GetData(SkysightCallType t, const TCHAR *const layer,
...
 }
 
+bool
+SkysightAPI::LoadFromFile(Path path, tstring &content)
+{
+  FileLineReader reader(path, Charset::AUTO);
+  content = "";
+  char *line;
+  while ((line = reader.ReadLine()) != NULL)
+    content += line;
```
MaxKellermann: What if this file has 100 gigabytes?

src/Weather/Skysight/SkysightAPI.cpp
```diff
@@ -531,6 +531,18 @@ SkysightAPI::GetData(SkysightCallType t, const TCHAR *const layer,
   return true;
 }
 
+bool
+SkysightAPI::LoadFromFile(Path path, tstring &content)
+{
+  FileLineReader reader(path, Charset::AUTO);
```
MaxKellermann: Why `Charset::AUTO`?

src/Weather/Skysight/SkysightAPI.cpp
```diff
@@ -531,6 +531,18 @@ SkysightAPI::GetData(SkysightCallType t, const TCHAR *const layer,
...
+  content = "";
+  char *line;
```
MaxKellermann: Wrong type

build/python/build/libs.py
```diff
@@ -281,3 +282,38 @@
...
+        '--disable-testsets',
+        '--disable-utilities',
+        '--disable-examples',
+        '--disable-doxygen',
+        '--disable-maintainer-mode',
+        '--disable-examples',
+        '--disable-shared', '--enable-static'
+    ],
+    patches=abspath('lib/netcdf/patches'),
+  ldflags='-Wl,--gc-sections'
```
MaxKellermann: Wrong indent

build/python/build/libs.py
```diff
@@ -281,3 +282,38 @@
...
+    'https://www.unidata.ucar.edu/downloads/netcdf/ftp/netcdf-cxx4-4.2.tar.gz',
+    '95ed6ab49a0ee001255eac4e44aacb5ca4ea96ba850c08337a3e4c9a0872ccd1',
+    'lib/libnetcdf_c++.a',
+    [
+        '--disable-shared', '--enable-static',
+        '--disable-large-file-tests',
+        '--disable-extra-tests',
+        '--disable-valgrind-tests'
+    ],
+  patches=abspath('lib/netcdfcxx/patches'),
```
MaxKellermann: Wrong indent

build/python/build/netcdfcxx.py
```diff
@@ -0,0 +1,6 @@
+from build.autotools import AutotoolsProject
+
+class NetcdfCxxProject(AutotoolsProject):
+    def get_make_args(self, toolchain):
+        # HELP TODO: Temp workaround. Will ANDROID_NDK alwats be set in env if in the default location?
```
MaxKellermann: No, but this whole thing looks completely wrong anyway. What are you trying to do here?

lib/netcdfcxx/patches/config.patch
```diff
@@ -0,0 +1,4826 @@
+diff --git a/config.guess b/config.guess
+index dc84c68..8ef92de 100755
+--- a/config.guess
++++ b/config.guess
```
MaxKellermann: Do not patch generated files!

lib/netcdfcxx/patches/cout.patch
```diff
@@ -0,0 +1,22 @@
...
+-#include <iostream>
++//#include <iostream>
+ #include "netcdfcpp.h"
+ 
+ #ifndef TRUE
+@@ -1645,7 +1645,7 @@ int NcError::set_err (int err)
+     // Check ncopts and handle appropriately
+     if(err != NC_NOERR) {
+ 	if(ncopts == verbose_nonfatal || ncopts == verbose_fatal) {
+-	    std::cout << nc_strerror(err) << std::endl;
```
MaxKellermann: WTF, what a shitty library will output messages to stdout? Are you sure you want to use such a library?

lib/netcdfcxx/patches/series
```diff
@@ -0,0 +1,3 @@
+config.patch
+#rem_iostream.patch
```
MaxKellermann: What's that?

src/Dialogs/Weather/SkysightDialog.cpp
```diff
@@ -82,6 +82,37 @@ class SkysightListItemRenderer
   std::shared_ptr<Skysight> skysight;
 };
 
+//TODO: Could extend this to extract Skysight data at point and display on list (see NOAAListRenderer::Draw(2)
+void
+SkysightListItemRenderer::Draw(Canvas &canvas, const PixelRect rc, unsigned i) {
+  const ComputerSettings &settings = CommonInterface::GetComputerSettings();
+  SkysightActiveMetric m = SkysightActiveMetric(skysight->GetActiveMetric(i));
+
+  tstring first_row = tstring(m.metric->name);
```
MaxKellermann: This always allocates memory and copies the string, even if nothing gets appended, for each repaint for each row. I don't like this kind of bloat.

src/Dialogs/Weather/SkysightDialog.cpp
```diff
@@ -82,6 +82,37 @@ class SkysightListItemRenderer
...
+
+  if (m.updating) {
+    second_row.Format("%s", _("Updating..."));
+  } else {
+    if (!m.from || !m.to || !m.mtime) {
+      second_row.Format("%s", _("No data. Press \"Update\" to update."));
+    } else {
+      uint64_t elapsed = BrokenDateTime::NowUTC().ToUnixTimeUTC() - m.mtime;
+
+      second_row.Format(_("_Data from %s to %s. Updated %s ago"),
```
MaxKellermann: Why does this string start with an underscore?
Will this be translatable to all languages? (grammar!)

src/Weather/Skysight/APIQueue.hpp
```diff
@@ -24,12 +24,14 @@ Copyright_License {
...
 
 #include "Request.hpp"
+#include "CDFDecoder.hpp"
 #include "Metrics.hpp"
 #include "Event/Timer.hpp"
 #include <vector>
 
 class SkysightAPIQueue final: public Timer {
   std::vector<std::unique_ptr<SkysightAsyncRequest>> request_queue;
+  std::vector<std::unique_ptr<CDFDecoder>> decode_queue;
```
MaxKellermann: Why a vector of unique_ptrs? Why not just a list?

src/Weather/Skysight/APIQueue.hpp
```diff
@@ -0,0 +1,55 @@
...
+#ifndef WEATHER_SKYSIGHTAPI_QUEUE_HPP
+#define WEATHER_SKYSIGHTAPI_QUEUE_HPP
+
+#include "Request.hpp"
+#include "Metrics.hpp"
+#include "Event/Timer.hpp"
+#include <vector>
+
+class SkysightAPIQueue final: public Timer {
+  std::vector<std::unique_ptr<SkysightAsyncRequest>> request_queue;
```
MaxKellermann: Why a vector of unique_ptrs? Why not just a list?

src/Weather/Skysight/CDFDecoder.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+void CDFDecoder::Tick() noexcept
+{
+  status = Status::Busy;
+  mutex.unlock();
+  Decode();
+}
+
+bool CDFDecoder::DecodeError() {
+  LogFormat("Decoder Error");
```
MaxKellermann: Who is ever going to read this message?

src/Weather/Skysight/CDFDecoder.hpp
```diff
@@ -0,0 +1,71 @@
...
+             status(Status::Idle) {};
+  ~CDFDecoder() {};
+
+  void DecodeAsync();
+  void DecodeSync();
+  void Done();
+  Status GetStatus();
+  
+private:
+  const AllocatedPath path;
```
MaxKellermann: Move fields to the top according to our coding style.

src/Weather/Skysight/CDFDecoder.cpp
```diff
@@ -0,0 +1,334 @@
...
+{
+  std::lock_guard<Mutex> lock(mutex);
+  CDFDecoder::Status s = status;
+  return s;
+}
+
+void CDFDecoder::Tick() noexcept
+{
+  status = Status::Busy;
+  mutex.unlock();
```
MaxKellermann: Never use unprotected mutex lock/unlock calls!

src/Weather/Skysight/CDFDecoder.cpp
```diff
@@ -0,0 +1,334 @@
...
+  Decode();
+}
+
+bool CDFDecoder::DecodeError() {
+  LogFormat("Decoder Error");
+  MakeCallback(false);
+  mutex.lock();
```
MaxKellermann: Never use unprotected mutex lock/unlock calls!

src/Weather/Skysight/CDFDecoder.cpp
```diff
@@ -0,0 +1,334 @@
...
+  status = Status::Error;
+  return true;
+}
+
+bool CDFDecoder::DecodeSuccess()
+{
+  MakeCallback(true);
+  mutex.lock();
```
MaxKellermann: Never use unprotected mutex lock/unlock calls!

src/Weather/Skysight/CDFDecoder.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+  data_var->get(&var_vals[0], (long)lat_size, (long)lon_size);
+  double fill_value = data_var->get_att("_FillValue")->values()->as_double(0);
+  float var_offset = data_var->get_att("add_offset")->values()->as_float(0);
+  float var_scale = data_var->get_att("scale_factor")->values()->as_float(0);
+
+  //Generate GeoTiff
+  TIFF *tf = XTIFFOpen(output_path.c_str(), "w");
+  if (!tf)
+    return DecodeError();
```
MaxKellermann: If an error occurs, throw an exception.

src/Weather/Skysight/CDFDecoder.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+bool CDFDecoder::DecodeSuccess()
+{
+  MakeCallback(true);
+  mutex.lock();
+  status = Status::Complete;
+  return true;
+}
+
+#ifdef ANDROID
```
MaxKellermann: Wait, two completely distinct implementations of this huge method? No redundant code, please!

src/Weather/Skysight/CDFDecoder.cpp
```diff
@@ -0,0 +1,334 @@
...
+  GTIFKeySet(gt, GTCitationGeoKey, TYPE_ASCII, 27, "Generated by XCSoar with data from skysight.io");
+  GTIFKeySet(gt, GeogLinearUnitsGeoKey, TYPE_SHORT, 1, Linear_Meter);
+  GTIFKeySet(gt, GeogAngularUnitsGeoKey, TYPE_SHORT, 1, Angular_Degree);
+
+  tsize_t linebytes = samplesperpixel * lon_size;
+  unsigned char *row;
+
+  row = (TIFFScanlineSize(tf) > linebytes) ?
+    (unsigned char *)_TIFFmalloc(linebytes) :
+    (unsigned char *)_TIFFmalloc(TIFFScanlineSize(tf));
```
MaxKellermann: What if memory allocation fails?

src/Weather/Skysight/CDFDecoder.cpp
```diff
@@ -0,0 +1,334 @@
...
+  if (row)
+    _TIFFfree(row);
+
+  GTIFFree(gt);
+
+  data_file.close();
+  File::Delete(path);
+  return (success) ? DecodeSuccess() : DecodeError();
+
+  return false;
```
MaxKellermann: return after return?

build/python/build/project.py
```diff
@@ -52,6 +52,10 @@ def unpack(self, toolchain, out_of_tree=True):
         path = untar(self.download(toolchain), parent_path, self.base)
         if self.patches is not None:
             push_all(toolchain, path, self.patches)
+        if os.path.isdir(path) is not True:
+            # rename wolfssl-4.0.0-stable to wolfssl-4.0.0
```
MaxKellermann: Bad and fragile kludge for exactly one special cae.

build/thirdparty.py
```diff
@@ -97,6 +97,7 @@ def __init__(self, tarball_path, src_path, build_path, install_prefix,
     ]
 elif target == 'ANDROID':
     thirdparty_libs = [
+	wolfssl,
```
MaxKellermann: Bad indent

src/Dialogs/Weather/SkysightDialog.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+public:
+  /* virtual methods from class Widget */
+  virtual void Prepare(ContainerWindow &parent,
+                       const PixelRect &rc) override;
+  virtual void Unprepare() override;
+
+protected:
+  /* virtual methods from ListItemRenderer */
+  virtual void OnPaintItem(Canvas &canvas, const PixelRect rc,
```
MaxKellermann: Redundant `virtual`

src/Weather/Skysight/Metrics.hpp
```diff
@@ -24,6 +24,91 @@ Copyright_License {
...
+  BrokenDateTime forecast_index;
+  DisplayedMetric() {
+    metric = nullptr;
+  };
+  DisplayedMetric(SkysightMetric *_metric, BrokenDateTime _fc_index):
+    metric(_metric), forecast_index(_fc_index) {};
+  void clear() {
+    metric = nullptr;
+  }
+  bool operator==(const DisplayedMetric &d) {
```
MaxKellermann: These operators look very wrong. `==` compares the id, but `<` and `>` compare the time stamp, while at the same time doing funny things with `IsPlausible`.

src/Weather/Skysight/Metrics.hpp
```diff
@@ -24,6 +24,91 @@ Copyright_License {
...
+  SkysightActiveMetric(SkysightMetric *_metric, uint64_t _from,
+		       uint64_t _to, uint64_t _mtime): 
+    metric(_metric), from(_from), to(_to), mtime(_mtime) {}
+  SkysightActiveMetric(const SkysightActiveMetric &m):
+    metric(m.metric), from(m.from), to(m.to), mtime(m.mtime),
+    updating(m.updating) {}
+};
+
+struct DisplayedMetric {
+  SkysightMetric *metric;
```
MaxKellermann: Who owns this pointer?

build/python/build/libs.py
```diff
@@ -127,7 +147,7 @@
         '--disable-manual',
         '--disable-threaded-resolver', '--disable-verbose', '--disable-sspi',
         '--disable-crypto-auth', '--disable-ntlm-wb', '--disable-tls-srp', '--disable-cookies',
-        '--without-ssl', '--without-gnutls', '--without-nss', '--without-libssh2',
+        '--without-ssl', '--with-wolfssl', '--without-gnutls', '--without-nss', '--without-libssh2',
```
MaxKellermann: This should be a separate commit.

src/Weather/Skysight/SkysightAPI.hpp
```diff
@@ -47,6 +47,7 @@ struct BrokenDateTime;
 class SkysightAPI final: public Timer {
   friend struct SkysightRequest;
   friend struct SkysightAsyncRequest;
+  friend class  CDFDecoder;
```
MaxKellermann: Two spaces? Don't draw ASCII art in code.

src/Weather/Skysight/Skysight.cpp
CazYokoyama: Agree. Will remove. I guess this is for debugging which I didn't do.

src/Weather/Skysight/Skysight.cpp
CazYokoyama: Will add logging message.

src/Weather/Skysight/Skysight.cpp
CazYokoyama: will remove.

src/Weather/Skysight/SkysightAPI.cpp
CazYokoyama: Will introduce maximum. Return true on success, false on exceed limit.

src/Weather/Skysight/SkysightAPI.cpp
CazYokoyama: Will introduce maximum. Return true on success, false on exceed limit.

build/python/build/libs.py
CazYokoyama: will fix.

build/python/build/libs.py
CazYokoyama: Will fix.

src/Weather/Skysight/CDFDecoder.cpp
CazYokoyama: Will remove.

src/Weather/Skysight/Skysight.cpp
MaxKellermann: But who will ever read the log?

CazYokoyama: remove try-catch.

src/Weather/Skysight/SkysightAPI.cpp
CazYokoyama: Correction. The function which have these lines is not used in anywhere. Will remove.

src/Weather/Skysight/SkysightAPI.cpp
CazYokoyama: Will remove the function.

src/Weather/Skysight/SkysightAPI.cpp
CazYokoyama: Will remove the function.

src/Weather/Skysight/SkysightAPI.cpp
CazYokoyama: Will remove the function.

build/python/build/netcdfcxx.py
CazYokoyama: I found it is not needed. AutotoolsProject is good enough. Will be removed.

lib/netcdfcxx/patches/cout.patch
CazYokoyama: Will remove the patch.

lib/netcdfcxx/patches/config.patch
CazYokoyama: No patches are neccessary by autogen=True

lib/netcdfcxx/patches/series
CazYokoyama: Will remove.

src/Dialogs/Weather/SkysightDialog.cpp
CazYokoyama: Could you confirm that your point is first_row? I.e. allocate first_row and copy m.metric->name to first_copy, correct?

src/Dialogs/Weather/SkysightDialog.cpp
CazYokoyama: No reason to start with an underscore. Will remove.
Skysight chart is available every 30 minutes. So second_row is "Data from 10:30 to 11:00. Updated 1 sec ago" for example. I may translate to Japanese as "10:30 kara 11:00 no de-ta. 1 byo maeni koushin". Is there any rule for localization?

build/thirdparty.py
CazYokoyama: It was by tab. Will replace it by spaces.

src/Weather/Skysight/SkysightAPI.hpp
CazYokoyama: remove a space.

src/Weather/Skysight/Metrics.hpp
CazYokoyama: I am not clear what you ask. Are you saying that this whole structure should be in the next commit, "convert nc -> tif for skysight chart" because it is not used for downloading skysight chart?

src/Weather/Skysight/Metrics.hpp
CazYokoyama: remove unused functions and operators include this.

src/Dialogs/Weather/SkysightDialog.cpp
CazYokoyama: remove virtual, i.e. virtual void  -> void.

Are you able to explain why it is redundant? I am not an expert of c++. I can remove virtual from all function in the class.

src/Weather/Skysight/CDFDecoder.cpp
CazYokoyama: will return false on allocation error.

src/Weather/Skysight/CDFDecoder.cpp
CazYokoyama: ANDROID:                       NcFile data_file(path.c_str(),               NcFile::FileMode::ReadOnly);
Non- ANDROID netCDF::NcFile data_file(path.c_str(), netCDF::NcFile::read);

I'll try to minimize difference.

CazYokoyama: 3 #if-#else-#endif are required until //Generate GeoTiff.

build/python/build/libs.py
CazYokoyama: This adds ssl authentication. i.e. https by wolfssl while openssl is disabled. It is required for skysight access. Are you sure you want it be a separate commit?

build/python/build/project.py
CazYokoyama: I agree. wolfssl source code is gotten from 'https://fossies.org/linux/misc/wolfssl-4.0.0.tar.gz' as a primary source. It no longer keeps the corresponding version. The second choice, https://github.com/wolfSSL/wolfssl/archive/v4.0.0-stable.tar.gz' is expanded as wolfssl-4.0.0-stable which does not match following processing. Do you have better idea?

src/Weather/Skysight/CDFDecoder.hpp
CazYokoyama: Because private variable status is Status which is public enum class. Move it and private variables in front of public enum classes and functions.

src/Weather/Skysight/APIQueue.hpp
CazYokoyama: I give up. Could you guide me how to convert a vector of unique_ptrs to a list?

src/Weather/Skysight/CDFDecoder.cpp
CazYokoyama: LoadGeoTiff(Path path) throws an exception as below. Do same?
 
    auto gtif = GTIFNew(tiff.Get());
    if (gtif == nullptr)
      throw std::runtime_error("Not a GeoTIFF file");

src/Weather/Skysight/CDFDecoder.cpp
CazYokoyama: Agree. nc -> tif conversion is done in separate thread. I'll update when I become clear on how mutex protests what.

CazYokoyama: The mutex guards status which is assigned on line 66. However, it is locked for a long time. Mutex must lock as short as possible. I'll guard status when it is accessed only in the next update.

CazYokoyama: In SkysightAPIQueue::Process(),  decode_queue has nc -> tif conversions. It serialize converson/decode jobs. It obtains job status. Depending on the status, it starts decoding thread by CDFDecoder::DecodeAsync(). CDFDecoder::DecodeAsync() locks mutex and calls Trigger(). Trigger() creates a thread which executes CDFDecoder::Tick(). Tick() sets the status to Busy and unlock mutex. While the thread is running, CDFDecoder::GetStatus(), CDFDecoder::DecodeError() and CDFDecoder::DecodeSuccess() checks and set status with locking and unlocking mutex.
You will see this scenario in the next patch update.

CazYokoyama:
Max, I update the patches. I implemented all your comments except for vector/list. My reply is

MaxKellermann 27 days ago Member
Why a vector of unique_ptrs? Why not just a list?
CazYokoyama 6 days ago Author Contributor
I give up. Could you guide me how to convert a vector of unique_ptrs to a list?

Please review the updated patches and give me your guide for vector/list.

CazYokoyama:
Max, could you review the updated patches? I integrate all your comment except for vector/list which I commented as 

MaxKellermann 27 days ago Member
Why a vector of unique_ptrs? Why not just a list?
CazYokoyama 6 days ago Author Contributor
I give up. Could you guide me how to convert a vector of unique_ptrs to a list?

MaxKellermann:
I'm still confused by the first commit and its description. What does it do? What does the description mean?

CazYokoyama:
> I'm still confused by the first commit and its description. What does it do? What does the description mean?

As I wrote in the commit message, the purpose of the commit is "type-in username. password, region for skysight". However, it is created after re-factoring. I move the code to latter commits as much as possible. But there are those a lot of files are included unfortunately. This is the one where there is no compile error. It will be beyond my skill if you ask to move more code and simplify.

MaxKellermann:
> > I'm still confused by the first commit and its description. What does it do? What does the description mean?
> 
> As I wrote in the commit message, the purpose of the commit is "type-in username. password, region for skysight".

Repeating the same confusing gibberish text doesn't magically make me understand it.

CazYokoyama:
I re-phrase the message of the first commit according to advice.

Plantain:
Hi Caz,
I think the kind of commit message Max is looking for is formatted with this template:
Subject line: <Component being touched>: <One sentence of what was done>
Body: <Why was it done and why was it done in this way>

For example, I might suggest below (these are not "Max Approved" but I think they will get you closer):

SSL: Include WolfSSL library
SSL support was previously dropped from XCSoar due to licencing issues. Import WolfSSL which has a compatible licence to allow connections to hosts via SSL again.

Weather/SkySight: Implement UI for user credentials
SkySight authentication requires a username and password, these are to be retained in the profile along with the user's preferred SkySight region.

Weather/SkySight: Implement weather data download
Download may be large so it is done asynchronously

Weather/SkySight: Convert NetCDF files to GeoTIFF
SkySight uses NetCDF's as it's data distribution format. Unpack and convert these files to GeoTIFF's in preparation for drawing.

Weather/SkySight: Overlay SkySight chart

CazYokoyama:
Thank you Plantain. I updated commit messages.

src/Weather/Skysight/CDFDecoder.cpp
CazYokoyama: mutex which guards status in line 75 is unlocked on line 76 in the next patch.

src/Weather/Skysight/CDFDecoder.cpp
CazYokoyama: mutex which guards status in line 83 is unlocked on line 84 in the next patch.

src/Weather/Skysight/APIQueue.hpp
CazYokoyama: I give up. Could you guide me how to convert a vector of unique_ptrs to a list?

src/DataGlobals.hpp
```diff
@@ -40,8 +41,10 @@ class RaspStore;
  * replaced in another program.
  */
 namespace DataGlobals {
-std::shared_ptr<RaspStore> GetRasp();
-void SetRasp(std::shared_ptr<RaspStore> rasp);
+  std::shared_ptr<RaspStore> GetRasp();
```
MaxKellermann: This changes the indent

src/Dialogs/Settings/Panels/WeatherConfigPanel.cpp
```diff
@@ -49,13 +58,31 @@ class WeatherConfigPanel final
   /* methods from Widget */
   virtual void Prepare(ContainerWindow &parent, const PixelRect &rc) override;
   virtual bool Save(bool &changed) override;
+
+  private:
```
MaxKellermann: Bad indent

src/Dialogs/Settings/Panels/WeatherConfigPanel.cpp
```diff
@@ -49,13 +58,31 @@ class WeatherConfigPanel final
...
+  /* methods from DataFieldListener */
+  virtual void OnModified(DataField &df) override {};
```
MaxKellermann: What's the point of implementing the interface if you leave the method empty?

src/Dialogs/Settings/Panels/WeatherConfigPanel.cpp
```diff
@@ -49,13 +58,31 @@ class WeatherConfigPanel final
...
+  df->Set(setting);
+  wp.RefreshDisplay();
+}
+
 void
 WeatherConfigPanel::Prepare(ContainerWindow &parent, const PixelRect &rc)
 {
   const auto &settings = CommonInterface::GetComputerSettings().weather;
-
+  WndProperty *wp;
```
MaxKellermann: Move variable declaration to where it's used

src/Weather/Skysight/Settings.hpp
```diff
@@ -0,0 +1,47 @@
...
+}
+*/
+
+#ifndef XCSOAR_WEATHER_SKYSIGHT_SETTINGS_HPP
+#define XCSOAR_WEATHER_SKYSIGHT_SETTINGS_HPP
+
+#include "Util/StaticString.hxx"
+
+struct SkysightSettings {
+	StaticString<64> email;
```
MaxKellermann: Bad indent

src/Weather/Skysight/Skysight.hpp
```diff
@@ -0,0 +1,58 @@
...
+}
+*/
+
+#ifndef WEATHER_SKYSIGHT_HPP
+#define WEATHER_SKYSIGHT_HPP
+
+#include "Weather/Skysight/SkysightAPI.hpp"
+#include "Blackboard/BlackboardListener.hpp"
+
+class Skysight final: private NullBlackboardListener {
```
MaxKellermann: What is this class? Documentation missing.

src/Weather/Skysight/Skysight.hpp
```diff
@@ -0,0 +1,58 @@
...
+public:
+  tstring region = "EUROPE";
```
MaxKellermann: Build failure here

src/Weather/Skysight/Skysight.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+  static void APIInited(const tstring &&details,  const bool success,
```
MaxKellermann: Confusing function name. Documentation?
Why const rvalue references? Why `const bool`?

src/Weather/Skysight/Skysight.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include "Blackboard/BlackboardListener.hpp"
+
+class Skysight final: private NullBlackboardListener {
+public:
+  tstring region = "EUROPE";
+
+  static void APIInited(const tstring &&details,  const bool success,  
+			const tstring &&layer_id,  const uint64_t time_index);
+
+  std::map<tstring, tstring> GetRegions() {
```
MaxKellermann: Why return by value? This creates a deep copy.

src/Weather/Skysight/Skysight.hpp
```diff
@@ -0,0 +1,58 @@
...
+    return api.regions;
+  }
+
+  tstring GetRegion() {
+    return api.region;
+  }
+
+  Skysight();
```
MaxKellermann: Bad coding style

src/Weather/Skysight/Skysight.hpp
```diff
@@ -0,0 +1,58 @@
...
+  ~Skysight();
+
+  void Init();
+
+protected:
+  SkysightAPI api;
```
MaxKellermann: Bad coding style

src/Weather/Skysight/SkysightAPI.hpp
```diff
@@ -0,0 +1,50 @@
...
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+#ifndef WEATHER_SKYSIGHTAPI_HPP
+#define WEATHER_SKYSIGHTAPI_HPP
+
+#include "Metrics.hpp"
+#include "Util/tstring.hpp"
+
+class SkysightAPI final {
```
MaxKellermann: Documentation?

src/Weather/Skysight/SkysightAPI.hpp
```diff
@@ -0,0 +1,50 @@
...
+public:
+  SkysightAPI() {
```
MaxKellermann: Pointless constructor

src/Weather/Skysight/SkysightAPI.hpp
```diff
@@ -0,0 +1,50 @@
...
+  };
+  ~SkysightAPI();
+  
+  bool IsInited();
```
MaxKellermann: What's that?

src/Weather/Skysight/SkysightAPI.hpp
```diff
@@ -0,0 +1,50 @@
...
+  
+  tstring region;
```
MaxKellermann: Bad coding style

src/Weather/Skysight/SkysightAPI.hpp
```diff
@@ -0,0 +1,50 @@
...
+  
+  std::map<tstring, tstring> regions;
+  
+  void Init(const tstring email, const tstring password,
```
MaxKellermann: Why an Init method? The constructor should initialize the object.

src/Weather/Skysight/SkysightAPI.cpp
```diff
@@ -0,0 +1,54 @@
...
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+
+#include "SkysightAPI.hpp"
+#include "SkysightRegions.hpp"
+
+SkysightAPI::~SkysightAPI() {
```
MaxKellermann: Pointless empty destructor

src/Weather/Skysight/SkysightAPI.cpp
```diff
@@ -0,0 +1,54 @@
...
+SkysightAPI::IsInited()
+{
+  return inited_regions;
+}
+
+void
+SkysightAPI::LoadDefaultRegions()
+{
+  for (auto r = skysight_region_defaults; r->id != nullptr; ++r)
+    regions.insert(std::pair<tstring, tstring>(r->id, r->name));
```
MaxKellermann: Use `emplace()`

src/Weather/Skysight/SkysightAPI.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+  if (regions.find(region) == regions.end()) {
+    region = "EUROPE";
```
MaxKellermann: This forced default on error will confuse users

build/python/build/autotools.py
```diff
@@ -44,6 +44,11 @@ def configure(self, toolchain, src=None, build=None):
             build = self.make_build_path(toolchain)
 
         if self.autogen:
+            # run autogen.sh in wolfssl
```
MaxKellermann: What's that, and why do you hide it in this commit?

build/python/build/libs.py
MaxKellermann: Yes, this has nothing to do with Skysight, even though it's a precondition for the Skysight code.

src/Weather/Skysight/APIGlue.hpp
```diff
@@ -0,0 +1,58 @@
...
+  const bool  success, 
+  const tstring &&layer,
+  const uint64_t time_index
+);
+
+enum class SkysightCallType {Login, Regions, Layers, LastUpdates, DataDetails,
+			       Data, Image};
+
+struct SkysightRequestArgs {
+  const tstring url;
```
MaxKellermann: Why is this const?

src/Weather/Skysight/APIGlue.hpp
```diff
@@ -0,0 +1,58 @@
...
+  const tstring path;
+  const bool to_file;
+  const SkysightCallType calltype;
+  const tstring region;
+  const tstring layer;
+  const uint64_t from;
+  const uint64_t to;
+  const SkysightCallback cb;
+  SkysightRequestArgs(const tstring &&_url, const tstring &&_path,
```
MaxKellermann: const rvalue references

src/Weather/Skysight/APIGlue.hpp
```diff
@@ -0,0 +1,58 @@
...
+		      const bool _to_file, const SkysightCallType _ct,
+		      const tstring &&_region, const tstring &&_layer,
+		      const uint64_t _from = 0, const uint64_t _to = 0,
+		      const SkysightCallback _cb = nullptr):
+    url(_url), path(_path), to_file(_to_file), calltype(_ct),
```
MaxKellermann: You declared the parameter as rvalue, but you chose not to use move semantics.

src/Weather/Skysight/APIGlue.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include "Operation/Operation.hpp"
+
+typedef void (*SkysightCallback) (
+  const tstring &&details,
+  const bool  success, 
+  const tstring &&layer,
+  const uint64_t time_index
+);
+
+enum class SkysightCallType {Login, Regions, Layers, LastUpdates, DataDetails,
```
MaxKellermann: Bad coding style

src/Weather/Skysight/APIGlue.hpp
```diff
@@ -0,0 +1,58 @@
...
+*/
+
+#ifndef WEATHER_SKYSIGHTAPIGLUE_HPP
+#define WEATHER_SKYSIGHTAPIGLUE_HPP
+
+#include "Util/tstring.hpp"
+#include "Operation/Operation.hpp"
+
+typedef void (*SkysightCallback) (
+  const tstring &&details,
```
MaxKellermann: const rvalue reference??

src/Weather/Skysight/APIQueue.hpp
```diff
@@ -0,0 +1,55 @@
...
+  tstring key;
+  uint64_t key_expiry_time = 0;
+  tstring email;
+  tstring password;
+
+public:
+  SkysightAPIQueue(): Timer() {};
+  ~SkysightAPIQueue();
+
+  void SetCredentials(const tstring &&_email, const tstring &&_pass);
```
MaxKellermann: And const rvalue reference again

src/Weather/Skysight/APIQueue.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include "Request.hpp"
+#include "Metrics.hpp"
+#include "Event/Timer.hpp"
+#include <vector>
+
+class SkysightAPIQueue final: public Timer {
+  std::vector<std::unique_ptr<SkysightAsyncRequest>> request_queue;
+  bool is_busy = false;
+
+  void Process();
```
MaxKellermann: Bad coding style

src/Weather/Skysight/Metrics.hpp
```diff
@@ -24,6 +24,71 @@ Copyright_License {
...
+  unsigned char Green;
+  unsigned char Blue;
+};
+
+struct SkysightMetric {
+  const tstring id;
+  const tstring name;
+  const tstring desc;
+  uint64_t last_update = 0;
+  std::map<float, LegendColor> legend;
```
MaxKellermann: Eh, using `float` as a map key - that's a WTF!

src/Weather/Skysight/Metrics.hpp
```diff
@@ -24,6 +24,71 @@ Copyright_License {
...
+
+public:
+  SkysightMetric(tstring _id, tstring _name, tstring _desc):
+    id(_id), name(_name), desc(_desc) {}
+  SkysightMetric(const SkysightMetric &m):
+    id(m.id), name(m.name), desc(m.desc), last_update(m.last_update),
+    legend(m.legend) {}
+};
+
+struct SkysightActiveMetric {
```
MaxKellermann: Documentation missing

src/Weather/Skysight/Metrics.hpp
MaxKellermann: No. Pointer ownership is a very basic C++ concept, you should learn about it and then fix your code.

src/Weather/Skysight/Request.cpp
```diff
@@ -0,0 +1,277 @@
...
+  if (!success) File::Delete(temp_path);
+
+  file = NULL;
+
+  if (success) {
+    if (!File::Delete(final_path) && File::ExistsAny(final_path)) {
+      File::Delete(temp_path);
+      return false;
+    }
+    std::rename(temp_path.c_str(), args.path.c_str());
```
MaxKellermann: This will fail to compile

src/DataGlobals.hpp
CazYokoyama: Yes, the indent is changed. However, contents of namespace DataGlobals  should be indented. See whole namespace DataGlobals. If you don't think so, I am glad to make it un-indented.

src/Dialogs/Settings/Panels/WeatherConfigPanel.cpp
CazYokoyama: Will fix. Thx.

src/DataGlobals.hpp
CazYokoyama: Disregard. Change as little as possible.

src/Dialogs/Settings/Panels/WeatherConfigPanel.cpp
CazYokoyama: When I remove it, I have following error. It makes DataFieldListener::OnModified() empty. But I am not clear. I am glad if someone tell me why.

sc/Dialogs/Settings/Panels/WeatherConfigPanel.cpp: In function ‘Widget* CreateWeatherConfigPanel()’:
src/Dialogs/Settings/Panels/WeatherConfigPanel.cpp:144:33: error: invalid new-expression of abstract class type ‘WeatherConfigPanel’
   return new WeatherConfigPanel();
                                 ^
src/Dialogs/Settings/Panels/WeatherConfigPanel.cpp:51:7: note:   because the following virtual functions are pure within ‘WeatherConfigPanel’:
 class WeatherConfigPanel final
       ^~~~~~~~~~~~~~~~~~
In file included from src/Dialogs/Settings/Panels/WeatherConfigPanel.cpp:33:0:
./src/Form/DataField/Listener.hpp:31:16: note: 	virtual void DataFieldListener::OnModified(DataField&)
   virtual void OnModified(DataField &df) = 0;
                ^~~~~~~~~~

src/Dialogs/Settings/Panels/WeatherConfigPanel.cpp
CazYokoyama: Move to the one it is assigned.

src/Weather/Skysight/Settings.hpp
CazYokoyama: will fix.

src/Weather/Skysight/SkysightAPI.hpp
CazYokoyama: Could you clarify more? Similar string is defined as "const TCHAR *" in RaspStore.hpp. Are you suggesting "const TCHAR*" instead of tstring?

src/Weather/Skysight/SkysightAPI.hpp
CazYokoyama: Move the code in Init() to the constructor. As a side effect,
-  SkysightAPI api;
+  SkysightAPI *api;

src/Weather/Skysight/SkysightAPI.cpp
CazYokoyama: Will remove.

src/Weather/Skysight/SkysightAPI.cpp
CazYokoyama: Compile and run fine when I replace insert to emplace.

src/Weather/Skysight/SkysightAPI.cpp
CazYokoyama: Remove line 51 and 52. Resulting default is Andes.

build/python/build/autotools.py
CazYokoyama: wolfssh source is built by
./autogen.sh
./configure
make
as documented in its INSTALL file instead of standard automake which is expected on autotools.py.

src/Weather/Skysight/APIGlue.hpp
CazYokoyama: Remove const.

CazYokoyama: Correction.
Because it is always literal such as "https://skysight.io/api/regions".

src/Weather/Skysight/APIGlue.hpp
CazYokoyama: remove && to make them simple tstring.

src/Weather/Skysight/APIGlue.hpp
CazYokoyama: remove && to make them simple tstring.

src/Weather/Skysight/APIGlue.hpp
CazYokoyama: Make Login, Regions, .. each line.

src/Weather/Skysight/APIGlue.hpp
CazYokoyama: Remove &&.

src/Weather/Skysight/APIQueue.hpp
CazYokoyama: remove &&.

src/Weather/Skysight/APIQueue.hpp
CazYokoyama: Which of is_busy or Process(0 are youi

CazYokoyama: Which of is_busy or Process() do you mention?

MaxKellermann: Mixing fields and methods.

CazYokoyama: True. Move Process() and OnTimer() below tstring password.

src/Weather/Skysight/Metrics.hpp
CazYokoyama: I dump the key when legend is created. It is 3.5, 2400, -600, 1.6, -3.5, ... It is really floating point. What is your point? Key should not be floating point?

src/Weather/Skysight/Metrics.hpp
CazYokoyama: /*
 * Skysight chart which is overlaid
 */

src/Weather/Skysight/Request.cpp
CazYokoyama: I build on Android, gcc 7.4.0, gcc 8.3.0. No error. Could you explain more?

CazYokoyama:
Max, I responded almost all your review comments include questions for you. Is there anything I can do for you? Just update the branch(i.e. force push)?

src/Dialogs/Settings/Panels/WeatherConfigPanel.cpp
MaxKellermann: I did not ask about the method. I asked about the interface.

src/Weather/Skysight/SkysightAPI.hpp
MaxKellermann: Mixed fields and methods is the bad coding style.

src/Weather/Skysight/APIGlue.hpp
MaxKellermann: Why remove &&?

src/Weather/Skysight/APIQueue.hpp
MaxKellermann: Why remove &&?

src/Weather/Skysight/Metrics.hpp
MaxKellermann: No, keys should not be floating point. Maybe a `map` is the wrong data structure here. I don't know, because you forgot to add documentation here, so nobody knows what you're trying to do.

src/Weather/Skysight/Metrics.hpp
MaxKellermann: I have no idea what this means.

src/Weather/Skysight/Request.cpp
MaxKellermann: Build Windows.

MaxKellermann:
Yes, always keep this rebased and force-push when you're done.

build/python/build/libs.py
CazYokoyama: I'll create a new commit as the first patch which has "add ssl authentication, https by wolfssl" as commit message.

src/Weather/Skysight/SkysightAPI.hpp
CazYokoyama: Will separate variables and functions.

CazYokoyama:
Sorry, I gave up. I can't give what Max wants due to 1) I am not a original author, 2) my limited skill on C++. I'll force push my latest code which does not fulfill his 5 requests/questions.
After a couple of weeks, I'll close this PR. I am glad if someone wants to continue to work.

MaxKellermann:
> I gave up.

No thanks for wasting my time. This whole exercise turns out to be pointless.

Hacking some quick'n'dirty code is easy and creates quick "wow" effects among users, which is exciting and makes you a rock star. But bringing that dirty code into shape is boring, and nobody will thank you. That test you failed.

CazYokoyama:
Max, I agree that I wasted your time. Sorry.

As a next step, there is another way to upstream which is "re-write code from scratch". I hope someone raise their hand or my guts is re-activated.

PR #301, opened 2019-12-11
Skysigh's nc file is converted to tif file by netcdf

CazYokoyama:


MaxKellermann:
I don't understand any of this.

CazYokoyama:
To solve build error on https://github.com/XCSoar/XCSoar/pull/298, i.e. build error is caused by missing those libraries. However, it should be done by something else. Will close.

PR #304, opened 2019-12-20
Read primary and secondary frequency from ACD to display in XCS infobox

Scumi:
As described in the title, this code change in the ACD driver makes use of the newest Infobox addition for the radio frequencies.

### DIFF
--- a/src/Device/Driver/AirControlDisplay.cpp
+++ b/src/Device/Driver/AirControlDisplay.cpp
@@ -57,8 +57,38 @@ ParsePAAVS(NMEAInputLine &line, NMEAInfo &info)
       auto qnh = AtmosphericPressure::Pascal(value);
       info.settings.ProvideQNH(qnh, info.clock);
     }
+  } else if (StringIsEqual(type, "COM")) {
+    /*
+    $PAAVS,COM,<CHN1>,<CHN2>,<RXVOL1>,<RXVOL2>,<DWATCH>,<RX1>,<RX2>,<TX1>
+     <CHN1> Primary radio channel;
+            25kHz frequencies and 8.33kHz channels as unsigned integer
+            values between 118000 and 136990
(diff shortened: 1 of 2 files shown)


MaxKellermann:
If your PR contains text, it means that your commit message is bad. And indeed it is:
> Reads primary and secondary frequency from ACD to display in XCS infobox

Reads, who reads something? What is ACD? What frequency? Is life really too short to write XCSoar instead of XCS?

What about: "Driver/AirControlDisplay: read radio frequency from PAAVS,COM sentence"
This is clearer **and** shorter than yours!

rawtaz:
Please, consider how you come across when writing messages to others.. Someone is trying to make a contribution, instead of telling them how *wrong* things they did are, focus on how the things they did can be *improved* (that is, with less focus on how bad/wrong it is).

There's also something to be said about e.g. using abbreviations. The first line of a commit message should be short and concise, so it's not unreasonable to use abbreviations in there (although I agree that XCSoar should be written out). If any abbreviation is too uncommon or need clarification for whatever reason, one can write a more elaborate description of the commit on the third line (i.e. with a blank line between the first line and the third line).

MaxKellermann:
@rawtaz apparently it's not enough to have you on my IRC ignore list, I have to block you on GitHub as well.

Scumi:
I don't know what happened between you two, for me Max` comments are fine. Will provide an update soon.

rawtaz:
@MaxKellerman I have no idea what you are talking about. I haven't done anything to harm you nor the project, and there is no reason what so ever for you to ignore me, on the IRC, GitHub or anywhere. This is a fact.

rawtaz:
@Scumi Me neither. There is really nothing between me and Max :) I consider him a great developer, I have nothing bad to say about him at all.

PR #306, opened 2019-12-24
introduce a set of 10 InfoBoxes to improve readability

bomilkar:
Improve readability of Title and Comment in InfoBoxes.
It is useful in particular for high and narrow screen orientation like OpenVario in portrait orientation.
To achieve this, I introduce a set of 10 InfoBoxes. This fits in between the existing layouts for 8 and 12 InfoBoxes. This gives a better balance/tradeoff between readability and number of InfoBoxes.
This patch introduces 3 additional screen layouts: TOP_LEFT_10, BOTTOM_RIGHT_10, SPLIT_10.


mlep:
@bomilkar, please, could you had this feature to the documentation?

bomilkar:
Good point! However, the manual is sufficiently vague so that it applies to the new layouts, too.
In chapter 11 it says:
[InfoBox geometry]  A list of possible InfoBox layouts. Do some trials to find the
  best for your screen size. The preceding number refer to the total number of
  InfoBoxes of this geometry.

I think that's good enough. Right, @mlep ?

mlep:
Right! Thank you for the followup.

bomilkar:
@mlep, can I ask you a question, since this is my first feature addition:
I see it has 2 branches: master and 6.8.x. Am I supposed to initiate a "pull-request" on the 6.8.x branch, too? I don't expect any conflicts in the code. I just want to do the right thing.

bomilkar:
@mlep, I want to add one more feature. I guess I should open a new branch. Probably on the XCSoar master. Right?
I hope it doesn't interfere with #306. In fact, it is supposed to be independent of #306.

mlep:
@bomilkar I am mainly contributing to the documentation, so I have no idea about the suitable place for code.
@lordfolken, please, could you provide advises?

Scumi:
According to the developer manual section 3.1.2:

> Developers should commit changes for the next minor release to the master branch. Bug fixes of the current minor release should be committed to the current minor branch, in preparation for the next patch release.

Since its not a bugfix, master-branch is the one to go.

bomilkar:
It just occurred to me that the values, labels, help texts, etc. should be **translated**. But I can't figure out how to do that. I can supply German and French, but who does all the other languages?
I found the stuff in the po folder. But that looks machine generated, at least the lines referring to source files and line numbers.
@mlep, @Scumi  please help.

MaxKellermann:
Your way of splitting the PR into 4 commits doesn't appear to make sense.
And it creates a `README.md` file describing the feature, which is the wrong place to add feature documentation.

bomilkar:
Sorry Max, this is my first feature and I got lost in the git flow between command line git and the browser based GitHub.

I probably missed the right way posting #311, too. This is meant to be a stand alone feature just like this here. But I guess I mixed it up because I didn't branch for 306 after the fork.

Is the first comment the right place to describe the intend? In that case README.md is redundant.
What exactly can I do to comply with rules. Is there a way to fix this here, or do you want me to close it altogether?
As far as I understand the communication happens on GitHub. But I'd like to edit, test, commit, on my Debian machine. Is that possible? Unfortunately the Developers' Guides (both 6.8. and 7.7) don't go into that level of detail. An example flow would be very useful (for the beginners).

PR #309, opened 2019-12-26
Chapter 4 xc tasks: Update based on XCSoar-testing

mlep:
Checking the manual against XCSoar-testing 7.0_preview14.
Removing some information and adding some others.

Plus some syntax improvements.

### DIFF
--- a/doc/manual/en/ch04_xc_tasks.tex
+++ b/doc/manual/en/ch04_xc_tasks.tex
@@ -1,3 +1,6 @@
+% !TeX encoding = utf8
+% !TeX spellcheck = en
+
 \chapter{Cross Country Tasks}\label{cha:tasks}
 
 XCSoar provides a full task management system, in which tasks can be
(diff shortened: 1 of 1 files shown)


doc/manual/en/ch04_xc_tasks.tex
```diff
@@ -46,9 +49,9 @@ \section{Editing tasks}
 editing prior to flight, and others allow tasks to be modified whilst
 in flight for casual cross-country touring.  Tasks can be saved to
 files and reloaded later, and can be transferred between any XCSoar
-platform (Android, Windows, Linux, ...).
+platform (Android, Windows, Linux\textellipsis).
```
MaxKellermann: This removes the comma. Did you really want to do that?

mlep: Yes it was. I think a comma is redundant before an ellispsis.
(it will not be the case in math, such as: 1, 2, 3, ... , 4, 5.)

PR #311, opened 2019-12-27
user controllable InfoBox vs. screen ratio

bomilkar:
Improve readability of InfoBoxes and still give maximum screen space to the Map

This feature makes the constant CONTROLHEIGHTRATIO in InfoBoxes/InfoBoxLayout.cpp user-accessible.


bomilkar:
It seems as if I had committed/pushed before switching to the proper branch. Therefor the commit which I expected here ended up in #306. If that's the case, can it be fixed?
I'm sorry if I messed it up.

PR #320, opened 2020-01-03
KRT2: fix crash when changing frequency from infobox

fb:
Crash is gone but changing frequency from infoboxes still only works rudimentary at most.

Tested with TARGET=UNIX, KRT-2 v8.2 connected via FT232. 

8679f44 is not directly related but was needed to find the bug.

@roeles fiy


MaxKellermann:
I don't understand your approach to fix the problem. The commit message says the bug is in the KRT2 driver, yet you change the API of a core interface and its callers.

fb:
As you pointed out, I was addressing the crash by changing the caller. Based on your feedback I rewrote the patch to get rid of the crash without changing too much elsewhere.

MaxKellermann:
Build fails:
```
  CXX     output/PC/dbg/src/Device/Driver/KRT2.o
src/Device/Driver/KRT2.cpp: In static member function ‘static void KRT2Device::GetStationName(char*, const TCHAR*)’:
src/Device/Driver/KRT2.cpp:382:14: error: cannot convert ‘const char [1]’ to ‘const TCHAR*’ {aka ‘const wchar_t*’} in assignment
       name = "";
              ^~
make: *** [build/compile.mk:108: output/PC/dbg/src/Device/Driver/KRT2.o] Error 1
```

roeles:
@fb one of the platforms, which is not in github-ci, wants an explicit conversion from char to TCHAR. 
I think I encountered that before, and that the solution was `name = _T("");`

fb:
I added the `_T` macro and it can be built with mingw. Thanks, @roeles.

PR #323, opened 2020-01-06
Update Flight Level infobox description

DanD222:
- Previous description implied that QNH needs to be set (this is not the case if barometric altitude is available -> if barometric altitude is available, the FL value is referenced to 1013 hPa)

- Description changed to take into account fallback FL calculation from GPS altitude (in this case the QNH does need to be set)

- Description changed to make user aware that red "FL" label indicates fallback GPS flight level calculation

### DIFF
--- a/doc/manual/en/ch10_infobox_reference.tex
+++ b/doc/manual/en/ch10_infobox_reference.tex
@@ -53,7 +53,7 @@ sea level obtained from the terrain file at the current GPS location.}
 \ibi{Height above take-off}{H T/O}{Height based on an automatic take-off reference 
 elevation (like a QFE reference).\footnotemark[1]}
 \ibi{Flight level}{Flight Level}{Pressure Altitude given as Flight Level. 
-Only available if barometric altitude available and correct QNH set.\footnotemark[1]}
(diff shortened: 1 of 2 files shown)


MaxKellermann:
If your PR contains text, it means that your commit messages are bad.

DanD222:
Thanks for the feedback - I've added the notes from the PR to the commit messages.

fb:
The commit messages are identical, yet there are two different commits. Just squash them into a single commit.

bomilkar:
@DanD222, I noticed "fallback FL calculation from GPS altitude" in the PR comment above.
GPS altitude and barometric altitude can differ a lot. Easily more than 100ft on a hot summer day, for instance. An airspace violation can get dangerous or at least expensive. I'd propose to not display the FL unless barometric altitude is available.

I know this PR is about description and documentation. But I'd like to open this discussion before we fix the description on a broken feature.

DanD222:
@bomilkar It's a valid point. My opinion is that the red FL label warns the user that the value may not be very accurate, and thus allows him to err on the side of caution. I think this is a good compromise and is more useful than having no value at all.

PR #324, opened 2020-01-12
add a set of 10 InfoBox layouts

bomilkar:
Improve readability of Title and Comment in InfoBoxes.
It is useful in particular for high and narrow screen orientation like OpenVario in portrait orientation.
To achieve this, I introduce a set of 10 InfoBoxes. This fits in between the existing layouts for 8 and 12 InfoBoxes. This gives a better balance/tradeoff between readability and number of InfoBoxes.
This patch introduces 3 additional screen layouts: TOP_LEFT_10, BOTTOM_RIGHT_10, SPLIT_10.

(This is a new attempt to submit this feature. Previously know as #306)


bomilkar:
Thanks Max.
I'll go ahead an post the 2nd PR (previously known as #311)

PR #325, opened 2020-01-12
user controllable InfoBox to screen ratio

bomilkar:
Improve readability of InfoBoxes and still give maximum screen space to the Map

This feature makes the constant CONTROLHEIGHTRATIO in InfoBoxes/InfoBoxLayout.cpp user-accessible.

(Previously filed as #311 which was closed)

### DIFF
--- a/src/InfoBoxes/InfoBoxLayout.cpp
+++ b/src/InfoBoxes/InfoBoxLayout.cpp
@@ -22,11 +22,10 @@ Copyright_License {
 */
 
 #include "InfoBoxes/InfoBoxLayout.hpp"
+#include "Interface.hpp"
 #include "Border.hpp"
 #include "Util/Macros.hpp"
 #include "Util/Clamp.hpp"
-
-static constexpr double CONTROLHEIGHTRATIO = 7.4;
 
 /**
  * The number of info boxes in each geometry.
@@ -409,18 +408,22 @@ InfoBoxLayout::ValidateGeometry(InfoBoxSettings::Geometry geometry,
   return geometry;
 }
 
-static constexpr unsigned
+static unsigned
 CalculateInfoBoxRowHeight(unsigned screen_height, unsigned control_width)
 {
-  return Clamp(unsigned(screen_height / CONTROLHEIGHTRATIO),
+  const double i2s = CommonInterface::GetUISettings().info_boxes.ib_to_screen_ratio;
+
+  return Clamp(unsigned(screen_height / i2s),
                control_width * 5 / 7,
                control_width);
 }
 
-static constexpr unsigned
+static unsigned
 CalculateInfoBoxColumnWidth(unsigned screen_width, unsigned control_height)
 {
-  return Clamp(unsigned(screen_width / CONTROLHEIGHTRATIO * 1.3),
+  const double i2s = CommonInterface::GetUISettings().info_boxes.ib_to_screen_ratio;
+
+  return Clamp(unsigned(screen_width / i2s * 1.3),
                control_height,
                control_height * 7 / 5);
 }
--- a/src/Profile/InfoBoxConfig.cpp
+++ b/src/Profile/InfoBoxConfig.cpp
@@ -123,6 +123,10 @@ Profile::Load(const ProfileMap &map, InfoBoxSettings &settings)
     break;
   }
 
+  map.Get(ProfileKeys::InfoBoxScreenRatio, settings.ib_to_screen_ratio);
+  if (settings.ib_to_screen_ratio < 5.99 || settings.ib_to_screen_ratio > 10.01)
+    settings.ib_to_screen_ratio = 7.4;
+
   map.Get(ProfileKeys::AppInverseInfoBox, settings.inverse);
   map.Get(ProfileKeys::AppInfoBoxColors, settings.use_colors);
 
--- a/doc/manual/de/ch11_configuration.tex
+++ b/doc/manual/de/ch11_configuration.tex
@@ -684,6 +684,11 @@ angepaßt werden.
 \begin{description}
 \item[\textit{InfoBox Geometrie}]  Eine Auswahlliste von mehreren Anordnungen der InfoBox Seiten Je nach Gerät und persöhnlichem
 Geschmack muß dies schlicht ausprobiert werden. Die angebene Zahl entspricht der Anzahl der InfoBoxen auf der entsprechenden InfoBoxSeite.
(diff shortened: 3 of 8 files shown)


bomilkar:
It just occurred to me that the values, labels, help texts, etc. should be translated. But I can't figure out how to do that. I can supply German and French, but who supplies all the other translations?
I found the stuff in the po folder. But that looks to me like machine generated, at least the lines referring to source files and line numbers.
Please help.

MaxKellermann:
> It just occurred to me that the values, labels, help texts, etc. should be translated. But I can't figure out how to do that. I can supply German and French, but who supplies all the other translations?

https://hosted.weblate.org/projects/xcsoar/

To be able to translate, we need to generate a new POT file (= PO template). This happens some time after your code got merged.

src/InfoBoxes/InfoBoxLayout.cpp
```diff
@@ -22,12 +22,11 @@ Copyright_License {
 */
 
 #include "InfoBoxes/InfoBoxLayout.hpp"
+#include "Interface.hpp"
```
MaxKellermann: Don't include this header. `InfoBoxLayout.cpp` is supposed to be a low-level library without too many dependencies, and `Interface.hpp` is maybe the heaviest dependency of all.
Instead, pass this new setting as another parameter.
For the same reason, the `InfoBoxSettings::Geometry` is passed as parameter instead of loading it from the global variables.

src/Profile/InfoBoxConfig.cpp
```diff
@@ -123,6 +123,10 @@ Profile::Load(const ProfileMap &map, InfoBoxSettings &settings)
     break;
   }
 
+  map.Get(ProfileKeys::InfoBoxScreenRatio, settings.ib_to_screen_ratio);
+  if (settings.ib_to_screen_ratio < 5.99 || settings.ib_to_screen_ratio > 10.01)
+    settings.ib_to_screen_ratio = 7.4;
```
MaxKellermann: Why set this arbitrary value if the configured value is out of range? Why not clamp?

MaxKellermann:
I'm not quite convinced that this setting is really necessary. Does anybody else have an opinion?

src/Profile/InfoBoxConfig.cpp
bomilkar: 7.4 was the original value (hard coded in InfoBoxes/InfoBoxLayout.cpp).
The addition here is meant to initialize the value to what the user had seen before.

MaxKellermann: That should sure be the default value until the user has configured something else. But that's not what your code expresses to do. It does that, but in a very obscure way.

bomilkar:
@MaxKellermann , the value becomes clear if it runs on a tall and narrow screen, like many new SmartPhones. If the InfoBox is 1/7.4 of the hight of the screen, then it consumes a lot of space at the expense of the map window.
See attached screenshot, one with InfoBoxScreenRatio=10 and one with 7.5. View it side by side.
InfoBoxScreenRatio=10
InfoBoxScreenRatio=7 5

MaxKellermann:
If that's the problem to be solved, then this is the wrong approach to do this!

Let's say you have chosen a larger value because your screen is very tall. Now you rotate your screen, swapping the aspect ratio. Now this large value is extremely inappropriate - for this display orientation, you'd need a very small value, not a very large one. So now you have a knob which is always wrong.

The whole idea of this `CONTROLHEIGHTRATIO` was wrong in the first place, but your code steps aside from fixing it, and instead adds a new complicated unintuitive knob. That's a kludge, not a solution.

The `CONTROLHEIGHTRATIO` should be replaced by a meaningful formula to calculate an InfoBox size which is always good, no matter how large your screen is and no matter its aspect ratio.

src/InfoBoxes/InfoBoxLayout.cpp
bomilkar: OK, I understand. I thought CommonInterface::GetUISettings().info_boxes.ib_to_screen_ratio would compile to not much more than a memory read.

I need to see how to fix it. May take some time.

MaxKellermann: > I thought CommonInterface::GetUISettings().info_boxes.ib_to_screen_ratio would compile to not much more than a memory read.

Your thought is correct. But this is not about runtime overhead, it's about API bloat (on the source code level, not at runtime). Adding a runtime dependency on `Interface.hpp` means we cannot write a unit test for this code, for example.

bomilkar:
I agree with you, @MaxKellermann .
Let me think about it.

Can we start with the assumtion that the InfoBoxs should never be :
higher than needed in portrait orientation
wider than needed in landscape orientation
So in portrait orientation use padding to adjust the width to the with of the screen.
And in landscape use padding to adjust the hight.
Is that right?

bomilkar:
I tried the above: I don't like it. It wastes sceen space to the detriment of the map window.

@MaxKellermann : your assumption ".. large value is extremely inappropriate - for this display orientation, you'd need a very small value, not a very large one" is incorrect. Larger numbers give more space to the map window, regardless of the orientation.

I keep coming back to my original idea, but pick a different name for it:
"InfoBox size" ranging from 50% to 150%
That's intuitive. Right?

In portrait the width of InfoBoxes is determined by the number in the row. The hight scales with the "InfoBox size".
In landscape vice versa.

src/InfoBoxes/InfoBoxLayout.cpp
bomilkar: Do you want me to add a 3rd parameter to InfoBoxLayout::Calculate() ??

bomilkar: Or just add a variable to the struct InfoBoxLayout::Layout ??
(But then I have to include  InfoBoxLayout into Profile/InfoBoxConfig.cpp which might be wash.)

src/Profile/InfoBoxConfig.cpp
bomilkar: OK.

MaxKellermann:
> I keep coming back to my original idea, but pick a different name for it:
> "InfoBox size" ranging from 50% to 150%
> That's intuitive. Right?

But ... why? You described a problem, and I understand the problem, but this problem can be solved with a better layout formula without asking for the user's opinion. So why do you still want to add a new configuration option?

MaxKellermann:
See #327 for continued discussion about this.

bomilkar:
Whatever the split between InfoBox area and map area is, different users will have different opinions on what looks better (on their device and configuration). Therefore the split should be user controllable (within reasonable limits) and cannot be fully automated. In my opinion a magic formula to please everybody doesn’t exist.

PR #327, opened 2020-01-15
user controllable InfoBox scaling

bomilkar:
Improve split between InfoBoxes and Map window

In essence, this replaces the constant CONTROLHEIGHTRATIO in InfoBoxes/InfoBoxLayout.cpp and introduces a config variable "InfoBox size" to control the free dimension of the InfoBox row.
There is no "golden" value since it depends a lot on user preference and taste.
It may not be considered a big improvement, but to me it's worth while.

This PR is an improved version of #325 
It addresses the issues Max had flagged (I hope)
It makes it more intuitive to the user to pic the right choice for the InfoBox size.

### DIFF
--- a/doc/manual/de/ch11_configuration.tex
+++ b/doc/manual/de/ch11_configuration.tex
@@ -684,6 +684,10 @@ angepaßt werden.
 \begin{description}
 \item[\textit{InfoBox Geometrie}]  Eine Auswahlliste von mehreren Anordnungen der InfoBox Seiten Je nach Gerät und persöhnlichem
 Geschmack muß dies schlicht ausprobiert werden. Die angebene Zahl entspricht der Anzahl der InfoBoxen auf der entsprechenden InfoBoxSeite.
+\item[\textit{InfoBox Grösse}] Die Größe der InfoBox in X- oder Y-Richtung als Prozentwert der normalen Größe.
+In der freien Richtung verschiebt sich die Grenze zwischen InfoBoxen und Kartenfenster.
+Bei sehr kleinen Werten kann, je nach Anzahl der InfoBoxen und Bildschirmorientierung, der Text in den InfoBoxen ineinander laufen.
+Voreinstellung ist 100% .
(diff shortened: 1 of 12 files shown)


MaxKellermann:
See https://github.com/XCSoar/XCSoar/pull/325#issuecomment-574612889
Your PR does not answer my question.

bomilkar:
Let me restate the intention and rational. I just don’t want to waste everybody’s time on something that isn’t perceived valuable.

1. The intention is to optimize the screen area allocation between the row(s) of InfoBoxes the map window. The improvement is not going to be huge, but large enough to make a difference. Do you agree?

2. Whatever the split between InfoBox area and map area is, different users will have different opinions on what looks better. Therefore the split should be user controllable (within reasonable limits) and cannot be fully automated. A magic formula to please everybody doesn’t exist. Do you agree?

3. To control the split between InfoBox and map area I want to introduce a setup parameter “InfoBox size” ranging from 50% to 150%. (Very much like the “Text size” parameter.) This should be intuitive to most users. The effect will be that it moves the free border between InfoBox and map area moves. (In one dimension there is a movable border. In the other dimension the InfoBox size is determined by the number of InfoBoxes and the size of the screen.) Do you agree with that user model?

MaxKellermann:
>     1. The intention is to optimize the screen area allocation between the row(s) of InfoBoxes the map window. The improvement is not going to be huge, but large enough to make a difference. Do you agree?

What do you want me to agree on? I don't agree with this code change.

>     2. A magic formula to please everybody doesn’t exist. Do you agree?

This is true for anything. There will always be somebody who isn't pleased, and this someone is always the loudest user. Just because one hard-coded formula doesn't please everybody doesn't mean we should make it configurable. With that reasoning, we must make everything configurable (and I mean everything everything), but we have to draw a line somewhere to reduce XCSoar's configuration complexity. Complexity will make more users unhappy. XCSoar is too complicated already.

>     3. [...] Do you agree with that user model?

No. I think we need a better formula which keeps InfoBoxes at roughly the same shape, independent of your screen's aspect ratio. But we don't need any user setting.

PR #329, opened 2020-01-17
fix InfoBox geometry for square layout

bomilkar:
Bugfix: it ignores the “InfoBox geometry” parameter when the screen is exactly square.
I eliminate the square case, which is now treated the same as portrait.
Observe the bug: running “xcsoar -square”, then resize the window to portrait or landscape. The square case has 5 InfoBoxes regardless of the “InfoBox geometry” setting.

### DIFF
--- a/src/InfoBoxes/InfoBoxLayout.cpp
+++ b/src/InfoBoxes/InfoBoxLayout.cpp
@@ -358,11 +358,8 @@ InfoBoxLayout::ValidateGeometry(InfoBoxSettings::Geometry geometry,
     case InfoBoxSettings::Geometry::TOP_8_VARIO:
       return InfoBoxSettings::Geometry::LEFT_6_RIGHT_3_VARIO;
     }
-  } else if (screen_size.cx == screen_size.cy) {
-    /* square */
-    geometry = InfoBoxSettings::Geometry::RIGHT_5;
   } else {
-    /* portrait */
+    /* portrait and square */
 
     switch (geometry) {
     case InfoBoxSettings::Geometry::SPLIT_8:


MaxKellermann:
If your PR has text, it means that your commit message is bad.

bomilkar:
Max, I don't understand your comment. I see the text as the first comment in this PR "Bugfix: it ignores the “InfoBox geometry” parameter.... ". And I see the commit message "infobox-fix-square-screen".
What is missing?

MaxKellermann:
Never attach text to a PR! A good PR with good commits with good commit messages doesn't need further explanation.

Unfortunately, your commit message isn't good. It says "fix InfoBox geometry for square layout". But it doesn't explain what was broken, what is fixed, how it was fixed and why.

Your PR text doesn't count. Don't bother write one, I won't read it (for good reasons).

bomilkar:
OK I understand. BTW: the PR comment does state what is broken and what I did to fix it.

Anyway, the changes in the code are simple enough to explain the change. However, I can't judge if the wired behavior was intended in the first place. And if it was, it was never documented.

DanD222:
@bomilkar I think the intention is to have documentation about what is being done and why stored in the commit message rather than in the pull request description, as the commit messages are what is kept in the project's code history.

bomilkar:
@DanD222 , thanks for your help. I really appreciate it.
I added a comment to the commit (I hope).

DanD222:
@bomilkar No problem, glad I can help. It doesn't seem like you've amended your commit message though: if you click on this link https://github.com/XCSoar/XCSoar/pull/329/commits/d506aa31d4ddec2471d213a62f0f7e97923a4214, you'll see there's still no text besides "_fix InfoBox geometry for square layout_" in the blue box, which is where your commit message should be.

I would recommend downloading a Git GUI client like Git Kraken and linking it to your GitHub account - you'll then be able to see your last commit and amend it, see the "Amending commits" section here: https://support.gitkraken.com/working-with-commits/commits/

For this change to manifest itself in your pull request, you'll then need to Force Push your locally-amended commit to your online repository. Your pull request will then be updated with your amended commit.

DanD222:
@bomilkar OK, it now looks like you have three separate commits in your pull request, so what you can do is Squash them into one, as per here: https://support.gitkraken.com/working-with-commits/squash/ 

Make sure to do Force Push afterwards, as this will delete the unwanted commits in your pull request that don't have the commit messages, which is what we want.

fb:
PR looks okay to me.

May I suggest to remove the "-square" option alltogether at least for the master branch? Does anybody use it?

bomilkar:
There is nothing wrong with the "-square" option. It doesn't hurt. Why not keep it?
Just to be clear: I'm suggesting a fix a bug in InfoBoxLayout::ValidateGeometry(), not to eliminate the "-square" option.

PR #330, opened 2020-01-17
WIP: Improve usabilty with hardware buttons as input device

fb:
developer manual 7.2.4 Usability:
> Ensure all dialogs are navigable using cursor keys only

On the PC version, I simulated things I would like to do in flight using a stick remote control.

At the moment there are lots of things that should be possible using only arrow keys, Return and ESC but do not work at the moment.

I hope to get some ideas and feedback from other users/developers about this. 

(incomplete) list of actions
* alternates dialog
* waypoint dialog
* Task editor: Swap waypoints
* interacting with infoboxes (#172, #191)
* FlarmTrafficWindow: <, >, Details (zooming works)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -57,6 +57,7 @@ The following people and organizations have contributed code to XCSoar:
  Peter F Bradshaw <pfb@exadios.com>
  Stefan Schumann <stefan.schumann@posteo.de>
  Roel Baardman <roel.baardman@gmail.com>
+ Fabian Bartschke <fabian@bartschke.de>
 
 Documentation:
 
--- a/src/Form/Button.cpp
+++ b/src/Form/Button.cpp
@@ -239,7 +239,7 @@ Button::OnPaint(Canvas &canvas)
 
   const bool pressed = down;
   const bool focused = HasCursorKeys()
-    ? HasFocus() || (selected && !HasPointer())
+    ? HasFocus() || selected
     : pressed;
 
   renderer->DrawButton(canvas, GetClientRect(),
--- a/src/Form/Button.hpp
+++ b/src/Form/Button.hpp
@@ -47,9 +47,8 @@ class Button : public PaintWindow {
   /**
    * This flag specifies whether the button is "selected".  The
    * "selected" button in a #ButtonPanel is the button that will be
-   * triggered by the #KEY_RETURN.  On some devices without touch
-   * screen, cursor keys left/right can be used to navigate the
-   * #ButtonPanel.
+   * triggered by the #KEY_RETURN. Cursor keys left/right can be used to
+   * navigate the #ButtonPanel.
    */
   bool selected;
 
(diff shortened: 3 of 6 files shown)


src/Dialogs/Task/AlternatesListDialog.cpp
```diff
@@ -124,7 +124,7 @@ AlternatesListWidget::Prepare(ContainerWindow &parent, const PixelRect &rc)
 void
 AlternatesListWidget::OnActivateItem(unsigned index)
 {
-  details_button->Click();
+  // details_button->Click();
 }
```
fb: This action will catch the is a key event when using the keyboard.
However by simply disabling it, clicking the list item with a pointing device is not possible anymore.

Clicking a list item should still open the details dialog.

DanD222: There's currently an issue with using the keyboard in the Alternates screen, in that you can only select the first or the last alternate in the list to navigate to using just the keyboard.

This is because there is no way to move to the Goto button from say the second or the third alternate in the list - I would expect the left / right arrow keys to do this, but instead they move the cursor to the first and last alternate in the list.

_(You can press the Enter key on an alternate to bring up the waypoint details, but then there's no Goto button in the waypoint details.)_


Am I right in assuming that your fix would allow the cursor to move between the Goto / Details / Close buttons and the alternates list using the arrow keys, from any position in the alternates list?

DanD222:
There's currently an issue with using the keyboard in the Alternates screen, in that you can only select the first or the last alternate in the list to navigate to using just the keyboard.

This is because there is no way to move to the Goto button from say the second or the third alternate in the list - I would expect the left / right arrow keys to do this, but instead they move the cursor to the first and last alternate in the list.

_(You can press the Enter key on an alternate to bring up the waypoint details, but then there's no Goto button in the waypoint details.)_

Am I right in assuming that your fix would allow the cursor to move between the Goto / Details / Close buttons and the alternates list using the arrow keys, from any position in the alternates list?

fb:
> Am I right in assuming that your fix would allow the cursor to move between the Goto / Details / Close buttons and the alternates list using the arrow keys, from any position in the alternates list?

Yes. *Left - Right* moves focus on the buttons. *Up - Down* scrolls the list. *Return* presses the focused button.

lordfolken:
Flight Setup Dialog
ballast press **down** you go to "wing loading" which is a non-changable value.
moving all the way down to the buttons:
you need to use up/down to change between dump|ok. Intutive would be left/right.

lordfolken:
Select waypoint dialog:
up/down selects the attributes/values ok
jumping to th elist is only possible by going down past "cancel" (this seems to be common for all the dialogs containing a list)

lordfolken:
F1 quick menu:
if an item is grayed out in a column (such as flarm radar) pressing down willl actually not skip that item, but jump to the central column.

mihu-ov:
Flight download from devices, it´s not possible to select the device and then download with arrow keys and return.

fb:
I will close this PR and summarize the disscussion in a new issue.

Issue #322, opened 2020-01-03
KRT2: Radio infboxes do not reflect changes made on the KRT


fb:
The current status is:
1. change standby frequency on KRT -- frequencies in infoboxes do not change
2. swap frequencies on KRT -- frequencies in infoboxes do not change

With https://github.com/XCSoar/XCSoar/pull/320/commits/ba75c2dd3d2be7ab1fca77b16a8dc6c7854004a3, the standby frequency will be updated when changed on the KRT.
In addition, the `C` command *Exchange of frequencies* would have to be handled by the driver. 

To be honest, I'm not sure if this issue is worth the effort. I could live without frequency infoboxes, so my personal solution would be to just remove the infoboxes :smiling_imp:

Everything I need is already working:
- I have a waypoint file containing frequency details.
- I can change to a frequency from that file with XCSoar without needing to touch the radio
- I can see the momentary value of active/standby frequency on KRT display

@roeles what are your thoughts?

roeles:
@fb Thanks for reporting!
I really want this functionality, because my KRT2 is somewhat hidden in my panel. It's hard to read, hence the need to mirror the current values on Infoboxes.

I want this code to remain in XCSoar and will invest my time into fixing/maintaining it.

**Question:** In an email to me you mentioned that setting the frequency from the Infobox panel did not work (or not all the time). Is that still true?

fb:
Okay. I understand that this feature is helpful if you can't see the display otherwise.

Regarding the first question see #321. I will add some details.

I do not have any load (or antenna) attached at the moment. Also no PTT button. To be on the safe side you could use a dummy in place of an antenna.

fb:
I think `SetComputerSettings().radio.active_frequency` should be only used to keep track of what is the momentary frequency setting.
Those properties were introduced together with the infoboxes here: https://github.com/XCSoar/XCSoar/commit/3481aa44a886119df1cf4461e719b295b58ec6ac#diff-eb136b0530683cd9dd72c7d6721e8227

`ActionInterface::SetActiveFrequency()` should not update `radio`. Device drivers  will take care of that (using `devices->PutActiveFrequency()`) making it a read only property from the user perspective.

- Change frequency directly on the radio: Device driver will update info boxes
- Change frequency from infobox: Frequency will be sent to radio. When it was sucessful, device driver will trigger info box update.
- Change frequency from waypoint menu: same

[...]

Issue #321, opened 2020-01-03
KRT2: Changing radio frequency from infobox does not work


fb:
I'm not able to change the frequency using the frequency infoboxes.

Tested setup: KRT-2 connected via RS232 to a computer running TARGET=UNIX.

fb:
In the current master, reading the frequencies from the KRT does not work at all. Without this feature the infoboxes will be empty and you simple won't be able to offset the frequency. I worked on that a bit and made some modifications. A part of this is already in PR #320.

The other problem is that we are trying to offset the frequencies in steps of 1000 KHz and 5 KHz. As long as the infobox is not empty and no radio is attached, adjusting the value in small or large steps will work. However this will result in invalid channels like 118.095 MHz.

In reality the frequencies are 8.33 KHz from each other which are rounded to 5 KHz in maps, ATC,  … So the gap can be 5 KHz or 10 KHz.    

```
  Freq   Channel
----------------
118.0750 118.080
118.0833 118.085
118.0916 118.090
118.1000 118.100
```

Maybe the `RadioFrequency` class could be extended to take care of this.

roeles:
I will think about a good way to solve it.
I've ordered a VHF antenna in order to test this on my bench with a real KRT2.

roeles:
@fb I'd like to start working on this. First I'll take a look at responding to the frequency exchange button on the KRT2, then I'll look at what to do with the frequencies.

However, I'm puzzled on what code I should base my changes.
First I saw a bugfix where you send data out as a response to receiving the 'S' character. Now I see nullptr checks in the same PR.
Can you point out to me what I can base my code on please?

roeles:
I think it makes sense to make the RadioFrequency carry two uint8_t members:
1. MHz
2. Counts of 8.33KHz

This should allow for all 8.33KHz channels as well as the old 25KHz channels as they are 3 multiples of 8.33.
A KHz-getter that returns a float could take care of the conversion from the true 8.33KHz frequencies to the rounded frequencies (which are multiples of 5KHz).

fb:
Am 2020-01-11 09:35, schrieb Roel Baardman:
> @fb I'd like to start working on this. First I'll take a look at
> responding to the frequency exchange button on the KRT2, then I'll
> look at what to do with the frequencies.
> 
> However, I'm puzzled on what code I should base my changes.
> First I saw a bugfix where you send data out as a response to
> receiving the 'S' character. Now I see nullptr checks in the same PR.
[...]

PR #338, opened 2020-01-21
Fix several bugs related to KRT2 remote control

roeles:
- Fix a bug where the frequency is re-set to a previous value after using the infobox to change it.
- Enable exchanging of frequencies from the KRT2
- Fix offsetting of frequencies, skip invalid 8.33kHz channels
- Clamp MHz values in KRT2 driver to values specified in the KRT2 user manual

### DIFF
--- a/src/ActionInterface.cpp
+++ b/src/ActionInterface.cpp
@@ -346,7 +346,7 @@ void ActionInterface::OffsetActiveFrequency(double offset_khz, bool to_devices)
 {
   RadioFrequency new_active_freq = SetComputerSettings().radio.active_frequency;
   if(new_active_freq.IsDefined()) {
-    new_active_freq.SetKiloHertz(new_active_freq.GetKiloHertz() + offset_khz);
+    new_active_freq.OffsetKiloHertz(offset_khz);
     if(new_active_freq.IsDefined()) {
       ActionInterface::SetActiveFrequency(new_active_freq, nullptr, to_devices);
     }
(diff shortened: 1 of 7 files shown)


roeles:
@fb FYI

fb:
I was able to test it today (c3cfd23). My KRT2 is "Vers. 8.2" (with dark color display).

- on startup, the correct frequencies will appear in the infoboxes
- set standby/active frequency works as expected
- I can play around on the KRT and the result in the UI will stay the same as on the radio

that means #322 is resolved. :+1: 

Regarding #321: Offsetting frequencies via infoboxes is also working (so technically, it's also fixed).

Now set standby to 118.00 and offset by -5 kHz. The infobox will show 117.990, the KRT will be set to 118.990.
Swapping frequencies on the KRT will put the 117.990 into the active frequency infobox.

Could be sufficient if KRT2Device::PutFrequency(… 117.990 …) returned false?

roeles:
Noticed that as well. The clamping is wrong (should be done on the entire
freq and not just on the mhz), but since there is no feedback mechanism
this is difficult.
Since in theory multiple radios might be connected and one could succeed
and another could fail, it is hard to define correct behavior.
Perhaps the driver should store the last successfully set frequency and
send that back as if it was sent by the radio with an stx frame.

roeles:
I'm going to close this request for now:
- The MHz check really belongs in RadioFrequency. The frequencies that KRT2 allows are the aviation voice frequencies. So, in the context of gliding (and thus XCSoar), anything outside of 118-136MHz is invalid.
- I had to solve a merge conflict, which results in an extra commit. I don't like that.

PR #300, opened 2019-12-11
self build on Raspbian/Raspberry Pi

CazYokoyama:
[WIP] Self build on
- Raspbian on Raspberry pi

Build command is 

make -j$(nproc) TARGET=UNIX DEBUG=n USE_CCACHE=y

Raspbian buster has gcc-8.3.0 as its default gcc.

PR #339, opened 2020-01-26
self build on Raspbian/Raspberry Pi

CazYokoyama:
[WIP] Self build on Raspbian on Raspberry pi
Build command is
make -j$(nproc) TARGET=UNIX DEBUG=n USE_CCACHE=y
Since Raspbian buster has gcc-8.3.0 as its default gcc, no need LOCAL_TCSUFFIX=-8.
This new PR is the follow-up of https://github.com/XCSoar/XCSoar/pull/300 which I mistakenly closed. Sorry.

### DIFF
--- a/build/detect.mk
+++ b/build/detect.mk
@@ -2,6 +2,7 @@
 
 UNAME_S := $(shell uname -s)
 UNAME_M := $(shell uname -m)
+OS_RELEASE := $(shell cat /etc/os-release)
 
 HOST_IS_LINUX := $(call string_equals,$(UNAME_S),Linux)
 HOST_IS_DARWIN := $(call string_equals,$(UNAME_S),Darwin)
@@ -9,6 +10,7 @@ HOST_IS_CYGWIN := $(call string_equals,$(UNAME_S),Cygwin)
 HOST_IS_MINGW := $(call string_contains,$(UNAME_S),MINGW)
 HOST_IS_WIN32 := $(call bool_or,$(HOST_IS_CYGWIN),$(HOST_IS_MINGW))
 HOST_IS_UNIX := $(call bool_not,$(HOST_IS_WIN32))
+HOST_IS_RASPBIAN := $(call string_contains,$(OS_RELEASE),Raspbian)
 
 HOST_IS_X86_32 := $(call bool_or,$(call string_contains,$(UNAME_M),i386),$(call string_contains,$(UNAME_M),i686))
 HOST_IS_X86_64 := $(call bool_or,$(call string_contains,$(UNAME_M),x86_64),$(call string_contains,$(UNAME_M),amd64))
--- a/build/targets.mk
+++ b/build/targets.mk
@@ -508,6 +508,12 @@ ifeq ($(TARGET_IS_KOBO),y)
   endif
 endif
 
+ifeq ($(TARGET),UNIX)
+  ifeq ($(HOST_IS_RASPBIAN),y)
+    TARGET_CPPFLAGS += -DSELFBUILD_RASPBIAN
+  endif
+endif
+
 ifeq ($(TARGET),ANDROID)
   TARGET_CPPFLAGS += -DANDROID
   CXXFLAGS += -D__STDC_VERSION__=199901L
--- a/src/Screen/X11/TopWindow.cpp
+++ b/src/Screen/X11/TopWindow.cpp
@@ -115,10 +115,24 @@ TopWindow::CreateNative(const TCHAR *text, PixelSize size,
     const Atom atoms[] = {
       XInternAtom(x_display, "_NET_WM_STATE_FULLSCREEN", false),
     };
+#ifdef SELFBUILD_RASPBIAN
+    XEvent e;
+    memset(&e, 0, sizeof(e));
+    e.xclient.type = ClientMessage;
+    e.xclient.message_type = XInternAtom(x_display, "_NET_WM_STATE", false);
+    e.xclient.display = x_display;
+    e.xclient.window = x_window;
+    e.xclient.format = 32;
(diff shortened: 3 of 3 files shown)


MaxKellermann:
This PR contains no correct commit. Only one large known-bad commit (already discussed) and several fixup commits. I can't even cherry-pick one of those. Can you please clean up this branch?

CazYokoyama:
Max, Yes, I will clean up.
I wondered which of clean-up or additional commit which integrate your review comment you prefer. Then I took latter way.
So, I'll clean-up and refactor the patches and forced-push this branch.
Thank you for reviewing. I appreciate your stamina which keeps reviewing.

CazYokoyama:
Max, could you give me your review comment or pulll? Another patch which makes full screen on Raspbian is waiting.

MaxKellermann:
I don't understand any of it. The commit message is confusing and doesn't explain anything. There are four parts in this patch:
1. Raspbian detection, setting a `make` variable
2. setting the preprocessor macro `SELFBUILD_RASPBIAN` which is never used
3. use OpenGL instead of GLES
4. adding `gcc_fallthrough`

I would be able to understand (1.) if there was a reason to do so, but there is none.
Since the preprocessor macro is never used, why would you set it?
The VideoCore GPU doesn't have OpenGL support, so this would effectively switch to painfully slow software rendering. Why?
So what remains is `gcc_fallthrough`, but this has nothing to do with Raspbian builds, and doesn't belong in this commit.

CazYokoyama:
> I would be able to understand (1.) if there was a reason to do so, but there is none.
> Since the preprocessor macro is never used, why would you set it?
> The VideoCore GPU doesn't have OpenGL support, so this would effectively switch to painfully slow software rendering. Why?
> So what remains is `gcc_fallthrough`, but this has nothing to do with Raspbian builds, and doesn't belong in this commit.
(1.) is for

CazYokoyama on Jan 2 Author Contributor
This is for fullscreen, i.e. do same as XInternAtom(x_display, "_NET_WM_STATE_FULLSCREEN", false). However, XInternAtom() does not give me fullscreen on Raspbian and Armbian.
@MaxKellermann
MaxKellermann on Jan 2 Member
But nothing of this is specific to Raspbian and it's unrelated to the rest of this commit and it's not documented in your commit message. You may propose such a change, but not hide it in an unrelated PR.
@CazYokoyama
CazYokoyama on Jan 2 Author Contributor

Move this code to another pull request.

Thank you for informing me OpenGL is slow while GLES is not. Raspbian supports GLES. I remember that the symbol, Font is defined in X11 and XCSoar, i.e. symbol conflict. I'll investigate.

I told in https://github.com/XCSoar/XCSoar/pull/300 that gcc of Raspbian which is 8.3.0 does not allow implicit fallthrough. Do you want me to create another pull request?

MaxKellermann:
> This is for fullscreen, i.e. do same as XInternAtom(x_display, "_NET_WM_STATE_FULLSCREEN", false). However, XInternAtom() does not give me fullscreen on Raspbian and Armbian.

I have no idea what you're talking about. I don't see any fullscreen code here. The only effective change this PR does is switch away from GLES to OpenGL, which is wrong.

> Thank you for informing me OpenGL is slow while GLES is not. Raspbian supports GLES.

But.... why did you decide to switch to OpenGL on Raspbian? But keep using GLES on non-Raspbian distributions on the Raspberry Pi? This seems so awkwardly wrong, that I cannot imagine why you would ever have such an idea.

> I remember that the symbol, Font is defined in X11 and XCSoar, i.e. symbol conflict. I'll investigate.

I have no idea what you're talking about.

> I told in #300 that gcc of Raspbian which is 8.3.0 does not allow implicit fallthrough. Do you want me to create another pull request?

Separate commit with a good commit message explaining why you're doing this (you don't need separate PRs, because even mixed inside one PR, I can cherry-pick good commits).

bomilkar:
> I told in #300 that gcc of Raspbian which is 8.3.0 does not allow implicit fallthrough.

Have you tied "__attribute__ ((fallthrough));" ?
It Indicates that the fall through from the previous case label is **intentional** and should not be diagnosed by a compiler that warns on fallthrough.

OscarKolsrud:
Hi @CazYokoyama I know this aint the right place to reach out, but i am trying to get xcsoar working on raspbian buster running on a pi 4. Have you gotten working, if so could you please take some time out of your day to explain to me how?

MaxKellermann:
There's still no explanation in the commit message.

CazYokoyama:
Combine the remaining 2 patches and have 1) what the patch does and 2) why as its commit message.

MaxKellermann:
Why combine? First you had one large commit, then I asked you to split it, and then you suddenly folded them again. That doesn't appear to make sense.
The commit message doesn't explain what the patch is doing.
All it explains about "why" is "XChangeProperty() does not work on Raspbian", but why does it not work? Is that all, did you do no research at all? (If you had done any research, then your commit message would explain the results.)

CazYokoyama:
> Why combine? First you had one large commit, then I asked you to split it, and then you suddenly folded them again. That doesn't appear to make sense.
> The commit message doesn't explain what the patch is doing.
> All it explains about "why" is "XChangeProperty() does not work on Raspbian", but why does it not work? Is that all, did you do no research at all? (If you had done any research, then your commit message would explain the results.)

Because the purpose of the patches become only one, i.e. "enable fullscreen on -fullscreen option on Raspbian". I believe "enable fullscreen on -fullscreen option on Raspbian" says what the patch does.

> why does it not work?
I investigated some. However, I lost my way in libX11 when I investigated. I'll add why XSendEvent() makes fullscreen as a workaround in Raspbian.

CazYokoyama:
Withdraw because I found another way to make xcsoar full screen.

PR #340, opened 2020-01-29
Fix various issues related to KRT2 and changing frequencies from infoboxes

roeles:
- Support exchanging channels
- Handle bug where frequencies could not be changed from the radio infoboxes
- Handle bug where invalid 8.33kHz frequencies could be send towards the radio

### DIFF
--- a/src/ActionInterface.cpp
+++ b/src/ActionInterface.cpp
@@ -346,7 +346,7 @@ void ActionInterface::OffsetActiveFrequency(double offset_khz, bool to_devices)
 {
   RadioFrequency new_active_freq = SetComputerSettings().radio.active_frequency;
   if(new_active_freq.IsDefined()) {
-    new_active_freq.SetKiloHertz(new_active_freq.GetKiloHertz() + offset_khz);
+    new_active_freq.OffsetKiloHertz(offset_khz);
     if(new_active_freq.IsDefined()) {
       ActionInterface::SetActiveFrequency(new_active_freq, nullptr, to_devices);
     }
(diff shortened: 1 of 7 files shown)


roeles:
Code was succesfully tested between a PC and a KRT2, connected through an FTDI RS232 cable.

MaxKellermann:
This commit is confusing and hard to read because it mixes several aspects in one commit.
For example, adding and using `OffsetKiloHertz()` should be one commit, and it has nothing to do with the KRT2 problems, does it? The `if` branch in `OffsetKiloHertz()` needs a code comment explaining it.
I don't understand the need for the `IsDefined()` change. I think that's the wrong thing to do. This doesn't check if the object is "defined", but if it holds a legal value.

roeles:
Thanks for taking the time to look at this pull request.

I will split the code into several separate pull requests.
`OffsetKiloHertz()` is radio-agnostic, correct. I will add a code comment.

Regarding your comment about the IsDefined() change...Anything outside the 118MHz-136MHz range can't exist, assuming that class `RadioFrequency` represents a VHF AirBand voice frequency.

I'll change the code.

roeles:
@fb If you'd like to test... I'll keep this branch and open new pull requests

MaxKellermann:
> Regarding your comment about the IsDefined() change...Anything outside the 118MHz-136MHz range can't exist, assuming that class `RadioFrequency` represents a VHF AirBand voice frequency.

True, but such a `RadioFrequency` instance shouldn't exist in the first place. Maybe as a temporary object while parsing some input, which should then call something like `IsValid()` (with your code) before propagating it.
The current `IsDefined()` check is simple, light and enough for checking "did somebody set a value". If yes, then it must be valid. (Welcome to add `assert()`s on this)

roeles:
I understand. 
I think I'll include the kHz check that I added as well, since for example 118.020 also is a non-existing frequency.

PR #346, opened 2020-02-17
Android: Aspect Ratio Fix

Caledonia91:
Currently on Android devices with aspect ratio's greater than 16:9, XCSoar does not use the full screen and instead will leave a black bar of unused space at the bottom.

As per this documentation, this should ensure the UI scales to fit the full screen:

https://android-developers.googleblog.com/2017/03/update-your-app-to-take-advantage-of.html

Note: I've been unable to test so this fix is speculative.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -24,6 +24,8 @@
 		 android:resource="@xml/accessory_filter" />
       <meta-data android:name="android.hardware.usb.action.USB_DEVICE_ATTACHED"
                  android:resource="@xml/usb_device_filter" />
+      <meta-data android:name="android.max_aspect"
+                 android:value="ratio_float"/>
     </activity>
 
     <service android:name=".MyService"/>


klujek:
It is important to not resize the screen to fit bigger aspect ratio and stretch all the elements.

Take benefit from wider screen and add infoboxes on booth sides of the screen and keep relatively plenty of space for map.

Caledonia91:
The problem is that the screen space is not recognised therefore you cannot place extra infoboxes at the bottom to utilise this space.

klujek:
> The problem is that the screen space is not recognised therefore you cannot place extra infoboxes at the bottom to utilise this space.

I know, so app need to know that there is some more space tu utilise.

MaxKellermann:
Cherry-picked to v6.8.x fb8370c2ed5df0e69a1055e3d53d922847ae2ea4
Note that your commit message was bad. I fixed it by copying text from this PR. Please do it properly next time.
Rule of thumb: if your PR contains text, your commit message is bad.

PR #347, opened 2020-02-19
Include – (minus sign) as a legal character in the parser for xci files.

bomilkar:
If the ‘-’ charter is not included, then strings are cut off at the position of the ‘-’ charter.

This affects for instance the following line in an xci file:
event=SendNMEAPort1 POV,C,RPO,0.000803,-0.009729,0.240054
It cuts off at this position ---------------------^

### DIFF
--- a/src/Input/InputParser.cpp
+++ b/src/Input/InputParser.cpp
@@ -215,7 +215,7 @@ ParseInputFile(InputConfig &config, TLineReader &reader)
           } else {
           #endif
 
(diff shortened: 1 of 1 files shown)


bomilkar:
The check failed with:
src/Input/InputParser.cpp:218:35: error: unknown escape sequence: '\-' [-Werror]

I need to investigate how to resolve this. (It didn't complain for TARGET=WIN64)

PR #348, opened 2020-02-19
Include – (minus sign) as a legal character in the parser for xci files.

bomilkar:
If the ‘-’ charter is not included, then strings are cut off at the position of the ‘-’ charter.

This affects for instance the following line in an xci file:
event=SendNMEAPort1 POV,C,RPO,0.000803,-0.009729,0.240054
It cuts off at this position ---------------------^


MaxKellermann:
Cherry-picked to v6.8.x 2805245e7e7548877a74a9d084395d88a7804451

PR #353, opened 2020-02-24
Implement OffsetKiloHertz for 8.33 kHz radio channels

roeles:
I'm not too sure about this, so pointers are more than welcome.

I am not sure if adding the two modulo-checks in `SetKiloHertz()` are a good idea. They look heavy. Yet it does prevent parsing clearly invalid frequencies.

### DIFF
--- a/src/RadioFrequency.hpp
+++ b/src/RadioFrequency.hpp
@@ -36,7 +36,8 @@ Copyright_License {
  */
 class RadioFrequency {
   static constexpr unsigned BASE_KHZ = 100000;
-  static constexpr unsigned MAX_KHZ = BASE_KHZ + 50000;
+  static constexpr unsigned MIN_KHZ = BASE_KHZ + 18000;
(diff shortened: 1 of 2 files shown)


src/RadioFrequency.hpp
```diff
@@ -81,12 +84,28 @@ class RadioFrequency {
...
+            (khz % 5 == 0) &&
+            (khz % 25 != 20)
       ? (khz - BASE_KHZ)
       : 0;
   }
 
+  void OffsetKiloHertz(int khz_offset) {
+    auto new_khz = GetKiloHertz() + khz_offset;
+    if ((new_khz % 25) == 20) {
+      new_khz += std::copysign(5, khz_offset);
```
MaxKellermann: This switches to floating point math.

roeles: Replaced std::copysign by a conditional operator.

Issue #244, opened 2019-08-21
CUP file is not displayed correctly


scls19fr:
Hello,

I wrote a small Python script to convert a CSV file with degree/min/sec coordinates of waypoints to CUP file.

```bash
$ cat poitiers_vfr.cup
name,latitude,longitude,desc
E,46°33’32”N,000°30’02”E,"St Julien - l’Ars : voie ferrée"
EA,46°35’30”N,000°23’40”E,"Usine Buxerolles (cheminée blanche, route)"
W,46°38’48”N,000°10’13”E,"Vouillé/route, rivière Auxance"
WA,46°37’43”N,000°15’51”E,"Rond-point de Moulinet"
```

and `convert_wpt_cup.py` script

https://gist.github.com/scls19fr/8cb7208fab72eb17ecd18a9fc97aeaad

This script is using Python Pandas and [aerofiles](https://github.com/Turbo87/aerofiles) from @Turbo87 

using this script like

```bash
$ python convert_wpt_cup.py poitiers_vfr.csv --name "%s@LFBI" --country "FR"
Loading poitiers_vfr.csv
  name    latitude    longitude                                        desc
0    E  46°33’32”N  000°30’02”E             St Julien - l’Ars : voie ferrée
1   EA  46°35’30”N  000°23’40”E  Usine Buxerolles (cheminée blanche, route)
2    W  46°38’48”N  000°10’13”E              Vouillé/route, rivière Auxance
3   WA  46°37’43”N  000°15’51”E                      Rond-point de Moulinet
4    S  46°26'55"N  000°16'10"E                             Réservoir d'eau
      name   latitude  longitude                                        desc shortname country
0   E@LFBI  46.559167   0.508333             St Julien - l’Ars : voie ferrée         E      FR
1  EA@LFBI  46.593056   0.389722  Usine Buxerolles (cheminée blanche, route)        EA      FR
2   W@LFBI  46.643889   0.169444              Vouillé/route, rivière Auxance         W      FR
3  WA@LFBI  46.626944   0.254167                      Rond-point de Moulinet        WA      FR
4   S@LFBI  46.440556   0.271111                             Réservoir d'eau         S      FR
Writing poitiers_vfr.cup
```

I'm getting the following cup file

```bash
$ cat poitiers_vfr.cup
name,code,country,lat,lon,elev,style,rwdir,rwlen,freq,desc
[...]

PR #359, opened 2020-03-06
DOC: Empty elevation field

scls19fr:
Related #244


MaxKellermann:
Bad commit message

scls19fr:
@MaxKellermann do you prefer this commit message?

PR #362, opened 2020-03-10
Panmode map scale render fix

CptFrikadel:
Second attempt at fixing issue #234. (Previous was PR #289 )

Added `GlueMapWindow::SetBottomMargin(unsigned int)` to be called whenever buttons are drawn (happens in `InputEvents::drawButtons(...)` ), so that `GlueMapWindow::DrawMapScale(..)` can avoid drawing in that area when in portrait mode.

Calling  `GlueMapWindow::SetBottomMargin(unsigned int)` triggers `QuickRedraw()`

### DIFF
--- a/src/Input/InputEvents.cpp
+++ b/src/Input/InputEvents.cpp
@@ -58,6 +58,8 @@ doc/html/advanced/input/ALL		http://xcsoar.sourceforge.net/advanced/input/
 #include "IO/LineReader.hpp"
 #include "Pan.hpp"
 #include "Dialogs/LockScreen.hpp"
+#include "Menu/MenuBar.hpp"
+#include "MapWindow/GlueMapWindow.hpp"
 
 #ifdef KOBO
 #include "Event/KeyCode.hpp"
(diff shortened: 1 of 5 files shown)


src/Input/InputEvents.cpp
```diff
@@ -198,6 +198,18 @@ InputEvents::drawButtons(Mode mode, bool full)
...
+  GlueMapWindow *map = CommonInterface::main_window->GetMapIfActive();
+  if (map != nullptr){
+      if (mode != MODE_DEFAULT){
+          map->SetBottomMargin(vertical_button_scale);
+          fprintf(stderr, "Setting margin!\n");
+      } else {
+          map->SetBottomMargin(0);
+          fprintf(stderr, "RESetting margin!\n");
+      }
+  }
```
MaxKellermann: What does this code do, and wat does it have to do with "moving scale magic number to const"?

src/Input/InputEvents.cpp
```diff
@@ -484,5 +496,5 @@ InputEvents::ProcessTimer()
 void
 InputEvents::eventLockScreen(const TCHAR *mode)
 {
-  ShowLockBox();  
+  ShowLockBox();
```
MaxKellermann: Whitespace change

src/Menu/MenuBar.hpp
```diff
@@ -28,6 +28,10 @@ Copyright_License {
 
 #include <tchar.h>
 
+
+const unsigned int vertical_button_scale = 6;
```
MaxKellermann: Should be `static constexpr` and needs documentation. And a better name - this is a global variable, after all.

src/Input/InputEvents.cpp
CptFrikadel: When the `MenuBar` is drawn, this will signal call `SetBottomMargin(...)`  with either a margin when a mode other than `MODE_DEFAULT` is active or a with zero to reset the margin when default mode is (re)entered.

This has indeed nothing to do with "moving scale magic number to const". It seems I made a mistake when cleaning up my branch. This should have been in the "Set bottom margin when buttons are drawn" commit instead of the removing of printfs that happened there.

CptFrikadel:
I cleaned the branch and gave `vertical_button_scale` a better name with documentation.

Issue #351, opened 2020-02-22
DisplayOrientation setting does not rotate the screen


kedder:
I intend to run XCSoar in bare KMS mode (i.e. not in X11 or Wayland window) and compiling it with `ENABLE_MESA_KMS=y` option. The display is portrait-oriented and I want to make XCSoar render the UI rotated 90 degrees on my screen. There is DisplayOrientation setting on xcsoar profile that seem to be supposed to do that, however it doesn't seem to do it correctly: the graphics seem to be rendered on portrait-oriented canvas, but then stretched on the landscape-oriented screen (see screenshot below).

The end goal is to run it in this mode on Openvario device (https://github.com/Openvario/meta-openvario/issues/2), but the issue doesn't seems to be specific to that device - I can reproduce it on my laptop as well.

To reproduce, I'm building the XCSoar on my laptop (Thinkpad T460 with Intel HD Graphics 520 GPU) with this command line:

```
make -j$(nproc) ENABLE_MESA_KMS=y
```

I have changed my `~/.xcsoar/default.prf` to include `DisplayOrientation="1"` line. To run XCSoar, I switch to linux console (Ctrl-Alt-F3) and start XCSoar normally: `output/UNIX/bin/xcsoar -fly`.

What I get as a result is on the photo below. I expect to see UI rotated 90 degrees instead, so it is usable if I orient display in portrait mode.

![IMG_20200222_125450](https://user-images.githubusercontent.com/59499/75091138-bf292980-5572-11ea-9e17-f00dd3d84972.jpg)

While looking into it, I've found this piece of code in `src/Screen/OpenGL/Init.cpp`:
```c
UnsignedPoint2D
OpenGL::SetupViewport(UnsignedPoint2D size)
{
  window_size = size;

  glViewport(0, 0, size.x, size.y);

#ifdef SOFTWARE_ROTATE_DISPLAY
  projection_matrix = glm::rotate(glm::mat4(1),
                                  (GLfloat)Angle::Degrees(OrientationToRotation(display_orientation)).Radians(),
                                  glm::vec3(0, 0, 1));
  OrientationSwap(size, display_orientation);
#endif
  projection_matrix = glm::ortho<float>(0, size.x, size.y, 0, -1, 1);
  UpdateShaderProjectionMatrix();

  viewport_size = size;

  return size;
}
```

[...]

PR #363, opened 2020-03-16
Multiply rotation (if there is one) with projection_matrix in OpenGL software rotation

roeles:
This solves issue #351.

Tested on TARGET=UNIX OPENGL=y, as discussed in issue #351.

### DIFF
--- a/src/Screen/OpenGL/Init.cpp
+++ b/src/Screen/OpenGL/Init.cpp
@@ -327,7 +327,7 @@ OpenGL::SetupViewport(UnsignedPoint2D size)
                                   glm::vec3(0, 0, 1));
   OrientationSwap(size, display_orientation);
 #endif
-  projection_matrix = glm::ortho<float>(0, size.x, size.y, 0, -1, 1);
+  projection_matrix *= glm::ortho<float>(0, size.x, size.y, 0, -1, 1);
   UpdateShaderProjectionMatrix();
 
   viewport_size = size;


MaxKellermann:
Bad commit message. Doesn't explain anything.

roeles:
@MaxKellermann thanks for taking the time to look at it.

Can you describe what you'd like to have addressed in the commit message?
Cause? Solution? Other information?

MaxKellermann:
The commit message describes what is changed, but not why, and what was wrong before. I don't understand the problem.
This PR has additional text (which is bad because it means your commit message is bad), but this text only refers to a bug report on GH. It should not be necessary to dig into GH to understand a commit.

kedder:
@roeles, I suggest to change it to:

```
OpenGL/Init: Fix screen rotation

Correctly calculate OpenGL projection matrix for software screen 
rotation according to DisplayOrientation setting.

Fixes #351. 
```

MaxKellermann:
Is this fix relevant for v6.8.x?

kedder:
No, the regression is introduced in 7.0-preview13 (67a9a193fa319029dc95b4f4c22c5e9d38a9884f to be precise).

MaxKellermann:
> No, the regression is introduced in 7.0-preview13 ([67a9a19](https://github.com/XCSoar/XCSoar/commit/67a9a193fa319029dc95b4f4c22c5e9d38a9884f) to be precise).

This should definitely be mentioned in the commit message! Such things are important for our future selves!

MaxKellermann:
Hm, is this really a regression, or was this an old bug which previously did not affect anybody, because nobody used GLSL?

roeles:
The line that I changed was unchanged for 6 years according to
https://github.com/XCSoar/XCSoar/blame/a397cf8d30d66edd9579bee8cf8218730b2dafc9/src/Screen/OpenGL/Init.cpp
My best guess is that nobody used software screen rotation with GLSL.

MaxKellermann:
I had to revert this change because it broke rendering on Android completely and caused a black screen. Tested on Pixel 2 and Cat S61.

kedder:
Okay, that's too bad. But I think we can figure it out.

PR #368, opened 2020-03-25
Added the CoPilot to the logger Settings

Luap-Paul:
next try, hope my pull request is now in a better shape.


MaxKellermann:
Don't update `po/`.
Where's the code for loading the configured setting from the profile?

MaxKellermann:
Build failure:
```
test/src/RunIGCWriter.cpp:51:56: error: too few arguments to function call, expected 9, have 8
                     _T("MM"), "FOO", driver_name, true);
```

MaxKellermann:
Say when you're done fixing all the problems.

Luap-Paul:
alright!

PR #370, opened 2020-03-30
adding a query option and reducing the traffic on the serial port.

bomilkar:
1. I have created a mechanism for the connected device (e.g. variod)
to read the relevant current settings of XCSoar. “$POV,?,3*45”
In response, the connected device responds with settings: bugs, ballast, polar, etc.
The connected device can do this at any time, especially
upon restart to sync up with XCSoar’s settings.

2. Since there is no method like "PutPolar" in class Device, I
continue to use OnCalculatedUpdate to transfer the polar data,
only that I check whether the coefficients have changed.
If not, then no $POV,C,RPO…. string is sent.

### DIFF
--- a/src/Device/Driver/OpenVario.cpp
+++ b/src/Device/Driver/OpenVario.cpp
@@ -54,11 +54,20 @@ private:
   		  OperationEnvironment &env);
   bool PutRealPolar(bool force, const DerivedInfo *calculated,
   		  OperationEnvironment &env);
-  // bool RepeatBugs(double bugs, OperationEnvironment &env);
-  // bool RepeatBallast(double ballast, OperationEnvironment &env);
+  bool RepeatBugs(OperationEnvironment &env);
+  bool RepeatBallast(OperationEnvironment &env);
+  bool RepeatMacCready(OperationEnvironment &env);
+
+  // remember the settings to be able to repeat the most recent values upon request
+  double _bugs = 1;
+  bool   _bugs_valid = false;
+  double _mc = 1;
+  bool   _mc_valid = false;
+  double _overload = 1;
+  bool   _overload_valid = false;
+  const DerivedInfo *_calculated = NULL;
 };
 
-static const DerivedInfo *_calculated = NULL;
 
 bool
 OpenVarioDevice::ComposeWrite(const char p_type,double a, double b, double c,
@@ -130,36 +139,60 @@ OpenVarioDevice::OnCalculatedUpdate(const MoreData &basic,
 }
 
 bool
-OpenVarioDevice::PutMacCready(double mc,OperationEnvironment &env)
+OpenVarioDevice::RepeatMacCready(OperationEnvironment &env)
 {
-  if (!EnableNMEA(env))
+  if ((!EnableNMEA(env)) || (!_mc_valid))
     return false;
   
   char buffer[20];
-  sprintf(buffer,"POV,C,MC,%0.2f", (float)mc);
+  sprintf(buffer,"POV,C,MC,%0.2f", (double)_mc);
+  return PortWriteNMEA(port, buffer, env);
+}
+
+bool
+OpenVarioDevice::PutMacCready(double mc, OperationEnvironment &env)
+{
+  _mc = mc;
+  _mc_valid = true;
+  return RepeatMacCready(env);
+}
+
+bool
+OpenVarioDevice::RepeatBallast(OperationEnvironment &env)
+{
+  if ((!EnableNMEA(env)) || (!_overload_valid))
+    return false;
+  
+  char buffer[20];
+  sprintf(buffer,"POV,C,WL,%1.3f",(float)_overload);
   return PortWriteNMEA(port, buffer, env);
 }
 
 bool
 OpenVarioDevice::PutBallast(double fraction, double overload, OperationEnvironment &env)
 {
-  if (!EnableNMEA(env))
+  _overload = overload;
+  _overload_valid = true;
+  return RepeatBallast(env);
+}
+
+bool
+OpenVarioDevice::RepeatBugs(OperationEnvironment &env)
+{
+  if ((!EnableNMEA(env)) || (!_bugs_valid))
     return false;
-  
-  char buffer[20];
-  sprintf(buffer,"POV,C,WL,%1.3f",(float)overload);
+ 
+  char buffer[32];
+  sprintf(buffer, "POV,C,BU,%0.2f",(float)_bugs);
   return PortWriteNMEA(port, buffer, env);
 }
 
 bool
-OpenVarioDevice::PutBugs(double bugs,OperationEnvironment &env)
+OpenVarioDevice::PutBugs(double bugs, OperationEnvironment &env)
 {
-  if (!EnableNMEA(env))
-    return false;
- 
-  char buffer[32];
-  sprintf(buffer, "POV,C,BU,%0.2f",(float)bugs);
-  return PortWriteNMEA(port, buffer, env);
+  _bugs = bugs;
+  _bugs_valid = true;
+  return RepeatBugs(env);
 }
 
 bool
@@ -205,9 +238,9 @@ OpenVarioDevice::POV(NMEAInputLine &line, NMEAInfo &info)
       case '?': {
 	  NullOperationEnvironment env;
 	  int sel = (int)(value+0.1);
-	  if (sel & 0x10) PutBallast(info.settings.ballast_fraction,info.settings.ballast_overload,env);
-	  if (sel & 0x08) PutBugs(info.settings.bugs,env);
-	  if (sel & 0x04) PutMacCready(info.settings.mac_cready,env);
+	  if (sel & 0x10) RepeatBallast(env);
+	  if (sel & 0x08) RepeatBugs(env);
+	  if (sel & 0x04) RepeatMacCready(env);
 	  if (sel & 0x02) PutIdealPolar(true, _calculated, env);
 	  if (sel & 0x01) PutRealPolar(true, _calculated, env);
 	  // TODO : what else might the connected device want to know ?
@@ -261,7 +294,6 @@ OpenVarioDevice::POV(NMEAInputLine &line, NMEAInfo &info)
 static Device *
 OpenVarioCreateOnPort(const DeviceConfig &config, Port &com_port)
 {
-  _calculated = NULL;
   return new OpenVarioDevice(com_port);
 }
 
--- a/src/Device/Driver/OpenVario.cpp
+++ b/src/Device/Driver/OpenVario.cpp
@@ -54,15 +54,11 @@ private:
   		  OperationEnvironment &env);
   bool PutRealPolar(bool force, const DerivedInfo *calculated,
   		  OperationEnvironment &env);
-  bool RepeatBugs(OperationEnvironment &env);
-  bool RepeatBallast(OperationEnvironment &env);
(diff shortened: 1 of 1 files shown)


bomilkar:
Check seem to fail because:
In file included from src/IO/DataFile.cpp:30:0:
./src/Util/StringCompare.hxx:41:23: fatal error: string_view: No such file or directory
 #include <string_view>

This is not a problem on my local repo. But I can't find a string_view.h anywhere in the tree.
Hm.... ??

MaxKellermann:
Bad check. The CI has an outdated toolchain which doesn't support C++17.

src/Device/Driver/OpenVario.cpp
```diff
@@ -46,65 +46,147 @@ class OpenVarioDevice : public AbstractDevice {
...
+  		  OperationEnvironment &env);
+  bool RepeatBugs(OperationEnvironment &env);
+  bool RepeatBallast(OperationEnvironment &env);
+  bool RepeatMacCready(OperationEnvironment &env);
 };
 
+static const DerivedInfo *_calculated = NULL;
+static double _bugs = 1;
+static double _overload = 1;
+static double _mc = 1;
```
MaxKellermann: I have no idea what this obscure commit does, but these global variables are not acceptable!

bomilkar: The static variables remember the previous settings, when e.g. PutMacCready() was called last time. The value will be used when the connected device (after a restart) wants to know what the current value was. It is used in RepeatMacCready.
Similarly for Ballast, Bugs, Polar.
I tried to keep the code self contained as much as I could.

MaxKellermann: You explained what those globals do, but that doesn't make them any better.
Maybe you understand it when you answer this question: who owns the memory pointed to by `_calculated`?
Another question: what if a user has two OpenVarios - who will protect those variables from clashing between the two instances?

bomilkar: I admit: I didn't think about multiple OpenVarios.
I need to figure out how to replace _calculated .
Can you provide a hint?

bomilkar:
@MaxKellermann There are 2 objectives for these changes:

1. I want to provide a connected device (external Vario, variod, …) with means to read relevant settings (MacCready, wing load, polar, polar degradation) from XCSoar. This is important, because a connected device may (re)start while XCSoar is already running for some time and thus the connected device misses out on all the changes, which had happened before the (re)start. So I implemented an NMEA-Sentence by which the connected device triggers XCSoar to repeat the most recent settings for MacCready, wing load, ideal polar, real polar, polar degradation. That sentence is “$POV,?,31" to receive all 5 settings.

2. the previous implementation of the OpenVario driver produced 2 polar string (ideal polar, real polar) every time OnCalculatedUpdate() is called. At a minimum this happens for every GPS position XCSoar receives. So it easily sends out 4 polar strings per second, even though nothing has changed in these polars for a while. So the change to address this is to check if the polar information has changed before it is sends it out. As a result polar strings are sent only when needed.

In the meantime I found an implementation which doesn’t use any global variable and supports multiple instances of the driver. I hope this addresses you initial concerns.

MaxKellermann:
You did not answer my question:
> who owns the memory pointed to by _calculated?

bomilkar:
This pointer part of the instance of OpenVarioDevice. In case of multiple OpenVario drivers, each should get it's own.
And it points to the same object as OnCalculatedUpdate was called with. Which is part of the data structure, XCSoar is working on. I assume this is a static reference. Isn't it?
I really don't know who owns it. I tried to find the place in the code where the instance of this DerivedInfo was instantiated but couldn't find it.
What can I do to resolve this?

MaxKellermann:
> And it points to the same object as OnCalculatedUpdate was called with.

But the reference pased to `OnCalculatedUpdate()` is only valid during that call. Out of this scope, it dangles, it may point anywhere.

> I assume this is a static reference. Isn't it?

This kind of question is too important to just "assume" something. In "unsafe" languages like C and C++, you must always know who owns a pointer and when it is safe to dereference it.

> What can I do to resolve this?

By not copying the pointer. I don't know why you copy it, and your code doesn't have a comment explaining it.

bomilkar:
I'm working on a different implementation: simply remembering the values of the polar coeffs, similar to wing load, bugs, MacCready.
I appreciate your feedback and thanks for the help. Just bear with me.

bomilkar:
@MaxKellermann , I think I've addressed all the issues.
I'm sorry for all these iterations.

src/Device/Driver/OpenVario.cpp
```diff
@@ -46,65 +46,162 @@ class OpenVarioDevice : public AbstractDevice {
...
+  double _overload = 1;
+  bool   _overload_valid = false;
+  PolarCoefficients _ideal_polar; 
+  bool _ideal_polar_valid = false;
+  PolarCoefficients _real_polar; 
+  bool _real_polar_valid = false;
 };
 
+bool
+OpenVarioDevice::IsEqual(const PolarCoefficients &p1, const PolarCoefficients &p2)
```
kedder, line 77: Since `IsEqual` is not comparing `OpenVarioDevice` instances, `IsPolarEqual` would be a more appropriate name

src/Device/Driver/OpenVario.cpp
```diff
@@ -46,65 +46,162 @@ class OpenVarioDevice : public AbstractDevice {
...
+  OperationEnvironment &env)
+{
+  if (!EnableNMEA(env))
+    return false;
+
+  char buffer[50];
+  // Compose Polar String
+  // TODO does the receiver understand %e format? Or do we have to use %f, even if
+  //      the number of significant digits is less then 3.
+  sprintf(buffer,"POV,C,%cPO,%e,%e,%e", p_type, p.a, p.b, p.c);
```
kedder, line 104: I'd be conservative here. What's the order of numbers we are dealing with here?

src/Device/Driver/OpenVario.cpp
```diff
@@ -46,65 +46,162 @@ class OpenVarioDevice : public AbstractDevice {
...
+OpenVarioDevice::IsEqual(const PolarCoefficients &p1, const PolarCoefficients &p2)
+{
+  if (p1.a != p2.a) return false;
+  if (p1.b != p2.b) return false;
+  if (p1.c != p2.c) return false;
+  return true;
+}
+
+void
+OpenVarioDevice::Copy(PolarCoefficients &dest, const PolarCoefficients &src)
```
kedder, line 86: Ditto, suggesting to name it `CopyPolar`

src/Device/Driver/OpenVario.cpp
```diff
@@ -131,13 +228,33 @@ OpenVarioDevice::POV(NMEAInputLine &line, NMEAInfo &info)
...
     char type = line.ReadOneChar();
     if (type == '\0')
       break;
 
+    if (type == '?') {
+      NullOperationEnvironment env;
+      char query_item[5];
+
+      for (int i=0;i < 10;i++) { // not more than 10 loops!
+	line.Read(query_item,sizeof(query_item));
```
kedder, line 246: Indentation for `for` loop is missing

src/Device/Driver/OpenVario.cpp
```diff
@@ -131,13 +228,33 @@ OpenVarioDevice::POV(NMEAInputLine &line, NMEAInfo &info)
...
+	if (strlen(query_item) == 0) return true;
+
+	if (StringIsEqual(query_item,"WL")) RepeatBallast(env);
+	if (StringIsEqual(query_item,"BU")) RepeatBugs(env);
+	if (StringIsEqual(query_item,"MC")) RepeatMacCready(env);
+	if (StringIsEqual(query_item,"IPO") && _ideal_polar_valid) ComposeWrite('I',_ideal_polar, env);
+	if (StringIsEqual(query_item,"RPO") && _real_polar_valid) ComposeWrite('R',_real_polar, env);
```
kedder, line 253: Something that bothers me here is: if polar is not available on request, we just keep silent and not answering anything. This might be confusing and make debugging difficult. I suggest responding with NMEA sentence like `POV,C,NA,IPO` in this case (and other cases too). This will make it obvious (for a human) that request cannot be fullfulled at this moment and will not confuse existing openvario nmea parsers.

src/Device/Driver/OpenVario.cpp
bomilkar: The operand types make it obvious what is being compared.

src/Device/Driver/OpenVario.cpp
bomilkar: %f format can easily end up with 2 significant digits or less. a is usually the coefficient with the smallest value.
As long as the receiving end uses atof() or scanf ("%f) it'll be safe. Is this a safe assumption??

src/Device/Driver/OpenVario.cpp
bomilkar: This is a formatting issue on this page. Maybe with the diff.
Look at the real file:
https://github.com/bomilkar/XCSoar/blob/restucture-and-improve-openvario-driver/src/Device/Driver/OpenVario.cpp
The indentation is OK. I think.

src/Device/Driver/OpenVario.cpp
bomilkar: That's a good point. But I'm not convinced n/a is a better choice. This might make the receiving end choke. BTW: the same happens for MC, BU, WL, it's hidden behind the Repeat... functions.

Anybody else an opinion?

src/Device/Driver/OpenVario.cpp
bomilkar: The operand types make it obvious what is being compared.
The appropriate name would be CopyPolarCoefficients which would be really redundant with the operand types.

src/Device/Driver/OpenVario.cpp
bomilkar: Here is an example:
a: -0.000177, b: 0.032300, c: -2.040927

src/Device/Driver/OpenVario.cpp
kedder: Oh, I guess that's because spaces and tabs are mixed for indentation. According to [xcsoar dev manual](http://download.xcsoar.org/releases/6.8.11/XCSoar-developer-manual.pdf), we should not use tabs for indentation:

> 79 columns, reasonable exceptions allowed. Indent 2 spaces, no tabs.

src/Device/Driver/OpenVario.cpp
kedder: Hm, ok, that makes sense.

src/Device/Driver/OpenVario.cpp
kedder: I think we can safely assume that receiver will ignore unknown/unrecognized sentences. We can pick different response format, but I believe we have to answer something.

src/Device/Driver/OpenVario.cpp
kedder: hm, ok, I don't have a strong opinion here because I don't know what these coefficients really mean. The only situation where "e" notation would be a PITA is on embedded code, where you don't have libc, but OTOH we probably don't want to lose precision here.

kedder:
@bomilkar, you probably want to squash these commits into one or two and also get rid of merge commit in the history (you can use interactive rebase on top of master)

bomilkar:
To clean it up, I combined the changes and created a new PR:
https://github.com/XCSoar/XCSoar/pull/378

PR #371, opened 2020-04-07
List nano downloadable flights in date descending order

cjwoolley:
This change lists nano flights in descending order thereby highlighting the latest flight to download by default, rather than the oldest flight.

### DIFF
--- a/src/Device/Driver/LX/NanoLogger.cpp
+++ b/src/Device/Driver/LX/NanoLogger.cpp
@@ -219,25 +219,22 @@ Nano::ReadFlightList(Port &port, RecordedFlightList &flight_list,
     return nflights == 0;
 
   env.SetProgressRange(nflights);
-
-  unsigned requested_tail = 1;
+  
+  unsigned requested_tail = 0;
   while (true) {
-    const unsigned room = flight_list.max_size() - flight_list.size();
-    const unsigned remaining = nflights - requested_tail + 1;
-    const unsigned nmax = std::min(room, remaining);
-    if (nmax == 0)
-      break;
 
-    /* read 8 records at a time */
-    const unsigned nrequest = std::min(nmax, 8u);
+    if (nflights == 0)
+      break;
 
     timeout = TimeoutClock(2000);
     if (!GetLogbookContents(port, reader, flight_list,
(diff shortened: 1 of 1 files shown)


MaxKellermann:
This PR contains text not present in the commit message.

cjwoolley:
> This PR contains text not present in the commit message.

Made changes to the commit message.

src/Device/Driver/LX/NanoLogger.cpp
```diff
@@ -219,25 +219,22 @@ Nano::ReadFlightList(Port &port, RecordedFlightList &flight_list,
...
 
-    /* read 8 records at a time */
-    const unsigned nrequest = std::min(nmax, 8u);
+    if (nflights == 0)
+      break;
 
     timeout = TimeoutClock(2000);
     if (!GetLogbookContents(port, reader, flight_list,
-                            requested_tail, nrequest, env, timeout))
+                            nflights, 1, env, timeout))
```
MaxKellermann, line 231: Previously, XCSoar did bulk transfer, and now you're doing one at a time (but you're still using the function for bulk transfers, just passing `n=1`). This may reduce performance and affects how timeouts are handled. Why?

cjwoolley: It was easier for me with my basic coding skills to implement the change this way.
I have not experienced any timeouts or noticeable performance reductions to date.
If required I'll have a go at reinstating the change using bulk transfers of 8 at a time.

MaxKellermann:
Dumb question: why do you believe the device driver should do the sorting? To me, this sounds like a pure user interface decision.

cjwoolley:
The device driver's protocol has the ability to list flights in reverse order albeit by selecting them one at a time. This is not ideal but allowed me with my basic C++ coding skills to implement the change. If this solution is not technically acceptable then I'll close this PR and create an enhancement request in its place.

MaxKellermann:
You did not answer my question, you did not even try.
With only "basic C++ coding skills", sorting a list is easier than what you did (rewrite the flight download code, change bulk to single download in a reversed loop). I think this is the wrong approach (it's too complicated, slow and the wrong place to implement a UI feature), therefore I close this PR.

PR #373, opened 2020-04-09
Fix software display rotation (again)

### LINKED ISSUE #351: DisplayOrientation setting does not rotate the screen

kedder:
Initial fix (8bd49afa25e112f3d5f377dfa1ad72656effc35d) caused rendering problems (black screen) on platforms in which `SOFTWARE_ROTATE_DISPLAY` is not defined (e.g. Android). It was reverted in 6ef5768a073f7fd07442dc6f897f4fe7220d460e.

Tested on Android, but was reproducible on regular UNIX as well. The problem with original `projection_matrix *= glm::ortho` fix is that `projection_matrix` is initialized with zero matrix.

Fixes #351.

### DIFF
--- a/src/Screen/OpenGL/Init.cpp
+++ b/src/Screen/OpenGL/Init.cpp
@@ -322,12 +322,19 @@ OpenGL::SetupViewport(UnsignedPoint2D size)
   glViewport(0, 0, size.x, size.y);
 
 #ifdef SOFTWARE_ROTATE_DISPLAY
-  projection_matrix = glm::rotate(glm::mat4(1),
-                                  (GLfloat)Angle::Degrees(OrientationToRotation(display_orientation)).Radians(),
-                                  glm::vec3(0, 0, 1));
   OrientationSwap(size, display_orientation);
 #endif
+
   projection_matrix = glm::ortho<float>(0, size.x, size.y, 0, -1, 1);
+
+#ifdef SOFTWARE_ROTATE_DISPLAY
+  glm::mat4 rot_matrix = glm::rotate(
+    glm::mat4(1),
+    (GLfloat)Angle::Degrees(OrientationToRotation(display_orientation)).Radians(),
+    glm::vec3(0, 0, 1));
+  projection_matrix = rot_matrix * projection_matrix;
+#endif
+
   UpdateShaderProjectionMatrix();
 
   viewport_size = size;


roeles:
Hm, sorry about breaking it.
Thanks for taking the time to fix the bug I introduced!

src/Screen/OpenGL/Init.cpp
```diff
@@ -322,12 +322,15 @@ OpenGL::SetupViewport(UnsignedPoint2D size)
...
-  projection_matrix = glm::rotate(glm::mat4(1),
-                                  (GLfloat)Angle::Degrees(OrientationToRotation(display_orientation)).Radians(),
-                                  glm::vec3(0, 0, 1));
+  glm::mat4 rot_matrix = glm::rotate(
+    glm::mat4(1),
+    (GLfloat)Angle::Degrees(OrientationToRotation(display_orientation)).Radians(),
+    glm::vec3(0, 0, 1));
   OrientationSwap(size, display_orientation);
+#else
+  glm::mat4 rot_matrix = glm::mat4(1);  // no rotation
```
MaxKellermann, line 331: This adds overhead for targets which have no software rotation, by multiplying this matrix which is a no-op. Can we avoid that overhead?

roeles:
I think that if you keep the code the same for #ifdef SOFTWARE_ROTATE_DISPLAY as in my patch, but keep the code identical to what it was before my patch for #if !defined(SOFTWARE_ROTATE_DISPLAY), that should be simple enough and optimal for all platforms?

kedder:
@MaxKellermann good point. 

I was choosing between introducing another `#ifdef SOFTWARE_ROTATE_DISPLAY` block and duplucating `glm::ortho<float>(0, size.x, size.y, 0, -1, 1);` sentence (as @roeles suggested) and picked the former option. 

The thing is `glm::ortho` needs `size` in right orientation, that is set up by `OrientationSwap` - it needs to be executed before `glm::ortho`.

kedder:
Is there anything I can improve here? Any concerns, suggestions?

PR #376, opened 2020-04-12
scalable InfoBox title font

bomilkar:
Introduce a user parameter "InfoBox title size" 50% ... 150% to scale
the font for the title and comment in the InfoBoxes.
The intention is to improve readability of InfoBox title and comment
in particular on tall and narrow displays. It allows the user to trade off
readability vs. length of text being displayed.
The main change is in Look/InfoBoxLook.cpp

### DIFF
--- a/src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp
+++ b/src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp
@@ -187,10 +187,7 @@ LayoutConfigPanel::Prepare(ContainerWindow &parent, const PixelRect &rc)
           _("A list of possible InfoBox layouts. Do some trials to find the best for your screen size."),
           info_box_geometry_list, (unsigned)ui_settings.info_boxes.geometry);
 
-  AddInteger(_("InfoBox title size"),
-             _("An InfoBox has 3 lines: the value in the middle, "
-               "the title in the top line and the comment in the bottom line. "
-               "This parameter scales title and comment."),
+  AddInteger(_("InfoBox title size"), nullptr,
              _T("%d %%"), _T("%d"), 80, 150, 5,
              ui_settings.info_boxes.scale_title_font);
   SetExpertRow(InfoBoxTitleScale);
--- a/src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp
+++ b/src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp
@@ -45,6 +45,7 @@ Copyright_License {
 enum ControlIndex {
   MapOrientation,
   AppInfoBoxGeom,
+  InfoBoxTitleScale,
   AppFlarmLocation,
   TabDialogStyle,
   AppStatusMessageAlignment,
@@ -186,6 +187,14 @@ LayoutConfigPanel::Prepare(ContainerWindow &parent, const PixelRect &rc)
           _("A list of possible InfoBox layouts. Do some trials to find the best for your screen size."),
           info_box_geometry_list, (unsigned)ui_settings.info_boxes.geometry);
 
+  AddInteger(_("InfoBox title size"),
+             _("An InfoBox has 3 lines: the value in the middle, "
+               "the title in the top line and the comment in the bottom line. "
+               "This parameter scales title and comment."),
+             _T("%d %%"), _T("%d"), 80, 150, 5,
+             ui_settings.info_boxes.scale_title_font);
+  SetExpertRow(InfoBoxTitleScale);
+
   AddEnum(_("FLARM display"), _("Choose a location for the FLARM display."),
           flarm_display_location_list,
           (unsigned)ui_settings.traffic.gauge_location);
@@ -249,6 +258,10 @@ LayoutConfigPanel::Save(bool &_changed)
   info_box_geometry_changed |=
     SaveValueEnum(AppInfoBoxGeom, ProfileKeys::InfoBoxGeometry,
                   ui_settings.info_boxes.geometry);
+
+  info_box_geometry_changed |=
+    SaveValueEnum(InfoBoxTitleScale, ProfileKeys::InfoBoxTitleScale,
+                  ui_settings.info_boxes.scale_title_font);
 
   info_box_geometry_changed |=
     SaveValueEnum(AppFlarmLocation, ProfileKeys::FlarmLocation,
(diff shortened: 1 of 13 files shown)


bomilkar:
I don't understand why the checks fail. "make TARGET=UNIX64" on the local branch ends w/o any issues.

MaxKellermann:
`InfoBoxLook` now uses a global variable, which it shouldn't.

bomilkar:
@MaxKellermann I guess you are referring to 
ui_settings.info_boxes.scale_title_font
I've been unable to come up with a different/better way to transport the information from the Setup
to where it's being used. The very same global variable is being set in
Dialogs/Settings/Panels/LayoutConfigPanel.cpp
What can I do?

MaxKellermann:
Pass it as parameter. Look at the rest of the class - the parameters passed to `Initialise()` are copied from `InfoBoxSettings`, for the same reasons.

bomilkar:
@MaxKellermann , @kedder , thanks for pointing me in the right direction. I must have been blind.

src/Look/InfoBoxLook.cpp
```diff
@@ -73,10 +75,11 @@ InfoBoxLook::Initialise(bool _inverse, bool use_colors,
 }
 
 void
-InfoBoxLook::ReinitialiseLayout(unsigned width)
+InfoBoxLook::ReinitialiseLayout(unsigned width, unsigned scale_title_font)
 {
   FontDescription title_font_d(8);
-  AutoSizeFont(title_font_d, width, _T("123456789012345"));
+  unsigned scaled_width = width * Clamp((int)scale_title_font,50,150) / 100;
```
kedder, line 81: Be careful, as you might be doing an integer division here. 150 / 100 == 1

MaxKellermann: More dangerously, `50/100 == 0` - what happens when `scaled_width==0`?

kedder: It is probably fine here though, since multiplication by width goes first.

src/InfoBoxes/InfoBoxSettings.hpp
```diff
@@ -124,6 +124,9 @@ struct InfoBoxSettings {
 
   } geometry;
 
+  /* scales the font for InfoBox titles and comments between 80% and 150% */
+  unsigned scale_title_font;
```
MaxKellermann, line 128: What does the value mean? This should be documented. Why is this an integer? The example values are `0.8` and `1.5` which cannot be represented with an `unsigned`.

src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp
```diff
@@ -186,6 +187,14 @@ LayoutConfigPanel::Prepare(ContainerWindow &parent, const PixelRect &rc)
           _("A list of possible InfoBox layouts. Do some trials to find the best for your screen size."),
           info_box_geometry_list, (unsigned)ui_settings.info_boxes.geometry);
 
+  AddInteger(_("InfoBox title size"),
+             _("An InfoBox has 3 lines: the value in the middle, "
+               "the title in the top line and the comment in the bottom line. "
```
MaxKellermann, line 192: I don't think this should be explained here. This only adds work for translators, translating redundant text, and users will not read it anyway because it's too long.

src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp
```diff
@@ -186,6 +187,14 @@ LayoutConfigPanel::Prepare(ContainerWindow &parent, const PixelRect &rc)
...
+               "This parameter scales title and comment."),
```
MaxKellermann, line 193: If it's called `scale_title_font`, then why does it scale the comment? This is confusing.

src/Look/InfoBoxLook.cpp
bomilkar: True, if width == 1.
Arithmetic expressions are evaluated left to right. I could add parentheses, though. Are you suggesting this?

MaxKellermann: No, @bomilkar you're right, this expression is ok after all. Truncation happens at the end.
What is questionable however is the cast to `int` - why did you do that?

src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp
bomilkar: I'll move this text to the documentation.

src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp
bomilkar: Because the same font is used for title and comment: title_font

src/Look/InfoBoxLook.cpp
bomilkar: Clamp wants either 3 int **or** 3 unsigned.
Clamp((int)scale_title_font,50,150)
Clamp(scale_title_font,(unsigned)50,(unsigned)150)

src/InfoBoxes/InfoBoxSettings.hpp
bomilkar: I'll change the documentation in the code.

src/Look/InfoBoxLook.cpp
MaxKellermann: That's what I thought. You're using the wrong integer literals, and you used a cast to the wrong type to work around that problem.
How about:
```
Clamp(scale_title_font, 50U, 150U)
```
It's pointless to use signed literals and then cast them to unsigned. Just use unsigned literals.
Even though an overflow during the cast is unlikely, it's always better to use the correct data types instead of casting forth and back.

bomilkar: Looks good! Learned something.
Will be in the next commit .

bomilkar:
@MaxKellermann , @kedder : are you waiting on me to do something?

MaxKellermann:
This PR contains a merge commit, and incremental fixup commits for a known-bad commit. Those things make further review very hard.

bomilkar:
Do you want me to close this PR and create a new one?

MaxKellermann:
No, never do that, that's confusing. Just fix this one.

bomilkar:
Do you want me to squash all commits into one?

MaxKellermann:
You can't squash a merge commit, you should remove that one completely, it should never be in a PR. But fixup commits should be squashed.

MaxKellermann:
btw. never "merge" from this repository - always "rebase".
To make the maintainer's work as easy as possible, you should always keep your PR branch rebased. At least rebase before you push a new revision.

bomilkar:
Started from scratch with a new PR.

PR #378, opened 2020-04-13
extending the NMEA protocol and reducing the traffic over the port

bomilkar:
There are 2 changes to the protocol, which are covered in the restructuring:

1.
I have created a mechanism for the connected device (e.g. variod)
to read the relevant current settings of XCSoar (MacCready, wing load,
polar, polar degradation). The connected device responds with these settings.
The connected device can do this at any time, especially upon restart
to sync up with XCSoar’s settings.

In case the requested setting is not available, e.g. the Pilot has not yet
set the Ballast, then the request will be answered with a "POV,C,NA,xx" sentence
where "xx" indicates the value.

2.
Since there is no method like "PutPolar" in class Device, I
continue to use OnCalculatedUpdate to transfer the polar data,
only that I check whether the coefficients have changed.
If not, then no string is sent.

The previous implementation of the OpenVario driver produced 2 polar
string (ideal polar, real polar) every time OnCalculatedUpdate() is called.
At a minimum this happens for every GPS position XCSoar receives.
So it easily sends out 4 polar strings per second, even though nothing has
changed in these polars for a while. So the change to address this is to
check if the polar information has changed before it is sends it out.
As a result polar strings are sent only when needed.

A detailed documentation of the NMEA protocol will be in the OpenVario Documentation.

### DIFF
--- a/src/Device/Driver/OpenVario.cpp
+++ b/src/Device/Driver/OpenVario.cpp
@@ -46,63 +46,203 @@ public:
   bool PutBugs(double bugs, OperationEnvironment &env) override;
   void OnCalculatedUpdate(const MoreData &basic,
                   const DerivedInfo &calculated) override;
-  static bool POV(NMEAInputLine &line, NMEAInfo &info);
+private:
+  bool POV(NMEAInputLine &line, NMEAInfo &info);
+  bool ComposeWrite(const char p_type,PolarCoefficients &p,
+                  OperationEnvironment &env);
+  bool InformUnavailable(const char *obj, OperationEnvironment &env);
+  bool RepeatIdealPolar(OperationEnvironment &env);
+  bool RepeatRealPolar(OperationEnvironment &env);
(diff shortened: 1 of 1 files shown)


bomilkar:
This PR replaces https://github.com/XCSoar/XCSoar/pull/370

MaxKellermann:
Does this implement a new OpenVario protocol version? Does the OpenVario already implement this change?

> A detailed documentation of the NMEA protocol will be in the OpenVario Documentation.

Documentation for a protocol change should be available *before* submitting code implementing it, or else the code cannot possibly be verified.

kedder:
> Does the OpenVario already implement this change?

It does not at the moment, however this is a backwards-compatible change (semantics of existing NMEA sentences did not change), so existing clients will not be affected.

MaxKellermann:
e2f1173383462d5927852611f1b9352fbee5d5e5

Issue #377, opened 2020-04-12
ANDROID: Permissions not in one go v6.8.13


lordfolken:
Hallo, 

Concering @v6.8.13
Observed on Android 7 and Android 8 devices:

If you install XCSoar.apk (6.8.13b) from the website, upon first launch XCSoar asks correctly for storage permissions. Once granted the dialog disappears and you are returned to your last application. XCSoar needs then be accessed from the notifications, after which it asks for location permissions. Once granted, you have to yet again access it from notifications and then it runs fine.

While not breaking, its highly confusing.

I have added a screen recording to document the issue:

![ScreenRecord-20200412-233615](https://user-images.githubusercontent.com/407371/79080710-7b9fa000-7d17-11ea-8334-3624f31ce7e9.gif)

Expected Behavior:
 - Launch XCSoar
 - Be asked for Storage Permission 1/2
 - Be asked for location Permission 2/2
 - Start XCSoar

dernatsch:
I decided to look into this a bit and noticed where this behavior is caused.
When the permission dialog is opened the Method onPause() of the XCSoar class is called. A few calls follow that end up [here](https://github.com/XCSoar/XCSoar/blob/master/src/Android/Main.cpp#L289-L294), where `exit(0)` causes the application to close, apparently because the event subsystem is not initialized.

lordfolken:
Thanks for looking into this. That what you describe may be an additional problem.

As for the permissions:
So you can call for all the permissions at once described here:
https://stackoverflow.com/questions/40169809/how-to-ask-and-accept-all-permissions-in-only-one-confirm-alert-dialog-in-androi

The requestPermissions function can take an array of permissions:
https://developer.android.com/reference/android/support/v4/app/ActivityCompat#requestPermissions(android.app.Activity
```java
int PERMISSION_ALL = 1; 
String[] PERMISSIONS = {Manifest.permission.READ_CONTACTS, Manifest.permission.WRITE_CONTACTS, Manifest.permission.SEND_SMS, Manifest.permission.ACCESS_FINE_LOCATION};

if(!hasPermissions(this, PERMISSIONS)){
    ActivityCompat.requestPermissions(this, PERMISSIONS, PERMISSION_ALL);
}
```
We currently request one after the other here in the requestAllPermissions function defined here:
https://github.com/XCSoar/XCSoar/blob/bc1dab4e72a0550f12772a46ff6db7d84ab5f0a2/android/src/XCSoar.java#L236

[...]

PR #392, opened 2020-05-10
request all permissions without manual restarts #377

dernatsch:
- `checkRequestPermission` is deleted and all
    permissions requested in a single call in `requestAllPermissions`
- permissions are checked in `hasAllPermissions`
- the pause when requesting the permissions does not cause an exit

### DIFF
--- a/android/src/XCSoar.java
+++ b/android/src/XCSoar.java
@@ -223,14 +223,16 @@ public class XCSoar extends Activity {
     ImmersiveFullScreenMode.enable(getWindow().getDecorView());
   }
 
-  private void checkRequestPermission(String permission) {
-    if (checkSelfPermission(permission) != PackageManager.PERMISSION_GRANTED) {
-      try {
-        this.requestPermissions(new String[]{permission}, 0);
-      } catch (Exception e) {
-        Log.e(TAG, "requestPermissions(" + permission + ") failed", e);
+  private final String[] NEEDED_PERMISSIONS = new String[] {Manifest.permission.ACCESS_FINE_LOCATION,
+                                                            Manifest.permission.WRITE_EXTERNAL_STORAGE};
+  private boolean hasAllPermissions() {
+    for (String p : NEEDED_PERMISSIONS) {
+      if (checkSelfPermission(p) != PackageManager.PERMISSION_GRANTED) {
+        return false;
       }
     }
+
+    return true;
   }
 
   private void requestAllPermissions() {
@@ -243,8 +245,13 @@ public class XCSoar extends Activity {
        permissions before using them; mentioning them in the manifest
        is not enough */
 
-    checkRequestPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE);
-    checkRequestPermission(Manifest.permission.ACCESS_FINE_LOCATION);
+    if (!hasAllPermissions()) {
+      try {
+        this.requestPermissions(NEEDED_PERMISSIONS, 0);
+      } catch (Exception e) {
+        Log.e(TAG, "could not request permissions: " + String.join(", ", NEEDED_PERMISSIONS));
+      }
+    }
   }
 
   @Override protected void onResume() {
--- a/android/src/XCSoar.java
+++ b/android/src/XCSoar.java
@@ -251,8 +251,8 @@ public class XCSoar extends Activity {
     if (!hasAllPermissions()) {
       try {
         this.requestPermissions(NEEDED_PERMISSIONS, 0);
-      } catch (Exception e) {
-        Log.e(TAG, "could not request permissions: " + String.join(", ", NEEDED_PERMISSIONS));
+      } catch (IllegalArgumentException e) {
+        Log.e(TAG, "could not request permissions: " + String.join(", ", NEEDED_PERMISSIONS), e);
       }
     }
   }
--- a/android/src/XCSoar.java
+++ b/android/src/XCSoar.java
@@ -123,11 +123,10 @@ public class XCSoar extends Activity {
     registerReceiver(batteryReceiver,
                      new IntentFilter(Intent.ACTION_BATTERY_CHANGED));
 
-    /* TODO: this sure is the wrong place to request permissions - it
-       will cause XCSoar to quit immediately; we should request
-       permissions when we need them, but implementing that is
-       complicated, so for now, we do it here to give users a quick
-       solution for the problem */
+    /* TODO: this sure is the wrong place to request permissions -
+       we should request permissions when we need them, but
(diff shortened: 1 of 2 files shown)


dernatsch:
Im wondering: What would be the best thing to happen if the permissions are not granted?
As of now the activity would simply continue regardless.

Maybe the internal GPS device should be "deactivated"?
Regarding the WRITE_EXTERNAL_STORAGE permission I have no idea.

lordfolken:
No ACCESS_FINE_LOCATION should not be an issue IMO. You could be feeding XCSoar's data via TCP/WIFI or an IOIO board, not using the Phone's integrated sensors.

lordfolken:
WRITE_EXTERNAL_STORAGE without this permission xcsoar crashes when one tries to download a map. Which isn't ideal behavior either... but i think should be the topic of another bug/pull request.

lordfolken:
I tested this on Android 9 & 7 and saw the expected behavior.

android/src/XCSoar.java
```diff
@@ -223,14 +223,16 @@ private void enableImmersiveModeIfSupported() {
...
   }
 
-  private void checkRequestPermission(String permission) {
-    if (checkSelfPermission(permission) != PackageManager.PERMISSION_GRANTED) {
-      try {
-        this.requestPermissions(new String[]{permission}, 0);
-      } catch (Exception e) {
-        Log.e(TAG, "requestPermissions(" + permission + ") failed", e);
+  private final String[] NEEDED_PERMISSIONS = new String[] {Manifest.permission.ACCESS_FINE_LOCATION,
+                                                            Manifest.permission.WRITE_EXTERNAL_STORAGE};
```
MaxKellermann, line 227: Line too long

android/src/XCSoar.java
```diff
@@ -223,14 +223,16 @@ private void enableImmersiveModeIfSupported() {
     ImmersiveFullScreenMode.enable(getWindow().getDecorView());
   }
 
-  private void checkRequestPermission(String permission) {
-    if (checkSelfPermission(permission) != PackageManager.PERMISSION_GRANTED) {
-      try {
-        this.requestPermissions(new String[]{permission}, 0);
-      } catch (Exception e) {
-        Log.e(TAG, "requestPermissions(" + permission + ") failed", e);
+  private final String[] NEEDED_PERMISSIONS = new String[] {Manifest.permission.ACCESS_FINE_LOCATION,
```
MaxKellermann, line 226: `static` missing

android/src/XCSoar.java
```diff
@@ -243,8 +245,13 @@ private void requestAllPermissions() {
        permissions before using them; mentioning them in the manifest
        is not enough */
 
-    checkRequestPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE);
-    checkRequestPermission(Manifest.permission.ACCESS_FINE_LOCATION);
+    if (!hasAllPermissions()) {
+      try {
+        this.requestPermissions(NEEDED_PERMISSIONS, 0);
+      } catch (Exception e) {
+        Log.e(TAG, "could not request permissions: " + String.join(", ", NEEDED_PERMISSIONS));
```
MaxKellermann, line 252: Bad log message - it doesn't show the error details, and instead shows a list of permissions, but is unable to point out which failed.

src/Android/Main.cpp
```diff
@@ -289,9 +289,10 @@ JNIEXPORT void JNICALL
 Java_org_xcsoar_NativeView_pauseNative(JNIEnv *env, jobject obj)
 {
   if (event_queue == nullptr || CommonInterface::main_window == nullptr)
-    /* pause before we have initialized the event subsystem does not
-       work - let's bail out, nothing is lost anyway */
-    exit(0);
+  {
```
MaxKellermann, line 292: Bad coding style

src/Android/Main.cpp
```diff
@@ -289,9 +289,10 @@ JNIEXPORT void JNICALL
...
+    /* event subsystem is not initialized, there is nothing to pause */
+    return;
```
MaxKellermann, line 294: Did you verify that this is safe to do?

android/src/XCSoar.java
dernatsch: > unable to point out which failed

As I understand it only the whole call can fail.
If a permission is denied no exception would be thrown, but `onRequestPermissionsResult` would be called with a corresponding grantResult entry.
An exception would only occur when requestCode is negative, which doesn't apply because it is fixed to `0`.

src/Android/Main.cpp
dernatsch: Again consulting the android [reference](https://developer.android.com/guide/components/activities/activity-lifecycle) I don't see why this would be unsafe.
The call to `onPause` seems to be a chance to stop or reduce activities that are not needed while the activity is not in the foreground. There is no direct mention of anything that would be unsafe to do while in this state. I think the activity would simply continue in the background.

MaxKellermann: The problem isn't Android - the problem is the big number of global variables which aren't initalized properly in XCSoar's very old code base. Doing `main()` again in a reused process may break lots of things, that's why we just `exit()` on Android to prevent reusing the process.
Of course, it would be a good idea to get rid of all global variables, and over the past 11 years (since I've joined), I have removed most of them, but not all.
I was asking because I thought maybe you were doing this in a corner case where recycling the process doesn't matter.

android/src/XCSoar.java
MaxKellermann: The code which you removed logs the Exception object. Yours does not.

dernatsch: Ah okay, I added the exception object back to the log message.

src/Android/Main.cpp
dernatsch: I see.
If I understand correctly, the first time `onResume` is called, `initSDL` initializes the NativeView. The case that this return is hit happens when the NativeView has been created but hasn't finished yet. So as soon as `nativeView` contains the object the second `onResume` does not call `initSDL` again.
So I would think that no global variables would be changed, as no "second start" took place.

dernatsch: I took another look today and one thing this causes is that the MainWindow is started while the Activity is in the paused state. When returning from the permissions dialog `CommonInterface::main_window->Resume()` is called.
This checks whether it has really been paused and just returns otherwise, so I don't see anything bad happening there either.

MaxKellermann: Agree, I think you're right.

MaxKellermann:
Cherry-picked to v6.8.x: a668ce78af7b27d1227720ecfb955d1db71b108d

PR #395, opened 2020-05-11
Added optional average needle instead of current climb needle

### LINKED ISSUE #178: include average climb vario needle for whole thermal
It would be great if there could be a  third vario needle indicator available for average climb over the whole actual thermal. may be a 90degree right rotatet "T" to indicate the average climb for the whole thermal (additional to the 30second average). This allows you to see extreamly fast if the 30s average is getting below the overall climb rate.
Additionally the color can indicate the tendency. Green for increasing overall climb rate and red for decreasing.

condoratberlin:
fixes #178 by showing the average thermal climb ratio instead of current climb ratio. The color bar shows unchanged the current thermal. While not climbing the last average climb ratio is shown. This view allows a fast assessment of the current thermal by comparing the 30s average and the average over thermal.

### DIFF
--- a/src/Dialogs/Settings/Panels/VarioConfigPanel.cpp
+++ b/src/Dialogs/Settings/Panels/VarioConfigPanel.cpp
@@ -36,6 +36,7 @@ enum ControlIndex {
   AppGaugeVarioBallast,
   AppGaugeVarioGross,
   AppAveNeedle,
+  AppAveThermalNeedle,
 };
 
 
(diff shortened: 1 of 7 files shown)


MaxKellermann:
These commits are not splitted properly. There are 4 commits, but it looks like the follow-up commits just fix bugs in previous commits. This is confusing and makes review hard.

condoratberlin:
You are right, sorry for that. I will rearrange the changes into more clear and separated commits.

condoratberlin:
I create a new pull request with merge commits.

PR #398, opened 2020-05-12
average climb needle color is not inverted

### LINKED ISSUE #397: 30s average Needle is not inverted by setting "Invert Info boxes"
Inverting the infoboxes gives them a black background. The current climb needle is inverted to white then, but the 30s average needle (trinagle) is not and can barely be seen as shown here:
XCSoar_Vario_black_30s_average_Needle

condoratberlin:
fixes #397 The average needle is painted with a pen instead a brush, therefore also the pen color is selected.


MaxKellermann:
Duplicate of https://github.com/XCSoar/XCSoar/pull/400 - that one will be merged to master.

PR #400, opened 2020-05-13
Fixes 30s average needle color when boxes are inverted for v6.8.xx

### LINKED ISSUE #397: 30s average Needle is not inverted by setting "Invert Info boxes"

condoratberlin:
fixes #397 The average needle is painted with a pen instead a brush, therefore also the pen color is selected for V6.8.xx.


MaxKellermann:
I merged this, even though the commit message is bad - it doesn't describe the problem.

PR #414, opened 2020-05-22
IGCParser: Add support for long-form `DTE` header

### LINKED ISSUE #412: IGCParser: Long-form HFDTE header not recognized
It looks like the [IGCParseDateRecord()](https://github.com/XCSoar/XCSoar/blob/719cedbf614b037c75081fa86f7be3735fb0daf4/src/IGC/IGCParser.cpp#L92) function is currently not able to read the long-form of the `HFDTE` header (e.g. `HFDTEDATE:170520,01`). According to the IGC spec this seems to be a valid way to write that header though.

Turbo87:
Resolves #412

### DIFF
--- a/src/IGC/IGCParser.cpp
+++ b/src/IGC/IGCParser.cpp
@@ -96,6 +96,13 @@ IGCParseDateRecord(const char *line, BrokenDate &date)
 
   line += 5;
 
+  if (strncmp(line, "DATE", 4) == 0) {
+    line += 4;
+  }
+  if (line[0] == ':') {
+    line += 1;
+  }
+
   char *endptr;
   unsigned long value = strtoul(line, &endptr, 10);
   if (endptr != line + 6)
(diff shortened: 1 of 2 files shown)


src/IGC/IGCParser.cpp
```diff
@@ -96,6 +96,13 @@ IGCParseDateRecord(const char *line, BrokenDate &date)
 
   line += 5;
 
+  if (memcmp(line, "DATE", 4) == 0) {
```
MaxKellermann, line 99: Buffer overflow if the line is shorter than 4 bytes, because `memcmp()` may choose any direction it likes.
Yes, much of XCSoar's existing code is wrong as well and needs to be fixed...

Turbo87: how can we avoid it? performing a `strlen()` check first?

kedder: should it be `strcmp` instead of `memcmp`?

MaxKellermann: `strncmp()`

Turbo87: alright, changed to `strncmp()`

PR #418, opened 2020-05-22
Show if files are out of date in file manager

dernatsch:
- Show "Update available" if files are out of date
- Add "Update All" button that downloads all updates

### DIFF
--- a/src/Dialogs/FileManager.cpp
+++ b/src/Dialogs/FileManager.cpp
@@ -100,6 +100,26 @@ CanDownload(const FileRepository &repository, const TCHAR *name)
   return FindRemoteFile(repository, name) != nullptr;
 }
 
+static bool
+UpdateAvailable(const FileRepository &repository, const TCHAR *name)
+{
+#ifdef HAVE_POSIX
+  const AvailableFile *remote_file = FindRemoteFile(repository, name);
+
+  if (remote_file == nullptr)
+    return false;
+
+  BrokenDate remote_changed = remote_file->update_date;
+
+  const auto path = LocalPath(name);
+  BrokenDate local_changed = static_cast<BrokenDate>(BrokenDateTime::FromUnixTimeUTC(
+                              File::GetLastModification(path)));
+
+  return local_changed < remote_changed;
+#else
+  return false;
+#endif
+}
 #endif
 
 class ManagedFileListWidget
@@ -123,12 +143,12 @@ class ManagedFileListWidget
     StaticString<32u> size;
     StaticString<32u> last_modified;
 
-    bool downloading, failed;
+    bool downloading, failed, out_of_date;
 
     DownloadStatus download_status;
 
     void Set(const TCHAR *_name, const DownloadStatus *_download_status,
-             bool _failed) {
+             bool _failed, bool _out_of_date) {
       name = _name;
 
       const auto path = LocalPath(name);
@@ -153,6 +173,8 @@ class ManagedFileListWidget
         download_status = *_download_status;
 
       failed = _failed;
+
+      out_of_date = _out_of_date;
     }
   };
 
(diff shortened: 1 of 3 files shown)


MaxKellermann:
src/Dialogs/FileManager.cpp:594:19: error: loop variable 'file' of type 'const ManagedFileListWidget::FileItem' creates a copy from type 'const ManagedFileListWidget::FileItem' [-Werror,-Wrange-loop-construct]

src/Repository/AvailableFile.hpp
```diff
@@ -51,6 +52,8 @@ struct AvailableFile {
 
   FileType type;
 
+  BrokenDate updated;
```
MaxKellermann, line 55: updated what? What does that mean? Maybe call it `update_date`. Does it need a time of day?

src/Dialogs/FileManager.cpp
```diff
@@ -452,9 +469,11 @@ ManagedFileListWidget::OnPaintItem(Canvas &canvas, const PixelRect rc,
     row_renderer.DrawRightFirstRow(canvas, rc, text);
   }
 
-  row_renderer.DrawRightSecondRow(canvas, rc, file.last_modified.c_str());
-
   row_renderer.DrawSecondRow(canvas, rc, file.size.c_str());
+
+  if (UpdateAvailable(repository, file.name)) {
```
MaxKellermann, line 477: What if that is not available - fall back to last modified?

src/Dialogs/FileManager.cpp
```diff
@@ -100,6 +100,23 @@ CanDownload(const FileRepository &repository, const TCHAR *name)
   return FindRemoteFile(repository, name) != nullptr;
 }
 
+static bool
+UpdateAvailable(const FileRepository &repository, const TCHAR *name)
+{
+  const AvailableFile *remote_file = FindRemoteFile(repository, name);
```
MaxKellermann, line 106: This whole function is rather expensive for the draw callback - it will be invoked for **every** file **every** time something on the screen gets drawn!
Don't do stuff in the draw callback!

src/Dialogs/FileManager.cpp
```diff
@@ -434,6 +436,7 @@ ManagedFileListWidget::UpdateButtons()
     download_button->SetEnabled(!items.empty() &&
                                 CanDownload(repository, items[current].name));
     cancel_button->SetEnabled(!items.empty() && items[current].downloading);
+    update_button->SetEnabled(!items.empty());
```
MaxKellermann, line 439: Enable it only if at least one update is available.

src/Repository/AvailableFile.hpp
dernatsch: But yeah, I can rename it of course.

MaxKellermann: If you give it bad names, at least add API documentation.

src/Dialogs/FileManager.cpp
dernatsch: You're right that's prob not a good idea. I'll see if I can compute it elsewhere.

dernatsch:
> src/Dialogs/FileManager.cpp:594:19: error: loop variable 'file' of type 'const ManagedFileListWidget::FileItem' creates a copy from type 'const ManagedFileListWidget::FileItem' [-Werror,-Wrange-loop-construct]

Strange, I don't encounter this at all (g++ 10.1.0). Is there some compile flag I didn't set?

MaxKellermann:
This is with clang. This is not about whether you see the warning, but about a warning about truly questionable code.

dernatsch:
> truly questionable code
I see. I made the iteration variable a reference variable, as in the other for-each loops.

MaxKellermann:
Can you fold those fixup changes into the commits? i.e. amend existing commits if they're wrong instead of adding fixup commits.

dernatsch:
> Can you fold those fixup changes into the commits?

Sure can!

MaxKellermann:
With `TARGET=PC`:
```
src/Dialogs/FileManager.cpp: In function ‘bool UpdateAvailable(const FileRepository&, const TCHAR*)’:
src/Dialogs/FileManager.cpp:114:70: error: ‘FromUnixTimeUTC’ is not a member of ‘BrokenDateTime’
  114 |   BrokenDate local_changed = static_cast<BrokenDate>(BrokenDateTime::FromUnixTimeUTC(
      |                                                                      ^~~~~~~~~~~~~~~
src/Dialogs/FileManager.cpp: In member function ‘void ManagedFileListWidget::RefreshList()’:
src/Dialogs/FileManager.cpp:414:76: error: ‘FromUnixTimeUTC’ is not a member of ‘BrokenDateTime’
  414 |         BrokenDate local_changed = static_cast<BrokenDate>(BrokenDateTime::FromUnixTimeUTC(
```

MaxKellermann:
And: please rebase every time you push corrections! That helps avoid merge conflicts.

dernatsch:
> With `TARGET=PC`:

Okay, protected by `HAVE_POSIX` now.
And rebased.

Issue #417, opened 2020-05-22
Repository files update availability


dernatsch:
On IRC @lordfolken suggested a system to notify about and automatically download files from the repository that are out of date.
This morning I put together a [solution](https://github.com/akafliegdarmstadt/XCSoar/tree/auto-file-update). It checks the update field in the repo vs the time of last change of the files.

Now what I came up with is deeply embedded into the file manager widget, so accessing it programmatically is near impossible.

Another, solution I can think of would be a kind of `RemoteResourceManager`.
An object that would completely manage the providing, downloading and updating of these files centrally. This would allow access from other parts of the program (i.e. updating the files automatically on start if the user wants it). But a lot of changes in the way these files are accessed would be necessary.

Maybe somebody has some input as to what would be the nicer solution here.

MaxKellermann:
I think your code is good, and we can merge it. Just a few things need further discussion:
- displaying file update time vs. showing when we updated the file in our list - I think we should tell the user when the file was current (e.g. which iteration of airspace changes this reflects), so he can compare it with official dates
- maybe we shouldn't really work with dates, but with checksums or version numbers? Many people have bad clock values (though rarely on Android), and using the system clock is often not the best idea and not so reliable.
- should this new feature really be merged in the stable branch?

dernatsch:
1. Certainly good for the user experience. But I think this would push a bit into the `Manager`(or however a more independent solution would look like) direction because you would need to look up this metadata for all of the locally stored files, in contrast to only probing the file update time.
2. I think the unreliability of the local clock is a good point.
A checksum would be nice because it can be calculated directly from the data, XCSoar could compute it on the fly. A version number would have to be managed/incremented somehow and would need a place to be stored (a separate file, the filename, the content itself?). In the end its a question of compute vs store I think.
3. I can rebase it of course.

dernatsch:
Two things I'm currently wondering about for sha256:

1. How to get the checksum the download generates. What comes to mind is passing it by value in `OnDownloadComplete` and adding the LocalRepository to the listeners, **or** making the DownloadManager save the hashes of the files instead of the LocalRepository (not a huge fan because of the DownloadManager being global).

2. How to persist the hashes. Recalculating the hashes could take a long time so saving them in a file could be a simple fix. But how? A CSV file? Or would it be viable to rehash the files concurrently and only providing the 'out of date'-functionality when it's done.

lordfolken:
[...]

PR #419, opened 2020-05-23
Add checksum field to Repository/AvailableFile

dernatsch:
Brief summary of the changes
----------------------------
- moved MD5 files to Util instead of Logger
- added `sha256_hash member to AvailableFile struct

Preparing a better solution for #417

### DIFF
--- a/build/test.mk
+++ b/build/test.mk
@@ -123,10 +123,16 @@ TEST_NAMES = \
 	TestIGCFilenameFormatter \
 	TestLXNToIGC \
 	TestLeastSquares \
+	TestHexString \
 	TestThermalBand
 
 
 TESTS = $(call name-to-bin,$(TEST_NAMES))
+
+TEST_HEX_STRING_SOURCES = \
+	$(TEST_SRC_DIR)/tap.c \
+	$(TEST_SRC_DIR)/TestHexString.cpp
+$(eval $(call link-program,TestHexString,TEST_HEX_STRING))
 
 TEST_CRC_SOURCES = \
 	$(SRC)/Util/CRC.cpp \
--- a/src/Repository/AvailableFile.hpp
+++ b/src/Repository/AvailableFile.hpp
@@ -54,6 +54,12 @@ struct AvailableFile {
 
   BrokenDate update_date;
 
+  /**
+  * The SHA256 hash of the contents of this file.
+  * Zeroed if no hash is available.
+  */
+  std::array<std::byte, 32> sha256_hash;
+
   bool IsEmpty() const {
     return name.empty();
   }
@@ -62,12 +68,22 @@ struct AvailableFile {
     return !name.empty() && !uri.empty();
   }
 
+  bool HasHash() const {
+    for (std::size_t i = 0; i < sha256_hash.size(); i++) {
+      if (sha256_hash[i] != std::byte{0})
+        return true;
+    }
+
+    return false;
+  }
+
   void Clear() {
     name.clear();
     uri.clear();
     area.clear();
     type = FileType::UNKNOWN;
     update_date = BrokenDate::Invalid();
+    sha256_hash.fill(std::byte{0});
   }
 
   const char *GetName() const {
--- a/src/Util/HexString.hpp
+++ b/src/Util/HexString.hpp
@@ -1,0 +1,72 @@
+/*
+Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2016 The XCSoar Project
+  A detailed list of copyright holders can be found in the file "AUTHORS".
+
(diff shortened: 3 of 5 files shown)


dernatsch:
The `md5_digest` member could also be a StaticString because it's size is known in advance. I don't know if that's better to use though.

MaxKellermann:
Are you sure our MD5 code is a standards-compliant implementation?

MaxKellermann:
And yes, it should not be dynamically allocated - or better, make it `std::array<std::byte>` (with the correct length) - store it in binary, not as hex string.

dernatsch:
> Are you sure our MD5 code is a standards-compliant implementation?

I haven't checked, so I wouldn't know. But I also didn't use it anywhere at the moment, so I could just change everything to sha256 easily.

MaxKellermann:
What use is this SHA256 commit? It does nothing but help parse a fixed-length hex string to an array of bytes. It has nothing to do with SHA256, except for converting one representation of SHA256 to another.

dernatsch:
I figured this function would practically only be used when handling SHA256, so I would group it there. The plan would be to use this namespace to do all of the interfacing to libsodium.
Of course I could put it into Repository/Parse, but then it couldn't be used elsewhere if the need arises.

dernatsch:
Another alternative I can think of would be to make `ParseHexString` a template function for it's length and provide it somewhere else in Util, then it would have nothing to do with SHA256.

MaxKellermann:
Good idea! Pass `std::string_view` to it, return the `std::array` and make it throw when parsing fails. No `std::optional` and no `bool` return values.

dernatsch:
> make it throw

I thought exceptions are a no-go as per the developer manual? Or are they alright in v7?

MaxKellermann:
Most of XCSoar's code is still not exception-safe, unfortunately, but I removed the "no exceptions" rule from v7, and existing old code should be fixed.

MaxKellermann:
I have added some libsodium integration for you: https://github.com/XCSoar/XCSoar/commit/553df61c75dcff00511d7e9e6d6639075db7ca6c

dernatsch:
Great, thanks!
I put the parsing stuff into Util/HexString now.

MaxKellermann:
Cherry-picked the MD5 move commit. Please rebase before you push next time.

build/libutil.mk
```diff
@@ -19,7 +19,7 @@ UTIL_SOURCES = \
 	$(UTIL_SRC_DIR)/StringCompare.cxx \
 	$(UTIL_SRC_DIR)/StringStrip.cxx \
 	$(UTIL_SRC_DIR)/StringUtil.cpp
-
+
```
MaxKellermann, line 22: Watch out for whitespace changes.

build/test.mk
```diff
@@ -123,11 +123,17 @@ TEST_NAMES = \
 	TestIGCFilenameFormatter \
 	TestLXNToIGC \
 	TestLeastSquares \
-	TestThermalBand
+	TestThermalBand \
+	TestHexString
```
MaxKellermann, line 127: To reduce the number of changed lines and improve "git blame", move the new line one up. Yes, this is an ugly problem.

src/Repository/AvailableFile.hpp
```diff
@@ -54,6 +54,12 @@ struct AvailableFile {
 
   BrokenDate update_date;
 
+  /**
+  * The SHA256 hash of the contents of this file.
+  */
+  std::array<std::byte, 32> sha256_hash;
+  bool has_hash;
```
MaxKellermann, line 61: This `bool` takes 8 bytes. Can we do better than that? Maybe consider the hash to be unavailable when it's all zeroes?

src/Util/HexString.hpp
```diff
@@ -0,0 +1,72 @@
...
+* Parses a string of hex data into a byte array.
+*
+* @param hex_str A string containing the data in hexadecimal form.
+* @return An array filled with the specified data.
+* @throws std::invalid_argument if parsing failed or the hex string has
+*         the wrong size for the array.
+*/
+template<std::size_t len>
+std::array<std::byte, len>
+ParseHexString(const std::string_view &hex_str)
```
MaxKellermann, line 56: Don't pass `string_view` by reference. This adds overhead for dereferencing, and cannot be optimized away sometimes.

src/Repository/AvailableFile.hpp
dernatsch: I thought about that too, but I wanted to be on the safe side.
Given how unlikely it is for a file to have a hash of all zeroes it should work though.

dernatsch: Went with testing for zeroes now, so I added a `HasHash` function.

PR #421, opened 2020-05-26
ide/docker: merge sandrinr/docker-xcsoar-buildenv into XCSoar

sandrinr:
Brief summary of the changes
----------------------------

- Use `delegated` volume modifier
- More concise notion of Docker commands
- Remove unnecessary "xcsoar" image path prefix
- Content and style improvements
- Support running XCSoar from within the Docker container

This PR is an effort to semantically merge https://github.com/sandrinr/docker-xcsoar-buildenv into XCSoar.

### DIFF
--- a/ide/docker/Dockerfile
+++ b/ide/docker/Dockerfile
@@ -16,3 +16,10 @@ WORKDIR /opt/xcsoar
 
 COPY ./docker/bin/* /usr/local/bin/
 RUN cd /usr/local/bin && chmod 755 ./*
+
+# Environment to support running XCSoar
+ENV DISPLAY=host.docker.internal:0
+ENV SDL_RENDER_DRIVER=software
+# We set the home directory to the project root such that XCSoar generated files
+# are retained.
+ENV HOME=/opt/xcsoar
--- a/ide/docker/README.md
+++ b/ide/docker/README.md
@@ -6,12 +6,13 @@ This Docker image when built, will compile XCSoar for several targets in a clean
 ## Currently Supported Targets
 
 - UNIX (linux & co)
+- UNIX-SDL (to run XCSoar from within the Docker container)
 - ANDROID
 - PC
 - KOBO
 - DOCS
 
-## Instructions
+## Instructions for building XCSoar
 
 The build container should be considered read only. When used as described below
 the output will appear in the XCSoar project directory the same way as when
@@ -47,3 +48,45 @@ docker run -it --rm -v $PWD:/opt/xcsoar:delegated xcsoar-build xcsoar-compile AN
 Note: the `delegated` volume modifier is used to speed up compilation on non Linux hosts 
 (macOS and Windows), it allows for eventual consistent filesystem change propagation 
 from the container to the host system.
+
+
+## Instructions for running XCSoar within the Docker container
+
+You can also use the Docker container to run the Linux build of XCSoar on a non-Linux
+system. This could be helpful to perform quick tests. It will use software rendering.
+Needless to say that the experience is inferior to running XCSoar on the host
+system directly. However, installing all build dependencies might also be overkill
+for small things.
+
+### macOS
+
+1. Install dependencies (socat and an X11 server) using Homebrew:
+    ```shell script
+    brew install socat
+    brew cask install xquartz
+    ```
+    Please ensure to restart your Mac after you have installed XQuartz, it is needed
+    for its environment to work correctly.
+
+1. To forward X11 network traffic out of the Docker container to the host X11-Server run
+`socat` on the host machine in a separate terminal (it will block):
+    ```shell script
+    socat TCP-LISTEN:6000,reuseaddr,fork UNIX-CLIENT:\"$DISPLAY\"
+    ```
+
+1. To compile XCSoar for this kind of usage, run:
+    ```shell script
+    docker run -it --rm -v $PWD:/opt/xcsoar:delegated xcsoar-build xcsoar-compile UNIX-SDL
+    ```
+
+1. After compilation you may run XCSoar if you have an X11
+server installed and running and the `socat` command from above running.
+    ```shell script
+    docker run -it --rm -v $PWD:/opt/xcsoar:delegated xcsoar-build output/UNIX/bin/xcsoar
+    ```
+
+### Windows
+
+*TODO*
+(The process should be the same as on macOS, however, we need someone to verify
+this.)
--- a/ide/docker/README.md
+++ b/ide/docker/README.md
@@ -1,6 +1,6 @@
 # XCSoar Docker Image
 
-This Docker Image when built, will compile XCSoar for several targets in a clean room environment.
+This Docker image when built, will compile XCSoar for several targets in a clean room environment.
 
 
 ## Currently Supported Targets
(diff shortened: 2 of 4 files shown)


.config/pulse/79372d8564c8545262975c46157f1482-runtime
```diff
@@ -0,0 +1 @@
+/tmp/pulse-PKdhtXMmr18n
```
MaxKellermann, line 1: What's this?

.xcsoar/Default.tsk
```diff
@@ -0,0 +1 @@
+<Task fai_finish="0" finish_min_height_ref="AGL" finish_min_height="0" start_max_height_ref="AGL" start_max_height="0" start_max_speed="0" start_requires_arm="0" aat_min_time="10800" type="RT"/>
```
MaxKellermann, line 1: What's this?

ide/docker/README.md
```diff
@@ -47,3 +48,45 @@ docker run -it --rm -v $PWD:/opt/xcsoar:delegated xcsoar-build xcsoar-compile AN
...
 (macOS and Windows), it allows for eventual consistent filesystem change propagation 
 from the container to the host system.
+
+
+## Instructions for running XCSoar within the Docker container
+
+You can also use the Docker container to run the Linux build of XCSoar on a non-Linux
+system. This could be helpful to perform quick tests. It will use software rendering.
+Needless to say that the experience is inferior to running XCSoar on the host
+system directly. However, installing all build dependencies might also be overkill
```
MaxKellermann, line 58: You don't need to install build dependencies to run XCSoar. Only the shared libraries XCSoar was linked with. Which most users probably already have.

ide/docker/bin/xcsoar-compile
```diff
@@ -19,13 +19,16 @@ case "${1}" in
...
+        make TARGET=UNIX OPENGL=n ENABLE_SDL=y USE_SDL2=y -j"${CPUS}" all
+      ;;
       "PC")
         make TARGET="${1}" -j"${CPUS}" all
       ;;
       "WIN64")
         make TARGET="${1}" -j"${CPUS}" all
       ;;
-      "*")
+      *)
```
MaxKellermann, line 31: How is this change related to running XCSoar inside docker?

.config/pulse/79372d8564c8545262975c46157f1482-runtime
sandrinr: Commit error on my side. Removed.

.xcsoar/Default.tsk
sandrinr: Commit error on my side. Removed.

ide/docker/README.md
sandrinr: If you are running macOS or Windows you need a whole build environment, such as compilers.

ide/docker/bin/xcsoar-compile
sandrinr: True, this is unrelated, however, it is still a bug in the current file. We can remove it here if you wish so.

rawtaz: +1 to removing it. Changes that are not related to this PR should not be in this PR. But feel free to open another PR for it if it makes sense to do so!

ide/docker/README.md
MaxKellermann: Not for running.

ide/docker/bin/xcsoar-compile
lordfolken: Please list the target UNIX-SDL also in the valid target lists on line 29

ide/docker/Dockerfile
```diff
@@ -16,3 +16,10 @@ WORKDIR /opt/xcsoar
 
 COPY ./docker/bin/* /usr/local/bin/
 RUN cd /usr/local/bin && chmod 755 ./*
+
+# Environment to support running XCSoar
+ENV DISPLAY=host.docker.internal:0
```
lordfolken, line 21: Hi, 
first of all thank you for following my call to action. :) +1 

Is this safe to hardcode for all possible runtime environments?

I saw for example this approach for a linux host: (the osx host will still require the above method)
https://medium.com/@SaravSun/running-gui-applications-inside-docker-containers-83d65c0db110

ide/docker/README.md
sandrinr: Ah, now I got your point. Sure thing, but the target audience here is someone wanting to make (small) changes to XCSoar and therefore needing a way to compile it also.

I was in this situation some time ago and after spending some time failing to get it to work natively on macOS. I decided it would be easier to just let it run on the Docker container.

MaxKellermann: Sure they can compile it in the container, but the point of this PR is to *run* it in the container. You can compile in the container which has all the development tools installed, but is running inside the container really worth it?
It is dangerous and complex and error prone and slow; for example the user has to disable authentication in his X server.
(I'm not convinced, but that doesn't mean anything - I'm not convinced of Docker at all, that's why I have never used it.)

ide/docker/bin/xcsoar-compile
sandrinr: done

ide/docker/README.md
lordfolken: Is it possible to mount in the Xauthority file on OSX just like in the medium.com example? that would make it safer for the enduser.

@MaxKellermann docker is dead. there are much saner run-times for running containers now called cri-o and podman as the pulling tool. Its a challenge to run/compile master on pretty much anything that isn't debian buster or a direct target.

lordfolken: @sandrinr in theory one cloud build a runtime container only, or one only containing the unix target. It would be much smaller than this multigigabyte monster with all the toolchains. But for that kind of runtime I'd prefeer snapper or static binaries.

sandrinr: @MaxKellermann You got a point. However, AFAIK it is not possible to build macOS binaries on Linux. I think you need Xcode for that which is only available on macOS. So building the Linux binaries and running them within the Linux container is the whole point. Besides, in XCSoar history the Linux target always worked well. Not sth that could be told about the OSX target.
However, I do agree the approach being complicated (re-routing of X11 traffic, ...). It was just simpler for me as opposed to fixing the OSX target. I guess there is no harm in deciding that this should not be part of XCSoar.

ide/docker/Dockerfile
sandrinr: I already tried this. The article you mention is targeted towards Linux. I was unable to get this to run within reasonable time on macOS. The fact, that Docker runs within a virtual machine on macOS complicates things. The `DISPLAY` variable holds the path to a UNIX socket when using XQuartz.

Now that I am thinking about this again, the solution I provided here is only really suitable for macOS and also only because the OSX target is currently not working. (PC target is working I guess.)

Maybe it is not worth to put this all into XCSoar just to work around a defunct OSX target. Better to fix the OSX target or to use my buildenv as an alternative as long as it is defunct.

ide/docker/README.md
rawtaz: @lordfolken Docker isn't dead, that's simply not true. It's used *a lot* all over the world. Even if there are better ones, they haven't hit mainstream use yet, and with that in mind having a docker container is better. It wouldn't hurt having others too, but let's not skip the docker one just because it's docker :)

MaxKellermann: So this is about building a Linux binary on Mac OS X and then running the Linux binary with Docker on Mac OS X? Does that even work? I thought Docker is a container manager, and containers share the host kernel, which isn't Linux... is Docker a VM on Mac OS X?

lordfolken: @MaxKellermann correct docker on os/x and windows use virtualization with very lightweight OSes (ramdisk/kernel)

rawtaz: Docker on macOS uses xhyve (practically bhyve) to run a VM one can enter into using `screen ~/Library/Containers/com.docker.docker/Data/vms/0/tty` to poke around:

~~~
docker-desktop:~# uname -a
Linux docker-desktop 4.19.76-linuxkit #1 SMP Thu Oct 17 19:31:58 UTC 2019 x86_64 Linux
docker-desktop:~# cat /etc/os-release 
PRETTY_NAME="Docker Desktop"
~~~

lordfolken:
Please have a look at #428

sandrinr:
We can close this since #428 was merged.

PR #427, opened 2020-06-04
XCSoar.rc: rename TESTING to XCSOAR_TESTING

fb:
restores red color scheme


MaxKellermann:
Oops, this was used there as well...

PR #432, opened 2020-06-08
Add FreeVario project driver to XCsoar

SashLi:
Add driver for FreeVario project to XCsoar. The documentation for created driver calls on https://freevario.de/en/communication-interface/

Brief summary of the changes
----------------------------

### DIFF
--- a/src/Device/Driver/FreeVario.cpp
+++ b/src/Device/Driver/FreeVario.cpp
@@ -59,9 +59,10 @@ FreeVarioDevice::PFVParser(NMEAInputLine &line, NMEAInfo &info, Port &port)
     char subCommand = line.ReadOneChar();
 
     if (type == '\0')break;
-    if (subCommand == '\0')break;
-
-       switch (type) {
+    if (subCommand == '\0')
+      break;
+    
+    switch (type) {
              case 'M': {
                if (subCommand == 'S'){
 
--- a/src/Device/Driver/FreeVario.cpp
+++ b/src/Device/Driver/FreeVario.cpp
@@ -1,0 +1,297 @@
+#include "Device/Driver/FreeVario.hpp"
+#include "Device/Driver.hpp"
+#include "Device/Util/NMEAWriter.hpp"
+#include "Message.hpp"
+#include "NMEA/Checksum.hpp"
+#include "NMEA/Info.hpp"
+#include "NMEA/Derived.hpp"
+#include "NMEA/MoreData.hpp"
+#include "NMEA/InputLine.hpp"
+#include "Units/System.hpp"
+#include "Operation/Operation.hpp"
+#include "LogFile.hpp"
+#include "Interface.hpp"
+#include "CalculationThread.hpp"
+#include "Protection.hpp"
+
+/*
+ * Commands via NMEA from FreeVario Device to XCSoar:
+ *
+ * $PFV,M,S,<double value> - set MC External to defined value
+ * $PFV,M,U                - set MC External up by 0.1
+ * $PFV,M,D                - set MC External down by 0.1
+ *
+ * $PFV,F,C                - set Vario FlightMode to Circle
+ * $PFV,F,S                - set Vario FlightMode to SpeedToFly
+ *
+ * $PFV,Q,S,<double value> - set QNH to defined value
+ * $PFV,B,S,<double value> - set Bugs to percent by capacity 0 - 100
+ *
+ */
+
+class FreeVarioDevice : public AbstractDevice {
+    Port &port;
+
+public:
+  explicit FreeVarioDevice(Port &_port):port(_port) {}
+  bool ParseNMEA(const char *line,NMEAInfo &info) override;
+  static bool PFVParser(NMEAInputLine &line, NMEAInfo &info,Port &port);
+  bool PutMacCready(double mc, OperationEnvironment &env) override;
+  bool PutBugs(double bugs, OperationEnvironment &env) override;
+  bool PutQNH(const AtmosphericPressure &pres,OperationEnvironment &env) override;
+  void OnCalculatedUpdate(const MoreData &basic, const DerivedInfo &calculated) override;
+  void OnSensorUpdate(const MoreData &basic) override;
+};
+
+/**
+ * Parse NMEA messsage and check if it is a valid FreeVario message
+ * Is true when a valid message or false if no valid message
+ */
+bool
+FreeVarioDevice::PFVParser(NMEAInputLine &line, NMEAInfo &info, Port &port)
+{
+  NullOperationEnvironment env;
+  bool validMessage = false;
+
+  while (!line.IsEmpty()) {
+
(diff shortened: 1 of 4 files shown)


SashLi:
Anything missing for this pull request ?

MaxKellermann:
Unfortunately, it is hard to track your progress because you created a new PR. Did you address all review comments?

MaxKellermann:
One thing I remember: you removed those strange error messages which were sent to the device. But I don't know why you removed them. The protocol documentation you linked in this PR doesn't mention this part of the protocol. So, did your old code violate the documented protocol?

SashLi:
I just removed it because it was debugging messages via the NMEA stream and they are not used in the project anymore. May add some debugging code in later commits and documentation. I just created this new PR to combine all to one commit as requeste by you before. Also changed the parts that were marked by you in the last PR

dernatsch:
You might have missed some wrong indents.

MaxKellermann:
No, you did not address all review comments. You did not answer a single question.
I had to look up your other PR, the one you closed for no reason at all. This wastes my time and makes it harder than necessary to discuss your code.
Silently removing code without an explanation and not answering questions makes it even harder.
All of this doesn't make me want to spend more time on reviews, because I feel like my time is being wasted.

SashLi:
- Added the link to the documentation to the PR .
- Added comments to the functions and changed the added explicit to the port . 
- Removed not needed buffer and NMEA debug output.
- May missed a bad indent here. 

Nobody wants to waste anybody’s time. We all do this for fun, so a little bit more kindness in communication would be great. C is not my main programming language so sorry when I missed something.

SashLi:
**### Feeedback request for old PR questions.**

I added a Q and A style in this comment to give you the requested feedback to all questions in the old PR. Hopefully I did not miss something.
Please give me some feedback to this, sorry for the closed PR before. 

----
- Q: Add explicit to avoid implicit conversions
**A:Added as requested**
```
explicit FreeVarioDevice(Port &_port):port(_port) {}
```
----
- Q: What does this function do? What does its return value mean? API documentation.
 **A: See Code comment**
```
/**
  * Parse NMEA messsage and check if it is a valid FreeVario message
  * Is true when a valid message or false if no valid message
  */
 bool
 FreeVarioDevice::PFVParser(NMEAInputLine &line, NMEAInfo &info, Port &port)
```
----
- Q: What kind of buffer is this?
**A: Buffer was removed, was for debugging messages via NMEA**
```
FreeVarioDevice::PFVParser(NMEAInputLine &line, NMEAInfo &info, Port &port)
{

   char buffer[80];
```
----
- Q: Since I can't find any protocol documentation, I can't verify this - but this protocol design looks strange!
**A: removed, was for debugging messages via NMEA**
```
 sprintf(buffer,"PFV,ERR,%s","Error MC IN value");
 PortWriteNMEA(port, buffer, env);
```
----
- Q: Bad indent
 **A: Which one of this do you mean ?**
```
if (subCommand == '\0')break;

        switch (type) {
             case 'M': {
```
----
- Q: Make it final 
 **A: Please give me a helping hand here.**
```
class FreeVarioDevice : public AbstractDevice {
```
----

-Q: Wait what, the FreeVario sends a command to XCSoar to increase the MacCready value? So it doesn't have its own MacCready setting?
**A: We use the same start MC as XCSoar and sync this with the external MC. We do not calculate one, we only increment or decrement the value up or down by 0.1 using a NMEA command when a rotary switch is turned in the menu. We save the last known value temporary in the FreeVario unit**
 ```
                   }
                }

                else if (subCommand == 'U'){

                    double mc = info.settings.mac_cready;
                    double newmc = mc + 0.1;
 ```                   

----

- Q: How can this happen?
**A: Removed condition, was only a safety check but value is read out of XCSoar and should not be < 0 because of sanity check in XCSoar**

```
                   double mc = info.settings.mac_cready;
                   double newmc = mc + 0.1;
                   // Check for upper and lower range
                   if (newmc <= 0){newmc = 0.0;}
```

----
- Q: Bad indent

**A: Changed indeed. Should be ok now**

```
                   else if (newmc > 5.0){newmc = 5.0;}
                    info.settings.ProvideMacCready(newmc,info.clock);
                }

                else if (subCommand == 'D'){

                    double  mc = info.settings.mac_cready;
                    double newmc = mc - 0.1;
                    // Check for lower Range
                    if (newmc <= 0){newmc = 0.0;}
                      info.settings.ProvideMacCready(newmc,info.clock);
                }
                break;
              }
              case 'B': {
                    if (subCommand == 'S'){

                                 double bugsIn;
```
----
- Q: This method sends all values on every CalculationThread step - wouldn't it be better to send only modified values?

**A: We need a very high update rate to give a smooth vario sound feedback and a streamline information on the value changed.**
```
void
 FreeVarioDevice::OnCalculatedUpdate(const MoreData &basic,const DerivedInfo &calculated)
 ``

**Comment by MaxKellermann** [comment]
> A: We need a very high update rate to give a smooth vario sound feedback and a streamline information on the value changed.

This part sounds like you don't understand my complaint. This has nothing to do with a high update rate. If you send old data again, it's not an update.

File: src/Device/Driver/FreeVario.cpp [hunk 1]
```diff
@@ -0,0 +1,297 @@
...
+
+  while (!line.IsEmpty()) {
+
+    char type = line.ReadOneChar();
+    char subCommand = line.ReadOneChar();
+
+    if (type == '\0')break;
+    if (subCommand == '\0')break;
+
+       switch (type) {
```
dernatsch, line 64: The indentation gives the impression that the switch is in the body of the preceding if clause, which it isn't.
```suggestion
    if (subCommand == '\0')
      break;
    
    switch (type) {
```

SashLi:
> > A: We need a very high update rate to give a smooth vario sound feedback and a streamline information on the value changed.
> 
> This part sounds like you don't understand my complaint. This has nothing to do with a high update rate. If you send old data again, it's not an update.

We decided to always send the complete data on the OnCalculatedUpdate task so we no not need to decide in the driver if the value was updated before and which information state the FreeVario have at this moment. Is there a better way to solve this or do you have another approach for this  ?

MaxKellermann:
> We decided to

Who is "we" and why did you decide that?

> Is there a better way to solve this or do you have another approach for this ?

Solve what?

"You" decided not to do it the way I suggested to, for reasons you did not explain, and now you ask for my approach. Which I told you already in the review comment.
If you have questions about my suggestion, ask, but only if you're really interested. Since it sounds like you already made up your mind, I'm not confident you're interested.

Blaubart:
What a strange communication. But I can help you to solve the most important question: I am we!
It could be so easy if you give short and clear instructions. We waste your time? Are you kidding me????

MaxKellermann:
Just letting you know: I've unsubscribed from this PR because I don't want to do more code reviews here.

Blaubart:
Thanks for your lovely cooperativeness

PR #435, opened 2020-06-10
Krt2backport

hjr:
This is a backport of the KRT2 driver from master to v6.8.x branch. It is ready to be mergerd into v6.8.x.


hjr:
So I have got positive feedback on the KRT2 driver working on the v6.8.x branch. There was actually no change needed with regard of the driver model, found just difference in mutex handling to master head. I will investigate in creating a nicer version of this backport branch and get back here.

MaxKellermann:
Let's reopen this when all issues have been resolved.

PR #437, opened 2020-06-12
Repo file checksums

dernatsch:
Brief summary of the changes
----------------------------

This puts the handling of repository files into a class FileRepository.
When the repository file is loaded another file `local_hashes` is loaded.
Upon downloading a file the hash that is denoted in the `repository` file is saved and persisted in the `local_hashes`.
Out-of-date-ness is checked by comparing the hashes, or falling back to the file modification timestamps.

Issue #417

### DIFF
--- a/src/Dialogs/FileManager.cpp
+++ b/src/Dialogs/FileManager.cpp
@@ -44,6 +44,7 @@ Copyright_License {
 #include "Util/Macros.hpp"
 #include "Repository/FileRepository.hpp"
 #include "Repository/Parser.hpp"
+#include "Repository/RepoFileManager.hpp"
 
 #ifdef HAVE_DOWNLOAD_MANAGER
 #include "Repository/Glue.hpp"
@@ -98,27 +99,6 @@ static bool
 CanDownload(const FileRepository &repository, const TCHAR *name)
 {
   return FindRemoteFile(repository, name) != nullptr;
-}
-
-static bool
-UpdateAvailable(const FileRepository &repository, const TCHAR *name)
-{
-#ifdef HAVE_POSIX
-  const AvailableFile *remote_file = FindRemoteFile(repository, name);
-
-  if (remote_file == nullptr)
-    return false;
-
-  BrokenDate remote_changed = remote_file->update_date;
-
-  const auto path = LocalPath(name);
-  BrokenDate local_changed = static_cast<BrokenDate>(BrokenDateTime::FromUnixTimeUTC(
-                              File::GetLastModification(path)));
-
-  return local_changed < remote_changed;
-#else
-  return false;
-#endif
 }
 #endif
 
@@ -191,7 +171,8 @@ class ManagedFileListWidget
   bool some_out_of_date;
 #endif
 
-  FileRepository repository;
+  RepoFileManager repomanager;
+  const FileRepository &repository = repomanager.repository;
 
 #ifdef HAVE_DOWNLOAD_MANAGER
   /**
@@ -355,6 +336,8 @@ ManagedFileListWidget::Unprepare()
     Net::DownloadManager::RemoveListener(*this);
 
   download_notify.ClearNotification();
+
+  repomanager.Close();
 #endif
 
   DeleteWindow();
@@ -381,11 +364,9 @@ try {
   }
 #endif
 
-  repository.Clear();
 
   const auto path = LocalPath(_T("repository"));
-  FileLineReaderA reader(path);
-  ParseFileRepository(repository, reader);
+  repomanager.Load(path);
 } catch (const std::runtime_error &e) {
 }
 
@@ -416,9 +397,8 @@ ManagedFileListWidget::RefreshList()
       bool is_out_of_date = false;
 #ifdef HAVE_POSIX
       if (file_exists) {
-        BrokenDate local_changed = static_cast<BrokenDate>(BrokenDateTime::FromUnixTimeUTC(
-                                    File::GetLastModification(path)));
-        is_out_of_date = (local_changed < remote_file.update_date);
+        is_out_of_date = repomanager.IsOutOfDate(remote_file.GetName());
+
 
         if (is_out_of_date)
           some_out_of_date = true;
@@ -528,16 +508,8 @@ ManagedFileListWidget::Download()
   assert(current < items.size());
 
   const FileItem &item = items[current];
-  const AvailableFile *remote_file_p = FindRemoteFile(repository, item.name);
-  if (remote_file_p == nullptr)
-    return;
 
-  const AvailableFile &remote_file = *remote_file_p;
-  const UTF8ToWideConverter base(remote_file.GetName());
-  if (!base.IsValid())
-    return;
-
-  Net::DownloadManager::Enqueue(remote_file.uri.c_str(), Path(base));
+  repomanager.Download(item.name.c_str());
 #endif
 }
 
@@ -608,11 +580,7 @@ ManagedFileListWidget::Add()
   assert((unsigned)i < list.size());
 
   const AvailableFile &remote_file = list[i];
-  const UTF8ToWideConverter base(remote_file.GetName());
-  if (!base.IsValid())
-    return;
-
-  Net::DownloadManager::Enqueue(remote_file.GetURI(), Path(base));
+  repomanager.Download(remote_file.GetName());
 #endif
 }
 
@@ -622,17 +590,8 @@ ManagedFileListWidget::UpdateFiles() {
   assert(Net::DownloadManager::IsAvailable());
 
   for (const auto &file : items) {
-    if (UpdateAvailable(repository, file.name)) {
-      const AvailableFile *remote_file = FindRemoteFile(repository, file.name);
-
-      if (remote_file != nullptr) {
-        const UTF8ToWideConverter base(remote_file->GetName());
-        if (!base.IsValid())
-          return;
-
-        Net::DownloadManager::Enqueue(remote_file->GetURI(), Path(base));
-      }
-    }
+    if (repomanager.IsOutOfDate(file.name.c_str()))
+      repomanager.Download(file.name.c_str());
   }
 #endif
 }
--- a/src/Repository/RepoFileManager.cpp
+++ b/src/Repository/RepoFileManager.cpp
@@ -1,0 +1,169 @@
+/*
+Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2016 The XCSoar Project
+  A detailed list of copyright holders can be found in the file "AUTHORS".
+
+  This program is free software; you can redistribute it and/or
+  modify it under the terms of the GNU General Public License
+  as published by the Free Software Foundation; either version 2
+  of the License, or (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+
+#include "Repository/RepoFileManager.hpp"
+#include "IO/BufferedOutputStream.hxx"
+#include "IO/FileOutputStream.hxx"
+#include "IO/KeyValueFileReader.hpp"
+#include "IO/KeyValueFileWriter.hpp"
+#include "LocalPath.hpp"
+#include "IO/FileLineReader.hpp"
+#include "Net/HTTP/DownloadManager.hpp"
+#include "OS/FileUtil.hpp"
+#include "Repository/FileRepository.hpp"
+#include "Repository/Parser.hpp"
+#include "Time/BrokenDate.hpp"
+#include "Time/BrokenDateTime.hpp"
+#include "Util/ConvertString.hpp"
+#include "Util/HexString.hpp"
+
+void
+RepoFileManager::Load(Path repofilepath)
+{
+  repository.Clear();
+
+  try {
+    FileLineReaderA reader(repofilepath);
+    bool success = ParseFileRepository(repository, reader);
+
+    if (!success) {
+      repository.Clear();
+    }
+  } catch (...) {
+    repository.Clear();
+    return;
+  }
+
+  LoadLocalHashes();
+}
+
+void
+RepoFileManager::Download(const char *name)
+{
+  const AvailableFile *remote_file = repository.FindByName(name);
+  if (remote_file == nullptr)
+    return;
+
+  const UTF8ToWideConverter base(remote_file->GetName());
+  if (!base.IsValid())
+    return;
+
+  Net::DownloadManager::Enqueue(remote_file->GetURI(), Path(base));
+
+  if (remote_file->HasHash()) {
+    local_hashes[remote_file->name] = remote_file->sha256_hash;
+  }
+}
+
+bool
+RepoFileManager::IsOutOfDate(const char *name)
+{
+  const AvailableFile *remote_file = repository.FindByName(name);
+
+  if (remote_file == nullptr)
+    return false;
+
+  if (remote_file->HasHash()) {
+    try {
+      std::array<std::byte, 32> local_hash = local_hashes.at(name);
+
+      return local_hash != remote_file->sha256_hash;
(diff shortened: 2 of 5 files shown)


dernatsch:
@lordfolken I know you said it would be better to calculate the checksums locally but I haven't found a way to do it nicely that also works on Android yet. So for now this is what you suggested first: just saving the hashes that are in the repo file.

MaxKellermann:
I don't like the first commit. The commit message doesn't explain why this change is necessary. And it adds a huge tail of dependencies to what was previously a very simple data structure.
Higher-level code like that should live in a high-level library.

src/Repository/FileRepository.hpp
```diff
@@ -24,6 +24,7 @@ Copyright_License {
 #ifndef XCSOAR_FILE_REPOSITORY_HPP
 #define XCSOAR_FILE_REPOSITORY_HPP
 
+#include "OS/Path.hpp"
```
MaxKellermann, line 27: This adds a heavy header dependency, but this should better be a forward declaration.

src/Repository/FileRepository.hpp
```diff
@@ -33,16 +34,25 @@ Copyright_License {
  * The announced contents of a remote file repository.  These are the
  * files that can be downloaded by the XCSoar user.
  */
-struct FileRepository {
+class FileRepository {
```
MaxKellermann: Your commit message says
> FileRepository is made a class because it exerts behaviour instead of only holding data

... but I don't understand this deduction.
Having behaviour is not what makes the difference between `struct` and `class`. A `struct` can have behavior as well - and in fact, the only difference between the two is the default member access.

src/Repository/FileRepository.hpp
```diff
@@ -33,16 +34,25 @@ Copyright_License {
...
+  void AddFile(AvailableFile &file);
+
+public:
   void Clear() {
     files.clear();
   }
 
+  void Load(Path repofile_path);
+  void Download(std::string name);
+  bool IsOutOfDate(std::string name);
```
MaxKellermann, line 60: This requires the caller to copy its name to a newly allocated `std::string`. Unnecessary bloat!

src/Repository/FileRepository.hpp
```diff
@@ -54,7 +64,9 @@ struct FileRepository {
   }
 
   gcc_pure
-  const AvailableFile *FindByName(const char *name) const;
+  const AvailableFile *FindByName(std::string name) const;
```
MaxKellermann, line 74: Again, unnecessary bloat - and you even changed an existing lean interface to a bloated one, for no reason.

src/Repository/FileRepository.hpp
```diff
@@ -54,7 +64,9 @@ struct FileRepository {
...
+
+
```
MaxKellermann: Extra empty lines - what for?

src/Repository/Parser.cpp
```diff
@@ -24,9 +24,16 @@ Copyright_License {
...
+#include "IO/FileLineReader.hpp"
 #include "Util/StringStrip.hxx"
 #include "Util/HexString.hpp"
+#include "LocalPath.hpp"
 
+/**
+ * Parses a line of the repository file.
+ * Each line is of the form `name = value`
+ * @returns A pointer to the value field.
+ */
```
MaxKellermann, line 36: Adding documentation is always good, but please make this a separate commit. I'd like to cherry-pick this change, but I can't.

src/Repository/Parser.cpp
```diff
@@ -47,23 +54,16 @@ ParseLine(char *line)
   return value;
 }
 
-static bool
-Commit(FileRepository &repository, AvailableFile &file)
+/**
+ * Reads the repository file at `repofile_path`.
+ */
```
MaxKellermann, line 59: API documentation of public/exported symbols belongs to the header.

MaxKellermann:
Your "local_hashes" code doesn't verify if the file was modified by the user. It will be wrong all the time, causing lots of sorrow.

dernatsch:
> I don't like the first commit. The commit message doesn't explain why this change is necessary. And it adds a huge tail of dependencies to what was previously a very simple data structure.

It is quite the change. My rationale behind this was providing an interface for interacting with the Repository instead of only holding the data it represents in a struct, which caused the FileManager Dialog to handle all of the stuff that maybe shouldn't be in UI code. Surely this could be an individual object but it would be largely the same. If that class would deal with a FileRepository object as it was it would merely be a shallow embedding of an AvailableFile-container, and I don't quite see the use in that.
If you have a more elegant suggestion I'm eager to implement it.

src/Repository/FileRepository.hpp
dernatsch: Yes it's not worded in the best way possible.
I was taught that in C++ structs should be primarily used for holding raw data and as soon as an abstraction of the underlying data or behavior is added (where private members are used) a class is more appropriate.
But I can understand that there are different schools of thought so I'm also good with a struct.

MaxKellermann: It's perfectly fine to convert struct to class if you want to wrap its implementation details in nicer public methods. But that was not your declared reason to do it, and it was hidden deep inside a complex patch, so I couldn't deduce your "real" reason from your code.

src/Repository/FileRepository.hpp
dernatsch: I tried to be consistent with using `std::string` as identifiers for the files.
But yeah they should references.

MaxKellermann:
`FileRepository` just maintains the list of files available in a remote repository. It *describes* the remote repository, but it *is not* the remote repository. We should keep that as a simple data structure, because doing so allows us to write simple unit tests for parsing repository listings.
With your change, such a unit test would require linking a shitload of more code, and requires complex initialization before using it.
Keep it simple!

None of the code you moved into `FileRepository` need to be a method. They "use" the `FileRepository`, but they do not need to *be* a `FileRepository`. If you really want to move the code out of `Dialogs/FileManager.cpp` (for reasons you never explained!), move them to a separate `.cpp` file and make them functions, passing a `FileRepository` reference to them (if they need to use one).

src/Repository/FileRepository.hpp
MaxKellermann: Why references? That's still bloat, just slightly less.

dernatsch:
> `FileRepository` just maintains the list of files available in a remote repository. It _describes_ the remote repository, but it _is not_ the remote repository. We should keep that as a simple data structure, because doing so allows us to write simple unit tests for parsing repository listings.
> With your change, such a unit test would require linking a shitload of more code, and requires complex initialization before using it.
> Keep it simple!

I agree with that being nicer for testing.
 
> None of the code you moved into `FileRepository` need to be a method. They "use" the `FileRepository`, but they do not need to _be_ a `FileRepository`. If you really want to move the code out of `Dialogs/FileManager.cpp` (for reasons you never explained!), move them to a separate `.cpp` file and make them functions, passing a `FileRepository` reference to them (if they need to use one).

The reason why I tried to pull this out of the FileManagerDialog is that this behavior could also be useful elsewhere. I already mentioned this in #417.
I disagree with it not _being_ a Repository. If there was no `local_hashes` object it would need to load the file every time it would need a hash, which is nonsense. The `local_hashes` object and a FileRepository it is needed by practically all the functionality I added and adding a bunch of static methods that get passed a reference/pointer to both of these things doesn't seem logical to me. It might as well be C then.

The `FileRepository` could still be a part of a new type that handles the abstractions. That would allow all of the testing to be done on the FileRepository itself, and provide a way to keep a `local_hashes` associated with it.

src/Repository/Parser.cpp
dernatsch: Split out.

dernatsch:
> Your "local_hashes" code doesn't verify if the file was modified by the user. It will be wrong all the time, causing lots of sorrow.

What should happen if a user modified a file?
If a user updates a file manually it wouldn't be out of date and if they just modified the old content somehow it would be. The only way to handle this that I see is computing the hash locally, which turns out to be kind of hard.

MaxKellermann:
> What should happen if a user modified a file?

I'm not sure, there are various ways to deal with this - what would you do, and is this what your code really does?
In any case, XCSoar shouldn't just overwrite files with "updated versions" if the user just happens to have a file with the same name in his XCSoarData. XCSoar should be confident that this file is really managed by the file manager.

The worst way to deal with this problem is to ignore it.

dernatsch:
> what would you do, and is this what your code really does?

I haven't encountered a use-case to modify a file manually so I didn't think of this problem at all.
If a user modifies a file and a new one gets releases the modified one would be out of date too, so showing the "Update available" would be the right thing I think.
If the user clicks download and and the file has been modified it could
- back up the modified file
- save the update under a different name
- show a dialog to warn the user his file will be overwritten
- probably a lot more

I think the best way would depend on the motivation to modify these files, which I'm not aware of at the moment.

MaxKellermann:
And what "if the user just happens to have a file with the same name" ... let's say he created `Germany.cup` which just happens to be the same name as in our repository. What happens? And what should happen?

dernatsch:
> What happens? And what should happen?

It gets overwritten but that also with the way that it is at the moment.
If a user created a file with the same name should shadow the repository file they would need to be separated, which would be quite a pain and not letting it show up in the FileManager or blocking it's download is unintuitive.
I think a warning that a user created file exists that would be overwritten would be the best way. There just needs to be a way to determine this reliably.
Do you have a suggestion as to what should happen?

dernatsch:
> I don't like the first commit. The commit message doesn't explain why this change is necessary. And it adds a huge tail of dependencies to what was previously a very simple data structure.
> Higher-level code like that should live in a high-level library.

I put my logic into a new object now and left the FileRepository alone. I think it came out a bit cleaner. What do you think?

MaxKellermann:
Fails to build:
```
./src/Util/HexString.hpp:75:1: error: implicit instantiation of undefined template 'std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>'
```

src/Dialogs/FileManager.cpp
```diff
@@ -191,7 +171,8 @@ class ManagedFileListWidget
   bool some_out_of_date;
 #endif
 
-  FileRepository repository;
+  RepoFileManager repomanager;
+  const FileRepository &repository = repomanager.repository;
```
MaxKellermann, line 175: What's the point of this field?

src/Repository/RepoFileManager.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+#include "IO/BufferedOutputStream.hxx"
+#include "IO/FileOutputStream.hxx"
+#include "IO/KeyValueFileReader.hpp"
+#include "IO/KeyValueFileWriter.hpp"
+#include "LocalPath.hpp"
+#include "IO/FileLineReader.hpp"
+#include "Net/HTTP/DownloadManager.hpp"
+#include "OS/FileUtil.hpp"
+#include "Repository/RepoFileManager.hpp"
```
MaxKellermann, line 32: A library's own header should be included first to verify the header includes.

src/Repository/RepoFileManager.cpp
```diff
@@ -0,0 +1,153 @@
...
+#include "Repository/FileRepository.hpp"
+#include "Repository/Parser.hpp"
+#include "Time/BrokenDate.hpp"
+#include "Time/BrokenDateTime.hpp"
+#include "Util/ConvertString.hpp"
+#include "Util/HexString.hpp"
+
+#define LOCAL_HASHES_FILE "local_hashes"
```
MaxKellermann, line 40: No macros.
And I don't like the file format. It's undocumented, but doesn't contain any management information but the file name and the digest.
No last-known file size and time stamp (for quick comparison), for example.
I've been thinking about integrating SQLite in XCSoar so we can put everything in a database...

src/Repository/RepoFileManager.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+void 
+RepoFileManager::Load(Path repofilepath)
+{
+  repository.Clear();
+
+  FileLineReaderA reader(repofilepath);
```
MaxKellermann, line 47: Who catches exceptions?

src/Repository/RepoFileManager.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Reads the hashes of files from the repo from the disk.
+ * Every line is an entry of the form `filename = hash`.
+ * This is done so we don't have to calculate the
+ * checksum of the local files on every start.
+ */
+void
+RepoFileManager::LoadLocalHashes() {
+  FileLineReaderA linereader(LocalPath(LOCAL_HASHES_FILE));
```
MaxKellermann, line 120: Who catches exceptions?

src/Repository/RepoFileManager.cpp
```diff
@@ -0,0 +1,153 @@
...
+  }
+}
+
+/**
+ * Writes the hashes of the files from the repo to the disk.
+ */
+void
+RepoFileManager::SaveLocalHashes() {
+  FileOutputStream file_os(LocalPath(LOCAL_HASHES_FILE),
+                           FileOutputStream::Mode::CREATE_VISIBLE);
```
MaxKellermann, line 142: Why `CREATE_VISIBLE`? Who will read the file while it's being written?

src/Repository/RepoFileManager.cpp
```diff
@@ -0,0 +1,153 @@
...
+  BufferedOutputStream buf_os(file_os);
+  KeyValueFileWriter kvwriter(buf_os);
+
+  for (const auto &pair : local_hashes) {
+    std::string name(pair.first);
+    std::string hash(RenderHexString(pair.second));
```
MaxKellermann, line 148: Why copy those strings to new allocations? Pointless overhead.

src/Repository/RepoFileManager.cpp
```diff
@@ -0,0 +1,153 @@
...
+    kvwriter.Write(name.c_str(), hash.c_str());
+  }
+
+  buf_os.Flush();
```
MaxKellermann, line 152: The `FileOutputStream` needs to be committed, or else the file will be deleted by its constructor.

src/Repository/RepoFileManager.hpp
```diff
@@ -0,0 +1,64 @@
...
+   * Checks if a local file has a newer version in the repository.
+   * If hashes are available they are compared, if not the file
+   * modification timestamp is older than the `update` field in
+   * the repository file.
+   */
+  bool IsOutOfDate(const char *name);
+
+private:
+  void LoadLocalHashes();
+  void SaveLocalHashes();
```
MaxKellermann, line 63: Who calls this?

src/Util/HexString.hpp
```diff
@@ -69,4 +69,22 @@ ParseHexString(const std::string_view hex_str)
 
   return raw_hash;
 }
+
+template<std::size_t len>
+std::string
```
MaxKellermann, line 74: Using `std::string` here is unnecessary bloat. You know the size already at compile time; just return `StringBuffer<len*2+1>`!

src/Util/HexString.hpp
```diff
@@ -69,4 +69,22 @@ ParseHexString(const std::string_view hex_str)
...
+RenderHexString(const std::array<std::byte, len> data)
+{
+  constexpr char digits[] = "0123456789ABCDEF";
+  std::string out;
+
+  for (std::size_t i = 0; i < len; i++) {
+    int upper = (static_cast<char>(data[i])&0xF0)>>4;
+    int lower =  static_cast<char>(data[i])&0x0F;
```
MaxKellermann, line 82: Why is this a signed integer?

src/Dialogs/FileManager.cpp
dernatsch: It prevents having `repomanager.repository` all over the place when iterating over the available files, which I think is a bit ugly.

src/Repository/RepoFileManager.cpp
dernatsch: > No macros.

constexpr instead?

> No last-known file size and time stamp (for quick comparison), for example.

I can put those into a CSV file and write some explanatory comments, if that's better.

> I've been thinking about integrating SQLite in XCSoar so we can put everything in a database...

I'd be a fan of that.

MaxKellermann:
One more idea... after all these years, it may be about time to manage "managed files" in a separate XCSoarData subdirectory, to avoid clashing with manually managed files. And have a separate subdirectory for each file type. Because some short-sighted people gave everything a `.txt` suffix.... *shudder*

That and SQLite support for structured data may be a good and mature solution.

MaxKellermann:
This PR looks like it has been abandoned by the author 3 years ago.

PR #439, opened 2020-06-27
build/targets: Remove obsolete "VFP register arguments" workaround

Turbo87:
This workaround seems to no longer be necessary and is now causing errors like:

```
clang++: error: -Wl,--no-warn-mismatch: 'linker' input unused [-Werror,-Wunused-command-line-argument]
```


MaxKellermann:
v6.8.x: 788c72e5d89abf6029d28b1328935b7fa94f4b5c

PR #447, opened 2020-07-06
Identify BLE devices when saving address

clydebarrow:
This is a partial fix for #149. It solves the reconnect problem, but does not preserve the name.

My previous PR to solve both problems was rejected as it used storage outside the native XCSoar profile. I spent some time revising code to store both the device type and name in the XCSoar profile, but the number of affected files was considerable.

This PR is more modest in its scope but solves the major problem of automatic reconnection without relying on any storage other than the existing profile.

### DIFF
--- a/android/src/BluetoothHelper.java
+++ b/android/src/BluetoothHelper.java
@@ -94,6 +94,9 @@ final class BluetoothHelper {
     if (adapter == null)
       return null;
 
+    if (address.startsWith("LE:")) {
+        address = address.substring(3);
+    }
     try {
       return adapter.getRemoteDevice(address).getName();
     } catch (Exception e) {
@@ -134,11 +137,16 @@ final class BluetoothHelper {
     if (adapter == null)
       return null;
 
+    boolean isLe = false;
+    if (address.startsWith("LE:")) {
+        isLe = true;
+        address = address.substring(3);
+    }
     BluetoothDevice device = adapter.getRemoteDevice(address);
     if (device == null)
       return null;
 
-    if (hasLe && BluetoothDevice.DEVICE_TYPE_LE == device.getType()) {
+    if (hasLe && (isLe || BluetoothDevice.DEVICE_TYPE_LE == device.getType())) {
       Log.d(TAG, String.format(
                                "Bluetooth device \"%s\" (%s) is a LE device, trying to connect using GATT...",
                                device.getName(), device.getAddress()));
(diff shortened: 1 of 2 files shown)


clydebarrow:
Reviewing comments on previous PR, it seems like you might prefer this fix on the 6.8.x branch initially. The affected code in that branch and master are different enough that a simple merge fails, but I can submit a PR for 6.8.x instead of or in addition to this if preferred.

MaxKellermann:
This breaks existing configurations, doesn't it?
Also your commit message is very bad.

clydebarrow:
It doesn't break any *working* configurations. Any existing addresses of BLE devices saved in the profile are already broken, and addresses of legacy BT devices will be unaffected.

MaxKellermann:
I wish your commit message would explain anything. You make it hard for me to understand what this is about.

clydebarrow:
I believe I gave a full explanation of the problem being solved in #149. This commit stores the *type* of the device along with the address enabling the correct connection sequence to be executed when attempting reconnection. It does so by munging the address of a BLE device only with the prefix "LE:". When XCSoar is restarted and a reconnection attempt is made, the presence of that prefix signals that the address is that of a BLE device which requires a GATT connection attempt rather than a legacy RFCOMM connection. The prefix is itself readily distinguishable from a valid, unadorned address since it contains the character "L" which cannot occur in a simple BT MAC address. Legacy BT device addresses are stored without the prefix, as previously, and connection to those is unaffected.

MaxKellermann:
Now this is the first explanation of your approach to the solution. But unfortunately, it's still not part of your PR branch.
If I would merge this PR, all of your explanations (of the problem and your solution) would be lost, and next year, we'd wonder what this strange code means. I want no more obscure code in XCSoar!
This is why I don't go looking for explanations when I do a code review. All I look at is the PR branch, i.e. your commits.

MaxKellermann:
Also check https://github.com/XCSoar/XCSoar/pull/152#issuecomment-456310187
One and a half year ago, I complained about your bad commit message.
And here, you had an idea: https://github.com/XCSoar/XCSoar/pull/152#issuecomment-456690537
... which is better than this PR, because it's a proper solution - while this PR implements a quick kludge to work around the problem.

clydebarrow:
Ok, I have created #452 against the 6.8 branch. This implements the  fix as I previously suggested in #152. Many more files and lines changed for the same effect. Closing this PR.

PR #451, opened 2020-07-08
Add -target option to TARGET_CPPFLAGS 

### LINKED ISSUE #450: Master branch fails to build libcurl for Android7
7.0_preview15

Expected behavior
-----------------
App can be built for Android on MacOs using `make TARGET=ANDROID7`

Actual behavior
---------------
Building fails during the curl library build. 

Run `make TARGET=ANDROID7` with the current master HEAD on MacOs.

The curl build fails as it does not set `HAVE_SOCKET`. This in turn is because the `configure` script checks for the presence of a prototype for `socket()` by running the preprocessor (CPP).

The Android build is performed with llvm, found in the Android NDK distribution. llvm implements CPP with `clang -E`, however `configure` runs this with arguments from `CPPFLAGS` which is different to `CFLAGS`. Since clang has to know the target architecture it requires a `-target` option, which is included in `CFLAGS` but currently not `CPPFLAGS`.  Without this the default include paths are not set up correctly, and in the test code `#include <sys/types.h>` fails.

Add a `-target` option to CPPFLAGS.

PR to follow.

clydebarrow:
Brief summary of the changes
----------------------------
Add -target option to TARGET_CPPFLAGS to ensure that when configure uses cpp to run checks that it gets the correct system include paths. Without clang (aka cpp) knowing the target architecture it cannot provide the correct include paths.

Closes #450

### DIFF
--- a/build/llvm.mk
+++ b/build/llvm.mk
@@ -23,7 +23,7 @@ endif
 ifeq ($(TARGET),ANDROID)
   TARGET_ARCH := $(filter-out -mthumb-interwork,$(TARGET_ARCH))
   TARGET_ARCH += -integrated-as
-  TARGET_CPPFLAGS += -DBIONIC -DLIBCPP_NO_IOSTREAM
+  TARGET_CPPFLAGS += -DBIONIC -DLIBCPP_NO_IOSTREAM $(TARGET_ARCH)
   TARGET_LDLIBS += -latomic
 endif # Android
 


MaxKellermann:
These are not preprocessor options, and thus don't belong in CPPFLAGS.
If your description in #450 is right, then this sounds like a CURL bug. You can't check for C functions in C headers by running the preprocessor - you need to run the C compiler.

PR #452, opened 2020-07-08
Add new device type for BLE serial ports

### LINKED ISSUE #149: Reconnect to BLE serial interface fails after Android Bluetooth restart

clydebarrow:
Brief summary of the changes
----------------------------
Add a new device type to represent Bluetooth Low Energy serial port devices. These must be distinguished from legacy Bluetooth serial devices using the RFCOMM profile since the connection sequence is different. The old code queried the Android BT system to try to determine the type of a device based solely on its address. This fails if the Android BT cache has been flushed (by device or BT system restart.) This commit adds a new device type - BLE_HM10 - representing a BLE device implementing a serial port on a specific characteristic.

The stored type is used to select the correct re-connection protocol.

Fixes #149

### DIFF
--- a/src/Dialogs/Device/DeviceEditWidget.cpp
+++ b/src/Dialogs/Device/DeviceEditWidget.cpp
@@ -236,13 +236,18 @@ FillAndroidBluetoothPorts(DataFieldEnum &df, const DeviceConfig &config)
 {
 #ifdef ANDROID
   JNIEnv *env = Java::GetEnv();
+  // list() returns an array of strings, 3 for each device, giving
+  //   mac address
+  //   name
+  //   type - either "BLE" or "CLASSIC"
+#define BLUETOOTH_LIST_STRIDE 3
   jobjectArray bonded = BluetoothHelper::list(env);
   if (bonded == NULL)
     return;
 
-  jsize n = env->GetArrayLength(bonded) / 2;
+  jsize n = env->GetArrayLength(bonded) / BLUETOOTH_LIST_STRIDE;
   for (jsize i = 0; i < n; ++i) {
-    jstring address = (jstring)env->GetObjectArrayElement(bonded, i * 2);
+    jstring address = (jstring)env->GetObjectArrayElement(bonded, i * BLUETOOTH_LIST_STRIDE);
     if (address == NULL)
       continue;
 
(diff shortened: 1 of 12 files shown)


MaxKellermann:
From a first glance, this looks good! I have some minor complaints, but first this one:
Why "Catch exception and return null rather than throwing"? By propagating exceptions, the caller can catch and obtain error details, e.g. to show it to the user.
Propagating errors and showing details to the user is something XCSoar is currently bad at. I've tried to improve 7.0 a lot, but we're not there, existing code is mostly bad.
But can we make your new code better, as an example of how to do it right?

clydebarrow:
I can see that you have changed the way in which exceptions are handled in V7, but for 6.8 I followed the same pattern as the existing code. 

I have done a similar patch for v7 with exception propogation, but there are issues: e.g. v7 has added a new port type "GLIDER_LINK" but it is incompletely implemented, which makes adding yet another port type problematic. I think I have sorted that out, and I can build v7 but it does not run well in several regards so I have not been able to properly test it.

MaxKellermann:
Following the same pattern as existing code is not a good idea, when existing code is known to be bad.
I don't understand your Glider Link problem. What issue number are you talking about?

clydebarrow:
> Following the same pattern as existing code is not a good idea, when existing code is known to be bad.

Well, the code in 6.8 is what it is, bad or not. I could not rewrite the whole exception handling in that branch - I had to make it work with what is there.

> I don't understand your Glider Link problem. What issue number are you talking about?

There is no issue raised AFAIK, but when `GLIDER_LINK` was added as an enum constant in `src/Device/Config.hpp` there should have been a corresponding entry added to the `port_type_strings` array in `src/Profile/DeviceConfig.cpp`. That broke my new code until I added that entry. Right now that's a moot problem since 7.0 is unusable anyway.

MaxKellermann:
> That broke my new code until I added that entry.

I don't see your PR for adding this missing entry.

> 7.0 is unusable anyway.

Which bug number is that?

src/Dialogs/Device/DeviceEditWidget.cpp
```diff
@@ -236,37 +236,51 @@ FillAndroidBluetoothPorts(DataFieldEnum &df, const DeviceConfig &config)
...
       ? env->GetStringUTFChars(name, NULL)
       : NULL;
 
-    AddPort(df, DeviceConfig::PortType::RFCOMM, address2, name2);
+    jstring devType = (jstring)env->GetObjectArrayElement(bonded, i * BLUETOOTH_LIST_STRIDE + 2);
+    const char *devType2 = devType != NULL
+      ? env->GetStringUTFChars(devType, NULL)
+      : NULL;
+
+    const DeviceConfig::PortType portType = strcmp("BLE",devType2) == 0 ? DeviceConfig::PortType::BLE_HM10 : DeviceConfig::PortType::RFCOMM;
```
MaxKellermann, line 268: As codacy correctly identified: this is a nullptr dereference crash bug.

MaxKellermann:
f64d9063434163407247f9e4964831a2458b694d

PR #453, opened 2020-07-09
Detect initial screen orientation

kedder:
On devices with fixed physical screens (such as Openvario) the orientation of initial screens, such as Fly/Simulator selection or profile selector might appear sideways (see the example screenshot below). 

This PR adds ability to autodetect the display orientation from the OS environment and implements detection from linux console orientation when XCSoar is run under DRM/KMS (Openvario case).

See the commit messages for more detailed description.

IMG_20200710_004529 resized

### DIFF
--- a/src/Hardware/DisplayGlue.cpp
+++ b/src/Hardware/DisplayGlue.cpp
@@ -43,6 +43,7 @@ Display::LoadOrientation(VerboseOperationEnvironment &env)
 
   DisplayOrientation orientation =
     CommonInterface::GetUISettings().display.orientation;
+
 #ifdef KOBO
   /* on the Kobo, the display orientation must be loaded explicitly
      (portrait), because the hardware default is landscape */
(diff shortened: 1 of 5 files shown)


MaxKellermann:
But what *is* the "initial display orientation"?
Your new function is not documented, so I had to guess what it does. And what it does is read the *current* display orientation.
I wonder why you call it "initial" and not "current".

kedder:
I came up with "initial" naming to distinguish this orientation from one, specified in the profile. That orientation would be used for "initial UIs" (shown before profile is selected), hence the name. 

Profile allows to set a completely different display orientation. From this point of view, the word "current" might be confusing too: orientation, specified in profile could be considered "current" as well, as soon as profile is selected, but it might be different from one returned by `Detect[Initial|Current]Orientation()`.

In any case, I don't feel strongly about that naming and I'm willing to change it to a better one. If we agree on the approach and terminology, I'll add documentation.

PR #458, opened 2020-07-23
Fix for ANDROID build - (OpenSSL, JAVA) on master-branch

Werner86:
When trying to do a fresh build of the master-branch for TARGET=ANDROID it currently fails because of two reasons:

**OpenSSL (libs.py) **
The file 
https://www.openssl.org/source/openssl-3.0.0-alpha4.tar.gz
has been moved to
https://www.openssl.org/source/ **old/3.0/** openssl-3.0.0-alpha4.tar.gz
Therefore the build breaks if this file has not been downloaded upfront / manually.

**JAVA (android.mk)**
The JAVA-build for android fails on newer JDKs because building JAVA 1.6 is not supported anymore. Therefore changed the build to 1.7

### DIFF
--- a/build/python/build/libs.py
+++ b/build/python/build/libs.py
@@ -54,8 +54,8 @@ libstdcxx_musl_headers = LibstdcxxMuslHeadersProject(
 )
 
 openssl = OpenSSLProject(
-    'https://www.openssl.org/source/openssl-3.0.0-alpha4.tar.gz',
-    'ftp://ftp.cert.dfn.de/pub/tools/net/openssl/source/openssl-3.0.0-alpha4.tar.gz',
+    'https://www.openssl.org/source/old/3.0/openssl-3.0.0-alpha4.tar.gz',
+    'ftp://ftp.cert.dfn.de/pub/tools/net/old/3.0/openssl/source/openssl-3.0.0-alpha4.tar.gz',
     'd930b650e0899f5baca8b80c50e7401620c129fef6c50198400999776a39bd37',
     'include/openssl/ossl_typ.h',
 )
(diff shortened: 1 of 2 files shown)


build/python/build/libs.py
```diff
@@ -54,8 +54,8 @@
 )
 
 openssl = OpenSSLProject(
-    'https://www.openssl.org/source/openssl-3.0.0-alpha4.tar.gz',
-    'ftp://ftp.cert.dfn.de/pub/tools/net/openssl/source/openssl-3.0.0-alpha4.tar.gz',
+    'https://www.openssl.org/source/old/3.0/openssl-3.0.0-alpha4.tar.gz',
+    'ftp://ftp.cert.dfn.de/pub/tools/net/openssl/source/old/3.0/openssl-3.0.0-alpha4.tar.gz',
```
MaxKellermann, line 58: I'd prefer updating to the latest version.

Werner86: i agree but the newest version is not working without adjustments so we should have a working built at first or?

Werner86: the commit-messages have been separated for the two different changes

MaxKellermann:
I just wanted to cherry-pick the Java 1.7 change, but it has a bad commit message. It doesn't explain why this change was done.

Werner86:
> I just wanted to cherry-pick the Java 1.7 change, but it has a bad commit message. It doesn't explain why this change was done.

you're right... was a bit too short - now it should be fine

MaxKellermann:
Java commit cherry-picked to v6.8.x: 4c369402dad57dd739c4974d3b60eb3e5aecce6a

Werner86:
> I'd prefer updating to the latest version.

There is currently an incompatability between the latest version of openssl and curl. In the latest alpha-version of openssl there has been removed the function SSL_get_peer_certificate (deprecated with version 3.0.0)
https://www.openssl.org/docs/manmaster/man3/SSL_get_peer_certificate.html
which is needed by curl (even in the latest version).
I recommend to either step back to a release-version of openssl or use the "old" version openssl-3.0.0-alpha4 which is in my commit.

MaxKellermann:
de746942d1e4d7a31c48a6a7eaa887d2694b2aca

PR #460, opened 2020-07-28
XCSoar new blue theme

Werner86:
This is a first try to get a modern theme for XCSoar replacing the "old-style" gray buttons with more modern layout and colors.

### DIFF
--- a/src/Dialogs/dlgQuickMenu.cpp
+++ b/src/Dialogs/dlgQuickMenu.cpp
@@ -84,11 +84,11 @@ QuickMenuButtonRenderer::DrawButton(Canvas &canvas, const PixelRect &rc,
     canvas.DrawFilledRectangle(rc, look.focused.background_color);
     canvas.SetTextColor(enabled
                         ? look.focused.text_color
-                        : look.button.disabled.color);
+                        : look.button.disabled.foreground_color);
   } else {
     if (HaveClipping())
       canvas.DrawFilledRectangle(rc, look.background_brush);
-    canvas.SetTextColor(enabled ? look.text_color : look.button.disabled.color);
+    canvas.SetTextColor(enabled ? look.text_color : look.button.disabled.foreground_color);
   }
 
   canvas.Select(*look.button.font);
--- a/src/Form/Form.cpp
+++ b/src/Form/Form.cpp
@@ -505,27 +505,12 @@ WndForm::OnPaint(Canvas &canvas)
 
     // JMW todo add here icons?
 
-#ifdef EYE_CANDY
-    if (!IsDithered() && is_active) {
-      canvas.SetBackgroundTransparent();
-      canvas.Stretch(title_rect.left, title_rect.top,
-                     title_rect.GetWidth(),
-                     title_rect.GetHeight(),
-                     look.caption.background_bitmap);
+    canvas.SetBackgroundColor(is_active
+                              ? look.caption.background_color
+                              : look.caption.inactive_background_color);
+    canvas.DrawOpaqueText(title_rect.left + Layout::GetTextPadding(),
+                          title_rect.top, title_rect, caption.c_str());
 
-      // Draw titlebar text
-      canvas.DrawText(title_rect.left + Layout::GetTextPadding(),
-                      title_rect.top, caption.c_str());
-    } else {
-#endif
-      canvas.SetBackgroundColor(is_active
-                                ? look.caption.background_color
-                                : look.caption.inactive_background_color);
-      canvas.DrawOpaqueText(title_rect.left + Layout::GetTextPadding(),
-                            title_rect.top, title_rect, caption.c_str());
-#ifdef EYE_CANDY
-    }
-#endif
   }
 
   if (dragging) {
--- a/src/Look/ButtonLook.cpp
+++ b/src/Look/ButtonLook.cpp
@@ -25,35 +25,29 @@ Copyright_License {
 #include "Colors.hpp"
 #include "Asset.hpp"
 
-void
-ButtonLook::Initialise(const Font &_font)
+#include "Look/Themes/BlueTheme.hpp"
+
+void ButtonLook::Initialise(const Font &_font)
 {
   font = &_font;
 
-  standard.foreground_color = COLOR_BLACK;
-  standard.foreground_brush.Create(standard.foreground_color);
-  standard.background_color = IsDithered() ? COLOR_WHITE : COLOR_LIGHT_GRAY;
-  if (IsDithered()) {
-    standard.CreateBorder(COLOR_BLACK, COLOR_BLACK);
-  } else if (!HasColors()) {
-    standard.CreateBorder(LightColor(COLOR_DARK_GRAY), COLOR_BLACK);
-  } else {
-    standard.CreateBorder(LightColor(standard.background_color),
-                          DarkColor(standard.background_color));
-  }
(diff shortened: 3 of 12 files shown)


kedder:
Some screenshots would be useful here

Werner86:
> Some screenshots would be useful here


yes, of course ... here you are:
image
image
image

(Sorry for the nav-bar on the bottom --- this is of my smartphone when taking screenshots)

kedder:
Looks very nice! :+1:

MaxKellermann:
A matter of personal taste - looks very ugly to me. Gaudy, but not modern. And non-standard, which may confuse people.

Werner86:
> A matter of personal taste - looks very ugly to me. Gaudy, but not modern. And non-standard, which may confuse people.

@MaxKellermann Please google pics for "button design" ... the world is gaudy 
Which ui-design do you like? Would be great to get some constructive feedback if you don't like it.
What do you mean with "non-standard"? -> All buttons are blue instead of gray --- that's the (new) standard - or do you think about the beveled borders of the gray buttons?

MaxKellermann:
What standard is this? I've never seen buttons like that.
Your buttons don't look like buttons at all. Nothing indicates that those areas can be pressed.
I can't give constructive feedback, because this is 90% opinion and 10% convention. And this violates all UI conventions I know.

kedder:
It is hard to argue on this topic because it is so subjective. That said, I think it can be claimed that xcsoar buttons and UI in general does not look native or standard on any platform xcsoar runs on these days. I don't see it as a problem that needs to be fixed though - there is no single standard anymore. Modern web and mobile apps are full of buttons of different styles, colors and often does not look like physical buttons at all. Still, users are able to interact with these apps.

In my personal opinion these blue buttons looks more fresh and modern, and I would prefer them over old grey ones. I realize that's only my opinion though.

I suggest a few things to consider:

1. Is the new style consistent throughout the app? 
2. Is it still ergonomic? E.g. indicate "pressed", "focused" state? Usable with finger, mouse, keyboard?
3. Does it look readable on black/white epaper screens (like Kobo)?
4. Does it look good on low-resolution screens (e.g. 96dpi)
5. What will be the answer to conservative users who would oppose any change in style/layout?

mlep:
For color displays, I see one (objective?) advantage to this design of the buttons: it is much easier to distinguish between active and deactivated buttons.

Werner86:
@MaxKellermann 
I'm trying to get your point about standards. Is the following picture a bit more like you think of? (I beveled the buttons to have more visibility about what's a button and what's not a button)
image

@kedder 
_Is the new style consistent throughout the app?_
All buttons look like this. Working on other style-elements as well, but not finished yet.

_Is it still ergonomic? E.g. indicate "pressed", "focused", "disabled" state? Usable with finger, mouse, keyboard?_
Yes

_Does it look readable on black/white epaper screens (like Kobo)?_
Not tested yet ... Can anybody support on this topic as i don't have a KOBO?

_Does it look good on low-resolution screens (e.g. 96dpi)_
In my opinion -> yes ;-)

_What will be the answer to conservative users who would oppose any change in style/layout?_
Good question... maybe the layout should be configurable ... so we could have multiple-color-designs (gray, blue, ... whatever)

@mlep 
_For color displays, I see one (objective?) advantage to this design of the buttons: it is much easier to distinguish between active and deactivated buttons._
yes, exactly

MaxKellermann:
> Modern web and mobile apps are full of buttons of different styles, colors and often does not look like physical buttons at all. Still, users are able to interact with these apps.

Many (most?) "modern" web and mobile apps are UI desasters.

XCSoar isn't exactly the most user-friendly application, and in fact it's still very complicated to use (OTOH it's not dumbed down to very few features like most apps - it's an extremely featureful and complex software). The worst thing about XCSoar is its inability to be used with keyboard only - you're lost quickly without a pointing device - bad bad bad! But this PR doesn't even try to fix that, this is only about the look.
But one thing XCSoar gets right: (most of the time) you know visually whether something can be clicked, can be interacted with.
One thing I changed long ago was the menu buttons - previously, they were green rectangles with no feedback, and now they look like any other button.

About "modern" - XCSoar's UI looks ugly and aged. Because it mostly imitates the 90ies GUI style (Windows 3/9x/NT/2k/7). You could argue whether XP's gaudy style was better or worse, but all of 3/9x/NT/2k/7 and XP were consistent. Something which cannot be said about "modern" "flat" designs (like Windows 10). Is that thing clickable? You don't know until you try.

And that's just about a mainstream operating system - if you look at web apps, the thing gets horrible very quickly. Where are the scroll bars? Every web site implements its own scroll bars, most of them unusable or only usable with the mouse, too thin for clicking. And I'm not yet talking about infinite scrolling...

So when we compare with other UI, let's compare with *good* and well-designed UIs, and not with other UI desasters. Just because most developers do bad UI design doesn't mean it should be regarded "normal" or "standard".

> Is the following picture a bit more like you think of?

It's less bad. There's now a darker rectangle on the right which looks like a shadow. You're approaching the old beveled look, but why not leave it that way?

mlep:
As a user, I agree with @MaxKellermann that well-designed GUI are very rare. I also recognize that XCSoar remains usable even if it has a lot of features (and this is already an achievement!).

I believe @Werner86 effort goes in the good direction: this PR could slightly improve the UI of XCSoar. And no, it does not redesigned the whole XCSoar GUI...

@MaxKellermann,  when you state _''let's compare with good and well-designed UIs,''_, please, could you give examples? I guess my question should be answered in a specific issue, so that not to mix things between the current PR (which mostly change the color) and the general subject of a good UI for XCSoar.

MaxKellermann:
> could you give examples?

Microsoft peaked Windows usability in Windows 2000, lost some of it in XP, and returned to good in Windows 7. 2000 and 7 look ugly (grey), but adding gaudy colors to XP hasn't made it any less ugly.
(For the last 10 years, they've been experimenting with flat designs, which was a big regression.)
https://docs.microsoft.com/en-us/windows/win32/uxguide/controls

Apple's OS X has always been very usable, and the buttons look very elegant (even though they're grey, only the focused button is highlighted in blue). You know exactly what is clickable.
https://developer.apple.com/design/human-interface-guidelines/macos/buttons/push-buttons/

However, that elegance is only visible on large desktop displays; on iOS, the user interface is flat, but clickable text is blue (like hyperlinks in browsers).
https://developer.apple.com/design/human-interface-guidelines/ios/controls/buttons/

We don't improve XCSoar by making interactive elements *less* visible. We can argue about colors and personal taste, but certainly every change must demonstrate that there is no regression in usability. This PR is a regression.

mlep:
Indeed, Apple's products have overall a good UI design (especially since their computers have a mouse with more than one button :wink:  ).

We were already discussing about the absence of regression (e.g. testing on Kobo).
But, at this point, I do not understand why you _already_ call this PR a regression.
Let's wait for the test results, no?

MaxKellermann:
It's an obvious regression already, even without possible Kobo-specific regressions, because the visual indication for "clickable button" has been changed from "beveled 3D look" (standard UI look which everybody recognizes) to "blue rectangle with dark blue line at the bottom", which I have never seen before in any other software. Just for the quick "oh look it's modern".

MaxKellermann:
> Indeed, Apple's products have overall a good UI design (especially since their computers have a mouse with more than one button wink ).

Note that this PR is only about UI *look*, not about UI *feel*. The number of mouse buttons is not about the look.

XCSoar's look is boring (but good enough) - XCSoar's feel is rather bad. We'd better work on the feel first!

PR #471, opened 2020-09-10
Select a list of airspace files

### LINKED ISSUE #425: add two more additional airspace fields
XCSoar version
--------------
8.8.14-Android May 14 2020

Please add two more additional entrie dialogues for airspace files in /XCSoarData 
Where the german language dialogue sais:
Standortdatei>Standortdatei>WeitereLufträume
change it to: 
Standortdatei>Standortdatei>WeitereLufträume1
Standortdatei>Standortdatei>WeitereLufträume2
Standortdatei>Standortdatei>WeitereLufträume3

  
Flying in the Alps alows us to cross several boarders in one flight. Currently we just concatanate the air space files to make them fit into the two boxes available. However, your interface for loading updates is so much nicer and it would require just two more enties for me to really use it.

CptFrikadel:
Brief summary of the changes
----------------------------

Change the airspace selection dialog so that the user can select a list of airspace files to use instead of the '`airspaces`' and '`more airspaces`' fields.

New `SiteConfigPanel` layout (with three airspace files selected):
2020-09-10-165004_890x480_scrot

When selected it shows the list of active airspace files and allows the user to add or remove files from the list:
2020-09-10-165010_890x480_scrot

When the user selects `add` the normal `FilePicker` opens and the user can pick a file to add to the active list or download a new one.

The selected files are saved in the profile as a comma separated list of paths under the `AirspaceFile` key. This sort of breaks backwards compatibility where a previous version will not know what to do with this new style of profile and simply not load the airspace file. The user then has to select their primary airspace file again. The other way around where an old profile is used with this new style will correctly read the set file, but will not load the additional airspace file. 

If this gets approval I will of course change the manual entries and NEWS.txt

Closes #425


src/Profile/PathValue.cpp
```diff
@@ -49,6 +49,28 @@ ProfileMap::GetPath(const char *key) const
...
+	std::vector<AllocatedPath> paths;
+	TCHAR buffer[MAX_PATH];
+
+	if(!Get(key, buffer, ARRAY_SIZE(buffer)))
+		return paths;
+	
+	if(StringIsEmpty(buffer))
+		return paths;
+
+	char *path = strtok(buffer, ",");
```
MaxKellermann, line 64: This will fail to compile on Windows.
And please, never use `strtok()`.
What if the filename has a comma?

src/Airspace/AirspaceGlue.cpp
```diff
@@ -88,13 +88,12 @@ ReadAirspace(Airspaces &airspaces,
   AirspaceParser parser(airspaces);
 
   // Read the airspace filenames from the registry
-  auto path = Profile::GetPath(ProfileKeys::AirspaceFile);
-  if (!path.IsNull())
-    airspace_ok |= ParseAirspaceFile(parser, path, operation);
+  std::vector<AllocatedPath> paths = Profile::GetMultiplePaths(ProfileKeys::AirspaceFile);
```
MaxKellermann, line 91: `auto`

src/Form/DataField/Base.hpp
```diff
@@ -46,6 +46,7 @@ class DataField
     ANGLE,
     ENUM,
     FILE,
+	MULTI_FILE,
```
MaxKellermann, line 49: Bad indent

src/Form/DataField/File.hpp
```diff
@@ -40,6 +40,8 @@ Copyright_License {
 class FileDataField final : public DataField {
   typedef StaticArray<StaticString<32>, 8> PatternList;
 
+  friend class MultiFileDataField;
```
MaxKellermann, line 43: Is this really necessary?

src/Dialogs/DataField.cpp
```diff
@@ -30,13 +30,16 @@ Copyright_License {
...
+#include "MultiFilePicker.hpp"
 
 bool
 EditDataFieldDialog(const TCHAR *caption, DataField &df,
                     const TCHAR *help_text)
 {
   if (df.GetType() == DataField::Type::FILE) {
     return FilePicker(caption, (FileDataField &)df, help_text);
+  } else if (df.GetType() == DataField::Type::MULTI_FILE) {
+	  return MultiFilePicker(caption,(MultiFileDataField&) df, help_text);
```
MaxKellermann, line 42: Bad indent

src/Dialogs/ListPicker.cpp
```diff
@@ -22,129 +22,11 @@ Copyright_License {
...
-#include "UIGlobals.hpp"
-#include "Language/Language.hpp"
-#include "Event/Timer.hpp"
-#include "Event/PeriodicTimer.hpp"
 
 #include <cassert>
 
 static constexpr int HELP = 100;
 
-class ListPickerWidget : public ListWidget, public ActionListener {
```
MaxKellermann, line 39: Why delete this?

src/Dialogs/MultiFilePicker.hpp
```diff
@@ -0,0 +1,31 @@
...
+#ifndef XCSOAR_DIALOGS_MULTI_FILE_PICKER_HPP
+#define XCSOAR_DIALOGS_MULTI_FILE_PICKER_HPP
+
+#include <tchar.h>
+
+class MultiFileDataField;
+
+bool
+MultiFilePicker (const TCHAR *caption, MultiFileDataField &df, 
+				 const TCHAR *help_text = nullptr);
```
MaxKellermann, line 29: Bad indent

src/Widget/ProfileRowFormWidget.cpp
```diff
@@ -55,6 +58,35 @@ RowFormWidget::AddFile(const TCHAR *label, const TCHAR *help,
   return edit;
 }
 
+WndProperty *
+RowFormWidget::AddMultipleFiles(const TCHAR *label, const TCHAR *help,
+                       const char *registry_key, const TCHAR *filters,
+                       FileType file_type,
+                       bool nullable)
```
MaxKellermann, line 65: Bad indent

src/Widget/ProfileRowFormWidget.cpp
```diff
@@ -141,3 +173,39 @@ RowFormWidget::SaveValueFileReader(unsigned i, const char *registry_key)
...
+		const auto contracted = ContractLocalPath(value);
+		if (contracted != nullptr)
+			value = contracted;
+
+		const WideToUTF8Converter value_to_add(value.c_str());
+		if (!value_to_add.IsValid())
+			continue;
+
+		strcat(new_output, value_to_add);
+		strcat(new_output, ",");
```
MaxKellermann, line 201: Stack buffer overflow bug here. Never use `strcat()`!

src/Profile/PathValue.cpp
CptFrikadel: There is an inline `StringToken()` in `StrinAPI.hxx`, but that just calls `strtok()` so I don't suppose I can use that? That also does not fix the issue when there is a comma in the filename.

Another alternative would be using `boost::tokenizer` with each file surrounded by quotes. Problem here is that `KeyValueFileReader::Read` stops reading after it finds a closing `"`. I don't really want to change that function simply to allow this niche functionality. 

Is simply disallowing commas in filenames a possibility?

src/Form/DataField/File.hpp
CptFrikadel: Yes, the `MultiFileDafield` needs access to `FileDataField`'s private `file` array and to protected member `EnsureLoaded()`.

src/Dialogs/ListPicker.cpp
CptFrikadel: This has been moved to `ListPicker.hpp` so that the `MultiFilePicker` can use it by including that file. If you want I can also move it to its own file.

src/Widget/ProfileRowFormWidget.cpp
CptFrikadel: I'll change that to `strncat()`

MaxKellermann: Do you know how `strncat()` works?
Do you know what it does when the buffer is too small?
(Spoiler: it omits the null terminator.)
Bottom line: don't use `strncat()` either. (Unless you really know what you're doing.)

src/Dialogs/ListPicker.cpp
MaxKellermann: It is impossible to review a commit which moves and modifies code at the same time.

src/Form/DataField/File.hpp
MaxKellermann: I'm not convinced that this is the best way to solve this. `friend` often means your API is badly designed, and explanations like "but I need to access this private and that protected" do not actually explain anything.

src/Profile/PathValue.cpp
MaxKellermann: No, please don't use that either. This is just a wrapper to reduce the pain of using it in legacy code.
You could use `IterableSplitString`, which needs just a string view, doesn't modify the string and is zero-copy.

MaxKellermann: You could disallow commas in filenames, but then your could should be robust, and have lots of additional checks. I don't see any of that in your code submission.
I wouldn't want to disallow commas. I don't think that's a good separator.

CptFrikadel:
Apologies for the long silence. I will try to address the comments within the next week or so. I will unmark it as draft once I believe it is ready for review.

CptFrikadel:
I've finally found some time to work on this again. Apologies for the long silence once more.

I have addressed most comments I believe. Because of the long time between opening this request and now, and because I think it makes reviewing a bit more straightforward I rebased the branch on top of the latest head. 

Most notable changes:
- I changed the path separator to a colon. This matches the 'nix environment variable style. 
- Changed `strtok()` to use `IterableSpliString()` 
- Use `std::string` instead of c-style strings, so no more `strcat()`. 

The manual and NEWS.txt stil require updating. I'll do that if this gets approval.

MaxKellermann:
Is the colon really a good separator? It's a standard character in absolute paths on Windows, e.g. `C:\Foo\Bar.txt`

MaxKellermann:
Your code breaks existing configurations with a secondary airspace file.

src/Profile/PathValue.cpp
```diff
@@ -49,6 +51,40 @@ ProfileMap::GetPath(const char *key) const
...
+		return paths;
+
+	if (StringIsEmpty(buffer))
+		return paths;
+
+	StringView string_view(buffer);
+
+	IterableSplitString path_splitstring(string_view, ':');
+
+	for (auto i = path_splitstring.begin(); i != path_splitstring.end(); ++i){
```
MaxKellermann, line 70: Too complicated!
```
for (auto i : IterableSplitString(buffer, ':'))
```
does the same and is simpler.

src/Profile/PathValue.cpp
```diff
@@ -49,6 +51,40 @@ ProfileMap::GetPath(const char *key) const
...
+
+		if ((*i).empty())
```
MaxKellermann, line 72: Even if an iterator were a good idea here: just do `i->empty()` instead of `(*i).empty()`. Don't make it more complicated that it needs to be!

src/Profile/PathValue.cpp
```diff
@@ -49,6 +51,40 @@ ProfileMap::GetPath(const char *key) const
...
+			continue;
+
+		AllocatedPath path(Path(std::string(*i).c_str()));
```
MaxKellermann, line 75: Doesn't compile on Windows. And too complicated.

src/Profile/PathValue.cpp
```diff
@@ -49,6 +51,40 @@ ProfileMap::GetPath(const char *key) const
...
+
+		if (!path.MatchesExtension("txt") && !path.MatchesExtension("air"))
```
MaxKellermann, line 77: Doesn't compile on Windows.

src/Form/DataField/Base.hpp
MaxKellermann: Indent is still bad

src/Form/DataField/File.hpp
MaxKellermann: This is still here

CptFrikadel: Yes it is indeed still here. The `MultiFileDataField` uses a `FileDataField` internally to keep track of the list of files and other housekeeping tasks. As in, the `MultiFileDataField` simply wraps around the `FileDataField` and keeps track of the selection of multiple files. To let the `FileDataField` keep track of the files, `MultiFileDataField` needs to be able to access the private member array `file`. For most of the `DataField`'s virtual methods it uses the `FileDataField`'s methods. To do that the `MultiFileDataField` needs to be able to call the protected `EnsureLoaded()` method.
A `MultiFileDataField`'s behavior is similar to, but it is not a `FileDataField`. So this way the `FileDataField`'s code is reused without inheriting from it. 
I'm not sure how else to solve this, other than rewriting the `MultFileDataField` as a standalone, with many methods identical to the `FileDataField`'s, which seems wrong to me.

CptFrikadel:
> Your code breaks existing configurations with a secondary airspace file.

Yes and no. A profile with a secondary airspace file will simply not load the secondary airspace file (i.e the user has to add it to the active files list). A new style profile used on an older build will just not load any airspace at all which should prompt the user to reconfigure their airspaces.  

I guess that maybe a different profile key for the list of airspace files is more desirable so that a profile can be used in old and new versions?

CptFrikadel:
>Is the colon really a good separator? It's a standard character in absolute paths on Windows, e.g. C:\Foo\Bar.txt

>Doesn't compile on Windows.

I completely forgot about windows. Colon indeed seems like a bad option as a separator. I think a multicharacter separator is the most robust option then. I.e. something like `:[]:` with colons at either end so that the string can still be easily split with an `IterableSplitString`.

src/Form/DataField/File.hpp
CptFrikadel: An alternative would be to use two `FileDataField`s in the `MultiFileDataField` one for active files and one for the inactive files. This has, I think, the only requirement that a `FileDataField::RemoveFile()` method be added.

CptFrikadel:
I believe all comments have been addressed: 

- The `MultiFileDataField` is no longer a friend of `FileDataField`.    
- The separator symbol is now a `'|'` pipe symbol. Windows explicitly disallows it in paths and filenames. *nix heavily discourages it. I think this is a reasonably compromise. 
- News.txt and the manual have also been updated

groundhogxc:
@CptFrikadel actually I think it could be nice if in "expert" mode, like before, there would be two fields (airspaces and more airspaces). Strictly not required, no, but quite often people use one set of "static" airspace files, and one set of "variable" ones; for example with some sets of airspaces that are sometimes active, sometimes not; or, people use this to display other types of information like thermal/weather data. So you would have a "air_force_restrictions_active.txt" file and a "..._not_active.txt" file. With two fields that would be nicer to manage because the overall configuration can be seen more easily. Also, consistent with the waypoint file entries entry int he config dialog.
But, like I said, not required for the functionality.

CptFrikadel:
@groundhogxc That is a valid point. I guess an additional "expert row" for volatile airspaces that change often could be beneficial. This does, in my opinion, add some bloat to the settings. For example, on my phone screen in portrait mode, the full name of the airspace file is not really readable from the menu anyway (longer names simply don't fit). So you would have to go in to the file picker to see it anyway. 

I'd like some more opinions on whether an extra row for "volatile" airspaces is desirable?

MaxKellermann:
Talking about UI - that configuration page gets more complicated by requiring one more step (one more modal dialog indirection) to do stuff. That's not good, that's tedious for users.
Maybe we can replace the whole page with a list, which contains all data files that are currently enabled? And have controls to enable (add) more (existing or newly downloaded) files, and a button to disable one file.

CptFrikadel:
Do you mean like a single list for all datafile categories at once? Wouldn't that make things really confusing? Having a long list of all active files would make it really hard to see which airspace files are in use for example. Especially if one uses files downloaded externally which might not have "airspace" or "waypoints" in the filename. I think at least some form of categorization would be better. 

How about a page that lists all files of a category with checkmark controls to enable/disable them and a control to add/download new files?

MaxKellermann:
> Do you mean like a single list for all datafile categories at once? Wouldn't that make things really confusing? Having a long list of all active files would make it really hard to see which airspace files are in use for example.

No I said "a list, which contains all data files that **are currently enabled**".

hperrot (commented):
The commit message still talks about a "colon
separated list"

MaxKellermann:
This PR looks like it has been abandoned by the author two years ago.

PR #474, opened 2020-09-11
Spelling: UART, I²C, etc.

comradekingu:


kerel-fs:
The english text in translatable strings is used as key for the translated texts in the [po files](https://github.com/XCSoar/XCSoar/tree/master/po). Changes in spelling of this key requires a patch in all translation files as well, otherwise the respective translations will be lost.

comradekingu:
@kerel-fs I am out and about and can't add that. Feel free to do so. Do the string-changes look OK?

kerel-fs:
The string changes look OK for me.
For some changes I can't find a reason for or against the new string, but I'm not a native speaker.

PR #478, opened 2020-09-17
Tracking/Skylines: Update Server IP

lordfolken:
We need to update the tracking ip address of the server for skylines. Its now 212.51.149.212


lordfolken:
@MaxKellermann could you please push a new release on play store?

Turbo87 (approved):

Issue #475, opened 2020-09-11
Unknown display dimensions in mm on Pi4


bomilkar:
XCSoar versions having and not having the problem
-------------------------------------------------
7.0 p15 on Raspberry Pi4 running raspbian OS 10.4.
It has a LCD display connected to Pi4’s DPI port and uses DRI/DRM VC4 V3D driver:
https://dri.freedesktop.org/docs/drm/gpu/vc4.html

Expected behavior
-----------------

XCSoar should not run into an Arithmetic Exception when the physical dimensions of the display in mm have not been set. Instead it should calculate the physical dimensions from other known parameters such as dots per inch and pixel width/hight

Actual behavior
---------------

XCSoar exits with a "Floating point exception" when the physical dimensions of the display in mm have not been set.

Steps to reproduce the behavior
-------------------------------

run XCSoar on Pi4 with "dtoverlay=vc4-fkms-v3d" in /boot/config.txt

Do you have any idea what may have caused this?
-----------------------------------------------

Here is a back trace from gdb:
(gdb) bt
#0  0xb6a948ec in raise (sig=8) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0xb68841f8 in __aeabi_ldiv0 () at /lib/arm-linux-gnueabihf/libgcc_s.so.1
#2  0x00135648 in MMToDPI(unsigned int, unsigned int) (pixels=1024, mm=0) at src/Hardware/DisplayDPI.cpp:73
#3  0x00135724 in Display::ProvideSizeMM(unsigned int, unsigned int, unsigned int, unsigned int)
    (width_pixels=1024, height_pixels=600, width_mm=0, height_mm=0) at src/Hardware/DisplayDPI.cpp:135
#4  0x0031db1c in TopCanvas::Create(PixelSize, bool, bool)
    (this=0x685880, new_size=..., full_screen=false, resizable=false) at src/Screen/EGL/TopCanvas.cpp:213
#5  0x0031bf78 in TopWindow::Create(char const*, PixelSize, TopWindowStyle) (this=
    0x63b000, text=0x4944ac "XCSoar", size=..., style=...) at src/Screen/Custom/TopWindow.cpp:59
#6  0x0015a824 in SingleWindow::Create(char const*, PixelSize, TopWindowStyle)
    (this=0x63b000, text=0x4944ac "XCSoar", size=..., style=...) at src/Screen/SingleWindow.hpp:60
[...]

PR #479, opened 2020-09-18
gracefully handle unknown size of the active area of the display

bomilkar:
Brief summary of the changes
----------------------------
If the active area is unknown usually the width and height in mm == 0.
With HAVE_DPI_DETECTION MMToDPI() runs into a divide by 0.

Instead use the DPI values known at this time. A good choice is what the user provides at the commend line.

https://github.com/XCSoar/XCSoar/issues/475


lordfolken:
Looks acceptable to me. Lets see what @MaxKellermann says.

MaxKellermann:
This is wrong. The function should only be called if the values are known and valid.

MaxKellermann:
c8a4e3f205ced25147e0585ad233e8597d07b474

Issue #123, opened 2018-07-02
Black terrain color ramp


georgebohnisch:
Using a device with an OLED panel, having a black terrain color ramp would greatly reduce battery consumption.

georgebohnisch:
To clarify on this, the base color would be black, so the majority of the terrain would be black (OLED displays do not expend power on pixels that are black/off) and the terrain would be lighter shades of gray/white.

MikeBravo15:
I can do that on my Nokia8 via colors inverse function.  Visibility  is OK but not as good as white as screen reflexion is pronounced. Issue is that sun heats the screen significantly so beware od that

lordfolken:
This would be fairly easy to do in:
https://github.com/XCSoar/XCSoar/blob/82bc641f038c8620ab534a8f203b8549bf26ff54/src/Terrain/TerrainRenderer.cpp

please contribute.

fberst:
i have implemented the black color ramp: https://github.com/fberst/xcsoar/commit/b6e8278360726a2bf7e5b13ea14bb18639438b74

lordfolken:
@fberst please create a pull request

lordfolken:
Closing as stale.

PR #488, opened 2020-10-18
TerrainRenderer: new ramp "Black"  #123

fberst:
Brief summary of the changes
----------------------------


MaxKellermann:
What's the point of that? This wastes CPU for rendering terrain, drains battery, only to make every pixel black.

fberst:
it's for OLED Displays, on these black pixels are off which should reduce the power consumption of the display

https://github.com/XCSoar/XCSoar/issues/123

I can see the point with the waste of CPU performance, but didn't take that into account with the ridiculous overpowered smartphones we have today, sorry....

MaxKellermann:
> the base color would be black [...] and the terrain would be lighter shades of gray/white

That's not what this PR does - it's all black without any gray/white. It effectively makes the terrain renderer result invisible.

fberst:
here a screenshoot:
Bildschirmfoto vom 2020-10-18 19-10-03

It's dark, but far away from all black and turning invisible, i admit that i don't understand why.
Also why do we already have a "white" ramp? Following the logic this should turn everything white, effectively disabling the rendering of the terrain, that would be the same as turning the terrain off in the settings, which would be also complete pointless.

MaxKellermann:
This is with hill shading. The sunny side gets somewhat gray (auto-calculated "sunny" color, a bit lighter than black), but the other side stays black, because it doesn't get blacker than black.
Is this shade of gray really visible on real screens during a flight, when the sun shines on the display?
... and just look at the black labels on black background - XCSoar just isn't ready for this color scheme.

fberst:
ok, i think we should close this pull request for now,,,

PR #490, opened 2020-10-20
PageActions: manage auto zoom state per page

lordfolken:
Brief summary of the changes
----------------------------------------
This is an extension to the distinct page zoom. With this you can have multiple map windows and enable auto zoom per window.

### DIFF
--- a/src/PageState.hpp
+++ b/src/PageState.hpp
@@ -48,9 +48,16 @@ struct PageState {
    */
   double circling_scale;
 
+  /**
+   * The last zoom mode on this page. This attribute is only used if
+   * PageSettings::distinct_zoom is enabled.
+   */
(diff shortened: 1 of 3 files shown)


src/PageState.hpp
```diff
@@ -48,6 +48,12 @@ struct PageState {
    */
   double circling_scale;
 
+  /**
+   * The last zoom mode on this page. This attribute is only used if
+   * PageSettings::distinct_zoom is enabled.
+   */
+  bool auto_zoom_enabled;
```
MaxKellermann, line 55: Who initializes this?

lordfolken: Its dynamically initialized by map_settings upon LeavePage(). I also added it to Clear() so its reset to a default value.

PR #491, opened 2020-10-28
Detect initial screen orientation for 6.8.x

August2111:
This Pull Request should insert the same code for detection of orientation (f.e. on OpenVario) like pr #453 from kedder on 23 Jul in branch v6.8.x, because the stable branch has the same problem!

Brief summary of the changes
----------------------------


MaxKellermann:
I merged this PR, but I don't like that it contains a fixup commit. This will break bisects. Next time, don't submit known-broken commits with fixups, and instead amend the known-broken commits.

PR #420, opened 2020-05-24
Scalable InfoBox title font

bomilkar:
Introduce a user parameter "InfoBox title size" 80% ... 150% to scale the font for the title and comment in the InfoBoxes.
The intention is to improve readability of InfoBox title and comment in particular on tall and narrow displays. It allows the user to trade off readability vs. length of text being displayed.

Added documentation in a separate commit.

PR #493, opened 2020-11-09
Larger InfoBox comments for smaller displays

mattreengage:
Brief summary of the changes
----------------------------

InfoBox comments on smaller displays are unreadable. This change enables larger text to be used if desired.
Some comment text may be truncated as a result.

This can be configured via Config/System/Look/Screen Layout

None

### DIFF
--- a/src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp
+++ b/src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp
@@ -51,6 +51,7 @@ enum ControlIndex {
   AppInverseInfoBox,
   AppInfoBoxColors,
   AppInfoBoxBorder,
+  AppInfoBoxLargeComment,
 #ifdef KOBO
   ShowMenuButton,
 #endif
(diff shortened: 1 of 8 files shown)


MaxKellermann:
I don't think this is a good change. This uses an arbitrary offset for exactly one text size, but font sizes should be chosen according to physical sizes to make them readable well enough. XCSoar already uses physical sizes for most fonts. If those sizes are too smal, there's a global "font size" setting to make everything larger (some people have better or worse eyes, or have the display very near or distant from their eyes).
If we merge this, everybody and his dog and cat will come and demand the same setting for each and every text in XCSoar. This doesn't scale, and makes the configuration too complicated.

mattreengage:
The global font size settings has no effect on the InfoBox text lines. The InfoBox text is auto scaled based on the size of the box.

On a smaller display, the Title line is almost unreadable in flight but can be read when configuring. This isn't a problem. The comment line is used as a 2nd info line in many InfoBoxes. It's for this situation that a larger font is desirable.

Possibly a better solution would be to have two types of InfoBox, one with the normal comment line and one with 2 lines of information, and then to update all InfoBoxes that use the comment line for the 2nd line of information

bomilkar:
@mattreengage this is a duplicate of PR #420 .
You might find it helpful.

MaxKellermann:
Seems to me d3ea258340d49b9cdf3be2bce34229321401cd61 is a better solution for this

PR #497, opened 2020-11-26
Preserve zoom level in TrafficWidget

### LINKED ISSUE #455: Flarm page defaults to 2 km range when switching pages
Which XCSoar version are you using when seeing the problem? latest stable and experimental versions
Which XCSoar version is the last one you know that did not show the problem? no version

Expected behavior
-----------------
 I would expect the Flarm page to stay on the latest zoom factor when I return to the flarm page after switching to another page. 

Actual behavior
---------------
When I set flarm zoom pag at 10 km and switch afterwards to another page, and then subsequently return to the Flarm page the zoom factor is back at 2 km default value.

a. setting Flarm page zoom at 10 km
b. switching page from Flarm to another page i.e. cruising,
c. returning to Flarm page, zoom fgactor is now suddenly 2 km default again

no

no

zinuzoid:
Brief summary of the changes
----------------------------
Preserver zoom level in FLARM radar when switching between pages
flarm

Closes #455

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -36,6 +36,8 @@ Version 7.0 - not yet released
   - enable SkyLines traffic display on Windows
   - add option to show SkyLines traffic names on the map
   - show thermals obtained from the XCSoar Cloud server
+* FLARM radar
+  - preserve zoom level when switch between pages
 * analysis
   - enhanced graphics: minor tics, color scheme, layout
   - key labels drawn on lines in several pages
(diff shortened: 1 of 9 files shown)


lordfolken (commented):
Does this survive the warning_mode logic? In warning mode XCsoar forces the zoom to 2km.

zinuzoid:
> Does this survive the warning_mode logic? In warning mode XCsoar forces the zoom to 2km.

Not sure, any suggestion how to reproduce that?

lordfolken:
So finally. in #925 there is a file for replaying flarm traffic. That way you can test. 
However I'm pretty sure that the alarm mode sets the zoom to 2 km.

lordfolken:
make output/UNIX/bin/FeedMEA
FeedNMEA tcp 4345 < pflaf01.txt

You find the file in the tests folder.

PR #499, opened 2020-11-28
Skip latitude,longitude=0.0,0.0 from XCTracer

zinuzoid:
Brief summary of the changes
----------------------------
While on the ground the XCTracer send lat,lon = 0.0,0.0
This is annoying as your location would jumping to the middle of the ocean (where 0.0/0.0 is).
It is hard to check your waypoint like this.
So we should just skip the 0.0 0.0 location from XCTracer

Here's after the fix, no GPS fix were accepted
Screenshot_20201127-184733_XCSoar_292x600|Screenshot_20201128-004137_XCSoar_292x600

### DIFF
--- a/src/Device/Driver/XCTracer/Parser.cpp
+++ b/src/Device/Driver/XCTracer/Parser.cpp
@@ -90,9 +90,22 @@ static int
 ReadCheckedLatitudeLongitude(NMEAInputLine &line, double &value_latitude, double &value_longitude)
 {
   int valid_fields = 0;
+  double latitude, longitude;
+  double epsilon = std::numeric_limits<double>::epsilon();
 
-  valid_fields += ReadCheckedRange(line, value_latitude, -90.0, 90.0);
-  valid_fields += ReadCheckedRange(line, value_longitude, -180.0, 180.0);
+  valid_fields += ReadCheckedRange(line, latitude, -90.0, 90.0);
+  valid_fields += ReadCheckedRange(line, longitude, -180.0, 180.0);
+
+  if(valid_fields == 2 && !(fabs(latitude) < epsilon && fabs(longitude) < epsilon)) {
+    value_latitude = latitude;
+    value_longitude = longitude;
+  } else {
+    /**
+     * While inactive on the ground XCTracer send the 0.0/0.0 location.
+     * We mark as invalid here, so map is not jumping to the 0.0/0.0
+     */
+    valid_fields = 0;
+  }
 
   return valid_fields;
 }
--- a/src/Device/Driver/XCTracer/Parser.cpp
+++ b/src/Device/Driver/XCTracer/Parser.cpp
@@ -86,6 +86,17 @@ ReadCheckedRange(NMEAInputLine &line,double &value_r, double min, double max)
   return true;
 }
 
+static int
+ReadCheckedLatitudeLongitude(NMEAInputLine &line, double &value_latitude, double &value_longitude)
+{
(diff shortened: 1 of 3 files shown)


lordfolken:
Is this situation on the ground, or without a valid fix?

A decent protocol reports whether the fix is valid.
If it at was NULL then it would work out of the box. 

While this probably solves the issue, i think we are fixing it on the wrong end.

@ThomasXBMC What do you think?

MaxKellermann:
The XCSoar source code has a link to the XCTracer protocol documentation, but there's no protocol documentation on that page: https://www.xctracer.com/en/tech-specs/?oid=1861
The XCSoar code comments suggest that the protocol was badly designed, and there's no way to indicate "we have no GPS fix".
Note that "0N 0E" is a valid position on earth's surface. Although unlikely, but an aircraft can be fly there.

So hardware vendors, seriously, please stop this shit, don't make "0N / 0E" your magic value for "I don't know"!

I don't like this PR because:
- it moves code around *and* changes it in one commit, which makes it hard to review
- it uses `==` for comparing floating point, which is broken

zinuzoid:
Addressed code comments

src/Device/Driver/XCTracer/Parser.cpp
```diff
@@ -90,9 +90,22 @@ static int
...
   int valid_fields = 0;
+  double latitude, longitude;
+  double epsilon = std::numeric_limits<double>::epsilon();
+
+  valid_fields += ReadCheckedRange(line, latitude, -90.0, 90.0);
+  valid_fields += ReadCheckedRange(line, longitude, -180.0, 180.0);
 
-  valid_fields += ReadCheckedRange(line, value_latitude, -90.0, 90.0);
-  valid_fields += ReadCheckedRange(line, value_longitude, -180.0, 180.0);
+  if(valid_fields == 2 && !(fabs(latitude) < epsilon && fabs(longitude) < epsilon)) {
```
zinuzoid, line 99: Not sure if this is what you expected but I wouldn't give any tolerant for this magic number.
Otherwise please advice.

ThomasXBMC:
> A decent protocol reports whether the fix is valid.
> Have you tried talking to the manufacturer?
> @ThomasXBMC What do you think?

Hi, I agree with your point of view -- the protocol should report whether the fix is valid, and this should be changed.
I've contacted the vendor, the proposal is to fix the protocol by using one of the three reserved fields in the XCTRC string to indicate "fix valid/invalid".
Also, the protocol spec would have to be updated and re-published with a valid link (the spec was available online  and the link in the code was OK when I submitted the original code a few years ago).
I got a quick initial response, they are looking into the topic, and I expect a final answer soon. Will follow-up as soon as I know more. 

NOTE: I wrote the driver because I use the device with xcsoar; I am not associated in any way with the vendor.

zinuzoid:
> the proposal is to fix the protocol by using one of the three reserved fields
@ThomasXBMC if you can convince the manufacturer then, I think the proper solution would be to send either NULL, or empty string in those fields rather than wasting another reserve field for this.

zinuzoid:
> the proposal is to fix the protocol by using one of the three reserved fields

@ThomasXBMC if you can convince the manufacturer then, I think the proper solution would be to send either NULL, or empty string in those fields rather than wasting another reserve field for this.

lordfolken:
Can we close this PR then?

zinuzoid:
Not much progress from the conversation with manufacturer, but I agree this should be fixed on the hardware side.

PR #500, opened 2020-12-02
Implement FANET '#FNF' service (type 4) sentences decoding

denibakovic:
Brief summary of the changes
----------------------------

This commit implements FANET driver for '#FNF' sentences of type 4(service) that contain current weather info and draws related wind arrow and info box with wind speed and wind gust to the screen.

Screenshot_20201203_000354_org xcsoar testing

### DIFF
--- a/src/Device/Driver/FANET/Device.hpp
+++ b/src/Device/Driver/FANET/Device.hpp
@@ -1,0 +1,142 @@
+/*
+Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2016 The XCSoar Project
+  A detailed list of copyright holders can be found in the file "AUTHORS".
+
+  This program is free software; you can redistribute it and/or
+  modify it under the terms of the GNU General Public License
+  as published by the Free Software Foundation; either version 2
+  of the License, or (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+
+#ifndef XCSOAR_FANET_DEVICE_HPP
+#define XCSOAR_FANET_DEVICE_HPP
+
+#include "Device/Driver.hpp"
+#include "FANET/List.hpp"
+#include "FANET/FanetStation.hpp"
+
+struct NMEAInfo;
+
+class FanetDevice final : public AbstractDevice {
+private:
+  enum FanetPacketType {
+    FANET_TYPE_ACK,
+    FANET_TYPE_TRACKING,
+    FANET_TYPE_NAME,
+    FANET_TYPE_MESSAGE,
+    FANET_TYPE_SERVICE,
+    FANET_TYPE_LANDMARKS,
+    FANET_TYPE_REMOTE_CONFIG,
+    FANET_TYPE_GROUND_TRACKING,
+    FANET_TYPE_HW_INFO,
+    FANET_TYPE_THERMAL
+  };
+
+  enum FanetServiceHeaderBits {
+    FANET_SERVICE_BIT_EHEADER,
+    FANET_SERVICE_BIT_SOC,
+    FANET_SERVICE_BIT_REMOTECFGSUPPORT,
+    FANET_SERVICE_BIT_PRESSURE,
+    FANET_SERVICE_BIT_HUMIDITY,
+    FANET_SERVICE_BIT_WIND,
+    FANET_SERVICE_BIT_TEMP,
+    FANET_SERVICE_BIT_INET
+  };
+
+  /**
+   * Checks if nth bit is set
+   *
+   * @return true if its set, false otherwise
+   */
+  static bool IsNthBitSet(uint8_t byte, int bit) { return byte & (1 << bit); };
+
+  /**
+   * Decodes FANET service(type 4) payload and updates provided station with
+   * decoded info
+   *
+   * @param payload to deode
+   * @param payload_length
+   * @param station to update
+   * @return true if successfully decoded, false otherwise
+   */
+  static bool DecodeServicePayload(uint8_t *payload, int16_t payload_length,
+                                   FanetStation *station);
+
+
+  /**
+   * Decodes first 6 bytes of provided payload to GeoPoint,
+   * does not check for length!
+   *
+   * @param buf payload(>= 6byte) to decode
+   * @return GeoPoint with decoded location if successfull,
+   *    GeoPoint::Invalid() otherwise
+   */
+  static GeoPoint Payload2GeoPoint(const uint8_t *buf);
+
+  /**
+   * Converts last bits 0-6 of byte to float. if bit 7 is set multiplies it
+   * with scale
+   *
+   * @param byte payload to convert
+   * @param scale scale to apply if bit 7 is 1
+   * @return value*scale if bit 7 set, value otherwise
+   */
+  static float Payload2Float(uint8_t byte, float scale);
+
+  /**
+   * Parses #FNF line
+   *
+   * @param stations_list list of stations to add parsed station to
+   * @param clock current time
+   * @return true if packet is parsed fully
+   */
+  static bool ParseRxPacket(char *cmd_str, StationsList &stations_list,
+                            double clock);
+
+  /**
+   * Updates existing station or allocates a new one in stations_list
+   *
+   * @param stations_list list to add station to
+   * @param clock current time
+   * @param station station to add
+   * @return true if station is successfully added
+   */
+  static bool AddStationToList(StationsList &stations_list, double clock,
+                               FanetStation station);
+
+  /**
+   * Replace first occurrence of newline character with \0 and moves pointer
+   * to the first non space character
+   *
+   * @param cmd_str line to replace newline character in
+   * @return stripped cmd_str
+   */
+  static char *RemoveExtraChars(char *cmd_str);
+
+  /**
+   * Check if cmd_str contains only hexadecimal chars & commas and is > 0 chars
+   *
+   * @param cmd_str line to check
+   * @return true if check is passed, false otherwise
+   */
+  static bool IntegrityCheck(char *cmd_str);
+
+public:
+  // Virtual method from class Device
+  bool ParseNMEA(const char *line, NMEAInfo &info) override;
+};
+
+#endif //XCSOAR_FANET_DEVICE_HPP
--- a/src/Device/Driver/FANET/Parser.cpp
+++ b/src/Device/Driver/FANET/Parser.cpp
@@ -1,0 +1,304 @@
+/*
+Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2016 The XCSoar Project
+  A detailed list of copyright holders can be found in the file "AUTHORS".
+
+  This program is free software; you can redistribute it and/or
+  modify it under the terms of the GNU General Public License
+  as published by the Free Software Foundation; either version 2
+  of the License, or (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+
+#include "Device.hpp"
+#include "NMEA/Info.hpp"
+#include "FANET/FanetAddress.hpp"
+
+bool
+FanetDevice::AddStationToList(StationsList &stations_list, double clock,
+                              const FanetStation station) {
+  FanetStation *fanet_slot = stations_list.FindStations(station.address);
+
+  if (fanet_slot == nullptr) {
+    fanet_slot = stations_list.AllocateStation();
+    if (fanet_slot == nullptr)
+      // no more slots available
+      return false;
+
+    fanet_slot->Clear();
+    fanet_slot->address = station.address;
+    fanet_slot->location = station.location;
+    fanet_slot->location_available = station.location_available;
+  }
+
+  fanet_slot->valid.Update(clock);
+  fanet_slot->Update(station);
+
+  return true;
+}
+
+bool
+FanetDevice::IntegrityCheck(char *cmd_str) {
+  if (strlen(cmd_str) == 0)
+    return false;
+
+  for(char *ptr = cmd_str; *ptr != '\0'; ptr++)
+  {
+    if((*ptr >= '0' && *ptr <= '9') ||
+       (*ptr >= 'A' && *ptr <= 'F') ||
+       (*ptr >= 'a' && *ptr <= 'f') ||
+       (*ptr == ',')) {
+      continue;
+    }
+    return false;
+  }
+  return true;
+}
+
+char *
+FanetDevice::RemoveExtraChars(char *cmd_str) {
+  char *ptr = strchr(cmd_str, '\r');
+  if(ptr == nullptr)
+    ptr = strchr(cmd_str, '\n');
+  if(ptr != nullptr)
+    *ptr = '\0';
+  while(*cmd_str == ' ')
+    cmd_str++;
+  return cmd_str;
+}
+
+GeoPoint
+FanetDevice::Payload2GeoPoint(const uint8_t *buf) {
+  if(buf == nullptr)
+    return GeoPoint::Invalid();
+
+  // Integer values
+  unsigned int lati = buf[2] << 16 | buf[1] << 8 | buf[0];
+  if(lati & 0x00800000)
+    lati |= 0xFF000000;
+  unsigned int loni = buf[5] << 16 | buf[4] << 8 | buf[3];
+  if(loni & 0x00800000)
+    loni |= 0xFF000000;
+
+  return GeoPoint(Angle::Degrees((float)loni / 46603.0f), Angle::Degrees((float)lati / 93206.0f));
+}
+
+float
+FanetDevice::Payload2Float(uint8_t byte, float scale) {
+  auto value = (float)(byte & 0x7F);
+  if(byte & 1 << 7)
+    return (value * scale);
+  else
+    return value;
+}
+
+bool
+FanetDevice::DecodeServicePayload(uint8_t *payload, int16_t payload_length, FanetStation *station) {
+  if(payload == nullptr || payload_length == 0)
+    return false;
+
+  uint16_t payload_pos = 1;
+
+  // Skip extended header
+  if(IsNthBitSet(payload[0], FANET_SERVICE_BIT_EHEADER))
+  {
+    // TODO: decode extended header byte if set, currently not needed/used
+    payload_pos++;
+  }
+
+  // Return if no location
+  if(payload_pos + 6 > payload_length)
+    return false;
+
+  station->location = Payload2GeoPoint(&payload[payload_pos]);
+  if( ! station->location.IsValid())
+    return false;
+  station->location_available = true;
+  payload_pos += 6;
+
+  /*
+   * Temperature (+1byte in 0.5 degree, 2-Complement)
+   */
+  if(IsNthBitSet(payload[0], FANET_SERVICE_BIT_TEMP))
+  {
+    if(payload_pos + 1 > payload_length)
+      return false;
+    station->temperature = (float)payload[payload_pos++] / 2.0f;
+  }
+
+  /*
+   * 	Wind (+3byte: 1byte Heading in 360/256 degree, 1byte speed and 1byte
+	 *  gusts in 0.2km/h (each: bit 7 scale 5x or 1x, bit 0-6))
+   */
+  if(IsNthBitSet(payload[0], FANET_SERVICE_BIT_WIND))
+  {
+    if(payload_pos + 3 > payload_length)
+      return false;
+
+    station->wind_dir_deg = (((float)payload[payload_pos++]) / 0.7083333334f);
+    station->wind_speed_kmph =
+        Payload2Float(payload[payload_pos++], 5.0f) * 0.2f;
+    station->wind_gust_kmph = Payload2Float(payload[payload_pos++], 5.0f) * 0.2f;
+  }
+
+  /*
+   * Humidity (+1byte: in 0.4% (%rh*10/4))
+   */
+  if(IsNthBitSet(payload[0], FANET_SERVICE_BIT_HUMIDITY))
+  {
+    if(payload_pos + 1 > payload_length)
+      return false;
+    station->rel_humidity = payload[payload_pos++] / 2.5f;
+  }
+
+  /*
+   * Barometric pressure normailized (+2byte: in 10Pa, offset by 430hPa,
+   * unsigned little endian (hPa-430)*10)
+   */
+  if(IsNthBitSet(payload[0], FANET_SERVICE_BIT_PRESSURE))
+  {
+    if(payload_pos + 2 > payload_length)
+      return false;
+    int lsb = payload[payload_pos++];
+    int msb = payload[payload_pos++];
+    station->preasure_hpa = (((msb << 8) | lsb) / 10.0) + 430.0;
+  }
+
+  /*
+   * State of Charge  +1byte
+   * (lower 4 bits: 0x00 = 0%, 0x01 = 6.666%, .. 0x0F = 100%)
+   */
+  if(IsNthBitSet(payload[0], FANET_SERVICE_BIT_SOC))
+  {
+    if(payload_pos + 1 > payload_length)
+      return false;
+    station->soc_percent =
+        Payload2Float(payload[payload_pos++], 1.0f) * 100.0f / 15.0f;
+  }
+
+  return true;
+}
+
+/*
+ *  Service(type 4 - weather info):
+ *  #FNF src_manufacturer,src_id,broadcast,signature,type,payloadlength,payload
+ */
+bool
+FanetDevice::ParseRxPacket(char *cmd_str, StationsList &stations_list,
+                           double clock) {
+  // Prepare the line
+  cmd_str = RemoveExtraChars(cmd_str);
+  if ( ! IntegrityCheck(cmd_str))
+    return false;
+
+  char *start_ptr = (char *)cmd_str;
+  char *end_ptr;
+
+  if(! start_ptr)
+    return false;
+
+  // Manufacturer(1 byte)
+  uint8_t manufacturer = strtol(start_ptr, &end_ptr, 16);
+  if(start_ptr == end_ptr)
+    return false;
+
+  start_ptr = strchr(start_ptr, ',') + 1;  // comma after src_manufacturer
+  if (start_ptr == nullptr)
+    return false;
+
+  // Id(2 bytes)
+  uint16_t id = strtol(start_ptr, &end_ptr, 16);
+  if(start_ptr == end_ptr)
+    return false;
+
+  /*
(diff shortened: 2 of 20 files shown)


src/Device/Driver/FANET/Device.hpp
```diff
@@ -0,0 +1,47 @@
...
+  // Virtual method from class Device
+  bool ParseNMEA(const char *line, NMEAInfo &info) override;
+
+  static bool ParseRxPacket(char *cmd_str, StationsList &stations_list, double clock);
+
+  static bool AddStationToList(StationsList &stations_list, double clock,
+                               const FanetStation &station);
+
+  static char *RemoveExtraChars(char *cmd_str);
+  static bool IntegrityCheck(char *cmd_str);
```
MaxKellermann, line 44: What do these methods do, and why are they public?

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  fanet_slot->valid.Update(clock);
+  fanet_slot->Update(station);
+
+  return true;
+}
+
+bool
+FanetDevice::IntegrityCheck(char *cmd_str) {
```
MaxKellermann, line 57: This method returns `true` given an empty string. Is that desired?

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,171 @@
...
+  return true;
+}
+
+char *
+FanetDevice::RemoveExtraChars(char *cmd_str) {
+  // Remove \r\n and any spaces
+  char *ptr = strchr(cmd_str, '\r');
+  if(ptr == nullptr)
+    ptr = strchr(cmd_str, '\n');
+  if(ptr != nullptr)
```
MaxKellermann, line 77: This will preserve all `\n` if at least one `\r` exists. Is that what you want?

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,171 @@
...
+  station.address = FanetAddress(manufacturer, id);
+
+  /*
+   * Skip broadcast(1 byte) and signature(2 bytes)
+   *  comma after src_id
+   *  comma after broadcast
+   *  comma after signature
+   */
+  for (int i = 0; i < 3; ++i)
+    start_ptr = strchr(start_ptr, SEPARATOR) + 1;
```
MaxKellermann, line 117: Null pointer dererefence bug (crash bug) here.

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+  // Type(1 byte)
+  station.type = strtol(start_ptr, nullptr, 16);
```
MaxKellermann, line 120: Another null pointer dererefence bug here.

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+  // Payload length(1 byte)
+  start_ptr = strchr(start_ptr, SEPARATOR) + 1;  // comma after type
```
MaxKellermann, line 123: Another null pointer dererefence bug here.

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,171 @@
...
+  station.payload_length = strtol(start_ptr, nullptr, 16);
```
MaxKellermann, line 124: Another null pointer dererefence bug here.

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,171 @@
...
+   * Standard limitation of LoRa buffers of 256 bytes and the maximum length
+   * of the FANET MAC header of 11 bytes reduces the permitted payload length
+   * to 245 bytes.
+   */
+  if(station.payload_length > 245)
+    return false;
+
+  // Payload(payload_length bytes)
+  station.payload = new uint8_t[station.payload_length];
+  start_ptr = strchr(start_ptr, SEPARATOR) + 1;  // comma after payloadlength
```
MaxKellermann, line 136: Another null pointer dererefence bug here.

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,171 @@
...
+  for(int i = 0; i < station.payload_length; i++)
+  {
+    char sstr[3] = {start_ptr[i * 2], start_ptr[i * 2 + 1], '\0'};
```
MaxKellermann, line 139: Another null pointer dererefence bug here.

src/Device/Driver/FANET/Parser.hpp
```diff
@@ -0,0 +1,33 @@
...
+*/
+
+#ifndef XCSOAR_FANET_PARSER_HPP
+#define XCSOAR_FANET_PARSER_HPP
+
+#define CMD_RX_PACKET			  "#FNF"
+#define CMD_WEATHER			    "#FNW"
+#define SEPARATOR			      ','
+
+#define TYPE_SERVICE        4
```
MaxKellermann, line 31: No macros.

src/Device/Driver/FANET/Register.cpp
```diff
@@ -0,0 +1,38 @@
...
+{
+    return new FanetDevice();
+}
+
+const struct DeviceRegister fanet_driver = {
+        _T("FANET"),
+        _T("FANET"),
+        0,
+        FanetCreateOnPort,
+};
```
MaxKellermann, line 38: Missing newline.

src/FANET/FanetAddress.hpp
```diff
@@ -0,0 +1,59 @@
...
+    return FanetAddress(UNDEFINED_VALUE, UNDEFINED_VALUE);
+  }
+
+  bool operator==(FanetAddress other) const {
+    return id == other.id && manufacturer == other.manufacturer;
+  }
+
+  bool operator<(FanetAddress other) const {
+    return id < other.id;
+  }
```
MaxKellermann, line 53: What's the point of this operator, and why does it ignore the manufacturer? (Unlike `operator==()`)

src/FANET/FanetStation.cpp
```diff
@@ -0,0 +1,146 @@
...
+void
+FanetStation::Update(const FanetStation &other)
+{
+  temperature = other.temperature;
+  wind_dir_deg = other.wind_dir_deg;
+  wind_speed_kmph = other.wind_speed_kmph;
+  wind_gust_kmph = other.wind_gust_kmph;
+  rel_humidity = other.rel_humidity;
+  preasure_hpa = other.preasure_hpa;
+  soc_percent = other.soc_percent;
```
MaxKellermann, line 35: What's the point of this, why not let the compiler generate this code for you?

src/FANET/FanetStation.hpp
```diff
@@ -0,0 +1,99 @@
...
+#define FANETFRAMESERVICE_INET			        7
+#define FANETFRAMESERVICE_TEMP			        6
+#define FANETFRAMESERVICE_WIND			        5
+#define FANETFRAMESERVICE_HUMIDITY		      4
+#define FANETFRAMESERVICE_PRESSURE		      3
+#define FANETFRAMESERVICE_REMOTECFGSUPPORT	2
+#define FANETFRAMESERVICE_SOC			          1
+#define FANETFRAMESERVICE_EHEADER           0
+
+#define IS_NTH_BIT_SET(byte, bit) byte & (1<<bit)
```
MaxKellermann, line 43: Macro hell

src/FANET/FanetStation.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+struct FanetStation {
+  uint8_t type;
```
MaxKellermann, line 46: What does this field mean?

src/FANET/FanetStation.hpp
```diff
@@ -0,0 +1,99 @@
...
+  float wind_dir_deg;
+  float wind_speed_kmph;
+  float wind_gust_kmph;
+  float rel_humidity;
+  double preasure_hpa;
+  float soc_percent;
+
+  /** Payload */
+  int16_t payload_length;
+  uint8_t *payload;
```
MaxKellermann, line 59: Who owns this pointer?
Why does this class allow copying the pointer?

MaxKellermann (changes requested):

src/Device/Driver/FANET/Device.hpp
denibakovic: Added comments and made them private

src/Device/Driver/FANET/Parser.cpp
denibakovic: No, fixed in recent amend

src/Device/Driver/FANET/Parser.cpp
denibakovic: This will insert \0 in place of first occurance of the \r, if \r is not found looks for \n and inserts it there. That sounds reasonable to me.

src/Device/Driver/FANET/Parser.cpp
denibakovic: Fixed

src/Device/Driver/FANET/Parser.cpp
denibakovic: Fixed

src/Device/Driver/FANET/Parser.cpp
denibakovic: Fixed

src/Device/Driver/FANET/Parser.cpp
denibakovic: Fixed

src/Device/Driver/FANET/Parser.cpp
denibakovic: Fixed

src/Device/Driver/FANET/Parser.cpp
denibakovic: Fixed

src/Device/Driver/FANET/Parser.hpp
denibakovic: Removed

src/Device/Driver/FANET/Register.cpp
denibakovic: Added

src/FANET/FanetAddress.hpp
denibakovic: Removed

src/FANET/FanetStation.hpp
denibakovic: Removed, replaced with enum and function

src/FANET/FanetStation.hpp
denibakovic: Removed, decoding is now done in Parser.cpp so its not needed

src/FANET/FanetStation.hpp
denibakovic: Removed, decoding is now done in Parser.cpp so its not needed

src/FANET/FanetStation.cpp
denibakovic: When station is allready present in the StationList this updates only weather data of that station from the data in other station. I'm not really sure how to "let the compiler generate this code for you", can you please advise?

denibakovic:
Fixed all build issues by adding missing imports to test.mk, checked with 'make everything'
Refactored FanetStation.cpp, moved most of it to Parser.cpp, keeping only data layer in the FanetStation
Removed all macros, replaced with enums and function
Added comments to functions
Proper private usage for Device.hpp functions end enums
Added missing newline in Register.cpp
IntegrityCheck empty string fix
Removed operator '<' in FanetAddress
Fixed all null pointer dereference issues
Comment cleanup
Added couple of basic tests for parser
Fixed broken windows build and broken windows unicode support
Tested on Linux, Windows, Android.

Should probably add option to disable showing stations and maybe change expire time and units according to user preferences

src/Device/Driver/FANET/Device.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+#include "Device/Driver.hpp"
+#include "FANET/List.hpp"
+#include "FANET/FanetStation.hpp"
+
+struct NMEAInfo;
+
+class FanetDevice final : public AbstractDevice {
+private:
+  enum FanetPacketType {
```
MaxKellermann, line 35: Those enums should be strictly-typed (`enum class`) so they are namespaced properly. That way, you can remove the `FANET_TYPE_` prefix.
But.... this enum isn't used at all - there isn't a single variable anywhere with this type!

src/Device/Driver/FANET/Device.hpp
```diff
@@ -0,0 +1,142 @@
...
+    FANET_TYPE_MESSAGE,
+    FANET_TYPE_SERVICE,
+    FANET_TYPE_LANDMARKS,
+    FANET_TYPE_REMOTE_CONFIG,
+    FANET_TYPE_GROUND_TRACKING,
+    FANET_TYPE_HW_INFO,
+    FANET_TYPE_THERMAL
+  };
+
+  enum FanetServiceHeaderBits {
```
MaxKellermann, line 48: Same here - should be strictly-typed, and this enum type is unused!

src/Device/Driver/FANET/Device.hpp
```diff
@@ -0,0 +1,142 @@
...
+#define XCSOAR_FANET_DEVICE_HPP
+
+#include "Device/Driver.hpp"
+#include "FANET/List.hpp"
+#include "FANET/FanetStation.hpp"
+
+struct NMEAInfo;
+
+class FanetDevice final : public AbstractDevice {
+private:
```
MaxKellermann, line 34: Redundant `private`. A `class` is `private` by default.

src/Device/Driver/FANET/Device.hpp
```diff
@@ -0,0 +1,142 @@
...
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+
+#ifndef XCSOAR_FANET_DEVICE_HPP
+#define XCSOAR_FANET_DEVICE_HPP
+
+#include "Device/Driver.hpp"
+#include "FANET/List.hpp"
+#include "FANET/FanetStation.hpp"
```
MaxKellermann, line 29: Don't include this, it's not used. Do a forward-declaration instead.

src/Device/Driver/FANET/Device.hpp
```diff
@@ -0,0 +1,142 @@
...
+    FANET_SERVICE_BIT_TEMP,
+    FANET_SERVICE_BIT_INET
+  };
+
+  /**
+   * Checks if nth bit is set
+   *
+   * @return true if its set, false otherwise
+   */
+  static bool IsNthBitSet(uint8_t byte, int bit) { return byte & (1 << bit); };
```
MaxKellermann, line 64: This should be `constexpr` and `noexcept`. The `bit` parmeter should not be signed.
And this shouldn't be a method. This doesn't belong in the class.

src/Device/Driver/FANET/Device.hpp
```diff
@@ -0,0 +1,142 @@
...
+  /**
+   * Decodes FANET service(type 4) payload and updates provided station with
+   * decoded info
+   *
+   * @param payload to deode
+   * @param payload_length
+   * @param station to update
+   * @return true if successfully decoded, false otherwise
+   */
+  static bool DecodeServicePayload(uint8_t *payload, int16_t payload_length,
```
MaxKellermann, line 75: Memory sizes should be `size_t`.

src/Device/Driver/FANET/Device.hpp
```diff
@@ -0,0 +1,142 @@
...
+                                   FanetStation *station);
```
MaxKellermann, line 76: Pass by reference, not by pointer, unless you allow the `nullptr` special value.

src/Device/Driver/FANET/Device.hpp
```diff
@@ -0,0 +1,142 @@
...
+                               FanetStation station);
+
+  /**
+   * Replace first occurrence of newline character with \0 and moves pointer
+   * to the first non space character
+   *
+   * @param cmd_str line to replace newline character in
+   * @return stripped cmd_str
+   */
+  static char *RemoveExtraChars(char *cmd_str);
```
MaxKellermann, line 127: Bad name. It doesn't say anything about what this function really does.

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+}
+
+/*
+ * https://github.com/3s1d/fanet-stm32/raw/master/fanet_module.pdf
+ * https://github.com/3s1d/fanet-stm32/blob/master/Src/fanet/radio/protocol.txt
+ */
+bool
+FanetDevice::ParseNMEA(const char *line, NMEAInfo &info)
+{
+  if(StringIsEqual(line, "#FNF", strlen("#FNF"))) {
```
MaxKellermann, line 298: Use `StringAfterPrefix()`, that's much easier and safer than your complicated pointer arithmetic with two `strlen()` calls!

src/Device/Driver/FANET/Parser.cpp
MaxKellermann: I'm not convinced, this sounds pretty broken. *If* you believe it is necessary to stop parsing at the next `\r` or `\n`, then leaving `\n` in there under certain conditions just isn't right.
But note that you were already given a single line to parse. So what's the point of this function?

src/Device/Driver/FANET/Device.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * Check if cmd_str contains only hexadecimal chars & commas and is > 0 chars
+   *
+   * @param cmd_str line to check
+   * @return true if check is passed, false otherwise
+   */
+  static bool IntegrityCheck(char *cmd_str);
```
MaxKellermann, line 135: Same here - bad name. "Integrity check" can mean anything.
But... I don't understand the point of this function. Why not check while parsing? Because this function doesn't parse anything. It is just a pre-processing step.

src/Device/Driver/FANET/Parser.cpp
MaxKellermann: Using `strlen()==0` is a pretty clumsy way to check for empty strings. If it's not empty, it needs to check each and every character in the string. The best way to check for empty strings in C is `*cmd_str == 0`.
But you don't even neeed that. You can just change the loop, e.g.:
```
do {
  if (!IsHexOrCommaChar(*cmd_str))
    return false;
} while (*++cmd_str != 0);
return true;
```
If the first character is a null byte, it will fail the test. But follow-up null bytes will instead just break out of the loop.
My example also avoids your quirky `continue` with `return` at the end of the loop scope... which I think is bad coding style.

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+  // Integer values
+  unsigned int lati = buf[2] << 16 | buf[1] << 8 | buf[0];
+  if(lati & 0x00800000)
+    lati |= 0xFF000000;
+  unsigned int loni = buf[5] << 16 | buf[4] << 8 | buf[3];
+  if(loni & 0x00800000)
+    loni |= 0xFF000000;
+
+  return GeoPoint(Angle::Degrees((float)loni / 46603.0f), Angle::Degrees((float)lati / 93206.0f));
```
MaxKellermann, line 94: Where do those "magic" values come from?

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+  auto value = (float)(byte & 0x7F);
+  if(byte & 1 << 7)
+    return (value * scale);
+  else
+    return value;
+}
+
+bool
+FanetDevice::DecodeServicePayload(uint8_t *payload, int16_t payload_length, FanetStation *station) {
+  if(payload == nullptr || payload_length == 0)
```
MaxKellermann, line 108: Why is it legal to call `DecodeServicePayload(nullptr)`? This should be considered a bug, and thus a runtime check is the wrong way to check - do `assert(payload != nullptr)` instead.

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+   * Barometric pressure normailized (+2byte: in 10Pa, offset by 430hPa,
+   * unsigned little endian (hPa-430)*10)
+   */
+  if(IsNthBitSet(payload[0], FANET_SERVICE_BIT_PRESSURE))
+  {
+    if(payload_pos + 2 > payload_length)
+      return false;
+    int lsb = payload[payload_pos++];
+    int msb = payload[payload_pos++];
+    station->preasure_hpa = (((msb << 8) | lsb) / 10.0) + 430.0;
```
MaxKellermann, line 175: This obscure code reads an unaligned 16 bit little-endian integer from the buffer. But we have a function for that already: `ReadUnalignedLE16()`.
Also note that you're using a signed integer here; you cast the unsigned byte to a signed integer. Why the ... would you do that?

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+                           double clock) {
+  // Prepare the line
+  cmd_str = RemoveExtraChars(cmd_str);
+  if ( ! IntegrityCheck(cmd_str))
+    return false;
+
+  char *start_ptr = (char *)cmd_str;
+  char *end_ptr;
+
+  if(! start_ptr)
```
MaxKellermann, line 208: Why is `ParseRxPacket(nullptr)` legal?

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+    return false;
+
+  // Manufacturer(1 byte)
+  uint8_t manufacturer = strtol(start_ptr, &end_ptr, 16);
+  if(start_ptr == end_ptr)
+    return false;
+
+  start_ptr = strchr(start_ptr, ',') + 1;  // comma after src_manufacturer
+  if (start_ptr == nullptr)
```
MaxKellermann, line 217: If there's no comma, then `start_ptr` will *not* be `nullptr`, but `nullptr+1`. This your code will still crash due to `nullptr` dereference!
Also I don't like this way of parsing at all. It looks very clumsy. You called `strtol()`, and you know where that hex number ends - at `end_ptr`. Still you start looking for the comma at `start_ptr`.

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+  char *start_ptr = (char *)cmd_str;
+  char *end_ptr;
+
+  if(! start_ptr)
+    return false;
+
+  // Manufacturer(1 byte)
+  uint8_t manufacturer = strtol(start_ptr, &end_ptr, 16);
+  if(start_ptr == end_ptr)
```
MaxKellermann, line 213: If you know it's one byte, why do you allow an arbitrary number of bytes here?

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+    return false;
+
+  char *start_ptr = (char *)cmd_str;
+  char *end_ptr;
+
+  if(! start_ptr)
+    return false;
+
+  // Manufacturer(1 byte)
+  uint8_t manufacturer = strtol(start_ptr, &end_ptr, 16);
```
MaxKellermann, line 212: Why signed integer?

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+  if(start_ptr == end_ptr)
+    return false;
+
+  /*
+   * Skip broadcast(1 byte) and signature(2 bytes)
+   *  comma after src_id
+   *  comma after broadcast
+   *  comma after signature
+   */
+  for (int i = 0; i < 3; ++i) {
```
MaxKellermann, line 231: Why signed integer?

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+      return false;
+  }
+
+  // Type(1 byte)
+  uint8_t type = strtol(start_ptr, &end_ptr, 16);
+  if(start_ptr == end_ptr)
+    return false;
+
+  start_ptr = strchr(start_ptr, ',') + 1;  // comma after type
+  if (start_ptr == nullptr)
```
MaxKellermann, line 243: Another `nullptr` dereference crash bug.

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+    return false;
+
+  /*
+   * Skip broadcast(1 byte) and signature(2 bytes)
+   *  comma after src_id
+   *  comma after broadcast
+   *  comma after signature
+   */
+  for (int i = 0; i < 3; ++i) {
+    start_ptr = strchr(start_ptr, ',') + 1;
```
MaxKellermann, line 232: Another `nullptr` dereference bug.

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+   * Standard limitation of LoRa buffers of 256 bytes and the maximum length
+   * of the FANET MAC header of 11 bytes reduces the permitted payload length
+   * to 245 bytes.
+   */
+  if(payload_length > 245)
+    return false;
+
+  // Payload(payload_length bytes)
+  auto* payload = new uint8_t[payload_length];
+  start_ptr = strchr(start_ptr, ',') + 1;  // comma after payloadlength
```
MaxKellermann, line 261: Another `nullptr` dereference bug.

src/Device/Driver/FANET/Parser.cpp
MaxKellermann: I don't see a fix - what did you change? This still crashes.
(btw. I was wrong - this isn't a nullptr dereference bug, it's a buffer overflow.)

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+  // Extract payload
+  for(int i = 0; i < payload_length; i++)
+  {
+    char sstr[3] = {start_ptr[i * 2], start_ptr[i * 2 + 1], '\0'};
+    if(strlen(sstr) != 2)
+      return false;
+
+    payload[i] = strtol(sstr,  &end_ptr,  16);
+    if(start_ptr == end_ptr)
```
MaxKellermann, line 273: This allows failing to parse the second character.

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+      return false;
+  }
+
+  switch (type)
```
MaxKellermann, line 277: `type` is declared to be a `uint8_t`, but you compare it with `enum FanetPacketType` here!

src/Device/Driver/FANET/Parser.cpp
```diff
@@ -0,0 +1,304 @@
...
+  {
+    // We only decode type 4 (service / weather data) for now
+    case FANET_TYPE_SERVICE:
+      FanetStation station;
```
MaxKellermann, line 281: You can't declare variables here.

src/FANET/FanetStation.cpp
MaxKellermann: The C++ compiler auto-generates an `operator=()` for all classes (with several exceptions of the rule). So if you want to update "only weather data of that station", then have "weather data of that station" should be a separate struct, not mixed with metadata you don't want to update.
Your way of writing this manually is cumbersome and fragile.

src/FANET/FanetStation.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+struct FanetStation {
+  /** Station info */
+  float temperature;
+  float wind_dir_deg;
+  float wind_speed_kmph;
+  float wind_gust_kmph;
+  float rel_humidity;
+  double preasure_hpa;
+  float soc_percent;
```
MaxKellermann, line 42: These need to be documented.
The variable names suggest that you're using non-standard units; that is error prone. XCSoar should use SI units everywhere, and special classes where we have them.
For pressure, we have a the `class AtmosphericPressure`. For angles, we have `Angle`.

MaxKellermann (changes requested):

MaxKellermann:
Lots and lots of problems, but no progress after 2 months. Closing until the fixes are available.

PR #517, opened 2021-01-30
Add cruse circling transition mode threshold setting

### LINKED ISSUE #501: Make CirclingComputer::CRUISE_CLIMB_SWITCH, CLIMB_CRUISE_SWITCH and MIN_TURN_RATE configurable
Currently the threshold before the screen change between cruise and climb mode is hardcoded. We should make these magic number configurable in settings.

I did some quickfix in my build https://github.com/zinuzoid/XCSoar/pull/5/files . But I think this could benefit the upstream if we make it configurable.

Just wanted to get your opinion before I'll working on it.

zinuzoid:
Brief summary of the changes
----------------------------

Currently, the threshold for the transition from Cruise to Circling mode is hardcoded (AUTO screen change to Circling).
I personally think that 15 sec in turn required before a change to Circling mode is too long.
So I expose the number here in the expert mode settings.

Close #501 

### Here's how it's looks
https://user-images.githubusercontent.com/496209/106370755-14f67c00-6355-11eb-89d4-b7d8745eda4d.mov


MaxKellermann:
You're bloating your commit numbers by splitting one change into 5 commits. But those changes really belong together - they can't be separated, they make no sense as separate commits.

zinuzoid:
> You're bloating your commit numbers by splitting one change into 5 commits. But those changes really belong together - they can't be separated, they make no sense as separate commits.

I definitely have no interest in bloat my commit numbers or whatever, it even increase unnecessary extra step to split the commits.
I've got your comment on #499 that you don't like too many things happen in one commit that's why I split it here.

zinuzoid:
Here's the squashed commit

MaxKellermann:
> I've got your comment on #499 that you don't like too many things happen in one commit that's why I split it here.

I think we misunderstood. This is not at all about "too many things in one commit". It's about "more than one *unrelated* thing in one commit".
In #499, you moved code around *and* modified it. That's bad because it makes reading your changes impossible.
In this PR, you splitted one change into several commits, which doesn't make any sense.

MaxKellermann:
The squashed commit has a bad commit message.

zinuzoid:
- updated commit message

PR #518, opened 2021-01-31
Fix android.max_aspect for Android device that aspect ratio greater than 16:9

zinuzoid:
In https://github.com/XCSoar/XCSoar/commit/fb8370c2ed5df0e69a1055e3d53d922847ae2ea4 there was an attempt to fix full screen size for Android device that aspect ratio greater than 16:9

But there's a bug in that fix, we suppose to replace `ratio_float` with the actual number based on the document: https://android-developers.googleblog.com/2017/03/update-your-app-to-take-advantage-of.html
So the fix wasn't actually works.

This should be a real fix.
before|after
------|-----
Screenshot_20210131-202158_300x650|Screenshot_20210131-201812_300x650

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -25,7 +25,7 @@
       <meta-data android:name="android.hardware.usb.action.USB_DEVICE_ATTACHED"
                  android:resource="@xml/usb_device_filter" />
       <meta-data android:name="android.max_aspect"
-                 android:value="ratio_float"/>
+                 android:value="2.1"/>
     </activity>
 
     <service android:name=".MyService"/>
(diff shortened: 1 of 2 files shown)


MaxKellermann:
Bad commit message, NEWS line should be added in the same commit, both AndroidManifests should be updated and should be merged to v6.8.x instead of master.

zinuzoid:
- Updated commit message
- Fixed both AndroidManifests
- Changed base branch

MaxKellermann:
The commit message is still bad, because it doesn't explain anything.

zinuzoid:
- updated commit message

Issue #519, opened 2021-02-01
In Docker some packages missing for an error free run


August2111:
XCSoar versions having and not having the problem
-------------------------------------------------
actually master branch 7.0 

Expected behavior
-----------------

Compiling XCSoar in the Docker environment it should no error happens

Actual behavior
---------------

Same Packages are missing: 
* overall build environment: cmake, ninja-build
* linux target: libglm-dev

Steps to reproduce the behavior
-------------------------------

Insert this package names in file ide/provisioning/install-debian-packages.sh
 

Do you have an idea how to solve the issue?
-------------------------------------------

I will make a PR!

August2111:
Add to the PR #520

PR #520, opened 2021-02-01
Add missing packages for a clean docker build run

August2111:
Brief summary of the changes
----------------------------

Solve Issue #519 'In Docker some packages missing for an error free run'
Add 3 Packages to the docker-provisioning-file: 
* overall build environment: cmake, ninja-build
* linux target: libglm-dev

The same packages are needed for a build in the native environment

### DIFF
--- a/ide/provisioning/install-debian-packages.sh
+++ b/ide/provisioning/install-debian-packages.sh
@@ -15,7 +15,7 @@ apt-get install $APTOPTS make \
   git quilt zip \
   m4 automake wget \
   ttf-bitstream-vera fakeroot \
-  pkg-config
+  pkg-config cmake ninja-build
 echo
 
 echo Installing Manual dependencies...


August2111:
Of course I don't need the issue tracking before, if I want to solve such an issue?

ide/provisioning/install-debian-packages.sh
```diff
@@ -46,7 +45,7 @@ apt-get install $APTOPTS make g++ \
   imagemagick gettext \
   mesa-common-dev libgl1-mesa-dev libegl1-mesa-dev \
   fonts-dejavu \
-  xz-utils
+  xz-utils  libglm-dev
```
MaxKellermann, line 48: This is certainly wrong.

MaxKellermann:
> Of course I don't need the issue tracking before, if I want to solve such an issue?

If you send a PR, you don't need to report an issue.

But I'm not happy with your commit message. It doesn't explain anything.

August2111:
Ok, I explain it much more on the issue site, but you are right, here is it much better...
Compiling XCSoar in the Docker environment it happens 3 stops regarding missing packages:
    * overall build environment: cmake, ninja-build - are missing and at this point the make process is stopping witch a failure, for example to build the c-ares in WIN64 target
    * linux target: libglm-dev - without this installed library I have every time the issue "glm/fwd.hpp" (missing in ui\canvas\opengl\Globals.hpp and in ui\canvas\opengl\Geo.hpp) if I want to compile the UNIX target.
 Maybe there is an other way to solve this, but in my case it was the easiest.

August2111:
Hopefully it is more clear now...

MaxKellermann:
GLM is still wrong. It's used as a git submodule, and thus the dev package is not needed. If you need it, you're doing it wrong.

August2111:
Ok, that's right...
I didn't get some submodule experience up to now ;-(
But it is very  strange, that XCSoar use so much possibilities to integrate 3rd party software in their package!

MaxKellermann:
It's documented in the developer manual.

August2111:
Now I had read this ;-)
But why you use such different methods?

I'm fighting a little bit with GitHub, in some minutes I will make an update of this PR!

Issue #505, opened 2020-12-28
Upload IGC to WeGlide


bomilkar:
XCSoar version
--------------

7.0

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------

We will see a new competition platform in the coming season, at least in Germany: DMSt is committed to switch to WeGlide.
I learned from the WeGlide team that there is an API for uploading IGC files.
https://api.weglide.org/docs#/igcfile/upload_flight_simple_v1_igcfile_post

**I'm proposing to integrate an Upload feature into XCSoar.**

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------

I'm envisioning adding the choice to upload the flight as part of the "Flight download" button/function in the logger dialog.
For example with FLARM as a connected logger:

- hit the "Flight download" button
- select a flight
- the download of the IGC file happens
- after the download  is complteted and **_before_** you see the message "Do you want to download another flight?"
- **insert the choice to upload the IGC file to WeGlide.**

It may need an extra page in Setup to define WeGlide Profile data. Alternatively it may collect/remember this information (Pilot ID, account info, ...) as part of the choice to upload to WeGlide.

Has anybody started to work on this already?
Maybe build this feature such that it can deal with other platforms, too. For instance SkyLines.

moldhouse:
Hey, just dropping a note from WeGlide. We have no familiarity with the XCSoar source code and can not state if this feature is sensible in any way.

If this feature is wanted for the XCSoar project, we are happy to provide help or adjust API endpoints. LXNAV and SeeYou also use our API to fetch tasks (cup or json) which users have planned.

MaxKellermann:
What is WeGlide? It has a `.org` domain, but it isn't an "organization". It is a commercial company, funded by tax money, it appears. If it was funded by the taxes I pay, I'd expect it to be a charitable/non-profit, but it is not. The code is proprietary and closed, isn't it?
This is unlike the OLC - which is a "gGmbH", which is a *non*-commercial company (officially). (Also proprietary and closed-source, but weird in another way - many people work for the company for free, while the company collects massive funds from the gliding association, effectively paid by the same people contributing their time for free.)

bomilkar:
I understand what you mean, @MaxKellermann . And you raising a valid point that I hadn't thought of. But I don't want to speculate about how WeGlide is funded or if anyone is making money off of it. I want to hear that first hand.
[...]

PR #523, opened 2021-02-06
TaskManager: Add the feature to download your declared Task from WeGlide

henrik-the:
Brief summary of the changes
----------------------------
This commit adds the feature to download your declared task from WeGlide. The Task must be publicly available (Authentification can be added later to also download private Tasks). This feature uses the WeGlide API: 
https://api.weglide.org/docs#/task/get_task_declared_v1_task_declaration__user_id__get

I also added a new Setting Parameter: WeGlide Pilot ID
It is located in Settings -> Logger
You can get your Pilot ID from your profile link. 
For example my ID is 67:
https://beta.weglide.org/profile/67

image

Here you can Download the Task and find it in the Browse Tab:
image

### DIFF
--- a/src/Dialogs/Task/Manager/TaskActionsPanel.cpp
+++ b/src/Dialogs/Task/Manager/TaskActionsPanel.cpp
@@ -37,6 +37,92 @@ Copyright_License {
 #include "Engine/Task/Ordered/OrderedTask.hpp"
 #include "Engine/Task/Factory/AbstractTaskFactory.hpp"
 #include "Engine/Waypoint/Waypoints.hpp"
+#include "net/http/DownloadManager.hpp"
+#include "LocalPath.hpp"
+#include "system/FileUtil.hpp"
+#include "LogFile.hpp"
+#include "../../WidgetDialog.hpp"
+#include "../../ProgressDialog.hpp"
+#include "net/http/Features.hpp"
+#include "Operation/ThreadedOperationEnvironment.hpp"
+#include "ui/event/Notify.hpp"
+#include "ui/event/PeriodicTimer.hpp"
+#include "UIGlobals.hpp"
+
+/**
+ * This class tracks a download and updates a #ProgressDialog.
+ */
+class WeGlideDownloadProgress final : Net::DownloadListener {
+  ProgressDialog &dialog;
+  ThreadedOperationEnvironment env;
+  const Path path_relative;
+
+  UI::PeriodicTimer update_timer{[this]{ Net::DownloadManager::Enumerate(*this); }};
+
+  UI::Notify download_complete_notify{[this]{ OnDownloadCompleteNotification(); }};
+
+  std::exception_ptr error;
+
+  bool got_size = false, complete = false, success;
+
+public:
+  WeGlideDownloadProgress(ProgressDialog &_dialog,
+                   const Path _path_relative)
+    :dialog(_dialog), env(_dialog), path_relative(_path_relative) {
+    update_timer.Schedule(std::chrono::seconds(1));
+    Net::DownloadManager::AddListener(*this);
+  }
+
+  ~WeGlideDownloadProgress() {
+    Net::DownloadManager::RemoveListener(*this);
+  }
+
+  void Rethrow() const {
+    if (error)
+      std::rethrow_exception(error);
+  }
+
+private:
+  /* virtual methods from class Net::DownloadListener */
+  void OnDownloadAdded(Path _path_relative,
+                       int64_t size, int64_t position) noexcept override {
+    if (!complete && path_relative == _path_relative) {
+      if (!got_size && size >= 0) {
+        got_size = true;
+        env.SetProgressRange(uint64_t(size) / 1024u);
+      }
+
+      if (got_size)
+        env.SetProgressPosition(uint64_t(position) / 1024u);
+    }
+  }
+
+  void OnDownloadComplete(Path _path_relative) noexcept override {
+          LogFormat("OnDownloadComplete");
+    if (!complete && path_relative == _path_relative) {
+      complete = true;
+      success = true;
+
+      download_complete_notify.SendNotification();
+      DirtyTaskListPanel();
+    }
+  }
+
+  void OnDownloadError(Path _path_relative,
+                       std::exception_ptr _error) noexcept override {
+    if (!complete && path_relative == _path_relative) {
+      complete = true;
+      success = false;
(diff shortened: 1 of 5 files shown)


src/Dialogs/Task/Manager/TaskActionsPanel.cpp
```diff
@@ -103,6 +107,23 @@ TaskActionsPanel::OnDeclareClicked()
...
+TaskActionsPanel::OnDownloadClicked()
+{
+  const ComputerSettings &settings = CommonInterface::GetComputerSettings();
+  char url[256];
+  char id[20];
+
+  strcpy(id, settings.logger.pilot_weglide_id.c_str());
+  snprintf(url, sizeof(url),"https://api.weglide.org/v1/task/declaration/%s?cup=false&tsk=true",id);
+  
+  const auto cache_path = MakeLocalPath(_T("weglide"));
```
MaxKellermann, line 120: This variable is never used. You should use it to build the final path name.

src/Dialogs/Task/Manager/TaskActionsPanel.cpp
```diff
@@ -103,6 +107,23 @@ TaskActionsPanel::OnDeclareClicked()
...
+  File::Delete(LocalPath(_T("weglide/weglide_declared.tsk")));
```
MaxKellermann, line 121: If the download fails, this loses data. Doesn't sound like a good way to handle errors.

src/Dialogs/Task/Manager/TaskActionsPanel.cpp
```diff
@@ -103,6 +107,23 @@ TaskActionsPanel::OnDeclareClicked()
...
+  Net::DownloadManager::Enqueue(url, Path(_T("weglide/weglide_declared.tsk")));
```
MaxKellermann, line 122: I don't like the hard-coded file name. How do you explain to users how this works? Your commit doesn't contain any documentation.

src/Dialogs/Task/Manager/TaskActionsPanel.cpp
```diff
@@ -103,6 +107,23 @@ TaskActionsPanel::OnDeclareClicked()
...
+  
+  DirtyTaskListPanel();
```
MaxKellermann, line 124: The "Enqueue" call is asynchronous. After it finishes, the file hasn't yet been downloaded. Download might not even have been started. So why call `DirtyTaskListPanel()`?
How do you present download errors to the user? There's no error handling at all here. You don't even know if deleting the file was successful.

src/Dialogs/Task/Manager/TaskActionsPanel.hpp
```diff
@@ -55,6 +55,7 @@ class TaskActionsPanel : public RowFormWidget {
   void OnBrowseClicked();
   void OnNewTaskClicked();
   void OnDeclareClicked();
+  void OnDownloadClicked();
```
MaxKellermann, line 58: New functions should be `noexcept` unless you know they can throw an exception.
Most of XCSoar's existing code doesn't do that, because we only allowed C++ exceptions in version 7.0. PRs for adding `noexcept` are very welcome, btw.

src/Profile/ProfileKeys.cpp
```diff
@@ -96,6 +96,7 @@ const char LanguageFile[] = "LanguageFile"; // pL
 const char InputFile[] = "InputFile"; // pL
 const char PilotName[] = "PilotName";
 const char CoPilotName[] = "CoPilotName";
+const char PilotWeGlideId[] = "PilotWeGlideId";
```
MaxKellermann, line 99: I'd rather name it `WeGlidePilotId`, because there may be other WeGlide specific settings in the future, and having a common prefix is a good thing.

MaxKellermann (changes requested):
This PR contains text, but it's missing in the commit message. If you ever feel the need to add text to a PR, it only means that your commit message is bad.

bomilkar:
@henrik-the I'd like to bring this Issue #505 to your attention.
Just want to let you know I'm working on it. Maybe we can align forces?

MaxKellermann:
Are you still working on this?

August2111:
Hello Henry, maybe I will insert your code to my work on issue #505, so from my point of view you can close this PR without any investigations... I will build a complete WeGlide component, hopefully there is containing all what you need. If you have any remarks on this please contact me directly

henrik-the:
@MaxKellermann Yes I am still working on it. I just have to cope with a few problems.

lordfolken:
There seems to be a compilation issue with the windows build.

August2111:
Hi Hendrik, 
It's nice to see your new approach for the 'Task Download' :) I also put a lot of work into it, although I took your 1st PR as a basis - and put my implementation on top of it - but it was completely different is. Most of the download actions take place in the download manager, in my 'TaskActionsPanel' there are only the code parts that actually affect the task... However, there is currently a small traffic jam when taking over the PRs - I would only want to submit my code once my PR #859 and a subsequent PR based on it have been adopted (this also applies to the selection of the WeGlide user ID and things like that). Unfortunately, our two approaches now get in each other's way ;-(
First of all, in my opinion, the developer flow has to get going again - then maybe we can discuss it again!

MaxKellermann:
This PR is a mess of code copied from `DownloadFilePicker.cpp`, inconsistent coding style, debug log left in the code, typos in user-visible messages ("Successfull").

Of the 149 code lines added by this PR, 74 were copied from `DownloadFilePicker.cpp`. That's 50% of this patch. Redundant code copies are not acceptable.

MaxKellermann:
I rewrote the core of this PR, the task downloader, from scratch. This is how to use it:
https://github.com/XCSoar/XCSoar/blob/d4454f065f0ce79483ac95e4ac82e0ea9a07bf37/test/src/RunWeGlideDownloadTask.cpp#L44
https://github.com/XCSoar/XCSoar/blob/d4454f065f0ce79483ac95e4ac82e0ea9a07bf37/test/src/RunWeGlideDownloadTask.cpp#L51-L53
This runs as a C++ coroutine, i.e. does not block anything, and doesn't use the DownloadManager, therefore omits the whole Net::DownloadListener complexity, doesn't create a file (returns an OrderedTask instead), and therefore omits all the messy things in this PR. It is simpler, faster, more robust and easier to use.

MaxKellermann:
b1b1b3ebe9da255bb923f7b9255226a5f5e151da

Issue #492, opened 2020-11-08
Can’t go back to previous task point


bomilkar:
XCSoar versions having and not having the problem
-------------------------------------------------

7.0 preview15

Since several previews of 7.0 I noticed the issue

Expected behavior
-----------------

For example, the following task points:
T0, T1, T2, T3, T4
Assume you have passed T0 and T1. Next turn point is now T2.
Use “Previous Turnpoint” to switch back to T1.

Actual behavior
---------------

In this example:
Assume you have passed T0 and T1. Next turn point is now T2.
Try to use “Previous Turnpoint” to switch back to T1. **That won’t work!**

It seems as if you cannot go back to turn points which you had already passed!

Steps to reproduce the behavior
-------------------------------

Attached is a testcase, consisting of a task file and a igc file for replay. You might want to use the waypoint file along with it. A map with southern part of Germany would be nice (although not in the testcase):

copy “Default.rsk” and “Airports_de_2020-05-10.cup” to XCSoar’s data directory
copy “2020-08-08-FLA-L6A-01.igc” to XCSoar’s log directory
start XCSoar
select waypoint file: “Config” → “System” → “Site Files” → “Waypoints”
replay the IGC file: “Config” → “Config” → “Replay”, you can speed up the replay.

Make sure one of your InfoBoxes shows “UTC time”.
Let it replay until 13:10:00
You have just passed turn point “Gammelsdorf”.
The next turn point is now “Regensburg”.
Try to switch back to “Gammelsdorf” using “Previous Turnpoint”. **This will fail!**

Using “Next Turnpoint” repeatedly you can switch to turn point N. Using “Previous Turnpoint” you can switch to turn point N-1. But only if you had not passed turn point N-1 already.

Do you have any idea what may have caused this?
-----------------------------------------------

Here is a testcase.
Please remove the .txt file extensions.

[2020-08-08-FLA-L6A-01.igc.txt](https://github.com/XCSoar/XCSoar/files/5506348/2020-08-08-FLA-L6A-01.igc.txt)
[Default.tsk.txt](https://github.com/XCSoar/XCSoar/files/5506347/Default.tsk.txt)
[Airports_de_2020-05-10.cup.txt](https://github.com/XCSoar/XCSoar/files/5506349/Airports_de_2020-05-10.cup.txt)

Do you have an idea how to solve the issue?
-------------------------------------------

I’ve tried to wrap my head around the code, Menu/ExpandMacros.cpp in particular, but w/o success.

[...]

PR #528, opened 2021-02-18
Fix adjust to previous task point, rev2

bomilkar:
When manually adjusting to the previous turn point it must forget that it had visited that task point before. If it finds this TP had been visited before it auto-increments to the next TP.

Brief summary of the changes
----------------------------

This fix addresses Issue #492 
A more detailed description of the issue is there.

### DIFF
--- a/src/Task/ProtectedTaskManager.cpp
+++ b/src/Task/ProtectedTaskManager.cpp
@@ -25,7 +25,8 @@ Copyright_License {
 #include "Task/RoutePlannerGlue.hpp"
 #include "Engine/Task/TaskManager.hpp"
 #include "Engine/Task/Ordered/OrderedTask.hpp"
-#include "Engine/Task/Ordered/Points/IntermediatePoint.hpp"
+#include "Engine/Task/Ordered/Points/ASTPoint.hpp"
+#include "Engine/Task/Ordered/Points/AATPoint.hpp"
 #include "Engine/Task/Points/TaskWaypoint.hpp"
 #include "Engine/Route/ReachResult.hpp"
 
@@ -85,13 +86,14 @@ ProtectedTaskManager::IncrementActiveTaskPointArm(int offset)
 {
   ExclusiveLease lease(*this);
   TaskAdvance &advance = lease->SetTaskAdvance();
-  IntermediateTaskPoint *nextwp = nullptr;
+  TaskWaypoint *twp = nullptr;
 
   switch (advance.GetState()) {
   case TaskAdvance::MANUAL:
-  case TaskAdvance::AUTO:
-    lease->IncrementActiveTaskPoint(offset);
-    nextwp = (IntermediateTaskPoint *)lease->GetActiveTaskPoint();
+  case TaskAdvance::AUTO: {
+      lease->IncrementActiveTaskPoint(offset);
+      twp = lease->GetActiveTaskPoint();
+    }
     break;
   case TaskAdvance::START_DISARMED:
   case TaskAdvance::TURN_DISARMED:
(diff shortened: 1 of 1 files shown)


src/Task/ProtectedTaskManager.cpp
```diff
@@ -84,29 +86,50 @@ ProtectedTaskManager::IncrementActiveTaskPointArm(int offset)
...
+  TaskWaypoint *twp = nullptr;
 
   switch (advance.GetState()) {
   case TaskAdvance::MANUAL:
-  case TaskAdvance::AUTO:
-    lease->IncrementActiveTaskPoint(offset);
+  case TaskAdvance::AUTO: {
+      lease->IncrementActiveTaskPoint(offset);
+      twp = lease->GetActiveTaskPoint();
+    }
```
MaxKellermann, line 96: This scope is pointless - just remove the curly braces.
Are you aware that this code applies to both `AUTO` and `MANUAL`?

src/Task/ProtectedTaskManager.cpp
```diff
@@ -84,29 +86,50 @@ ProtectedTaskManager::IncrementActiveTaskPointArm(int offset)
...
+    switch (twp->GetType()) {
+    case TaskPointType::AST : {
+      ASTPoint *p = (ASTPoint *)twp;
+      if (p->HasEntered()) p->Reset();
+      }
+      break;
+    case TaskPointType::AAT : {
+      AATPoint *p = (AATPoint *)twp;
+      if (p->HasEntered()) p->Reset();
+      }
```
MaxKellermann, line 127: Duplicate code.
Why check `HasEntered()`?

MaxKellermann (changes requested):

src/Task/ProtectedTaskManager.cpp
bomilkar: Will remove the curly braces if you prefer.

It's a fallthrought and the compiler would flagged if it wasn't empty. I know.
Do you prefer not to use empty fallthroughs?

MaxKellermann: Empty fallthroughs are ok, I was just wondering whether you were aware that this also covers `MANUAL`.

src/Task/ProtectedTaskManager.cpp
bomilkar: Duplicate code: Can I assume classes ASTPoint and AATPoint would do the same???? I didn't want to make that assumption.

Agreed: I don't have to check HasEntered(). But doesn't it improve the readability?

MaxKellermann:
"Update ProtectedTaskManager.cpp" is a bad commit message. I have no idea what this commit does.

src/Task/ProtectedTaskManager.cpp
MaxKellermann: > Can I assume classes ASTPoint and AATPoint would do the same?

They don't do the same. But you do the same with both types. And that's what your code is about: what you do *with* those objects, not what *they* do.

> Agreed: I don't have to check HasEntered(). But doesn't it improve the readability?

Does it? Unnecessary checks don't improve readability. They reduce readability by having useless code, and they reduce performance.

Maybe the check is necessary - I did not check - I was just asking what was your intention why you wrote it.

MaxKellermann:
This PR contains a commit which appears to get reverted by the same PR. I don't merge known-bad commits; remove them from the PR instead.

bomilkar:
I start with a new PR.

MaxKellermann:
Don't. This will only cause confusion. Instead, fix this PR.

PR #535, opened 2021-03-03
Change of icons for v6.8.x

MikeBravo15:
Same change of icons as #526, 529


MaxKellermann:
v6.8.x is abandoned now, and this backport is obsolete.

PR #540, opened 2021-03-08
Reactivation PC-Met access

August2111:
Brief summary of the changes
----------------------------

More then 2-3 years ago Flugwetter.de changed a lot of URI fields for accessing the DWD content like Radar, Sat images and so on. Unfortunately there was no maintainance of this component since more then 4 years. 
In this PR I changed all related things, that this access and all the views are functional. For example on GDI this couldn't display anything, because without gdiplus (or maybe an other lib) there is no possibiltiy to read JPEG files (anf PNG, GIF,... and so on) 

With this functional component there are a lot of things possible to made:
* automatic update of the last current sat image
* a view with zoom in, zoom out and move... (the complete picture doesn't have the best resolution...)
* a (short) video of last pictures

At least this feature is functional again...

### DIFF
--- a/src/Startup.cpp
+++ b/src/Startup.cpp
@@ -97,6 +97,10 @@ Copyright_License {
 #include "Units/Units.hpp"
 #include "Formatter/UserGeoPointFormatter.hpp"
 #include "thread/Debug.hpp"
+#if defined(_WIN32) && defined(USE_GDI)
+#  include "ui/canvas/gdi/GdiPlusBitmap.hpp"
+#endif
+
 
 #include "lua/StartFile.hpp"
 #include "lua/Background.hpp"
@@ -457,6 +461,10 @@ Startup()
 #endif
 #endif
 
+#if defined(_WIN32) && defined(USE_GDI)
+  GdiStartup();
+#endif
+
   assert(!global_running);
   global_running = true;
 
@@ -483,6 +491,9 @@ Shutdown()
   // Log shutdown information
   LogFormat("Entering shutdown...");
 
+#if defined(_WIN32) && defined(USE_GDI)
+  GdiShutdown();
+#endif
   main_window->BeginShutdown();
 
   StartupLogFreeRamAndStorage();
--- a/src/Weather/PCMet/Images.cpp
+++ b/src/Weather/PCMet/Images.cpp
@@ -37,8 +37,7 @@ Copyright_License {
 #include <string.h>
 #include <stdio.h>
 
-//#define PCMET_URI "https://www.flugwetter.de"
-#define PCMET_URI "http://www.flugwetter.de"
+#define PCMET_URI "https://www.flugwetter.de"
 
 static constexpr PCMet::ImageArea rad_lokal_areas[] = {
   { "pro", _T("Prötzel") },
@@ -84,10 +83,21 @@ static constexpr PCMet::ImageArea rad_areas[] = {
 };
 
 static constexpr PCMet::ImageArea sat_areas[] = {
-  { "ceu_hrv", _T("Mitteleuropa HRV") },
-  { "ceu_rgb", _T("Mitteleuropa RGB") },
-  { "dlnw_hrv", _T("Deutschland Nordwest HRV") },
-  { "dlnw_rgb", _T("Deutschland Nordwest RGB") },
+  { "vis_hrv_eu", _T("Mitteleuropa HRV") },
+  { "ir_rgb_eu", _T("Mitteleuropa RGB") },
+  { "ir_108_eu", _T("Mitteleuropa IR") },
+  { "vis_hrv_ce", _T("Mitteleuropa HRV") },
+  { "ir_rgb_ce", _T("Mitteleuropa RGB") },
+  { "ir_108_ce", _T("Mitteleuropa IR") },
+  { "vis_hrv_mdl", _T("Deutschland HRV") },
+  { "ir_rgb_mdl", _T("Deutschland RGB") },
+  { "ir_108_mdl", _T("Deutschland IR") },
+  { "vis_hrv_ndl", _T("Deutschland Nord HRV") },
+  { "ir_rgb_ndl", _T("Deutschland Nord RGB") },
+  { "ir_108_ndl", _T("Deutschland Nord IR") },
+  { "vis_hrv_sdl", _T("Deutschland Süd HRV") },
+  { "ir_rgb_sdl", _T("Deutschland Süd RGB") },
+  { "ir_108_sdl", _T("Deutschland Süd IR") },
   { nullptr, nullptr }
 };
 
@@ -147,6 +157,8 @@ PCMet::DownloadLatestImage(const char *type, const char *area,
                                          UTF8ToWideConverter(name));
 
   if (!File::Exists(path)) {
+    // URI for a single page of a selected 'Satellitenbilder"-page with link
+    // to the latest image and the namelist array of all stored images
     snprintf(url, sizeof(url), PCMET_URI "%s", src);
 
     Net::DownloadToFileJob job2(session, url, path);
@@ -154,6 +166,7 @@ PCMet::DownloadLatestImage(const char *type, const char *area,
     if (!runner.Run(job2))
       return Bitmap();
   }
+  assert(File::Exists(path));
 
   Bitmap bitmap;
   try {
--- a/src/ui/canvas/gdi/Bitmap.cpp
+++ b/src/ui/canvas/gdi/Bitmap.cpp
@@ -22,6 +22,8 @@ Copyright_License {
 */
 
 #include "ui/canvas/Bitmap.hpp"
+#include "ui/canvas/gdi/GdiPlusBitmap.hpp"
+
 #include "Screen/Debug.hpp"
 #include "system/Path.hpp"
 
@@ -36,7 +38,11 @@ Bitmap::Bitmap(Bitmap &&src)
 bool
 Bitmap::LoadFile(Path path)
 {
-  return false;
+  bitmap = GdiLoadImage(path.c_str());
+  bool success = IsDefined();
+  assert (success);
+
+  return success;
 }
 
 void
--- a/src/ui/canvas/gdi/GdiPlusBitmap.cpp
+++ b/src/ui/canvas/gdi/GdiPlusBitmap.cpp
@@ -1,0 +1,69 @@
+/*
+Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2021 The XCSoar Project
+  A detailed list of copyright holders can be found in the file "AUTHORS".
+
+  This program is free software; you can redistribute it and/or
+  modify it under the terms of the GNU General Public License
+  as published by the Free Software Foundation; either version 2
+  of the License, or (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+
+#include "GdiPlusBitmap.hpp"
+
+#if defined(_MSC_VER)
+# include <algorithm>
+using std::min;  // to avoid the missing 'min' in the gdiplush headers
+using std::max;  // to avoid the missing 'max' in the gdiplush headers
+#endif  // _MSC_VER
+
+#include <assert.h>
+#include <unknwn.h>
+#include <gdiplus.h>
+
+static ULONG_PTR gdiplusToken;
+
+//----------------------------------------------------------------------------
+void
+GdiStartup()
+{
+  Gdiplus::GdiplusStartupInput gdiplusStartupInput;
+  Gdiplus::GdiplusStartup(&gdiplusToken, &gdiplusStartupInput, NULL);
+}
+
+//----------------------------------------------------------------------------
+void
+GdiShutdown()
+{
+  Gdiplus::GdiplusShutdown(gdiplusToken);
+}
+
+//----------------------------------------------------------------------------
+// can load: BMP, GIF, JPEG, PNG, TIFF, Exif, WMF, and EMF
+HBITMAP
+GdiLoadImage(const TCHAR* filename)
+{
+  HBITMAP result = nullptr;
+#ifdef _UNICODE  // TCHAR has to be WCHAR in GdiPlus
+  Gdiplus::Bitmap bitmap(filename, false);
+  Gdiplus::Color color = Gdiplus::Color::White;
+#ifndef NDEBUG
+  Gdiplus::Status status =
+#endif
+    bitmap.GetHBITMAP(color, &result);
+  assert(status == Gdiplus::Status::Ok);
+#endif  // _UNICODE
+  return result;
+}
--- a/src/ui/canvas/gdi/GdiPlusBitmap.hpp
+++ b/src/ui/canvas/gdi/GdiPlusBitmap.hpp
@@ -1,0 +1,34 @@
+/*
+Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2021 The XCSoar Project
+  A detailed list of copyright holders can be found in the file "AUTHORS".
+
+  This program is free software; you can redistribute it and/or
+  modify it under the terms of the GNU General Public License
+  as published by the Free Software Foundation; either version 2
+  of the License, or (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+
+#ifndef XCSOAR_UI_CANVAS_GDI_GDIBITMAP_HPP
+#define XCSOAR_UI_CANVAS_GDI_GDIBITMAP_HPP
+
+#if defined(_WIN32) && defined(USE_GDI)
+#include <windows.h>
+HBITMAP GdiLoadImage(const TCHAR* filename);
+void GdiStartup();
+void GdiShutdown();
+#endif  // _WIN32 && USE_GDI
+
+#endif  // XCSOAR_UI_CANVAS_GDI_GDIBITMAP_HPP
--- a/build/screen.mk
+++ b/build/screen.mk
@@ -200,6 +200,7 @@ SCREEN_SOURCES += \
 	$(CANVAS_SRC_DIR)/gdi/Pen.cpp \
 	$(CANVAS_SRC_DIR)/gdi/Brush.cpp \
 	$(CANVAS_SRC_DIR)/gdi/Bitmap.cpp \
+	$(CANVAS_SRC_DIR)/gdi/GdiPlusBitmap.cpp \
 	$(CANVAS_SRC_DIR)/gdi/ResourceBitmap.cpp \
 	$(CANVAS_SRC_DIR)/gdi/RawBitmap.cpp \
 	$(CANVAS_SRC_DIR)/gdi/Canvas.cpp \
@@ -207,7 +208,7 @@ SCREEN_SOURCES += \
 	$(CANVAS_SRC_DIR)/gdi/PaintCanvas.cpp
 GDI_CPPFLAGS = -DUSE_GDI
 GDI_CPPFLAGS += -DUSE_WINUSER
-GDI_LDLIBS = -luser32 -lgdi32 -lmsimg32
+GDI_LDLIBS = -luser32 -lgdi32 -lmsimg32 -lgdiplus
 
 ifeq ($(TARGET),PC)
 GDI_LDLIBS += -Wl,-subsystem,windows


src/Dialogs/Airspace/dlgAirspaceWarnings.cpp
```diff
@@ -330,6 +330,7 @@ AirspaceWarningListWidget::OnPaintItem(Canvas &canvas,
   auto [text_rc, altitude_rc] =
     text_altitude_rc.VerticalSplit(text_altitude_rc.right - (padding + altitude_width));
   text_rc.right -= padding;
+  altitude_rc.bottom = altitude_rc.bottom;  // unused variable?
```
MaxKellermann, line 333: Doesn't belong in this commit

src/Weather/PCMet/Images.cpp
```diff
@@ -101,8 +111,7 @@ const PCMet::ImageType PCMet::image_types[] = {
 Bitmap
 PCMet::DownloadLatestImage(const char *type, const char *area,
                            const PCMetSettings &settings,
-                           JobRunner &runner)
-{
+                           JobRunner &runner) {
```
MaxKellermann: Doesn't belong in this commit

src/Weather/PCMet/Images.cpp
```diff
@@ -154,7 +163,6 @@ PCMet::DownloadLatestImage(const char *type, const char *area,
     if (!runner.Run(job2))
       return Bitmap();
   }
-
```
MaxKellermann, line 157: Why?

src/ui/canvas/Bitmap.hpp
```diff
@@ -106,7 +108,8 @@ class Bitmap final
 #elif defined(USE_MEMORY_CANVAS)
   WritableImageBuffer<BitmapPixelTraits> buffer = WritableImageBuffer<BitmapPixelTraits>::Empty();
 #else
-  HBITMAP bitmap = nullptr;
+  static std::unordered_map<HBITMAP, size_t> bitmap_map;
```
MaxKellermann, line 111: Hidden global variable, and it's not thread-safe. This will crash XCSoar.

src/ui/canvas/gdi/Bitmap.cpp
```diff
@@ -22,36 +22,64 @@ Copyright_License {
...
+
 }
 
 bool
 Bitmap::LoadFile(Path path)
 {
-  return false;
+  hbitmap = GdiLoadImage(path.c_str());
+  if (!IsDefined()) {
+    LogFormat(_T("Bitmap couldn't load:  %d!"), GetLastError());
```
MaxKellermann, line 54: Throw exception instead of logging

src/Startup.cpp
```diff
@@ -140,13 +144,14 @@ AfterStartup()
   } catch (...) {
       LogError(std::current_exception());
   }
-
+  LogFormat("next: processGlideComputer");
```
MaxKellermann, line 147: Useless log message

src/Startup.cpp
```diff
@@ -140,13 +144,14 @@ AfterStartup()
...
   if (is_simulator()) {
     InputEvents::processGlideComputer(GCE_STARTUP_SIMULATOR);
   } else {
     InputEvents::processGlideComputer(GCE_STARTUP_REAL);
   }
 
+  LogFormat("next: LoadDefaultTask");
```
MaxKellermann, line 154: Useless log message

src/Startup.cpp
```diff
@@ -159,11 +164,15 @@ AfterStartup()
     protected_task_manager->TaskCommit(*defaultTask);
   }
 
+  LogFormat("next: TaskManager->Resume");
```
MaxKellermann, line 167: Useless log message

src/Startup.cpp
```diff
@@ -159,11 +164,15 @@ AfterStartup()
...
   task_manager->Resume();
 
+  LogFormat("next: InfoBoxManager::SetDirty()");
```
MaxKellermann, line 170: Useless log message

src/Startup.cpp
```diff
@@ -159,11 +164,15 @@ AfterStartup()
...
   InfoBoxManager::SetDirty();
 
+  LogFormat("next: ForceCalculation");
```
MaxKellermann, line 173: Useless log message

src/Startup.cpp
```diff
@@ -159,11 +164,15 @@ AfterStartup()
...
   ForceCalculation();
+  LogFormat("end: AfterStartup");
```
MaxKellermann, line 175: Useless log message

src/Startup.cpp
```diff
@@ -465,7 +478,8 @@ Startup()
   operation.Hide();
 
   main_window->FinishStartup();
-
+  
+  LogFormat("end: Startup");
```
MaxKellermann, line 482: Useless log message

src/ui/canvas/gdi/Bitmap.cpp
```diff
@@ -26,7 +26,9 @@ Copyright_License {
 
 #include "Screen/Debug.hpp"
 #include "system/Path.hpp"
+#include "LogFile.hpp"
 
+// #include <cassert>
```
MaxKellermann: Why?

src/ui/canvas/gdi/Bitmap.cpp
```diff
@@ -26,7 +26,9 @@ Copyright_License {
 
 #include "Screen/Debug.hpp"
 #include "system/Path.hpp"
+#include "LogFile.hpp"
```
MaxKellermann, line 29: Why?

src/ui/canvas/gdi/GdiPlusBitmap.cpp
```diff
@@ -0,0 +1,72 @@
...
+// can load: BMP, GIF, JPEG, PNG, TIFF, Exif, WMF, and EMF
+HBITMAP GdiLoadImage(const TCHAR* filename)
+{
+  HBITMAP result =nullptr;
+  Bitmap bitmap(filename, false);
+  bitmap.GetHBITMAP(0, &result);
+  return result;
+}
+
+#if 0
```
MaxKellermann, line 46: Huh?

src/ui/canvas/gdi/GdiPlusBitmap.hpp
```diff
@@ -0,0 +1,30 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+// #define min std::min
+// #define max std::max
```
MaxKellermann, line 24: Huh?

src/ui/canvas/gdi/GdiPlusBitmap.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include <windows.h>
+// namespace Gdi {
```
MaxKellermann, line 26: Huh?

src/net/http/Request.cpp
```diff
@@ -41,6 +41,11 @@ Net::Request::Request(Session &_session, ResponseHandler &_handler,
   handle.SetUserAgent(user_agent);
   handle.SetFailOnError(true);
   handle.SetWriteFunction(WriteCallback, this);
+#if defined(_WIN32)
+  // otherwise under Windows it is no access to https urls per SSL
+  // maybe the better option is to adjust the openssl component!
```
MaxKellermann, line 46: No, we should fix the underlying issue instead of doing such a kludge.

MaxKellermann (changes requested):

src/Dialogs/Airspace/dlgAirspaceWarnings.cpp
August2111: Ok, I would move it in an extra commit. Is it possible to do this in this PR?

src/Weather/PCMet/Images.cpp
August2111: I will remove this!

src/ui/canvas/Bitmap.hpp
August2111: This I don't like
What do you prefere? An use counter? I don't like the member setting to null in the source at a object copy, see line 32
Bitmap::Bitmap(Bitmap &&src)
  :bitmap(src.bitmap)
{
  src.bitmap = nullptr;
}
This means, after a copy you cannot access to the source object?

src/ui/canvas/gdi/Bitmap.cpp
August2111: Because LogFormat lines, - if they not necessary, this can be removed!

src/ui/canvas/gdi/GdiPlusBitmap.cpp
August2111: :( my mistake

src/ui/canvas/gdi/GdiPlusBitmap.hpp
August2111: :( my mistake too

src/net/http/Request.cpp
August2111: This needs much more investigations, I thought, a workaround is better as a complete failure: In the moment under Windows XCSoar has NO access to any https page yet

August2111:
I will resolve the issues and I want to make an update of the PR as soon as possible!

src/ui/canvas/Bitmap.hpp
MaxKellermann: This is not a copy constructor, it's a move constructor, and it's supposed to make the moved-from object invalid.
Why a counter?

src/Dialogs/Airspace/dlgAirspaceWarnings.cpp
MaxKellermann: Yes.

src/Weather/PCMet/Images.cpp
```diff
@@ -99,10 +109,11 @@ const PCMet::ImageType PCMet::image_types[] = {
 };
 
 Bitmap
-PCMet::DownloadLatestImage(const char *type, const char *area,
-                           const PCMetSettings &settings,
-                           JobRunner &runner)
+PCMet::DownloadLatestImage(const char* type, const char* area,
+  const PCMetSettings& settings,
+  JobRunner& runner)
```
MaxKellermann, line 114: Don't reformat code, unless you submit a commit which does *only* that.

src/Weather/PCMet/Images.cpp
```diff
@@ -99,10 +109,11 @@ const PCMet::ImageType PCMet::image_types[] = {
...
 {
+  Bitmap bitmap;
```
MaxKellermann, line 116: Why?

src/Weather/PCMet/Images.cpp
```diff
@@ -118,44 +129,46 @@ PCMet::DownloadLatestImage(const char *type, const char *area,
...
-  char *img = strstr(buffer, "<img name=\"bild\" src=\"/");
+  char* img = strstr(buffer, "<img name=\"bild\" src=\"/");
   if (img == nullptr)
-    return Bitmap();
+    return bitmap;
 
-  char *src = img + sizeof(img_needle) - 2;
-  char *end = strchr(src + 1, '"');
+  char* src = img + sizeof(img_needle) - 2;
+  char* end = strchr(src + 1, '"');
```
MaxKellermann, line 142: More code reformatting

src/Dialogs/Weather/PCMetDialog.cpp
```diff
@@ -157,6 +156,7 @@ class ImageTypeListWidget final : public TextListWidget {
   }
 };
 
+//-----------------------------------------------------------------------------
```
MaxKellermann, line 159: ??

src/Dialogs/Weather/PCMetDialog.cpp
```diff
@@ -44,11 +44,10 @@ Copyright_License {
...
 {
-  TWidgetDialog<ViewImageWidget>
-    dialog(WidgetDialog::Full{}, UIGlobals::GetMainWindow(),
-           UIGlobals::GetDialogLook(),
-           _T("pc_met"));
-  dialog.SetWidget();
+  ViewImageWidget widget(bitmap);
+  WidgetDialog dialog(WidgetDialog::Full{}, UIGlobals::GetMainWindow(),
+                      UIGlobals::GetDialogLook(),
+                      _T("pc_met"), &widget);
```
MaxKellermann, line 50: Why? Your commit message doesn't explain.

src/Dialogs/Airspace/dlgAirspaceWarnings.cpp
```diff
@@ -330,6 +330,7 @@ AirspaceWarningListWidget::OnPaintItem(Canvas &canvas,
   auto [text_rc, altitude_rc] =
     text_altitude_rc.VerticalSplit(text_altitude_rc.right - (padding + altitude_width));
   text_rc.right -= padding;
+  altitude_rc.bottom = altitude_rc.bottom;  // unused variable
```
MaxKellermann, line 333: This isn't an unused variable. `altitude_rc` is a variable, but `altitude_rc.bottom` is not.

August2111: My compiler says that altitude_rc is an unused variable, and so i made this fix ... :-( 
Or should I only change the comment?

MaxKellermann: See, it's about `altitude_rc`, not about `altitude_rc.bottom`.
If the variable is unused, remove it instead of applying a workaround.

August2111:
Excuse me I'm not complete ready yet with my code - tomorrow I want to make it really clean, f.e. also test the previous dialog with  TWidgetDialog too... Maybe I copied this from an other state or file ...
I want to compare every line with the repo - and than I hope you are happy with .

The part of curl I want to exclude from this PR: This has nothing to do with PC-Met - and have an impact of the access of all https sites under Windows. 

I have a solution for this - not a kludge - but probably the curl component itself needs an other building setup on mingw ;-(
In the meanwhile with MSVC I have the communication to SSL-Sites running WITH PeerVerification, but the setup for mingw isn't ready yet so far. Hopefully on monday latest I will have start a PR with this!

src/ui/canvas/Bitmap.hpp
August2111: Unfortunately I didn't know the move constructor ;-( I learned, this is defined in C++11. With such a functional possibility their is no ref counter necessary, of course...

src/Dialogs/Weather/PCMetDialog.cpp
August2111: Changed back

src/Weather/PCMet/Images.cpp
August2111: I have to avoid the changes by my editor!

src/Dialogs/Airspace/dlgAirspaceWarnings.cpp
August2111: Done

August2111:
Now I'm ready with my review yet!
Hopefully it is ok now.
Thanks for your hints, Max

src/Weather/PCMet/Images.cpp
MaxKellermann: This can crash XCSoar due to TOCTOU bug.

src/Weather/PCMet/Images.cpp
MaxKellermann: > Creates Bitmap() once - and return with this instead of creation on every return place

Again: why? This whole commit seems rather pointless.

src/ui/canvas/gdi/GdiPlusBitmap.hpp
```diff
@@ -0,0 +1,27 @@
...
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+
```
MaxKellermann, line 23: Include guard is missing

src/ui/canvas/gdi/GdiPlusBitmap.hpp
```diff
@@ -0,0 +1,27 @@
...
+#include <windows.h>
+extern HBITMAP GdiLoadImage(const TCHAR* filename);
+extern void GdiStartup();
+extern void GdiShutdown();
```
MaxKellermann, line 27: `extern` not necessary in C++

src/Startup.cpp
```diff
@@ -457,6 +461,10 @@ Startup()
 #endif
 #endif
 
+#ifdef USE_GDI
+  GdiStartup();
```
MaxKellermann, line 465: This needs to be called in all programs using the UI toolkit library - i.e. all those programs in `test/src/`. You need a better approach for this, e.g. in `class ScreenGlobalInit` which is used by all relevant programs.

src/ui/canvas/gdi/GdiPlusBitmap.cpp
```diff
@@ -0,0 +1,58 @@
...
+}
+*/
+
+#include "GdiPlusBitmap.hpp"
+
+#if defined(_MSC_VER)
+# include <algorithm>
+# define min std::min
+# define max std::max
+#endif  // _MSC_VER
```
MaxKellermann, line 30: Why?

src/ui/canvas/gdi/GdiPlusBitmap.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <gdiplus.h>
+
+using namespace Gdiplus;
+
+// can load: BMP, GIF, JPEG, PNG, TIFF, Exif, WMF, and EMF
+HBITMAP GdiLoadImage(const TCHAR* filename)
+{
+  HBITMAP result =nullptr;
+  Bitmap bitmap(filename, false);
+  bitmap.GetHBITMAP(0, &result);
```
MaxKellermann, line 42: Does this even compile?
https://docs.microsoft.com/en-us/windows/win32/api/gdiplusheaders/nf-gdiplusheaders-bitmap-gethbitmap
> Status GetHBITMAP(
>  const Color & colorBackground,
>  HBITMAP       *hbmReturn
>);

How can you pass `0` for the `const Color&`?

src/ui/canvas/gdi/GdiPlusBitmap.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <unknwn.h>
+#include <gdiplus.h>
+
+using namespace Gdiplus;
+
+// can load: BMP, GIF, JPEG, PNG, TIFF, Exif, WMF, and EMF
+HBITMAP GdiLoadImage(const TCHAR* filename)
+{
+  HBITMAP result =nullptr;
+  Bitmap bitmap(filename, false);
```
MaxKellermann, line 41: Which constructor is that? I can't find it at https://docs.microsoft.com/en-us/windows/win32/api/gdiplusheaders/nl-gdiplusheaders-bitmap

src/ui/canvas/gdi/Bitmap.cpp
```diff
@@ -45,10 +51,7 @@ Bitmap::Reset()
   if (bitmap != nullptr) {
     assert(IsScreenInitialized());
 
-#ifndef NDEBUG
-    bool success =
-#endif
-      ::DeleteObject(bitmap);
+    bool success = ::DeleteObject(bitmap);
```
MaxKellermann, line 54: Why?

MaxKellermann (changes requested):

src/ui/canvas/gdi/GdiPlusBitmap.hpp
August2111: I know, but for me it looks better...
I will remove it!

src/ui/canvas/gdi/GdiPlusBitmap.hpp
August2111: Ok, of course...

src/ui/canvas/gdi/Bitmap.cpp
August2111: With a defined NDEBUG you have an assert test with an undefined variable in the next line:
assert(success)
This makes no sense for me - and the macro switch has to be removed - or the assert have to chambered with  the same definition!

MaxKellermann: With `NDEBUG`, all `assert()` calls are removed.

August2111: Ok, I found this in the assert.h yet. 
So 'success' is an unused variable without NDEBUG ;-( 
Without this backround knowledge it is diffecult to read (and understand).
As I said: I'm very surprised how much new things I have to learn, before I can follow your style and intentions completely.

MaxKellermann: `assert()` exists since at least C89, i.e. 32 years ago.

MaxKellermann:
Please rebase before you push again to your PR branch; I have already cherry-picked several of your commits, and rebasing will drop them automatically from the PR branch.

src/Weather/PCMet/Images.cpp
MaxKellermann: Still problem still exists

August2111:
How can I mark the 'requested changes' as ready? Only with the push to my origin? Or it is possible to block your visiting by another way? Before you make a deeper look inside I want to do a review by myself - but you was faster the me...
Before I would set a 'ready' flag I want to do the rebasing :) (Also a workflow that is complete new for me)

src/Weather/PCMet/Images.cpp
August2111: I didn't know anything about this bug - I didn't changed here the old code. But I will look for this Bug - and hopefully I can solve!

MaxKellermann: You added an `assert()`.

August2111: 'Time-of-Check-to-Time-of-Use-Problem' - ok, I will check it

August2111:
The last push was wrong, I have to try to setup it again ;-(

src/Weather/PCMet/Images.cpp
August2111: Ok, I changed it. But why? I had learned that an assert is operative only in Debug mode (With NDEBUG its only 0)?

src/ui/canvas/gdi/GdiPlusBitmap.cpp
August2111: Inside the gdiplus headers the 'min' and 'max' are using very often, but with the define NOMINMAX they are not available. So you have define it explicitly. The recommandations found in stackoverflow (https://stackoverflow.com/questions/4913922/possible-problems-with-nominmax-on-visual-c) shows exactly this problem, but they are using the 'using' directive, not the #define - and so I changed it in this way.

August2111:
From my point of view I have done it - hopefully in the correct way...
I had also prepared a PR with a change of the curl component - but in your last commits I found the whole changes that I would proposed... It's poosible to use the openssl but here you need an additional certification file - and this also like a kludge for me. The use uf SCHANNEL ( = the native WinSSL) looks for me much better :) So I can delete my 'curl' branch - I think all is fine yet!

src/Weather/PCMet/Images.cpp
MaxKellermann: True, and it will crash only debug builds, but it's still a debug.

MaxKellermann:
I merged this manually, with several bug fixes.

src/Startup.cpp
August2111: Ok, I forgot this move to  class ScreenGlobalInit  - your way is much better!

PR #541, opened 2021-03-09
Fix adjust to previous turnpoint in a task

bomilkar:
Brief summary of the changes
----------------------------
When manually adjusting to the previous turn point it must forget that it had visited that turn point before. If it finds this TP had been visited before it auto-increments to the next TP.

Issue #492 

This was previously addressed in PR #528 .
I misread Max' comment and started with a new PR. But when Max told me to fix the old PR it was already too late and I had delete the old branch. Sorry for this.

### DIFF
--- a/src/Task/ProtectedTaskManager.cpp
+++ b/src/Task/ProtectedTaskManager.cpp
@@ -25,6 +25,7 @@ Copyright_License {
 #include "Task/RoutePlannerGlue.hpp"
 #include "Engine/Task/TaskManager.hpp"
 #include "Engine/Task/Ordered/OrderedTask.hpp"
+#include "Engine/Task/Ordered/Points/OrderedTaskPoint.hpp"
 #include "Engine/Task/Points/TaskWaypoint.hpp"
 #include "Engine/Route/ReachResult.hpp"
 
@@ -84,11 +85,13 @@ ProtectedTaskManager::IncrementActiveTaskPointArm(int offset)
 {
   ExclusiveLease lease(*this);
   TaskAdvance &advance = lease->SetTaskAdvance();
+  OrderedTaskPoint *nextwp = nullptr;
 
   switch (advance.GetState()) {
   case TaskAdvance::MANUAL:
   case TaskAdvance::AUTO:
     lease->IncrementActiveTaskPoint(offset);
+    nextwp = dynamic_cast<OrderedTaskPoint *>(lease->GetActiveTaskPoint());
     break;
   case TaskAdvance::START_DISARMED:
   case TaskAdvance::TURN_DISARMED:
@@ -96,17 +99,22 @@ ProtectedTaskManager::IncrementActiveTaskPointArm(int offset)
       advance.SetArmed(true);
     } else {
       lease->IncrementActiveTaskPoint(offset);
+      nextwp = dynamic_cast<OrderedTaskPoint *>(lease->GetActiveTaskPoint());
     }
     break;
   case TaskAdvance::START_ARMED:
   case TaskAdvance::TURN_ARMED:
     if (offset>0) {
       lease->IncrementActiveTaskPoint(offset);
+      nextwp = dynamic_cast<OrderedTaskPoint *>(lease->GetActiveTaskPoint());
     } else {
       advance.SetArmed(false);
     }
     break;
   }
+
+  // forget that we have visited that waypoint already
+  if(nextwp && nextwp->HasEntered()) nextwp->Reset();
 }
 
 bool 


src/Task/ProtectedTaskManager.cpp
```diff
@@ -84,29 +85,36 @@ ProtectedTaskManager::IncrementActiveTaskPointArm(int offset)
...
     } else {
       lease->IncrementActiveTaskPoint(offset);
+      nextwp = (IntermediateTaskPoint *)lease->GetActiveTaskPoint();
     }
     break;
   case TaskAdvance::START_ARMED:
   case TaskAdvance::TURN_ARMED:
     if (offset>0) {
       lease->IncrementActiveTaskPoint(offset);
+      nextwp = (IntermediateTaskPoint *)lease->GetActiveTaskPoint();
```
MaxKellermann, line 109: How do you know that these casts are legal?

bomilkar: Diving through the src tree, lot of "grep" and paper.
Not a simple task and I hope I didn't make a mistake. I wish I had a static code analyzer, for instance to follow inheritance. But I don't. All I have (and know) is gdb.

MaxKellermann: So all your code diving made you confident that `GetActiveTaskPoint()` will only ever return `IntermediateTaskPoint` instances? And never a `FinishPoint`?

bomilkar: Good point. But it never moves past the FinishPoint.

`OrderedTaskPoint` might be more appropriate.
I need to think about it.

bomilkar: Here is the result of my thinking, @MaxKellermann :

Prerequisite:

1. `class ScoredTaskPoint` has the methods I need: `Reset()` and `HasEntered()` .

2. `class OrderedTaskPoint` inherits from `ScoredTaskPoint` and `TaskWaypoint` , and from others.
```
class OrderedTaskPoint
  :public TaskLeg,
   public TaskWaypoint,
   public ScoredTaskPoint,
   public ObservationZoneClient
{ .....
```

3. In `IncrementActiveTaskPointArm()` I have a `TaskWaypoint*` which points to the object I want to work on.

**Question**: Can I simply type cast a `TaskWaypoint*` to `OrderedTaskPoint*` if I want access to the methods described in #1 ?
When I think about what a compiler can do then I guess the answer is YES.

MaxKellermann: You can always cast everything to everything. The compiler will "allow" everything. The compiler doesn't know if it's a good idea. Casts are however only a good idea if you really know the object is really an instance of the target type. This you need to make sure before you cast.
If you're dealing with an `OrderedTask`, all points will always be `OrderedTaskPoint`, which is the base class for all other point types in ordered tasks. Therefore, after you have ensured that this is an `OrderedTask`, casting to `OrderedTaskPoint` is legal.

MaxKellermann:
You still havn't made sure that the cast is legal. This will crash with all task types but OrderedTask.

bomilkar:
`GetActiveTaskPoint()` is a member of class `OrderedTask` and it returns a `TaskWaypoint*` .
If that doesn't do it, then I don't know.

MaxKellermann:
> `GetActiveTaskPoint()` is a member of class `OrderedTask` and it returns a `TaskWaypoint*` .
> If that doesn't do it, then I don't know.

No, `GetActiveTaskPoint()` is a pure virtual method defined in `class TaskInterface`:
https://github.com/XCSoar/XCSoar/blob/2b1a9202c015023fff55b476adb112650873fdba/src/Engine/Task/TaskInterface.hpp#L75
That's why it returns `TaskWaypoint` and not `OrderedTaskPoint`.
`OrderedTask` implements this pure virtual method, but it cannot change the return type. And you didn't even call that method - you called `TaskManager::GetActiveTaskPoint()`:
https://github.com/XCSoar/XCSoar/blob/2b1a9202c015023fff55b476adb112650873fdba/src/Engine/Task/TaskManager.hpp#L126
implemented here:
https://github.com/XCSoar/XCSoar/blob/2b1a9202c015023fff55b476adb112650873fdba/src/Engine/Task/TaskManager.cpp#L135-L139
... into this field:
https://github.com/XCSoar/XCSoar/blob/2b1a9202c015023fff55b476adb112650873fdba/src/Engine/Task/TaskManager.hpp#L71
... which is not an `OrderedTask`! Maybe it is, maybe not. It is the active task.

bomilkar:
I give up. This is beyond my C++ skills.

MaxKellermann:
"I give up" is not a valid matching sentence for "This is beyond my C++ skills". A proper sentence for that is "I'll learn this and extend my C++ skills".

Hint: the easiest trick is `dynamic_cast`:
```
nextwp = dynamic_cast<OrderedTaskPoint *>(lease->GetActiveTaskPoint());
```
This verifies the object type at runtime, and results in `nullptr` if the type does not match. See https://en.cppreference.com/w/cpp/language/dynamic_cast

Or check the task type before casting (we used this method years ago when we compiled XCSoar without RTTI, which is necessary for `dynamic_cast`):
```
  TaskWaypoint *nextwp = nullptr;
  [...]
  if (lease->IsMode(TaskType::ORDERED))
     // now the cast is legal
     ((OrderedTaskPoint *)nextwp)->Reset();
```

bomilkar:
I agree: I'll learn this and extend my C++ skills
It'll take some time, though. And thanks for mentioning RTTI. Now I know what to search for.

MaxKellermann:
Next time, please rebase every time you push to the PR branch!

PR #545, opened 2021-03-15
ACD driver fix for setting frequency and volume

Scumi:


MaxKellermann:
Next time, please write a better commit message. It should at least explain what was wrong.

Scumi:
Will do, thank you.

Issue #555, opened 2021-03-29
Alternate list (simple mode): priority not respected


francescoRubini:
XCSoar versions having and not having the problem
-------------------------------------------------
last version (6.8.17-Android / Windows)

Expected behavior
-----------------
When i check alternate list (is set to "simple" mode) the first must be a reachable airport, then airfield

Actual behavior
---------------
in simple mode the prioriry is not respected. (the first is an airfield but there is an reachable airport)

Steps to reproduce the behavior
-------------------------------

For example i'm on this situation:
![image](https://user-images.githubusercontent.com/6067265/112877622-4a0a1b00-90c7-11eb-9861-fee3e6b7116b.png)

The first alternate is Brivio (an Airfield) not the airport (kong)
![image](https://user-images.githubusercontent.com/6067265/112877721-63ab6280-90c7-11eb-87fd-df3953d0cd1d.png)

If I abort task then the alternates is correct:
![image](https://user-images.githubusercontent.com/6067265/112877810-7faf0400-90c7-11eb-9caf-a614b86a3259.png)

lordfolken:
I think this isn't sufficiently explained in the docs or help. The task abort mode will always prioritize an airport over an out landing site if able. The alternates list when not it task abort, will show you the nearest landing possibility. The main idea is to show good airman-ship and implement a risk averse approach.

lordfolken:
Would you mind enhancing the help text and/or documentation in the manual?

francescoRubini:
ok.
I understand the mistake. In the italian version of application there is a error in help.

Now I check the english version.

![image](https://user-images.githubusercontent.com/6067265/113691087-5fd19e80-96cc-11eb-84dc-1fb9289cf74f.png)
 (sorted by reachable airport, then airfield and nearest)

francescoRubini:
I opened a pull request to correct the description.

I change english and Italian (also on weblate).

But in weblate there is the old key (https://hosted.weblate.org/translate/xcsoar/translations/en/?offset=1&q=The+alternates+will+only&sort_by=-priority%2Cposition&checksum=). I can't change this.

I hope my pull request is correct.

PR #568, opened 2021-04-06
Correct the description in SafetyFactorsConfigPanel for simple mode alternates

francescoRubini:
Brief summary of the changes
----------------------------
On the configuration panel about safety factor on "alternates mode" input there are 3 options.
On first option (simple) the description was "The alternates will only be sorted by waypoint type (airport/outlanding field) and arrival height." but is not correct.
The correct is "The alternates will only be sorted by arrival height."

https://github.com/XCSoar/XCSoar/issues/555

### DIFF
--- a/po/bg.po
+++ b/po/bg.po
@@ -7206,8 +7206,7 @@ msgstr ""
 #: src/Dialogs/Settings/Panels/SafetyFactorsConfigPanel.cpp:80
 #, no-c-format
 msgid ""
-"The alternates will only be sorted by waypoint type (airport/outlanding "
-"field) and arrival height."
+"The alternates will only be sorted by arrival height."
 msgstr ""
 
 #: src/Dialogs/Settings/Panels/SafetyFactorsConfigPanel.cpp:82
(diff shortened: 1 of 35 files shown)


MaxKellermann:
You shouldn't update any translation file - those will be regenerated automatically with texts picked up from the source code.

francescoRubini:
> You shouldn't update any translation file - those will be regenerated automatically with texts picked up from the source code.

ok, and how I can do it?
I changed msgid because the english version of the sentence changed.
Consequently I had to change all the translation files using the new ID.

Can you explain to me how I should have done? Thanks and sorry

MaxKellermann:
All PO files are now updated. You can edit Italian translations here: https://hosted.weblate.org/projects/xcsoar/translations/it/

francescoRubini:
> All PO files are now updated. You can edit Italian translations here: https://hosted.weblate.org/projects/xcsoar/translations/it/

I translated into Italian, but the English version continues to be wrong. The correct is: "The alternates will only be sorted by arrival height.".
Please read this: https://github.com/XCSoar/XCSoar/issues/555

PR #570, opened 2021-04-06
Upgrade Provisioning script to Android SDK 29.0.3

groundhogxc:
Brief summary of the changes
----------------------------

Upgrade the Provisioning script to the currently used Android SDK 29.0.3 / Tools 29


MaxKellermann:
This PR removes lib/glm. Why?

groundhogxc:
That is because it seems I do not understand how the submodules work. That was of course a mistake. I will figure out how to change it. Sorry!

groundhogxc:
OK, so I researched this, and it seems history changes of already-pushed branches are a big "no-no". So according to what I understand, it seemed the cleanest way was to do:
* a full `git revert` commit of the entire previous commit
* then re-apply the change and commit that via `git commit --amend` in order to make this one single fix-up commit
I hope this was correct? If there is a better way / a specific way we do this here, please let me know.

I tried to do the revert itself as an --amend, but that creates an error on pushing due to the change in history; which I did not want to override.

MaxKellermann:
> OK, so I researched this, and it seems history changes of already-pushed branches are a big "no-no".

That's wrong. PR branches can be rewritten as you like, and it's a good idea to do so. Only official branches (which other people depend on) must never be rewritten.

>     * a full `git revert` commit of the entire previous commit

No. A `revert` creates a new commit. This would imply that I merge the wrong commit and another commit (the revert). But I don't want to merge any of those. I want to merge clean commits.

>     * then re-apply the change and commit that via `git commit --amend` in order to make this one single fix-up commit
>       I hope this was correct? If there is a better way / a specific way we do this here, please let me know.

Just `--amend` the original commit.
To manage patches, I use [stgit](https://stacked-git.github.io/guides/tutorial/). Try it. I can't live without it.

> I tried to do the revert itself as an --amend, but that creates an error on pushing due to the change in history; which I did not want to override.

Yes, override it. Do `git push --force` with PR branches if they were wrong.

groundhogxc:
> That's wrong. PR branches can be rewritten as you like, and it's a good idea to do so. Only official branches (which other people depend on) must never be rewritten.
Oh, I see then. So I made it more complicated than necessary. (So, this means PR branches are effectively always ever used by one person, and not meant to be pulled by others without them creating their own branch from it?)

> Yes, override it. Do `git push --force` with PR branches if they were wrong.
That makes things a lot simpler. Wilco.

MaxKellermann:
> So, this means PR branches are effectively always ever used by one person, and not meant to be pulled by others without them creating their own branch from it?

Not quite. The PR branch is meant to be pulled into the repository where you submitted the PR, after it has passed review. Obviously. That's the whole point of submitting a PR.

But other than the target repository where you want to have your work merged, it is not supposed to be used by anybody, and your branch should be deleted after it was merged.

groundhogxc:
Yeah, that's what I meant. Thanks for the coaching! A lot of pain for a single-character-commit! :-))) But never did a PR before...

Issue #480, opened 2020-09-24
Display relative altitude for traffic on map


DavisChappins:
XCSoar version
--------------

6.8.16

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------

Traffic drawn on the map (not the FLARM radar popup or the FLARM specific radar page) shows aircraft rate of climb, if positive, below the traffic target. Relative altitude should also be shown above the target. Relative altitude should be shown in hundreds of feet (or meters) with an up or down arrow preceeding the numeric field. It can be capped at 99 if using two digits or a third can be displayed for targets greater than 9,900 ft (or meters) difference in relative altitude. Both are acceptable.

This is an EFB industry standard way of displaying traffic and relative altitude is the most important traffic attribute to be displayed.
Currently I have to press on the traffic target to see its information to get it's altitude.

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------

Display of relative altitude above traffic targets should be shown on the map. This information is already gathered and shown on the FLARM radar page but since that maximum range is 2km you end up with bunched up targets overlapping information when looking for relative altitudes.

Sample image here: https://i.imgur.com/V4ZFkNb.jpg taken by me driving down the road, displaying ADS-B traffic. The two red circles indicate traffic targets that I have added relative altitude to.

For people who don't wish to see relative altitude on the map, this function could be enabled/disabled by a config inside Map Display > Elements > FLARM traffic. Currently FLARM traffic is selected to **on** or **off** . It could be modified to **on**, **on with relative**, **off**.

bomilkar:
I started a discussion on displaying relative altitude:
https://forum.xcsoar.org/viewtopic.php?f=2&t=3753

DavisChappins:
Your link talks about color coding based on relative altitude. This is different.

Every EFB and TCAS display lists the relative altitude of the target above (or sometimes below) the target icon. This simply brings XCSoar to the same display standard.

lordfolken:
I'd love a pr for this.

bomilkar:
I have a patch which does this. Will submit a PR in an hour or so.
Is there a consensus on what's a reasonable relative altitude difference inside which the warning color (orange) should be used? I'm using +/- 50m and it seems OK. But sometimes I think +/- 75m would be a better choice.
BTW: does anybody know whether FLARM reports the relative altitude of the traffic based on GPS or baro? How accurate is this?

PR #572, opened 2021-04-08
Color code relative altitude of FLARM traffic in map window

bomilkar:
Brief summary of the changes
----------------------------

Addresses Issue #480 

It color codes FLARM traffic symbols on the map (little triangles). Traffic 50m and deeper below is coded green. Traffic 50m and above is coded blue. Traffic in the +/-50m range is coded orange.
A FLARM alert overrides this color coding: the symbol is red (unchanged behavior).

### DIFF
--- a/src/Look/TrafficLook.cpp
+++ b/src/Look/TrafficLook.cpp
@@ -34,10 +34,12 @@ void
 TrafficLook::Initialise(const Font &_font)
 {
   safe_color = Color(0x1d,0x9b,0xc5);
+  safe_below_color = Color(0x1d,0xc5,0x10);
   warning_color = Color(0xfe,0x84,0x38);
   alarm_color = Color(0xfb,0x35,0x2f);
 
   safe_brush.Create(safe_color);
+  safe_below_brush.Create(safe_below_color);
   warning_brush.Create(warning_color);
   alarm_brush.Create(alarm_color);
 
--- a/src/Look/TrafficLook.hpp
+++ b/src/Look/TrafficLook.hpp
@@ -33,10 +33,12 @@ class Font;
 
 struct TrafficLook {
   Color safe_color;
+  Color safe_below_color;
   Color warning_color;
   Color alarm_color;
 
   Brush safe_brush;
+  Brush safe_below_brush;
   Brush warning_brush;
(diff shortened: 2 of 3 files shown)


MaxKellermann:
This PR contains text which is missing in the commit message.
You're using the `warning_brush` for traffic which is not warned about by the FLARM. I'm not sure if that is a good idea, because that color is reserved for warnings.
I'm not even sure if adding more colors is a good idea at all.

bomilkar:
I can think of two ways to signal relative altitude on the map:
1. color code the traffic symbols (little triangles) depending on the relative altitude
2. add numbers to the traffic symbols

I don't think adding more text to the busy map is a good idea.
Hence I vote for option 1.
ColorCoded_FLARMTraffic_5

MaxKellermann:
This PR does two changes in one commit - the green color and the orange color. The latter is questionable because it (ab)uses an existing color for a new meaning. The former is a useful feature, just the color is a matter of taste. But I can't merge one without the other. Do one thing in one commit!

bomilkar:
Let me restate the intent because I’m not sure I made myself clear.
I want to identify the FLARM targets in the same altitude range as me (+/- 50m). I’m using color to visualize this as opposed to using text. So I need 2 colors: color1 for in range, color2 for outside of altitude range.
As a cheap by-product I can differentiate _below range_ and _above range_. That splits color2 into two: color2_above and color2_below.

One can debate forever which real colors to pick. I’m suggesting the following:
color1: orange, same color as FLARM Warning color. As far as I can tell `warning_brush` (which uses that `warning_color`) is defined but never really used.
color2_above: blue as the sky, and the same color as it currently uses for FLARM traffic symbols
color2_below: green as the grass. That’s a newly defined color.
Unchanged remains the use of alarm_color.

bomilkar:
With the above in mind, I suggest I split this into 2 commits:

1.  identify the FLARM targets in the same altitude range as me.
I can use a different brush and color if we don’t want to (re)use the otherwise unused FLARM Warning color. But then: what color?

2. split the _outside_range_ in two: _below range_ and _above range_.
I guess here I can use green and blue.

Would that be OK, @MaxKellermann ?

Issue #569, opened 2021-04-06
Custom sprint trace time


Scumi:
The german DMSt now also hosts a sprint league. Unlike OLC, DMSt runs it with a 2 hour window instead of 2.5. 
The sprint trace in XCSoar (see https://github.com/XCSoar/XCSoar/blob/4fca62a5e2e131f1999837b2c439878e678498fd/src/Computer/TraceComputer.cpp#L45) has a fixed 9000 sec (=> 2.5 hours) input. I didnt find a good way to make this somewhat configurable to create a new contest solver for DMSt sprint league.

XCSoar version
--------------
7.3

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------
Make sprint trace configurable in some way so a contest can define a custom time

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------
Contest solvers to define their own sprint time.

PR #574, opened 2021-04-08
Engine/Contest/Solvers/Weglide*: WeGlide contests

Scumi:
Brief summary of the changes
----------------------------
Weglide contest solvers added.

loosely related: #569

### DIFF
--- a/src/Engine/Contest/ContestManager.cpp
+++ b/src/Engine/Contest/ContestManager.cpp
@@ -39,7 +39,11 @@ ContestManager::ContestManager(const Contest _contest,
    dhv_xc_free(trace_full, true),
    dhv_xc_triangle(trace_triangle, predict_triangle, true),
    sis_at(trace_full),
-   net_coupe(trace_full)
+   net_coupe(trace_full),
+   weglide_free(),
+   weglide_distance(trace_full),
+   weglide_fai(trace_triangle, predict_triangle),
+   weglide_or(trace_full)
 {
   Reset();
 }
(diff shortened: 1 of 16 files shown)


MaxKellermann:
Doesn't compile:
```
src/Renderer/FlightStatisticsRenderer.cpp:176:29: error: label at end of compound statement: expected statement
  case Contest::WEGLIDE_FAI:
```

The source code (or at least the commit message) should contain a link to the official WeGlide rules, so interested developers can look them up and verify your code.
Add a test to `test/src/RunOLCAnalysis.cpp` (which is misnamed these days) so developers can test this on existing IGC files without needing to run replay manually.

MaxKellermann:
Also: rebase every time right before you push a PR branch.

MaxKellermann:
Also: rebase every time right before you push a PR branch.

Scumi:
> test/src/RunOLCAnalysis.cpp (which is misnamed these days)

Yes, its pretty confusing seeing real OLC contests and other decentral ones intermixed when working the code. Ideas for a naming update? Happy to provide a PR.

src/Engine/Contest/ContestManager.cpp
```diff
@@ -39,7 +39,11 @@ ContestManager::ContestManager(const Contest _contest,
    dhv_xc_free(trace_full, true),
    dhv_xc_triangle(trace_triangle, predict_triangle, true),
    sis_at(trace_full),
-   net_coupe(trace_full)
+   net_coupe(trace_full),
+   weglide_free(),
```
MaxKellermann, line 43: Just a note: this kind of explicit constructor call is redundant and useless.
The other constructor calls should be done in-line where the fields are declared, but the other (existing) fields aren't initialised that way either; that's a chance code code modernisation.

MaxKellermann:
> Yes, its pretty confusing seeing real OLC contests and other decentral ones intermixed when working the code. Ideas for a naming update? Happy to provide a PR.

Many years ago, we called the engine "Contest", but that isn't a good name either, too generic. But it's a lot better than "OLC".

Naming stuff is something I sometimes procrastinate the most with, because it's more difficult than writing code. I don't know what's the best name for this feature. Scoring? Also too generic. Not sure. Maybe standardise on "Contest" for now, just to get rid of "OLC"?

PR #575, opened 2021-04-09
Contest renaming

Scumi:
Brief summary of the changes
----------------------------
When working with the code, it can be very confusing to separate functions and calculations that belong to the actual OLC and the ones for other decentral contests (like WeGlide or NetCoupe) that arised in the recent years. This commit renames all occurences of OLC or On-Line Contest if not directly linked to the OLC.

#574


src/Profile/ContestProfile.cpp
```diff
@@ -29,7 +29,8 @@ Copyright_License {
 void
 Profile::Load(const ProfileMap &map, ContestSettings &settings)
 {
-  if (map.GetEnum(ProfileKeys::OLCRules, settings.contest)) {
+  if (map.GetEnum(ProfileKeys::OLCRules, settings.contest) ||
```
MaxKellermann, line 32: This means that the old config always wins, i.e. changing the setting never has an effect.
But let's not rename the ProfileKey for now. This is error prone as you see; let's concentrate on renaming stuff in the code, things that can't break any code.

Scumi: I agree, I had some reluctance there as well. I just left a note in the ProfileKey now.

Issue #580, opened 2021-04-11
Broken link to actual version of openssl in build.py


fraktal-sb:
XCSoar versions having and not having the problem
-------------------------------------------------

OK  6.8.17
NOK  7.4

Expected behavior
-----------------

compile KOBO version should use the actual version of openssl

Actual behavior
---------------

compile error; missing link

![1](https://user-images.githubusercontent.com/69430537/114295324-dc85c380-9aa4-11eb-9447-6f18a2e7b07c.png)

Steps to reproduce the behavior
-------------------------------

make TARGET=KOBO TCSUFFIX=-8 DEBUG=n libs

Do you have any idea what may have caused this?
-----------------------------------------------

link in libs.py no mor available

![3](https://user-images.githubusercontent.com/69430537/114295269-7600a580-9aa4-11eb-8863-9a9a5ed0699d.png)

![2](https://user-images.githubusercontent.com/69430537/114295276-857fee80-9aa4-11eb-851f-9de66d752caf.png)

Do you have an idea how to solve the issue?
-------------------------------------------

use this one

![4](https://user-images.githubusercontent.com/69430537/114295291-97fa2800-9aa4-11eb-8e59-c69740974066.png)

lordfolken:
sadly the openssl3.0.0alpha14 is broken for android.
https://github.com/openssl/openssl/issues/14804
potential fix:
https://github.com/openssl/openssl/pull/14812

lordfolken:
https://github.com/XCSoar/XCSoar/commit/58bd0576101d9008925a89fcbb70839e30f6ee3b Updates the ssl lib to current 3.0.0alpha14. Still broken though.

MaxKellermann:
Works here.

CazYokoyama:
https://www.openssl.org/source/ has alpha15 and no longer has 14.

lordfolken:
fixed in 5743209f4db78eacfec0cfe51b2231cfae680387

PR #583, opened 2021-04-11
gh-workflow: build unix target for PRs/Push

lordfolken:
Compiling the UNIX target from our own defined environment. This would deprecate circle-ci.


lordfolken:
I also have android / win64 / pc targets ready. but will propose merge when #580 is solved.

PR #585, opened 2021-04-14
Color code relative altitude of FLARM traffic in map window

bomilkar:
New attempt to address Issue #480

FLARM traffic symbols are colored Magenta if the traffic is within +/- 50m of my FLARM's altitude.

Brief summary of the changes
----------------------------
I want to identify the FLARM targets in the same altitude range as me (+/- 50m). I’m using color to visualize this as opposed to using text. So I need 2 colors: color1 for in range, color2 for outside of altitude range.
For color1 I picked Magenta
Color2 remains unchanged: Blue

Later in an additional commit, I'd like to differentiate the "outside altitude range" in two: "below altitude range" and "above altitude range" each with its own color.

### DIFF
--- a/src/Look/TrafficLook.cpp
+++ b/src/Look/TrafficLook.cpp
@@ -35,10 +35,12 @@ TrafficLook::Initialise(const Font &_font)
 {
   safe_color = Color(0x1d,0x9b,0xc5);
   warning_color = Color(0xfe,0x84,0x38);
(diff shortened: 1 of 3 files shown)


bomilkar:
This is now split in 2 commits as requested in PR #572.

MaxKellermann:
Cherry-picked two commits because this PR branch is broken.

PR #586, opened 2021-04-14
enable SSL in curl for Android and Kobo

August2111:
Brief summary of the changes
----------------------------

With changes of commit 585e7e3384431 to enabling the WinSSL for Win64 and PC Targets unfortunately the curl link to openssl for Android  and Kobo toolchains was switched off!
Maybe this PR can help to bring the SSL in this curl libs to life.

Byside with the 2nd commit I changed the Android SDK setup for the docker build to the newest one regarding to the needs of the make run.

Maybe this can help to overcome the issue #580, but unfortunately I don't have a Kobo hardware to testing this.

### DIFF
--- a/ide/provisioning/install-android-tools.sh
+++ b/ide/provisioning/install-android-tools.sh
@@ -3,23 +3,28 @@
 set -e
 
 # set ANDROID variables
-ANDROID_SDK_TOOLS_VERSION=3859397
+ANDROID_SDK_TOOLS_VERSION=6858069_latest
 ANDROID_BUILD_TOOLS_VERSION=29.0.3
-ANDROID_PLATFORM_VERSION=26
+ANDROID_PLATFORM_VERSION=29
 ANDROID_NDK_VERSION=r22b
 ANDROID_REPO_URL=https://dl.google.com/android/repository
 ANDROID_SDK_DIR=~/opt/android-sdk-linux
 ANDROID_NDK_DIR=~/opt/android-ndk-${ANDROID_NDK_VERSION}
 
-if [ -d ${ANDROID_SDK_DIR} ]
+if [ -d ${ANDROID_SDK_DIR}\build-tools\${ANDROID_BUILD_TOOLS_VERSION} ]
 then
-  echo Not installing Android NDK, because ${ANDROID_SDK_DIR} exists already
+  echo Not installing Android SDK, because ${ANDROID_SDK_DIR}\build-tools\${ANDROID_BUILD_TOOLS_VERSION} exists already
 else
+  if [ -d ${ANDROID_SDK_DIR} ]
+  then
+    echo Remove old Android SDK  '${ANDROID_SDK_DIR}'...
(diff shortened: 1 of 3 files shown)


August2111:
Maybe I'm a little late, I saw this problem some days ago ;-( But I couldn't check it - and so I want to wait until I can it prove. Now I think this could be the reason for issue #580 - hopefully it helps yet!

lordfolken:
Hey!
Thank you for your contribution & Time. 
To get this merged a few points:
 * please make it 2 PRs as the issues are not fundamentally related.
 * change the commit text to: "lib/curl: enable openssl for non-windows targets"
 * add yourself to AUTHORS.txt
 * add an entry to NEWS.txt und the current non-deployed version.

MaxKellermann:
>     * please make it 2 PRs as the issues are not fundamentally related.

No, that's OK. Splitting commits is good, splitting PRs is just cumbersome extra work.

MaxKellermann:
What's the point of the first commit? OpenSSL is enabled by default, and this change is not necessary and doesn't do anything, does it?

ide/provisioning/install-android-tools.sh
```diff
@@ -32,10 +37,12 @@ else
 fi
 echo
 
-cd ~/opt/android-sdk-linux
+cd ${ANDROID_SDK_DIR}
 echo Installing Android SDK packages...
-tools/bin/sdkmanager "build-tools;${ANDROID_BUILD_TOOLS_VERSION}" \
-    "platforms;android-${ANDROID_PLATFORM_VERSION}"
+echo cmdline-tools/bin/sdkmanager --sdk_root=${ANDROID_SDK_DIR} \
+    "build-tools;${ANDROID_BUILD_TOOLS_VERSION}" "platforms;android-${ANDROID_PLATFORM_VERSION}"
```
MaxKellermann, line 43: This is rather pointless. Anybody who wishes to see the commands can use `-x`.

MaxKellermann:
I have no idea about the second commit. I don't know what this file does and therefore can't review the changes.

ide/provisioning/install-android-tools.sh
August2111: Sorry for that - I'm coming from windows - and didn't know the option  -x. Of course this seems to be much better (readable).

August2111:
1st Commit: I didn't find the default option for OpenSSL in the Curl cmake, but a deeper look in the CMakeLists.txt show this point:
```
set(openssl_default ON)
if(WIN32 OR CMAKE_USE_SECTRANSP OR CMAKE_USE_SCHANNEL OR CMAKE_USE_MBEDTLS OR CMAKE_USE_NSS OR  CMAKE_USE_WOLFSSL)
      set(openssl_default OFF)
endif()
option(CMAKE_USE_OPENSSL "Use OpenSSL code. Experimental" ${openssl_default})
```
So it looks like you explain - and so we can forgot this PR (The 2nd commit should be a switch to a newer android sdk only, but this isn't really necessary yet

PR #589, opened 2021-04-18
gh-workflow: other buildtargets

lordfolken:


MaxKellermann:
I cherry-picked WIN64 and KOBO; Android was already merged.
295854b1e8a862e9fe65ac446b6e83d6c725b399
I dropped CUBIE because that target requires a Cubieboard sysroot; and I don't think that's worth the trouble, because I consider the Cubieboard obsolete, after the Raspberry Pi has caught up.

DanD222:
> I cherry-picked WIN64 and KOBO; Android was already merged. [295854b](https://github.com/XCSoar/XCSoar/commit/295854b1e8a862e9fe65ac446b6e83d6c725b399) I dropped CUBIE because that target requires a Cubieboard sysroot; and I don't think that's worth the trouble, because I consider the Cubieboard obsolete, after the Raspberry Pi has caught up.

Max, even though there may be instances of individuals running OpenVario on RPi, as far as I’m aware Cubieboard is still the only hardware OpenVario is intended to be run on. Not sure what the implications are with regards to this PR; I just wanted to note that Cubieboard is probably not considered obsolete by the OpenVario project as of yet.

lordfolken:
This PR has currently no implication as the openvarios are supported by the openvario project. 

There is currently this issue preventing curent builds of xcsoar on openvario:
https://github.com/Openvario/meta-openvario/issues/138

I would also say that its too early to deprecate cubieboards as there are at least the 650+ that stelfy sold, and probably another 400 or so that people self built. However the platform itself isn't an xcsoar problem. As long as its some form of (current) linux, xcsoar will run.

MaxKellermann:
Those who wish to have Cubieboard support shall volunteer to become maintainer of the port, and for example help write a GH Actions workflow to do CI tests. Stuff like that doesn't complete by itself, there's always a need for people taking care of problems. Don't expect that existing developers of XCSoar will silently do it all for you. Never say "keep supporting the Cubieboard", instead say "I'll help ensure that XCSoar will always run perfectly on the Cubieboard".

PR #592, opened 2021-04-19
(Re-)Enabling MSVC

August2111:
Brief summary of the changes
----------------------------

All compile steps for XCSoar are done in Linux - with cross compiling for Windows, Android, Kobo,...
Because as a developer I'm coming from windows and mostly my work is on Visual Studio I prefare tto enable this platform for XCSoar too.
From my point of view ther are only some small steps to change, but without I cannot compile the project. 

Maybe there are some other guys with the same intention - I think with enabling MSVC the XCSoar project could attract additional volonteers ;-)

### DIFF
--- a/src/Geo/ConvexHull/GrahamScan.cpp
+++ b/src/Geo/ConvexHull/GrahamScan.cpp
@@ -22,6 +22,7 @@
 
 #include "GrahamScan.hpp"
 #include "Geo/SearchPointVector.hpp"
+#include <iterator>  // necessary for std::back_inserter
 
 static bool
 sortleft
--- a/src/LogFile.cpp
+++ b/src/LogFile.cpp
@@ -98,7 +98,7 @@ LogFormat(const char *fmt, ...) noexcept
   va_list ap;
 
   va_start(ap, fmt);
-  vsprintf(buf, fmt, ap);
+  vsnprintf(buf, sizeof(buf) - 1, fmt, ap);
   va_end(ap);
 
   LogString(buf);
--- a/src/Math/Trig.hpp
+++ b/src/Math/Trig.hpp
@@ -24,6 +24,10 @@ Copyright_License {
 #ifndef XCSOAR_MATH_TRIG_HPP
 #define XCSOAR_MATH_TRIG_HPP
 
+#if defined(_MSC_VER)
+# include "util/Compiler.h"
+#endif
+
 #include <utility>
 
 #include <math.h>
@@ -35,6 +39,9 @@ sin_cos(const double thetha) noexcept
   double s, c;
 #ifdef __APPLE__
   __sincos(thetha, &s, &c);
+#elif defined(_MSC_VER)  // STL and MSVC have no sincos...
+  s = sin(thetha);
+  c = cos(thetha);
 #else
   sincos(thetha, &s, &c);
 #endif
--- a/src/net/http/ToBuffer.cpp
+++ b/src/net/http/ToBuffer.cpp
@@ -89,6 +89,7 @@ public:
     memcpy(buffer + received, data.data, data.size);
     received += data.size;
 
+    buffer[received] = 0;  // end of (string) buffer!
     env.SetProgressRange(received);
   }
 
--- a/src/util/Compiler.h
+++ b/src/util/Compiler.h
@@ -21,8 +21,12 @@ Copyright_License {
 }
 */
 
-#ifndef COMPILER_H
-#define COMPILER_H
+#ifndef UTIL_COMPILER_H
+#define UTIL_COMPILER_H
+
+#ifdef _MSC_VER
+#   include "util/msvc"
+#else  // _MSC_VER
 
 #define GCC_MAKE_VERSION(major, minor, patchlevel) ((major) * 10000 + (minor) * 100 + patchlevel)
 
@@ -141,6 +145,8 @@ Copyright_License {
 
 #endif
 
+#endif  // _MSC_VER
+
 #if CLANG_OR_GCC_VERSION(4,3)
 
 #define gcc_hot __attribute__((hot))
@@ -195,4 +201,4 @@ Copyright_License {
 #define gcc_unreachable()
 #endif
 
-#endif
+#endif  // UTIL_COMPILER_H
--- a/src/util/TriState.hpp
+++ b/src/util/TriState.hpp
@@ -33,6 +33,7 @@
 #include <cstdint>
 
 #ifdef _WIN32
+#include <winsock2.h>
 #include <windows.h>
 #elif defined(__APPLE__)
 #import <Foundation/Foundation.h>
--- a/src/util/msvc.h
+++ b/src/util/msvc.h
@@ -1,0 +1,81 @@
+/*
+Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2021 The XCSoar Project
+  A detailed list of copyright holders can be found in the file "AUTHORS".
+
+  This program is free software; you can redistribute it and/or
+  modify it under the terms of the GNU General Public License
+  as published by the Free Software Foundation; either version 2
+  of the License, or (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+
+#ifndef UTIL_MSVC_COMPILER_H
+#define UTIL_MSVC_COMPILER_H
+
+#ifdef _MSC_VER
+#if __cplusplus >= 201700L  // with C++17 it's this deprecated
+                            // struct not available anymore
+#   include "util/xcs_functional.hpp"
+#endif
+#   include <io.h>
+#   include "corecrt_math_defines.h"
+
+#define STRING2(x) #x
+#define STRING(x) STRING2(x)
+
+#define CLANG_OR_GCC_VERSION(x, y)  0
+#define GCC_CHECK_VERSION(x, y)     0
+#define GCC_OLDER_THAN(x, y)        1  // This isn't a (new!) GCC
+#define CLANG_CHECK_VERSION(x, y)   0
+
+#define DT_UNDERLINE 0            // not avalable in WinUser.h!
+
+#define __attribute__(x)
+
+#define gcc_const
+#define gcc_deprecated
+#define gcc_may_alias
+#define gcc_malloc
+#define gcc_noreturn
+#define gcc_packed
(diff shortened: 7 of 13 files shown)


src/net/http/ToBuffer.cpp
```diff
@@ -89,6 +89,7 @@ class DownloadToBufferHandler final : public CurlResponseHandler {
     memcpy(buffer + received, data.data, data.size);
     received += data.size;
 
+    buffer[received] = 0;  // end of (string) buffer!
```
MaxKellermann, line 92: Why oh why? This API is not documented to null-terminate the buffer; it's not even about strings, it can hold arbitrary data, even null bytes.
And your code adds a buffer overflow bug. You're writing past the end of the buffer.

src/LogFile.cpp
```diff
@@ -98,7 +98,7 @@ LogFormat(const char *fmt, ...) noexcept
   va_list ap;
 
   va_start(ap, fmt);
-  vsprintf(buf, fmt, ap);
+  vsnprintf(buf, sizeof(buf) - 1, fmt, ap);
```
MaxKellermann, line 101: Off-by-one bug. Why `- 1`?

src/Math/Trig.hpp
```diff
@@ -24,6 +24,10 @@ Copyright_License {
 #ifndef XCSOAR_MATH_TRIG_HPP
 #define XCSOAR_MATH_TRIG_HPP
 
+#if defined(_MSC_VER)
+# include "util/Compiler.h"
+#endif
```
MaxKellermann, line 29: Why? What does this have to do with sincos or MSVC?

src/util/TriState.hpp
```diff
@@ -33,6 +33,7 @@
 #include <cstdint>
 
 #ifdef _WIN32
+#include <winsock2.h>
```
MaxKellermann, line 36: sincos???

src/util/Compiler.h
```diff
@@ -21,8 +21,12 @@ Copyright_License {
 }
 */
 
-#ifndef COMPILER_H
-#define COMPILER_H
+#ifndef UTIL_COMPILER_H
+#define UTIL_COMPILER_H
+
+#ifdef _MSC_VER
+#   include "util/msvc"
```
MaxKellermann, line 28: Missing `.h`

src/util/Compiler.h
```diff
@@ -21,8 +21,12 @@ Copyright_License {
 }
 */
 
-#ifndef COMPILER_H
-#define COMPILER_H
+#ifndef UTIL_COMPILER_H
+#define UTIL_COMPILER_H
```
MaxKellermann, line 25: Why???

src/util/msvc.h
```diff
@@ -0,0 +1,81 @@
...
+#ifndef UTIL_MSVC_COMPILER_H
+#define UTIL_MSVC_COMPILER_H
+
+#ifdef _MSC_VER
+#if __cplusplus >= 201700L  // with C++17 it's this deprecated
+                            // struct not available anymore
+#   include "util/xcs_functional.hpp"
+#endif
+#   include <io.h>
+#   include "corecrt_math_defines.h"
```
MaxKellermann, line 33: Why these headers?

src/util/msvc.h
```diff
@@ -0,0 +1,81 @@
...
+
+#define STRING2(x) #x
+#define STRING(x) STRING2(x)
+
+#define CLANG_OR_GCC_VERSION(x, y)  0
+#define GCC_CHECK_VERSION(x, y)     0
+#define GCC_OLDER_THAN(x, y)        1  // This isn't a (new!) GCC
```
MaxKellermann, line 40: Wrong. Your implementation says "this is a GCC older than the version specified", but it's not a GCC at all, thus cannot be a GCC older than that.

src/util/msvc.h
```diff
@@ -0,0 +1,81 @@
...
+#define CLANG_CHECK_VERSION(x, y)   0
+
+#define DT_UNDERLINE 0            // not avalable in WinUser.h!
```
MaxKellermann, line 43: Doesn't belong here.

src/util/msvc.h
```diff
@@ -0,0 +1,81 @@
...
+
+#define __attribute__(x)
```
MaxKellermann, line 45: Doesn't belong here.

src/util/msvc.h
```diff
@@ -0,0 +1,81 @@
...
+#define gcc_aligned(n)
+#define gcc_visibility_hidden
+#define gcc_visibility_default
+#define gcc_always_inline
+
+// unuse deprecated functions:
+#define strncasecmp(a, b, n)  _strnicmp(a, b, n)
+#define strcasecmp(a, b)      _stricmp(a, b)
+#define strdup(a)             _strdup(a)
+#define wcsdup(a)             _wcsdup(a)
```
MaxKellermann, line 72: Doesn't belong here.

src/util/msvc.h
```diff
@@ -0,0 +1,81 @@
...
+
+// TODO(August2111): where is this defined (on GCC)?
+#ifndef ssize_t
+typedef size_t   ssize_t;
+#endif  // ssize_t
```
MaxKellermann, line 77: Doesn't belong here. And TODO comments don't belong in a PR which is ready for review. And your name doesn't belong in code comments.

src/Geo/ConvexHull/GrahamScan.cpp
```diff
@@ -22,6 +22,7 @@
 
 #include "GrahamScan.hpp"
 #include "Geo/SearchPointVector.hpp"
+#include <iterator>  // necessary for std::back_inserter
```
MaxKellermann, line 25: Not needed anymore since commit 94c4c865f4458b78c47937a30621487b4a1af736

MaxKellermann:
I have cherry-picked 2 good commits; the rest is questionable or obsolete - I commented on these.

src/net/http/ToBuffer.cpp
August2111: Ok, my fault ;-( Mostly this are strings, and I want to avoid the 'closing'of the strings after this... For data it's don't make sense, but there does no harm, if the length of the buffer is ok. And so you would need an additional check - but with difference of string and data buffer - and so it is better to leave it as it is

src/LogFile.cpp
August2111: Because the ending 0, but this is vsnprintf doing by itself, I think!

MaxKellermann: > The functions snprintf() and vsnprintf() write at most size bytes (including the terminating null byte ('\0')) to str.

src/net/http/ToBuffer.cpp
MaxKellermann: There are various callers to this function, and some of them download binary data, but some of them download text, and if they want to null-terminate it, they do it like this:
https://github.com/XCSoar/XCSoar/blob/074243526049e79b68cd2c050533da813b7ab0d7/src/Weather/NOAADownloader.cpp#L164-L168
Passing `sizeof(buffer)-1` reserves space for the null terminator, which is applied after successful completion.
But if the API doesn't document that it null-terminates the buffer, you shouldn't null-terminate it. This violates the API contract.

src/util/TriState.hpp
August2111: Ok, I have to split the commit! 
This is because you have to inlude winsock2.h before windows.h.

src/util/Compiler.h
August2111: Of course!
I don't know, how this happens ;-(

src/util/TriState.hpp
MaxKellermann: > This is because you have to inlude winsock2.h before windows.h.

Is that a documented WIN32 API requirement?

src/util/Compiler.h
August2111: In my job we use cpplint (https://en.wikipedia.org/wiki/Cpplint) as sytle checker - we use an adapted Google style for the projects. In this you can find:

# The #define Guard

All header files should have #define guards to prevent multiple inclusion. The format of the symbol name should be <PROJECT>_<PATH>_<FILE>_H_.

To guarantee uniqueness, they should be based on the full path in a project's source tree. For example, the file foo/src/bar/baz.h in project foo should have the following guard:

```
#ifndef FOO_BAR_BAZ_H_
#define FOO_BAR_BAZ_H_
...
#endif  // FOO_BAR_BAZ_H_
```
We don't use the Project name as head for this to make it a little bit shorter, but with path name with underlines. So the define should be XCSOAR_UTIL_COMPILER_H - as you defined it for example in the Math folder (FastMath.hpp:  
```
#ifndef XCSOAR_MATH_FASTMATH_HPP
#define XCSOAR_MATH_FASTMATH_HPP
```
But I think, you want to do this in the same way (with XCSOAR_ as head), but you mean to do this renaming in an own commit - not byside from an other explonation, isn't it?

src/util/msvc.h
```diff
@@ -0,0 +1,81 @@
...
+}
+*/
+
+#ifndef UTIL_MSVC_COMPILER_H
+#define UTIL_MSVC_COMPILER_H
+
+#ifdef _MSC_VER
+#if __cplusplus >= 201700L  // with C++17 it's this deprecated
+                            // struct not available anymore
+#   include "util/xcs_functional.hpp"
```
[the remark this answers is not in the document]
August2111: ... and missing header file "util/xcs_functional.hpp" :-(
In C++11 the std:: binary_function and std::unary_function are deprecated, but in C++17 this structs are removed (like the standard says). I need this additional definitions to avoid this problem at least with msvc, but maybe this should be solved for all targets. I used a proposal from Stackoverflow as a workaround. Maybe it is clearer if I add the file ;-)

src/util/msvc.h
August2111: So  I have to use at every time where this is ask:
```
#if GCC_OLDER_THAN(x, y)  || defined(_MSC_VER)
```
Is this ok so ? Or did you have a better idea?

src/util/msvc.h
August2111: for the 3rd and 4th definition I disabled the warnings, this I can remove. But the both first: Either I change the code on this place ( maybe with the std-function) or I make a define there. Do you mean I should do this in the cpp or hpp file itself where it happens?

August2111:
Ok, after rebase and some changing I will try it again!

src/util/Compiler.h
MaxKellermann: Whether this is an improvement or not, and whether changing this for existing code or not is necessary - this doesn't matter, because you did not propose to change the names of include guards. You hid this change inside an unrelated commit.

src/util/msvc.h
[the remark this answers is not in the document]
MaxKellermann: Don't just copy'n'paste code from Stackoverflow without thinking. Stackoverflow is mostly filled with bad code samples.

src/util/msvc.h
MaxKellermann: Why would you want to do that?

src/util/msvc.h
MaxKellermann: I have no idea what you're trying to do here.

PR #594, opened 2021-04-27
Forward ExchangeFrequency action to devices

Scumi:
Brief summary of the changes
----------------------------
Allow forwarding the ExchangeRadioFrequencies action to devices.

Special emphasis on a188b12 since it alters the previous behaviour (even though it was unused in the code). 

n/a


MaxKellermann:
A driver which doesn't support the "exchange" operation should still be able to exchange the frequencies using the less efficient method we had before.

MaxKellermann:
I cherry-picked the commit "Input/InputEvents: add frequency exchange event", because wiring this to a user-reachable event is a good idea. (Documentation is missing.)
The rest should be refactored to fall back properly.

Scumi:
> A driver which doesn't support the "exchange" operation should still be able to exchange the frequencies using the less efficient method we had before.

I disagree. XCSoar forces an exchange that is maybe not intended or supported as an external event by the device. To me this is a hack and no different (especially with no documentation on the side of the device manufaturer) to a vario driver sending data with wrong units and XCSoar rightfully not supporting a correction on its side.

Will see what I can do to refactor, didn't come up with anything useful yet.

MaxKellermann:
> XCSoar forces an exchange that is maybe not intended or supported as an external event by the device.

What does that mean? XCSoar knows the current active/standby frequency, and then XCSoar does a very simple thing: set new active to old standby, and set new standby to old active. That is a simple fallback - not the most efficient one, but does the right thing. What is wrong with that?

Scumi:
In my view the device should support the command directly.

For example, the ACD device prevents a direct setting of the active frequency by an external source (required by regulation according to the manufacturer), so the simple fallback for ExchangeRadioFrequencies doesn't work. However, it supports a frequency swap command. Now one could create a workaround so XCSoar is still able to set active frequency directly by swapping -> setting new passive -> swapping back, which would violate the intention of the manufacturer (or the regulating authority).

Anyways I see what I can do wrt the refactor of the ExchangeRadioFrequencies method.

MaxKellermann:
> In my view the device should support the command directly.

We agree on this.

But this statement of yours is missing the point. This is not about whether we should support native swapping; this is about whether we should have a fallback if the device doesn't support swapping natively.

src/Device/MultipleDevices.hpp
```diff
@@ -90,7 +90,7 @@ class MultipleDevices final : PortListener {
                           OperationEnvironment &env);
   void PutStandbyFrequency(RadioFrequency frequency, const TCHAR *name,
                            OperationEnvironment &env);
-  void ExchangeRadioFrequencies(OperationEnvironment &env);
+  bool ExchangeRadioFrequencies(OperationEnvironment &env);
```
MaxKellermann, line 93: What does the return value mean? Documentation!

src/ActionInterface.cpp
```diff
@@ -366,6 +366,13 @@ void ActionInterface::OffsetStandbyFrequency(double offset_khz, bool to_devices)
 
 void ActionInterface::ExchangeRadioFrequencies(bool to_devices)
 {
+  /* send to external devices */
+  if (to_devices) {
+    MessageOperationEnvironment env;
+    if (device_blackboard->ExchangeRadioFrequencies(env))
+      return;
```
MaxKellermann, line 373: Has this failed due to an error or because the driver doesn't support it?
If there was an error, there should not be a fallback.

Scumi: It returns if ExchangeRadioFrequencies was successful, thus not executing the fallback. I could reverse the statement so the fallback gets called directly on unsuccessful frequency exchange.

src/Device/MultipleDevices.hpp
Scumi: Will add docs.

src/ActionInterface.cpp
Scumi: Maybe I misunderstood your question.
The previous implementation failed due to one of the drivers (and its device) does not support a direct setting of the active frequency, but it supports an exchange frequency command (1ac47d1a3c).
Because of our previous discussion if XCSoar should support a fallback or not, I've implemented a way to directly exchange the frequencies in the driver and fallback if the method is not implemented by it (or fails).

src/Device/MultipleDevices.cpp
```diff
@@ -106,11 +106,14 @@ MultipleDevices::PutStandbyFrequency(RadioFrequency frequency,
...
 }
 
-void
+bool
 MultipleDevices::ExchangeRadioFrequencies(OperationEnvironment &env)
 {
-  for (DeviceDescriptor *i : devices)
-    i->ExchangeRadioFrequencies(env);
+  for (DeviceDescriptor *i : devices) {
+    return i->ExchangeRadioFrequencies(env);
```
MaxKellermann, line 113: This calls only the first device and ignores the others.
Have you ever tested your code?

src/ActionInterface.cpp
MaxKellermann: I have no idea what you are talking about, but none of this answers my question.
Your code will never use the fallback code.
Did you test this? Did it work? Please demonstrate that I'm wrong.

src/Device/Driver.cpp
```diff
@@ -92,6 +92,12 @@ AbstractDevice::PutStandbyFrequency(RadioFrequency frequency,
   return true;
 }
 
+bool
+AbstractDevice::ExchangeRadioFrequencies(OperationEnvironment &env)
+{
+  return true;
```
MaxKellermann, line 98: Why return success here? Your code did not actually exchange any frequencies, did it?

MaxKellermann (changes requested):

src/Device/MultipleDevices.cpp
Scumi: Agreed, that doesn't make sense at all.

src/ActionInterface.cpp
Scumi: Yes, I tested it and it works for me. I'll have a look into your comments to investigate the flaws.

I thought about this again in general and I don't see (or maybe understand) the usefulness of the fallback at all. All devices I know and checked have different capabilities. Some have allow exchange commands, some allow setting one frequency, some both, some allow activating dual watch. The best place to cope with the commands and respective error handling is in the drivers themselves. Its another thing of course for stuff like QNH or baro altitude, which XCSoar needs as well for internal calculations, but XCSoar has no use for radio frequencies without an attached device.

MaxKellermann: > The best place to cope with the commands and respective error handling is in the drivers themselves.

That is not what your code submission does. Your code submission breaks the feature for existing drivers.

MaxKellermann:
Is this going anywhere? 6 weeks ago, I pointed out a few bugs in this PR, but there is no progress.

src/ActionInterface.cpp
Scumi: That is correct. Unfortunately, you cherry-picking parts of the PR broke it for one of the devices as well.

I see two options:
1. create a fall-back in XCSoar to handle the feature according to the devices capabilities (i'm not content with that approach).
2. get all drivers to implement ExchangeRadioFrequencies themselves. I either need the original developers for that or all four devices that drivers exist for to test the implementation.

Scumi:
Yes, it is not forgotten, I just got not time to work on it (and my testing device is flying). You can close if you want, I can re-open once time allows working on it.

MaxKellermann:
> Unfortunately, you cherry-picking parts of the PR broke it for one of the devices as well.

Uh, what? Please give more details.

If that is true, then your PR was broken in the first place. Each commit in a PR *must* not introduce a regression. Cherry-picking must be possible, or the PR is broken.

Maybe I should revert the cherry-picks, and wait for you to resubmit bug-fixed commits in this PR.

src/ActionInterface.cpp
MaxKellermann: Fallback in XCSoar core code is fine; I don't think it's a good idea to duplicate code in each driver.

Scumi: The drivers wouldn't duplicate code. Each would implement the exchange of frequencies in a way that is supported by its device, same way setting frequencies is handled by the device drivers.
Unfortunately I cannot check what the other radio devices are capable of because I cannot find any documentation on the protocols.

Scumi:
The ExchangeFrequency function was broken before this PR for some devices, but the commit you cherry-picked made the function publicy accessible (it wasnt used before). Maybe I should've left it out until the other issues are fixed. 

I tried to fix the function in a way it made sense to me; by moving the responsibility for exchanging frequencies into the drivers (where it belongs in my view).

src/ActionInterface.cpp
MaxKellermann: Sigh, now you understand my strict "merge driver code only if vendor publishes documentation for free" policy which I'm often criticized for... disappearing documentation is obviously another problem. There were vendors who used copyright to prevent us to mirror their documentation. 10 years later, this is still sad.

So, if you take care for implementing this in all drivers, then yes I'd agree to your idea moving everything into the drivers.

src/ActionInterface.cpp
```diff
@@ -366,11 +366,10 @@ void ActionInterface::OffsetStandbyFrequency(double offset_khz, bool to_devices)
...
 {
-  const auto radio_settings = SetComputerSettings().radio;
-
-  const auto old_active_freq = radio_settings.active_frequency;
-  const auto old_active_freq_name = radio_settings.active_name;
+  /* send to external devices */
 
-  ActionInterface::SetActiveFrequency(radio_settings.standby_frequency, radio_settings.standby_name, to_devices);
-  ActionInterface::SetStandbyFrequency(old_active_freq, old_active_freq_name, to_devices);
+  if (to_devices) {
```
roeles, line 371: You're not updating our own internal administration here, which appears incorrect to me. When is this updated?
Previously it happened because as a result of the call to `SetActiveFrequency` and `SetStandbyFrequency`.
I can't comment on the other devices, but the KRT2 driver doesn't give any feedback to XCSoar after exchanging the frequencies.
I suspect this code would cause the radio and xcsoar (in the infoboxes I added) to display two different things.

src/Device/Driver.cpp
roeles: @MaxKellermann I'm not saying you're wrong, but returning true appears to be the norm in this file. 
The argument you're making could also be made for `PutStandbyFrequency` above, and many others.

roeles:
I was asked to test the KRT2 code. I have not had time for this yet, since it requires me ejecting the canopy of my glider and attaching my laptop to my KRT2.

The PR looks a bit large to me. From my experience with adding functionality for KRT2 in the past, I think proper testing with a real device is a must.
I would therefore recommend to make functionality backwards-compatible (devices/drivers that do not yet natively support swapping frequencies use my `ActionInterface` function). And then improve device per device.

src/ActionInterface.cpp
Scumi: Yes, I saw that as well in the documentation I received. 

PR is WIP (it was marked as such, don't know who removed it).

Scumi:
All devices where drivers exist for are compatible for exchanging the frequency directly. Special care has to be taken with the subsequent read-out of the frequencies (not all devices communicate the change). I will continue working on that part.

src/Device/Driver.cpp
MaxKellermann: This method is different, because the caller needs to know the difference between "has succeeded", "has failed" and (new third option) "not implemented", or else the caller cannot decide whether to use the fallback code.

hjr:
What is the status here? I would like to see frequency transfer to ACD working. What does it need to push?

Scumi:
Secondary Frequency push works already. Primary Frequency change via external device is not intended by the manufacturer.

Other than that I have two out of three devices on my desk and need some time to update the code. Not sure how far I will get, the protocols of the devices appear to lack some functionalities.

hjr:
Ok, that's cool. I understood 7.23 just does not have this functionality, but hopefully the next release.
I had contact with Marc Förderer. Garrecht intend to change the behavior, so that also the active frequency can be changed. ETA of the ACD update with that feature will be mid 2023.
For me changing the standby frequency would be awesome. Actually that is the standard procedure of ACD. You always change the standby frequency. To actually use your newly set frequency you have the swap them. That is just one button press on the device. Arguable procedure, I know.
Can I support you with anything?

hjr:
And BTW Garrecht support alias Marc offered to answer any question to the ACD protocol.

Scumi:
> I had contact with Marc Förderer. Garrecht intend to change the behavior, so that also the active frequency can be changed. ETA of the ACD update with that feature will be mid 2023.

Okay, great. Happy to implement once the update is there.

> For me changing the standby frequency would be awesome. Actually that is the standard procedure of ACD. You always change the standby frequency. To actually use your newly set frequency you have the swap them. That is just one button press on the device. Arguable procedure, I know.

I guess you mean active frequency in your first sentence. :-) Yes, it would be nice and I thought about a workaround back when I created the driver, but I personally didn't want to violate the intend from the manufacturer.

> Can I support you with anything?

If you have one of the other radios and can help implementing some features there, yes. ACD is pretty simple and most of the functionality is available already. I'll let you know if something can be done.

> And BTW Garrecht support alias Marc offered to answer any question to the ACD protocol.

I know, I was in contact with him several times during the driver development :-)

MaxKellermann:
This draft PR looks unmaintained.

PR #626, opened 2021-05-15
use curl for download on android

lordfolken:


MaxKellermann:
This doesn't solve the long-term problem that by the end of the year, XCSoar will not be allowed to use XCSoarData anymore. This PR just addresses the short-term problem that the DownloadManager already refuses to use it now. Once we have a solution for the former, we can easily use the DownloadManager, which has advantages: downloads continue even if XCSoar isn't running, and will be done whenever the user is on a Wifi network. The user's global download settings will be obeyed. XCSoar's own download code isn't aware of those settings.

lordfolken:
@MaxKellermann I completley with that statement, and i didn't mean for this to be the final fix. 
However moving to the api level 30 scoped storage concept will take quite some planning. 
* We need to move the userdata to the new location, 
* announce the move ahead of time, as there are a lot of 3rd party tools that currenlty depend on this path. (CoTaCO for example) 
* Make the process transparent to the user (Dialog, Handling code etc)

RIght now XCsoar is hard to use for a lot of people in the android eco system. About 40% of the users use android 10 or higher. This is meant to address this issue in the short term. 

Could you please, prepare another release with the current workaround?

PR #630, opened 2021-05-24
Settings/TrackingConfigPanel: same tracking intervals for Livetrack24 as for SkyLines

sandrinr:
Brief summary of the changes
----------------------------

Use the existing tracking intervals list for SkyLines also for LiveTrack24.
Currently LiveTrack24 allows tracking intervals between 5 seconds and 1 hour using 5 second increments.

Intent: Allow LiveTrack24 tracking intervals smaller than 5 seconds.

none

Caveats
--------

Users with an existing setting of `TrackingInterval` outside of the values specified in the `tracking_intervals` array (e.g. "1800") will have their tracking interval set to 1 second.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -1,4 +1,6 @@
 Version 7.9 - not yet released
+* settings
+  - allow LiveTrack24 tracking intervals smaller than 5 seconds
 
 Version 7.8 - 2021/06/02
 * don't delete old IGC files automatically
--- a/src/Dialogs/Settings/Panels/TrackingConfigPanel.cpp
+++ b/src/Dialogs/Settings/Panels/TrackingConfigPanel.cpp
@@ -86,6 +86,8 @@ public:
 private:
   /* methods from DataFieldListener */
   void OnModified(DataField &df) noexcept override;
+
+  unsigned int FindClosestTrackingInterval(unsigned int) noexcept;
 };
 
 #ifdef HAVE_SKYLINES_TRACKING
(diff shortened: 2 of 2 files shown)


MaxKellermann:
> Users with an existing setting of TrackingInterval outside of the values specified in the tracking_intervals array (e.g. "1800") will have their tracking interval set to 1 second.

That's not acceptable. Since the LT24 protocol is heavily bloated, this will cause big amounts of mobile traffic.

sandrinr:
> That's not acceptable. Since the LT24 protocol is heavily bloated, this will cause big amounts of mobile traffic.

Understandable. What would be your minimal acceptance criteria?
Initializing it to a value close to the original setting value?
Would you like it more me creating a new enum for LT24 (instead of using the one for SkyLines) holding all 5 second increments up to 60 minutes in order to be compatible with every possible old setting value?

More generally, do you have another idea to allow sub 5 second intervals not involving an enum?

MaxKellermann:
It is not acceptable to break existing configurations, even less so if it means the user has to pay money for unwanted internet traffic. All changes to configuration must be migrated properly; I guess rounding slightly up or down to the nearest supported value would be acceptable, but not resetting the configuration.
Neither the integer DataField nor the EnumDataField are a good choice for this setting. Maybe we need a new DataField class for configuring a time interval, with a number of suggested presets, but if an existing value is not in those presets, it will appear anyway in the list.

sandrinr:
Updated the branch. Went the route of adding some more values to `tracking_intervals` and added a method to find the closest tracking interval setting given an input setting.

PR #635, opened 2021-05-28
Basic Kobo Nia support - WiFi and USB Storage export.

vnks:
Brief summary of the changes
----------------------------

Adds Kobo Nia support for WiFi and USB Storage export features. Note that XCSoar 7 does not currently boot on the Kobo Nia due to missing /dev/ entries, which needs to be addressed to actually support the Nia.


lordfolken (approved):
Looks good IMO.

PR #639, opened 2021-05-31
New Gesture 'C' for Checklist

lordfolken:


MaxKellermann:
What is the justification for this shortcut? You need this feature once per flight, so my idea is that it doesn't deserve this kind of shortcut.

lordfolken:
I have multiple checklists besides, rigging, takeoff etc.  enroute i use it for ATC callouts (i fly often in congested airspace), transponder operation, and for landing, as well as turbo start procedure. I have also some notes on the many frequencies used for air2air comms (not in an airspace file).I kind of like it because its quicker than digging out paper checklists out of the side pocket.

fb:
Same for me. I use my checklists/notes multiple times in each flight and I have something similar to the proposed PR in my personal configuration.

At least for me, with a touchscreen, gestures are a lot easier accessible than the double-tab menu system. (For the same reason I have a gesture for the QuickMenu.)

PR #647, opened 2021-06-14
Pilot Event (PEV) start

### LINKED ISSUE #646: Support PEV (Pilot Event) start procedure

Since the 19 November 2020 new start procedure is added to the [rules for world and continental
gliding championships](https://www.fai.org/sites/default/files/sc3a_2020a.pdf): _Pilot Event (PEV) start_ (see page 28, 29).

> When the PEV Start Option is in use, pilots must record a Pilot Event in the
> Primary Flight Recorder (“press the PEV”) before crossing the Start Line or
> Ring. Pressing the PEV creates an interval in which a penalty-free start may
> be made. The interval begins PEV Wait Time minutes after pressing PEV,
> and it ends PEV Start Window minutes later.
> 
> PEV may be pressed at any time, up to a total of three times (the Scorer will
> ignore PEVs after the third one). Each press of PEV cancels the previous
> Start window and restarts the Wait Time timer.
> 
> Multiple PEVs recorded within 30 seconds will be treated as a single PEV at
> the time of the first PEV of the cluster.
>
> Failure to record a PEV, or a valid start outside the penalty-free interval will be
> penalized.

XCSoar should have a first class support for this start procedure, because it is already being used for international and regional contests.

This is a request for comments, but I imagine the feature would require these modifications:

* Add 2 new options in "Task Defaults > Task Rules" configuration screen, both containing a value expressed in minutes:
  * PEV Wait Time
  * PEV Start Window
* Add a "PEV" button to (perhaps) NAV menu

When PEV button is pressed, this should happen:
* NMEA message `$PFLAI,PILOTEVENT*` is announced on all active NMEA devices. Alternatively, this might be a device-driver specific function.
* `EXXXXXX PEV` message is written to .igc log. [Spec](https://xp-soaring.github.io/igc_file_format/igc_format_2008.html#link_E).
* Start open time is set to "Current Time" +  "PEV Wait Time"
* Start close time is set to "Current Time" +  "PEV Wait Time" + "PEV Start Window"

kedder:
Brief summary of the changes
----------------------------

This adds support for Pilot Event start procedure, recently introduced for use in gliding competitions. See  [rules for world and continental gliding championships](https://www.fai.org/sites/default/files/sc3a_2020a.pdf), page 28, 29 for more details.

This PR introduces a new button, available under Nav -> Nav -> Pilot Event. When pressed, a time window for a valid start is configured according to two new user-configurable settings for the task. This feature takes advantage of existing "start open time" and "start close time" parameters as well as existing "Start open" and "Start reach" infoboxes to aid the pilot.

Expected mode of operation: when "Pilot Event" start is used in a competition day, pilot will set "PEV start wait time" and "PEV start window" task properties according to values, announced during the briefing. Later, in the air, in order to make a valid start, the pilot will press the "Pilot Event" button in XCSoar (same way they would press the button on the physical logger hardware), wait the specified amount of time and then cross the start line while start window is open. Once "Pilot Event" is pressed, "Start Open" infobox will start the countdown to the new start open time.

The pilot event is announced on connected devices (by the driver-specific NMEA message) - this is useful in case the logger device is physically unreachable by the pilot (e.g. behind the instrument panel).

This PR is ready for review, but not ready to merge yet. I'd like to make a test flight in order to confirm the integration with FLARM is working as expected.

Closes #646.

### DIFF
--- a/src/Dialogs/Settings/Panels/TaskRulesConfigPanel.cpp
+++ b/src/Dialogs/Settings/Panels/TaskRulesConfigPanel.cpp
@@ -39,6 +39,9 @@ enum ControlIndex {
   spacer_2,
   FinishMinHeight,
   FinishHeightRef,
+  spacer_3,
+  PEVStartWaitTime,
+  PEVStartWindow,
 };
 
 class TaskRulesConfigPanel final : public RowFormWidget {
@@ -116,6 +119,24 @@ TaskRulesConfigPanel::Prepare(ContainerWindow &parent,
           altitude_reference_list,
           (unsigned)task_behaviour.ordered_defaults.finish_constraints.min_height_ref);
   SetExpertRow(FinishHeightRef);
+
+  AddSpacer();
+  SetExpertRow(spacer_3);
+
+  AddTime(_("PEV start wait time"),
+          _("Wait time in minutes after Pilot Event and before start gate opens. "
+            "0 means start opens immediately."),
+          0, 30*60, 60,
+          task_behaviour.ordered_defaults.start_constraints.pev_start_wait_time);
+  SetExpertRow(PEVStartWaitTime);
+
+  AddTime(_("PEV start window"),
+          _("Number of minutes start remains open after Pilot Event and PEV wait time."
+            "0 means start will never close after it opens."),
+          0, 30*60, 60,
+          task_behaviour.ordered_defaults.start_constraints.pev_start_window);
+  SetExpertRow(PEVStartWindow);
+
 }
 
 
@@ -151,6 +172,14 @@ TaskRulesConfigPanel::Save(bool &_changed) noexcept
 
   changed |= SaveValueEnum(FinishHeightRef, ProfileKeys::FinishHeightRef,
                            otb.finish_constraints.min_height_ref);
+
+  changed |= SaveValue(PEVStartWaitTime,
+                       ProfileKeys::PEVStartWaitTime,
+                       otb.start_constraints.pev_start_wait_time);
+
+  changed |= SaveValue(PEVStartWindow,
+                       ProfileKeys::PEVStartWindow,
+                       otb.start_constraints.pev_start_window);
 
   _changed |= changed;
 
--- a/src/Input/InputEventsActions.cpp
+++ b/src/Input/InputEventsActions.cpp
@@ -84,6 +84,8 @@ doc/html/advanced/input/ALL		http://xcsoar.sourceforge.net/advanced/input/
 #include "MapWindow/GlueMapWindow.hpp"
 #include "Simulator.hpp"
 #include "Formatter/TimeFormatter.hpp"
+#include "Operation/MessageOperationEnvironment.hpp"
+#include "Device/MultipleDevices.hpp"
 
 #include <cassert>
 #include <tchar.h>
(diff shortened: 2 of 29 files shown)


src/Dialogs/Settings/Panels/TaskRulesConfigPanel.cpp
```diff
@@ -116,6 +119,24 @@ TaskRulesConfigPanel::Prepare(ContainerWindow &parent,
...
           (unsigned)task_behaviour.ordered_defaults.finish_constraints.min_height_ref);
   SetExpertRow(FinishHeightRef);
+
+  AddSpacer();
+  SetExpertRow(spacer_3);
+
+  AddFloat(_("PEV start wait time"),
+             _("Wait time in minutes after Pilot Event and before start gate opens. "
+               "0 means start opens immediately."),
+             _T("%.0f min"), _T("%.0f min"), 0, 30, 1, false, UnitGroup::ALTITUDE,
```
MaxKellermann, line 129: `UnitGroup::ALTITUDE` sounds wrong

src/Dialogs/Settings/Panels/TaskRulesConfigPanel.cpp
```diff
@@ -116,6 +119,24 @@ TaskRulesConfigPanel::Prepare(ContainerWindow &parent,
...
+             task_behaviour.ordered_defaults.start_constraints.pev_start_wait_time);
+  SetExpertRow(PEVStartWaitTime);
+
+  AddFloat(_("PEV start window"),
+             _("Number of minutes start remains open after Pilot Event and PEV wait time."
+               "0 means start will never close after it opens."),
+             _T("%.0f min"), _T("%.0f min"), 0, 30, 1, false, UnitGroup::ALTITUDE,
```
MaxKellermann, line 136: Same here

src/Input/InputEventsActions.cpp
```diff
@@ -193,6 +195,41 @@ InputEvents::eventMarkLocation(const TCHAR *misc)
...
+    new_start = new_start + delta;
+  }
+
+  if (start.pev_start_window > 0) {
+    new_end = new_start + RoughTimeDelta::FromMinutes(start.pev_start_window);
+  }
+  const RoughTimeSpan ts = RoughTimeSpan(new_start, new_end);
+  start.open_time_span = ts;
+
+  protected_task_manager->SetOrderedTaskSettings(ots);
```
MaxKellermann, line 223: This is a very broad call, when all you need is a way to change the task's `open_time_span` attribute. To avoid developers to do stupid/dangerous things with this method, let's make a dedicated method for just setting `open_time_span`.

MaxKellermann (changes requested):

src/Dialogs/Settings/Panels/TaskRulesConfigPanel.cpp
kedder: I agree. I forgot I did this hack. 

I hesitated to add another UnitGroup for `MINUTES` just for this use case though. Would you recommend introducing `UnitGroup::MINUTE` or rather use `UnitGroup::NONE` here? I don't remember exactly what was wrong now, but  `UnitGroup::NONE` didn't work here out of the box. I can collect more details on this if that's helpful.

MaxKellermann: No new UnitGroup. XCSoar's units are for values where the user may choose different display units, but the code internally uses the base SI unit.
The first thing that's somewhat "wrong" with your code is that it doesn't use SI units; the SI unit for duration is seconds, not minutes. I'm not 100% sure if I want to demand you to switch to SI units - there may be good reasons to stay with minutes, but using SI units is never wrong, because it avoids a whole class of bugs.
The second thing that's wrong - if you're not going to use a UnitGroup, don't use the AddFloat overload that takes a UnitGroup parameter. Don't use UnitGroup::NONE.

src/Input/InputEventsActions.cpp
kedder: Changed that to `SetStartTimeSpan` instead.

src/Dialogs/Settings/Panels/TaskRulesConfigPanel.cpp
kedder: I have discovered `AddTime()`, which works with seconds. This was a good enough incentive to switch from minutes to seconds.

MaxKellermann: I think you may need to adjust the help text. "in minutes" is redundant because the UI already displays the unit.

MikeBravo15:
Unfortunately, new Pilot Event texts can not be translated. Can you please have a look on this issue?

kedder:
@MikeBravo15 from what I can tell, strings are marked as translatable (wrapped in `_()`). Is there one I missed or you cannot translate any of them? How do you expect the translation process to work?

MikeBravo15:
Not a big expert on the code, it seems to be OK, but text strings do not appear in Weblate for translation

kedder:
I see, perhaps the translation template needs to be updated (somehow). Will try to figure it out.

MaxKellermann:
The POT file has not been regenerated yet.

kedder:
@MaxKellermann will it haappen automatically or someone should do something to regenerate it?

MaxKellermann:
Manually. `make po/xcsoar.pot` - I'll do it.

PR #653, opened 2021-06-25
Replaced goofy character in menu string.

hjr:
Brief summary of the changes
----------------------------

Minimal fix of some french non ascii character slipped into some GUI string.

### DIFF
--- a/src/InfoBoxes/Content/Factory.cpp
+++ b/src/InfoBoxes/Content/Factory.cpp
@@ -1011,7 +1011,7 @@ static constexpr MetaData meta_data[] = {
   // CONTEST_SPEED
   {
     N_("Contest speed"),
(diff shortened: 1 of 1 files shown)


lordfolken (changes requested):
This is used as a string match in translations. If we change it it needs to be changed in all occurrences.

hjr:
That is correct, but the typo is actually visible for English and makes just no sense. All translations will stay the same and it is just a mechanical Ok click for the given suggestion from the translation AI system to get the translations aligned again. It is just a sooner or later...

MaxKellermann:
fee707de890dc3d3d17d579696b1506b68a57d36

hjr:
Ok, thx!

PR #679, opened 2021-08-06
read barometric pressure from an experimantal BLE peripheral.

bomilkar:
Brief summary of the changes
----------------------------
I want to get started with BLE since this is new in 7.11
The intend is to learn from the experiment. Any comment is highly appreciated!
The code changes may not make it into the production code.
At this time the codes does not behave as intended. See commit message for more detail.

### DIFF
--- a/src/Device/AndroidSensors.cpp
+++ b/src/Device/AndroidSensors.cpp
@@ -179,6 +179,7 @@ DeviceDescriptor::OnBarometricPressureSensor(float pressure,
   NMEAInfo &basic = *e;
 
   basic.UpdateClock();
+  basic.alive.Update(basic.clock);
   basic.ProvideNoncompVario(ComputeNoncompVario(kalman_filter.GetXAbs(),
                                                 kalman_filter.GetXVel()));
   basic.ProvideStaticPressure(


bomilkar:
More detail on what I'm trying to do can be found here:
https://forum.xcsoar.org/viewtopic.php?f=3&t=3934#p22464
This post includes the Arduino Sketch for the BLE peripheral.

@MaxKellermann : your review and feedback is very welcome. I'd like to find out why it's not working as expected.

MaxKellermann:
Your BLE specific code looks correct.
Try adding
```
  basic.alive.Update(basic.clock);
```
to `DeviceDescriptor::OnBarometricPressureSensor()` - this may be the missing piece here, I'm not sure.
If that doesn't work, check if your `readBaro()` function ever gets called (by adding logging calls to the Java source).

bomilkar:
Thanks Max!

Adding
`basic.alive.Update(basic.clock);`
to `DeviceDescriptor::OnBarometricPressureSensor()` does the trick.
Now the Device list shows "Connected, Baro" under my BLE device. :-)

One minor issue in my code:
`listener.onBarometricPressureSensor()`
wants the pressure in hPa, not in Pa.

Shall I close this PR w/o merge?
Or do you want to keep it?

MaxKellermann:
I'd like to have the (separate) commit that adds this one line.
The driver for your new toy - I'll merge it if you think it's useful for others. For example, if you make it a project that others can build as well, or if you build it and sell it. You should think about what should happen to it.

bomilkar:
Here is the one-liner as you suggested, @MaxKellermann .

I don't know where I can put the Arduino Sketch for this BLE peripheral. This would be the only piece of interest because there is nothing new in my additions to BluetoothSensor.java and BluetoothUuids.java. For now I think the ZIP-File in the forum has all the stuff required to start a project. Agreed?

MaxKellermann:
What do you want this project to go? Is this just an one-off experiment which will disappear in your drawer, or will you actually put it in your glider and help turn this into a full vario, maybe will the paraglider community pick it up as a cheap DIY vario?
If you want to make it a project, you could make all the stuff needed to build one including the firmware into a git repository (which we could put below https://github.com/XCSoar/), and then you can resubmit your XCSoar driver code with additional code comments explaining where to find more information about it.

bomilkar:
I'll think about it.
Initially I wanted to learn the BLE stuff including a peripheral device. That I accomplished. (And found a bug on the way.)
I need to check with paragliders what they would need besides a pressure sensor which is probably in their smartphone already.

PR #728, opened 2021-09-23
Fix for LXNAV Varios

### LINKED ISSUE #688: Not possible to declare task to LXNAV S100

XCSOAR version 7.11 on android
S100 version 8.02 (is not working)
nano 3.00 (is working)
nano3 3.02 (is working)

Expected behavior
-----------------

task -> declare should result in declaring the task into the S100.
All other communication with the S100 is working (connection over bluetooth)
Similary, the task declarition to an lxnav nano and nano3 is working.
I think download of igc files is also not working with the s100 (other but similar issue)

Actual behavior
---------------

task is not declared, after a while an error is displayed "task not declared"

connect with BT to lxnav S100
in task->manage press declare task

much be some (recent?) change in the LXNAV protocol.
previously (i heard from somebody) it used to be working.

contact lxnav for latest protocol

HeroicLemon:
Brief summary of the changes
----------------------------
Fixes issues with declaring tasks and other comms with LXNAV S series varios.

Based on original work from LXNAV located here: https://github.com/XCSoar/XCSoar/pull/220/

Tested with an LXNAV S100 vario. No testing has yet been performed with other LXNAV Varios or their passthrough capability with the Nano.

Closes #688 

Update to indicate test steps taken with S100 vario:

1. NAV->Task->Manage->Declare; XCSoar indicates success, S100 receives task.
2. Config->Flight; made changes to ballast and bugs. Vario recevies updated values. 
3. Changed manual MC value; vario receives the updated value.
4. Changed MC, ballast and bugs at vario; XCSoar receives the updated values.
5. Config->Devices->LXNAV->Manage; the vario management dialog appears.
6. Config->Devices->Flight Download; flight list is read with all flights displayed. Selecting a flight and clicking Download causes the flight to be downloaded.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -4,6 +4,8 @@ Version 7.19 - not yet released
   - fix crash bug in terrain loader
   - support short name in CUP files
   - search waypoints via short name
+* devices
+  - improved support for LXNAV S8x/S10x varios, including task declaration.
 * user interface
   - consistent progress bar during startup
 * Android
--- a/build/main.mk
+++ b/build/main.mk
@@ -26,8 +26,8 @@ DIALOG_SOURCES = \
 	$(SRC)/Dialogs/Device/ManageFlarmDialog.cpp \
 	$(SRC)/Dialogs/Device/BlueFly/BlueFlyConfigurationDialog.cpp \
 	$(SRC)/Dialogs/Device/ManageI2CPitotDialog.cpp \
-	$(SRC)/Dialogs/Device/LX/ManageV7Dialog.cpp \
-	$(SRC)/Dialogs/Device/LX/V7ConfigWidget.cpp \
+	$(SRC)/Dialogs/Device/LX/ManageLXNAVVarioDialog.cpp \
+	$(SRC)/Dialogs/Device/LX/LXNAVVarioConfigWidget.cpp \
 	$(SRC)/Dialogs/Device/LX/ManageNanoDialog.cpp \
 	$(SRC)/Dialogs/Device/LX/NanoConfigWidget.cpp \
 	$(SRC)/Dialogs/Device/LX/ManageLX16xxDialog.cpp \
--- a/src/Device/Descriptor.cpp
+++ b/src/Device/Descriptor.cpp
@@ -719,7 +719,7 @@ DeviceDescriptor::IsManageable() const
 
     if (StringIsEqual(driver->name, _T("LX")) && device != nullptr) {
       const LXDevice &lx = *(const LXDevice *)device;
-      return lx.IsV7() || lx.IsNano() || lx.IsLX16xx();
+      return lx.IsManageable();
     }
   }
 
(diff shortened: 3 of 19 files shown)


MaxKellermann:
Why those renames in the first commit?
The second commit is still a big mess. It mixes several independent changes, but the commit message just says "Fix for LXNAV Varios", but doesn't explain what exactly was wrong and what exactly is being fixed.
The third commit should not be a separate commit; the commit that actually fixes a bug should add the NEWS line.

There is an obvious bug in the second commit that was **not** present in #220: it misinterprets the return value of `strcmp()`, it's the wrong way around. This cannot possibly work! This completely breaks product detection.
Can you explain why this change was made at all?

This is really confusing. Maybe a little less confusing than #220 but still bad enough to reject merging it.

HeroicLemon:
The renaming was done to make the classes less specific to the V7 as they are applicable to the Sxx varios as well. 
The inversed `strcmp()` was present in the previous PR, but you are right that it's incorrect and I will fix it.
I'll roll the second and third commits together, and improve the commit message to describe what was going wrong.

MaxKellermann:
> The renaming was done to make the classes less specific to the V7 as they are applicable to the Sxx varios as well.

That explanation should be part of the commit message.

I think the names you chose are bad. This changes "V7" to "LX", but "LX" usually means "LX Navigation", which this is exactly **not**.

> The inversed `strcmp()` was present in the previous PR

There's no single `strcmp()` in #220! It uses method `equals()` which is less confusing to use than `strcmp()`, which I tend to avoid for that very reason.

>  I'll roll the second and third commits together, and improve the commit message to describe what was going wrong.

That won't do. There's a lot of stuff in the second commit which makes it very hard to read, which doesn't belong there.

HeroicLemon:
The original PR changed the names from V7 to Vario, which I thought was a little *too* generic. I can use LXNAVVario if that's a better description.

In #220, I'm seeing `src/Device/Driver/LX/Internal.hpp` lines 152-155 utilizing the `strcmp()` function. That's where I pulled it from.

I'll see if I can break down the fixes a little more logically. Looks like I neglected updating the tests as well.

MaxKellermann:
> The original PR changed the names from V7 to Vario, which I thought was a little _too_ generic. I can use LXNAVVario if that's a better description.

True, that was also bad. LXNAVVario is clunky and ugly but ... is correct.

Also I'd like to avoid all those `IsV7() || IsSVario()` checks - maybe these should be replaced with `IsLXNAVVario()`? And `IsNano() || IsSVario()` becomes `IsLXNAVLogger()`? Do not list each and every product everywhere; do semantic checks instead, these are easier to understand.

> In #220, I'm seeing `src/Device/Driver/LX/Internal.hpp` lines 152-155 utilizing the `strcmp()` function. That's where I pulled it from.

Oh yes you're right, it was a separate commit (just as buggy) - you folded it into the main commit.

MaxKellermann:
Looks very good now, thanks for taking the time to refactor the PR!

HeroicLemon:
Great! I'll message the LXNAV folks and let them know, maybe they'll have a sec to check out their other equipment and test the comms.

PR #732, opened 2021-09-26
Update Analysis buttons description

MikeBravo15:
Change of the description of the Analysis buttons, used to call out Flight Setup menu to better describe called menu to make it more simmilar to other buttons (Set Wind) + typo correction.

### DIFF
--- a/src/Dialogs/dlgAnalysis.cpp
+++ b/src/Dialogs/dlgAnalysis.cpp
@@ -463,7 +463,7 @@ AnalysisWidget::Update()
     dialog.SetCaption(sTmp);
     BarographCaption(sTmp, glide_computer.GetFlightStats());
     info.SetText(sTmp);
-    SetCalcCaption(_("Settings"));
+    SetCalcCaption(_("Set Flight"));
     break;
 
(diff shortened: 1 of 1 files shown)


MaxKellermann:
I don't see why "Set Flight" is an improvement. What does "Set Flight" mean? "Settings" is something I can understand: it lets the user change settings. Admittely, it does not spell out which settings, that depends on the context where this button is located; but since this is a button label, and button labels must be short, I think that's ok. But "Set Flight"??
What typo correction anyway?

MikeBravo15:
Typo correction is in Task (c)Calc to keep it uniform in all menus. For translation purposes I would like to separate Setting as now it is used for Map settings and also for Analysis pages and Weblate has one expression for all 5 iterations. Set Flight is IMHO good solution as it is calling Flight Setup page, exactly the same as Flight button of the main menu. If you are nOK with this, please fix at least the typo in Task Calc

MaxKellermann:
If you had included the typo fix as a separate documented commit instead of mixing it inside another commit, I could have cherry-picked it.

MikeBravo15:
OK, will close this now

Issue #457, opened 2020-07-21
Kobo : battery level in the launcher


xbug42:
XCSoar version
--------------
6.8.15 on Kobo

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------
Display the battery level somewhere in the Kobo launcher (the menu with the FLY / SIM and Nickel / Tools / Network / System / Poweroff options)

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------
I would like to have a quick and simple way to know the battery level without starting XCSoar

lordfolken:
Should be possible. You could try to implement something here: https://github.com/XCSoar/XCSoar/blob/master/src/Kobo/System.cpp

PR #733, opened 2021-09-26
Adds a small indicator to the poweroff screen to show the battery lev…

ulipo:
…el at the time the Kobo was shut down

Brief summary of the changes
----------------------------
I added a small indicator in the top right corner of the power off screen to show at a glance how much battery the Kobo had when shut down, so one can make his considerations in order to see if recharging is needed without the hassle of turning it on and launching XCSoar.

kinda closes issue #457

### DIFF
--- a/src/Kobo/PowerOff.cpp
+++ b/src/Kobo/PowerOff.cpp
@@ -37,6 +37,8 @@ Copyright_License {
 #include "io/FileLineReader.hpp"
 #include "Resources.hpp"
 #include "Model.hpp"
+#include "Hardware/Battery.hpp"
+#include "Hardware/PowerInfo.hpp"
 
 #include <algorithm>
 #include <stdexcept>
@@ -84,10 +86,14 @@ DrawBanner(Canvas &canvas, PixelRect &rc)
   canvas.DrawText({x, rc.top + int(banner_height - normal_font.GetHeight()) / 2},
                   website);
 
-  const TCHAR *const comment = _T("powered off");
-  canvas.DrawText({rc.right - (int)canvas.CalcTextWidth(comment) - padding, rc.top + padding},
-                  comment);
+  TCHAR comment[30] = _T("powered off");   
+  const auto power_info = Power::GetInfo();
+  if (power_info.battery.remaining_percent) {
+    snprintf ( comment+strlen(comment), 30-strlen(comment), _T(" - battery %d%%"), *power_info.battery.remaining_percent);  
+  }
 
+  canvas.DrawText({rc.right - (int)canvas.CalcTextWidth(comment) - padding, rc.top + padding},
+                          comment);
   rc.top += banner_height + 8;
 }
 
--- a/build/kobo.mk
+++ b/build/kobo.mk
@@ -52,6 +52,7 @@ KOBO_POWER_OFF_SOURCES = \
 	$(SRC)/Hardware/RotateDisplay.cpp \
 	$(SRC)/Hardware/DisplayDPI.cpp \
 	$(SRC)/Hardware/DisplaySize.cpp \
+	$(SRC)/Hardware/Battery.cpp \
 	$(SRC)/Screen/Layout.cpp \
 	$(SRC)/Logger/FlightParser.cpp \
 	$(SRC)/Renderer/FlightListRenderer.cpp \


lordfolken:
Hi the only concern I have is the hardcoded path in there. I foresee breakage in different kobi models. Isn't there a generic interface we could query?

MaxKellermann:
> Hi the only concern I have is the hardcoded path in there. I foresee breakage in different kobo models. Isn't there a generic interface we could query?

He actually copied code from Hardware/Battery.cpp, so that aspect ok. But creating redundant copies of existing code is not ok.
Even inside this PR, there's redundant code - the `DrawText()` line is there twice for no reason.

src/Kobo/PowerOff.cpp
```diff
@@ -84,9 +85,18 @@ DrawBanner(Canvas &canvas, PixelRect &rc)
   canvas.DrawText({x, rc.top + int(banner_height - normal_font.GetHeight()) / 2},
                   website);
 
-  const TCHAR *const comment = _T("powered off");
-  canvas.DrawText({rc.right - (int)canvas.CalcTextWidth(comment) - padding, rc.top + padding},
-                  comment);
+  TCHAR battery[4];
+  if (File::ReadString(Path("/sys/class/power_supply/mc13892_bat/capacity"), battery, sizeof(battery))){
```
MaxKellermann, line 89: This duplicates code from `Hardware/Battery.cpp` - if you need this, better use that library; the library is hard to use, because it writes to global variables, which is bad, but you can move the Kobo specific part to a separate library e.g. `Hardware/KoboBattery.cpp` in a separate commit, and amend this PR's existing commit to use that library.

src/Kobo/PowerOff.cpp
```diff
@@ -84,9 +85,18 @@ DrawBanner(Canvas &canvas, PixelRect &rc)
...
+          TCHAR comment[] = _T("powered off - battery ");
+          strcat(comment, battery);
+          strcat(comment, "%");
```
MaxKellermann, line 92: Stack buffer overflow bug here, may crash.

src/Kobo/PowerOff.cpp
```diff
@@ -84,9 +85,18 @@ DrawBanner(Canvas &canvas, PixelRect &rc)
...
+          canvas.DrawText({rc.right - padding - (int)canvas.CalcTextWidth(comment), rc.top + padding}, 
+                          comment);
+  }else{
+          const TCHAR *const comment = _T("powered off");
+          canvas.DrawText({rc.right - (int)canvas.CalcTextWidth(comment) - padding, rc.top + padding},
+                          comment);
```
MaxKellermann, line 98: Duplicate code.

MaxKellermann (changes requested):

src/Kobo/PowerOff.cpp
ulipo: I'm afraid this is way over my skills :-(

src/Kobo/PowerOff.cpp
ulipo: Defining a size for the variable "comment" should be enough, right? Like:
`TCHAR comment[30];`

MaxKellermann: If you believe that's enough, then yes - and please use `snprintf()` instead of `strcat()` - the latter is a bad API.

src/Kobo/PowerOff.cpp
MaxKellermann: OK I'll do that.

src/Kobo/PowerOff.cpp
```diff
@@ -84,10 +85,13 @@ DrawBanner(Canvas &canvas, PixelRect &rc)
   canvas.DrawText({x, rc.top + int(banner_height - normal_font.GetHeight()) / 2},
                   website);
 
-  const TCHAR *const comment = _T("powered off");
+  TCHAR comment[30] = _T("powered off");
+  TCHAR battery[4];
+  if (File::ReadString(Path("/sys/class/power_supply/mc13892_bat/capacity"), battery, sizeof(battery))){
+          snprintf ( comment+strlen(comment), 30, " - battery %s%%", battery);
```
MaxKellermann, line 91: This is still a stack buffer overflow, because you pretend the buffer is 30 bytes **plus** the current length of the comment.

MaxKellermann:
I've refactored the existing battery code to a function without global variables - now you can do:
```
#include "Power/Battery.hpp"
#include "Power/Info.hpp"
// ...
const auto power_info = Power::GetInfo();
if (power_info.battery.remaining_percent_valid)
  // ...
```

ulipo:
> I've refactored the existing battery code to a function without global variables - now you can do:
> 
> ```
> #include "Power/Battery.hpp"
> #include "Power/Info.hpp"
> // ...
> const auto power_info = Power::GetInfo();
> if (power_info.battery.remaining_percent_valid)
>   // ...
> ```
I guess I'm missing some commits, I have no trace of Power/Battery.hpp or Power/Info.hpp...

i tried to do something similar to SystemStatusPanel.cpp but it's not working (and I admit I'm too ignorant to understand why):
`  TCHAR comment[30] = _T("powered off");   
  const auto power_info = Power::Info();
  if (power_info.battery.remaining_percent) {
    snprintf ( comment+strlen(comment), 30-strlen(comment), _T(" - battery %d%%"), *power_info.battery.remaining_percent);  
  }
`
Like that it outputs just "powered off", if I comment out the IF block I get "powered off - battery 255%" (should be 100%)

MaxKellermann:
Your code creates a default-initialised `Power::Info` instance, and that creates an instance where `remaining_percent` is unset. Now by removing the "valid" check, you dereference the value, which is undefined behaviour, and thus gives you an arbitrary value. That cannot possibly work.
Instead of using a default-initialised value, call `Power::GetInfo()`, which attempts to load real values.

ulipo:
I'm trying this:
```
#include "Hardware/Battery.hpp"
#include "Hardware/PowerInfo.hpp"
[...]
  const auto power_info = Power::GetInfo();
  if (power_info.battery.remaining_percent) {
    snprintf ( comment+strlen(comment), 30-strlen(comment), _T(" - battery %d%%"), *power_info.battery.remaining_percent);  
  }
```
but it seems I can't get my legs out of it, I only get this:
`src/Kobo/PowerOff.cpp:93: error: undefined reference to 'Power::GetInfo()'`

Sorry for my noobness...

MaxKellermann:
You have to add `src/Hardware/Battery.cpp` to `KOBO_POWER_OFF_SOURCES` in `build/kobo.mk`.

ulipo:
> You have to add `src/Hardware/Battery.cpp` to `KOBO_POWER_OFF_SOURCES` in `build/kobo.mk`.

I feel even _noober_ :-)
Thanks!

PR #743, opened 2021-10-04
WaypointListBuilder: Ensure uniqueness of waypoints in dialogs

xtophe80:
Brief summary of the changes
----------------------------

Ensure waypoint uniqueness in dialogs.
With both shortname and name in the tree, it could be returned twice.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -1,4 +1,6 @@
 Version 7.20 - not yet released
+*user interace
+ - ensure uniqueness in waypoint picker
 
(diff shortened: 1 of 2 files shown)


MaxKellermann:
This scales `O(n²)` which is *very* bad. We need to have a better solution.

xtophe80:
Closing as fixed more eleguantly in https://github.com/XCSoar/XCSoar/commit/98a76d315f38ebc3cbfd560605f9da7ad458bcc9

PR #766, opened 2021-10-23
Extend driver to sync also QNH to XCVario device

iltis42:
Brief summary of the changes:
add PutQNH() method to XCVario device driver in order to sync QNH value set in XCS to the device.

Related issues and discussions:
Meanwhile MC, Ballast and Bugs were supported from inherited CAI302 commands since begin, here comes the missing part for QNH sync with a new implementation in recent XCVario code:
https://github.com/iltis42/XCVario/commit/86b99b0a15636247bd31b195a081ffad75a216a6

### DIFF
--- a/src/Device/Driver/XCVario.cpp
+++ b/src/Device/Driver/XCVario.cpp
@@ -27,6 +27,7 @@ Copyright_License {
 #include "Units/System.hpp"
 #include "NMEA/Checksum.hpp"
 #include "NMEA/Info.hpp"
+#include "Device/Port/Port.hpp"
 #include "NMEA/InputLine.hpp"
 #include "util/Clamp.hpp"
 #include "Atmosphere/Pressure.hpp"
@@ -44,6 +45,7 @@ public:
   bool PutMacCready(double mc, OperationEnvironment &env) override;
   bool PutBugs(double bugs, OperationEnvironment &env) override;
   bool PutBallast(double fraction, double overload, OperationEnvironment &env) override;
+  bool PutQNH(const AtmosphericPressure &pres, OperationEnvironment &env) override;
 };
 
 /*
@@ -157,6 +159,20 @@ XVCDevice::ParseNMEA(const char *String, NMEAInfo &info)
   else
     return false;
 }
+
+// For documentation refer to chapter 10.1.3 Device Driver/XCVario in mulilingual handbook: https://xcvario.de/handbuch
+
+bool 
+XVCDevice::PutQNH(const AtmosphericPressure &pres, OperationEnvironment &env)
+{
+  /* the XCVario understands "!g,q<NNNN>" command for QNH updates with recent builds */
+  char buffer[32];
+  unsigned qnh = uround(pres.GetPascal()/100);
+  int msg_len = sprintf(buffer,"!g,q%u\r", std::min(qnh,(unsigned)2000));
+  return port.FullWrite(buffer, msg_len, env, std::chrono::seconds(2) );
+}
+
+
 
 bool
 XVCDevice::PutMacCready(double mac_cready, OperationEnvironment &env)


hjr:
Phillipp, could you please kick the Codacy check again. Cannot believe that there is a real reason for failing.

hjr:
Ok, seems the strlen Codacy dont like. Here my suggestion avoiding the strlen() and get things moving ..
 ...
  int msg_len = sprintf(buffer,"!g,q%u\r", std::min(qnh,(unsigned)2000));
  if ( msg_len > 0 ) {
    return port.FullWrite(buffer, msg_len, env, std::chrono::seconds(2) );
  }
  return false;

iltis42:
> Ok, seems the strlen Codacy dont like. Here my suggestion avoiding the strlen() and get things moving .. ... int msg_len = sprintf(buffer,"!g,q%u\r", std::min(qnh,(unsigned)2000)); if ( msg_len > 0 ) { return port.FullWrite(buffer, msg_len, env, std::chrono::seconds(2) ); } return false;

green now, thanks Helmut !
And all changes as requested are in, so kindly request re-review.

src/Device/Driver/XCVario.cpp
```diff
@@ -158,6 +160,18 @@ XVCDevice::ParseNMEA(const char *String, NMEAInfo &info)
...
 }
 
+bool 
+XVCDevice::PutQNH(const AtmosphericPressure &pres, OperationEnvironment &env)
+{
+  /* the XCVario understands "!g,q<NNNN>" command for QNH updates with recent builds */
+  char buffer[16] = { 0 };
+  unsigned qnh = round(pres.GetPascal()/100);
+  sprintf(buffer, "!g,q%u\r", qnh);
+  return port.FullWrite(buffer, strnlen(buffer,16), env, std::chrono::seconds(2) );
```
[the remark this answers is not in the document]
iltis42: Codacy warning solved by proposal from @hjr

src/Device/Driver/XCVario.cpp
```diff
@@ -158,6 +160,18 @@ XVCDevice::ParseNMEA(const char *String, NMEAInfo &info)
     return false;
 }
 
+bool 
+XVCDevice::PutQNH(const AtmosphericPressure &pres, OperationEnvironment &env)
+{
+  /* the XCVario understands "!g,q<NNNN>" command for QNH updates with recent builds */
+  char buffer[16] = { 0 };
```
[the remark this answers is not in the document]
iltis42: codacy red flag fixed by proposal from @hjr what works perfectly.

src/Device/Driver/XCVario.cpp
```diff
@@ -158,6 +160,23 @@ XVCDevice::ParseNMEA(const char *String, NMEAInfo &info)
...
+// For documentation refer to chapter 10.1.3 Device Driver/XCVario in mulilingual handbook: https://xcvario.de/handbuch
+
+bool 
+XVCDevice::PutQNH(const AtmosphericPressure &pres, OperationEnvironment &env)
+{
+  /* the XCVario understands "!g,q<NNNN>" command for QNH updates with recent builds */
+  char buffer[32];
+  unsigned qnh = uround(pres.GetPascal()/100);
+  int msg_len = sprintf(buffer,"!g,q%u\r", std::min(qnh,(unsigned)2000));
+  if ( msg_len > 0 ) {
```
MaxKellermann, line 172: What's the point of this check?

src/Device/Driver/XCVario.cpp
```diff
@@ -158,6 +160,23 @@ XVCDevice::ParseNMEA(const char *String, NMEAInfo &info)
...
+    return port.FullWrite(buffer, msg_len, env, std::chrono::seconds(2) );
+  }
+  return false;
```
MaxKellermann, line 175: How is this reachable?

src/Device/Driver/XCVario.cpp
iltis42: Avoid to send once formatting to buffer fails.

MaxKellermann: Under what circumstances will this fail?

src/Device/Driver/XCVario.cpp
iltis42: The idea behind this was to check formatting errors. I agree, in this trivial case this should not happen for possible range of an integer 0..2000 , the sprintf %u should of course always work. We can of course remove this double check, okay ?

src/Device/Driver/XCVario.cpp
iltis42: I see the point, we can remove it, see also my other comment, okay ?

hjr: Then lets rely on sprintf. No problem, remove the extra check.

src/Device/Driver/XCVario.cpp
MaxKellermann: What kind of formatting error would make `sprintf()` fail? I ask because maybe you have knowledge about this that I don't have, and then we need to fix a thousand `printf()` calls in XCSoar. If you can't explain how it can fail, then we don't need the check.

Of course, I know how printf can fail - but such failures remain undetected; for example, if the buffer is too small. `sprintf()` will overwrite the buffer bounds, which can lead to all kinds of behavior, but *not* to a negative return value. `snprintf()` on the other hand will silently truncate the output, which also doesn't return a negative value. There are various other kinds of mistakes, like not enough arguments, wrong argument types etc. - but none of these lead to a negative return value, but to other random behavior, including crashes. All of these are not runtime errors, but code bugs, therefore a runtime check is the wrong way to approach this.

iltis42: Great explanation, you had already won, change is addressed in here:
https://github.com/XCSoar/XCSoar/pull/766/commits/b79114fc91f61eabcb363d4a2b44a65b7489bbc1

MaxKellermann: This isn't about winning - when I see code that looks strange, I want to know the author's motivation, to see whether they had something in mind that I was not aware about. I want to learn and self-improve, that's why I ask seemingly stupid questions. When it turns out that the code was just strange and unnecessarily complex with no motivation/reason at all, like here, I havn't "won". Oh yeah, I've been right after all, but who cares. Only you have "won", because it was your chance to learn something, you got your takeaway, but I got nothing.

iltis42: Me and @hjr appreciate the way you help not only us to improve things and understand how to get out of the way of bugs, issues or simply performance limiting unneeded code, by asking the right questions. The next pull request shall consider all the lessons learned!

MaxKellermann:
There are now 10 commits, none of which comes with a proper commit message, mostly incremental fixups for mistakes in earlier commits in this PR - but it looks like this whole PR is just one single change, and should not be splitted. Can you please fold the fixup commits? I think in the end, only one commit will remain.

iltis42:
> There are now 10 commits, none of which comes with a proper commit message, mostly incremental fixups for mistakes in earlier commits in this PR - but it looks like this whole PR is just one single change, and should not be splitted. Can you please fold the fixup commits? I think in the end, only one commit will remain.

Here the wrapped up commit with a proper message and branch reset before incremental development:
https://github.com/XCSoar/XCSoar/pull/766/commits/a46f2eaceb13883fbef508514b77e0d509687a69

src/Device/Driver/XCVario.cpp
```diff
@@ -158,6 +160,20 @@ XVCDevice::ParseNMEA(const char *String, NMEAInfo &info)
...
 }
 
+// For documentation refer to chapter 10.1.3 Device Driver/XCVario in mulilingual handbook: https://xcvario.de/handbuch
+
+bool 
+XVCDevice::PutQNH(const AtmosphericPressure &pres, OperationEnvironment &env)
+{
+  /* the XCVario understands "!g,q<NNNN>" command for QNH updates with recent builds */
+  char buffer[32];
+  unsigned qnh = uround(pres.GetPascal()/100);
```
MaxKellermann, line 170: Look at the implementation of `GetPascal()`:
https://github.com/XCSoar/XCSoar/blob/c3997f5458039805f9a09727a89eae11132c8084/src/Atmosphere/Pressure.hpp#L90-L92
If you want hPa, and there's a method called `GetHectoPascal()`, it's most likely the most efficient way to obtain the hectopascal value.
I'll merge this code now, but I'd also welcome a patch which optimizes this line. It would also make the code more readable!

MaxKellermann: Btw. usually, the compiler would not be allowed to optimize this `value / 100 * 100` away into just `value`; but since XCSoar uses `-ffast-math`, we allow the compiler to do such unsafe optimizations. It's not guaranteed that this optimization actually happens, though.

iltis42: Sure, we have meanwhile another change in the pipeline, more precision of ballast to cover this problem ( https://github.com/iltis42/XCVario/issues/74 ), that we needed to address as well in XCS/XCVario.cpp, so i will move as well to GetHectoPascal() in here of course.

PR #774, opened 2021-11-05
Fix inverted cruise/circling mode in Borgelt and XCVario protocol

iltis42:
Brief summary of the changes
----------------------------
Invert Cruise/Circling mode in NMEA parsing of Borgelt B50 driver and same issue in XCVario code that took over wrong logic from there.

Actually the Cruise Mode is transmitted as a "1" in the corresponding bit in native Borgelt B50 protocol, even the comment in the code indicates this, hence the parsing in the corresponding driver does exactly vice and CIRCLING is chosen once there is a one reading in the 'G' field for cruise, indicated by a circling arrow in the Vario gauge in XCSoar. 

See here the manual from B50, and also the  
// inclimb/incruise 1=cruise,0=climb, OAT
comment in the source file which also says 1=cruise.

B50 Manual (see last page):
https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwiZuIGl8YH0AhVGqaQKHVX5CUsQFnoECAgQAQ&url=http%3A%2F%2Fwww.borgeltinstruments.com%2FB50man.pdf&usg=AOvVaw1SGtyWW7LyiW7bRSWJBlM7

### DIFF
--- a/src/Device/Driver/BorgeltB50.cpp
+++ b/src/Device/Driver/BorgeltB50.cpp
@@ -104,11 +104,11 @@ PBB50(NMEAInputLine &line, NMEAInfo &info)
   // inclimb/incruise 1=cruise,0=climb, OAT
   switch (line.Read(-1)) {
   case 0:
-    info.switch_state.flight_mode = SwitchState::FlightMode::CRUISE;
+    info.switch_state.flight_mode = SwitchState::FlightMode::CIRCLING;
     break;
 
   case 1:
-    info.switch_state.flight_mode = SwitchState::FlightMode::CIRCLING;
+    info.switch_state.flight_mode = SwitchState::FlightMode::CRUISE;
     break;
   }
 
--- a/src/Device/Driver/XCVario.cpp
+++ b/src/Device/Driver/XCVario.cpp
@@ -102,11 +102,11 @@ PXCV(NMEAInputLine &line, NMEAInfo &info)
   // inclimb/incruise 1=cruise,0=climb, OAT
   switch (line.Read(-1)) {
   case 0:
-    info.switch_state.flight_mode = SwitchState::FlightMode::CRUISE;
+    info.switch_state.flight_mode = SwitchState::FlightMode::CIRCLING;
     break;
 
   case 1:
-    info.switch_state.flight_mode = SwitchState::FlightMode::CIRCLING;
+    info.switch_state.flight_mode = SwitchState::FlightMode::CRUISE;
     break;
   }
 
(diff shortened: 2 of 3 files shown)


MaxKellermann:
Bad commit message. Please do not write PR text - instead, write better commit messages! PR text is almost never useful for anything.

iltis42:
> Bad commit message. Please do not write PR text - instead, write better commit messages! PR text is almost never useful for anything.

Actually it was the other way around, github offers the commit message, if there is just one commit as default, what i kept here. Not clear what you expect to improve in the commit message, there is not much more about that to fix the inverted sense of the cruise bit. May i kindly ask for advise what exactly to improve in this case?

Any option to retrigger the checks ? Unix compile failed here, not explainable from the code change and on may PC i can perfectly compile for unix.

MaxKellermann:
It would be enough to move the PR text into the commit message. The PR text is good, but text in PRs will be lost after merging. Therefore, PR text is useless, and everything about your commits should be in the commit message.
I have no idea what "github offers the commit message" means.

iltis42:
> have no idea what "github offers the commit message" means.

I meant, when opening a new PR, and there is only one change, github uses the commit comment from this one changeset as a default for the PR title message.

In this case the commit comment was "Fix inverted cruise/circling mode in Borgelt and XCVario protocol", and exactly same as the PR text, so IMHO there should nothing be lost. Correct me when i am wrong, thanks !

MaxKellermann:
> I meant, when opening a new PR, and there is only one change, github uses the commit comment from this one changeset as a default for the PR title message.

True, but at that point it was already too late - your commit message was insufficient. And you, too, thought it was insufficient, which is why I assume you wrote more text in this PR. That text should have been in the commit message.

> In this case the commit comment was "Fix inverted cruise/circling mode in Borgelt and XCVario protocol", and exactly same as the PR text, so IMHO there should nothing be lost. Correct me when i am wrong, thanks !

I did not say commit messages get lost. I said PR texts get lost.

iltis42:
> And you, too, thought it was insufficient,
yes, true also ;)  i see now your point, addressed now here: https://github.com/XCSoar/XCSoar/pull/774/commits/4fface91deb418719c5a125b76ed8d43e3a3cc4b

iltis42:
See now the issue on unix compile, the test apparently needs an adapation:
#     Failed test (test/src/TestDriver.cpp:TestBorgeltB50() at line 356)

iltis42:
Test adapted, and ready to merge from my point of view.

PR #778, opened 2021-11-08
Engine/Task: Increase default Safety MC from 0.5 to 2

DanD222:
When flying at optimal glide speed, a MC setting of 2 should give a pilot a bigger safety margin than 0.5 in case of unexpected sink.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -1,6 +1,7 @@
 Version 7.39 - not yet released
 * Android
   - fix Bluetooth device list on Android versions older than 12
+  - do not request permission WRITE_EXTERNAL_STORAGE on Android 11 or newer
 
 Version 7.38 - 2023/07/27
 * WeGlide
--- a/android/src/XCSoar.java
+++ b/android/src/XCSoar.java
@@ -104,7 +104,11 @@ public class XCSoar extends Activity implements PermissionManager {
     registerReceiver(batteryReceiver,
                      new IntentFilter(Intent.ACTION_BATTERY_CHANGED));
 
-    requestPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE, null);
+    /* WRITE_EXTERNAL_STORAGE has no effect on Build.VERSION_CODES.R
+       (Android 11 or newer); we request it on older versions so users
+       can keep using /sdcard/XCSoarData */
+    if (android.os.Build.VERSION.SDK_INT < Build.VERSION_CODES.R)
+      requestPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE, null);
   }
 
   private void quit() {


MaxKellermann:
This sounds like replacing one arbitrary value with another arbitrary value. Since that is configurable, I don't see a point of this change. Why 2, why not 4? Why not 5 or 10? What makes 2 more suitable than 1.5 or 2.5?

DanD222:
That is correct. It's probably however a slightly safer value for someone who is not yet very knowledgeable and doesn't change the default value. I don't have an answer as to the correct value here; I only aimed for an incremental improvement. One thing I wanted to avoid was to give the impression to an starting-off pilot that gliding close to MC 0 is somehow safe, "because XCSoar has it in its default settings".

So feel free to select a different value if you prefer, or to reject this; it's just my suggestion at what I see as an incremental improvement.

lordfolken:
We have the STF Risk Factor algorithm that adds a safety range to the MC when far away and close to the glide path.

In the manual Section:
6.7 Speed to fly with risk 

The very interesting paper to read on that is here: 
https://www.johnhcochrane.com/s/newmcred.pdf

Do you feel that adequatly covers your case of concern?

DanD222:
I don’t know. Reading the manual, it seems to me that the STF Risk Factor is related to the speed to fly commander, whereas the SafetyMC governs the reach calculations (providing that SafetyMC is selected for the Reach polar calculations). So I can still envision someone inexperienced painting themselves into a corner due to not understanding the implications and being given an impression that 0.5 is somehow safe. I would still tend to err on the side of caution in terms of reach, not only speed to fly. 


> “The STF risk factor reduces the MacCready setting used to calculate speed to fly as the glider gets low, in order to compensate for risk. Set to 0.0 for no compensation, 1.0 scales MC linearly with height (with reference to height of the maximum climb). If considered, 0.3 is recommended. See Section 6.7 for more details.”

> “When safety MC is enabled, this MacCready setting is used for reach calculations, task abort, alternates and for determining arrival altitude at airfields.”

> “When calculating the arrival heights of landable fields (for map dis-play purposes and in abort mode), a safety MacCready value can bespecified in the configuration settings. This safety value is set to zero by default. Larger values make the arrival height calculation more conservative”

lordfolken:
There is the flip side of this coin. setting it to 2.0 will remove quite a few landing options (non,green) from the landables dialog. This can also lead a novice to make a decision to fly to an outlandable field even though an airfield would be in reach.  

However i see your point. The problem is that the crowd that "just downloads the app" and goes flying with it is growing. Reading manuals and studying equipment is out. This however requires a different app design than we currently have.

An idea  to lessen this point a bit:  An intro dialog where you can select what kind of flying you do (pg/sailplane) and what kind of profile you fly: (novice,cross-country,competition). Depending on that selection this and other values could be adjusted. 

In the mountains we use 1:25 regardless of the glider you fly. Simply because of wind... so mc 2.0 won't even be enough there.

DanD222:
Yes. My concern is whether the inexperienced pilot who doesn’t change his default Safety MC will be able to reach those airfields safely. The answer is probably “probably yes”. The proposed tradeoff is changing “probaby yes” to “most likely yes”, but for a smaller group of landable sites. 

I would say that my issue here is that the value of 0.5 is for a parameter which is called “Safety MC”. If it’s intended as a safety parameter, then I would like to increase the probability of safely reaching the planned destination.

In this case where one would want to know the margins of what one can reach the answer would be of course to set the safety height to zero, but this is a parameter that requires knowledge of the implications (and even then, I landed out because I forgot to decrease my task MC for reach calculations in deteriorating conditions).

Mountains of course require a very specific approach, like you say. The proposed change is targeted at the other, more general-use cases.

Different experience profiles are probaby not a bad idea; what this PR aims for though is a “less than optimal solution now”, rather than a “more optimal solution some time in the future”.

groundhogxc:
I agree with Dan. Something with "safety" in the name should typically provide significant additional safety over the typical use setting. Which in the case of an MC value means significantly higher. There are not many circumstances where MC 0.5 provides significant "additional(!)" safety over even the lowest value (MC 0).

I would also argue to set the default higher. If not MC 2, maybe 1.5 (in the mountains I would not even count that as "safety MC"). Even 1 is better as a default than 0.5. Personally I never set a value smaller than 0.5 for the "regular" MC value other than in exceptional circumstances (late night glide home in case of tailwind being one of the few exceptions).

PR #785, opened 2021-11-22
Android background location prominent disclosure dialog

### LINKED ISSUE #618: Android 11 - New Way of requesting background location access
XCSoar is nothing short of brilliant, but the latest Android update has caused a real issue:  Android 11 seems to have restricted xc's access to the GPS sensor, when it runs in the background.

This means it can't produce a track or position unless you keep xc running in the foreground.  Which you can, but..

It stops you using xc just purely as an automatic flight tracker, with the screen off, which was brilliant for logging a day's local flights.  It also loses the gps if you for instance, access the camera app while running xc.

And it's not even XCSoar's fault...

zinuzoid:
Closes #618

Here's the video for Google Play Store: https://www.youtube.com/watch?v=h14BWgzDD0w

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -3,6 +3,8 @@ Version 7.21 - not yet released
   - Support for AF/AR frequency fields in OpenAir
 * devices
   - fix inverted cruise/circling mode in borgelt and xcvario driver
+* Android
+  - re-enable background location and comply with Google Play Store policy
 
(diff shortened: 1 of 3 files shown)


MaxKellermann:
I have submitted your video the Google. I hope it'll pass the next release upload. Thanks!

PR #786, opened 2021-11-23
io/FileOutputStream: merge win32 function Cancel

### LINKED ISSUE #782: XCSoar not saving igc file since version 7.xx

Which XCSoar version are you using when seeing the problem?  
7.19-PC, 7.20-PC
7.4 WIN64
7.2 WIN64
7.1 WIN64

Which XCSoar version is the last one you know that did not show the problem? 
6.8.17-PC

Expected behavior
-----------------
At the end of a task (in my case, virtual gliding flight) XCSoar should save an .igc file

Actual behavior
---------------
For version 6.xx, the file was saved correctly, in the XCSoarDat/Logs folder.  
For version 7.xx, I can see the igc file in the folder, while the program is running (while the flight is in progress), but as soon as XCSoar shuts down, the file disappears.

To reproduce the error, load a task (.tsk) file; run the simulation (fly the task).  While the flight/task is in progress, monitor the XCSoarData/Logs folder - the igc file is visible.
At the end of the flight/task, close XCSoar - file disappears!

This might give us ideas we haven't thought of, feel free to suggest causes. NO,  It worked in 6.xx (and still does); stopped working in 7.xx

lordfolken:
Win32 had a separate implementation for FileOutputStream::CANCEL Logic
was added to the other platform function, but forgotten in this one.

This merges the functions into one, as only the call for file deletion
is platform specific.

closes #782

### DIFF
--- a/src/io/FileOutputStream.cxx
+++ b/src/io/FileOutputStream.cxx
@@ -145,16 +145,6 @@ FileOutputStream::Commit()
 	Close();
 }
 
-void
-FileOutputStream::Cancel() noexcept
-{
-	assert(IsDefined());
-
-	Close();
-
-	DeleteFile(GetPath().c_str());
-}
-
 #else
 
 #include <cerrno>
@@ -274,6 +264,8 @@ FileOutputStream::Commit()
 	}
 }
 
+#endif
+
 void
 FileOutputStream::Cancel() noexcept
 {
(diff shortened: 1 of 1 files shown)


src/io/FileOutputStream.cxx
```diff
@@ -288,6 +280,8 @@ FileOutputStream::Cancel() noexcept
 #endif
 #ifdef __linux__
 			unlinkat(directory_fd.Get(), GetPath().c_str(), 0);
+#elif _WIN32
+	  DeleteFile(GetPath().c_str());
```
MaxKellermann, line 284: Bad indent

MaxKellermann:
789c4807108d529ca736a2da75a16462e30c5efc

Issue #611, opened 2021-05-06
Kobo Clara HD not anymore supported (touchscreen not working)


wolkenjaeger:
XCSoar versions having and not having the problem
-------------------------------------------------

6.8.17

Expected behavior
-----------------

Touchscreen input shall work

Actual behavior
---------------

Touchscreen inputs are not working. However, the physical button works to turn Kobo on and off.

Steps to reproduce the behavior
-------------------------------

Install Kobo 7.6 on a Kobo Clara HD

Do you have any idea what may have caused this?
-----------------------------------------------

Kobo Clara HD is not defined in the source code. 

Do you have an idea how to solve the issue?
-------------------------------------------

Add relevant definitions in the source code, maybe it is just a driver issue during boot. I am able to help with basic testing.

MaxKellermann:
This can be analyzed only if a developer has a Kobo Clara HD.

lpojzl:
@MaxKellermann: Max, I bought Kobo ClaraHD just for XCsoar. As it's currently not working, I am open for a 1/2 year loaner to you. Are you willing to have a look at the ClaraHD related issues? It would be great having it working before next season. ;-)
The device would come flashed with latest Kobo software and XCsoar build installed, together with non-intrusive serial2usb cable for console debugging.
![ClaraHD](https://user-images.githubusercontent.com/67234454/130317565-d1c7d0da-dfef-40b4-a78a-0e6030786a57.jpg)

sepi:
I also have the same issue on the Clara HD with the the 7.x versions. Will see if I can debug somehow. For now, version 6.8.11 works ok.

fagat73:
Hello,
Any news regarding this issue ? I've installer the latest XCSoar 7.20 version on my Kobo Clara HD and the touch screen not working anymore.
Thank you.

MaxKellermann:
@fagat73 of course not. If there were news, you would be able to read them here. No developer has a Kobo Clara HD, and there cannot ever be any progress until a developer has one.

fagat73:
I've got the Kobo but I'm not a developer on the projet. Anyway let me know if I can be of any help with resolving this issue.

lordfolken:
You can organize a crowd funding for a device for @MaxKellermann that way the project can support the platform. Without hardware it will be very difficult.

MaxKellermann:
You can help by either
- becoming a developer and solve the issue, or
- donating a Kobo Clara HD to the XCSoar project

[...]

Issue #515, opened 2021-01-17
Touchscreen inactive on Kobo Glo HD


druide862:
XCSoar versions having and not having the problem
-------------------------------------------------
Versions having the problem:
- 6.8.17
- 6.8.16
- 6.8.15
- 6.8.11

Expected behavior
-----------------
While in "Fly" mode, when I press the big **M** button, I expect the menu to show.

Actual behavior
---------------
When I start the Kobo, I first have the splash screen with FLY/SIM. I hit the FLY button, and at this time, the touchscreen is responding as expected. Then XCSoar loads lots of things (Maps, terrain, etc...) and eventualy displays the map with the gages and the big **M** button. Then, when I hit the big M button, nothing appends. Can't neither zoom in or out the map. Looks like the touchscreen is inactive (but it worked at the FLY/SIM splash screen). The kobo and xcsoar are working correctly otherwise: gages values are evolving, the map evolves as I move, etc.

So, I then power off the kobo by holding the power button 10s., and I start it again. Hit the FLY button, and then the touchscreen works as expected: **M** button shows the menu, etc...

If I power off/on the kobo within few minutes after the "hard" power off, the touchscreen works as expected **BUT**, if I let the kobo powered off a few hours, and power it on again, the problem comes back: Touchscreen inactive until I "hard" power it off.

Tested on 4 different kobo Glo HD with blue fly vario module.

Steps to reproduce the behavior
-------------------------------
- Install xcsoar 6.8.17 on a kiobo Glo HD with bluefly vario module
- power it off and let it powered off a few hours
- power it on
- hit the FLY button
- There you are: Touchscreen inactive
- Hold power button for 10s to hard power off the kobo
- Power it on again
- Hit the fly button
- Touchscreen now active (till next long powered off period...)

Do you have any idea what may have caused this?
-----------------------------------------------
No.

Do you have an idea how to solve the issue?
-------------------------------------------
Neither

lordfolken:
@GeekGoneOld this is now fixed?

GeekGoneOld:
> @GeekGoneOld this is now fixed?

As I mention in the commit, it may be fixed but I don't have a Glo HD so I can't be sure.  The particular problem mentioned in this seems to be somewhat different from any other reports so I doubt this fixes it.  It is quite possible that the problem is specific to the author of this issue and not experienced by anyone else.

@druide862 can you try the latest version of XCSoar and see if this issue is fixed?  If not, let us know if the symptoms have changed.

[...]

Issue #317, opened 2020-01-01
Kobo Clara HD blank screen


aa11bb22cc33dd44:
When turned on, the start screen is blank (but when you press the screen in the right place, menu options such as FLY and SIM still work)

lordfolken:
Version of XCSoar? Platform?

MarkusBagge:
Can confirm this with v6.8.13. After pressing the power-(hard)button, screen becomes empty/white. Drag your finger from bottom to top brings up the FLY/SIM startup screen. Easy if you know. Not easy if you don't :)

cefn:
I can confirm this also with v6.8.13 and v6.8.15 on Kobo Clara HD. Blank screen on startup but successfully triggering any button or interaction which makes the screen re-render presents the normal landing screen correctly. All subsequent writes to the screen and general XCSoar features seem to work for this model. 

It's as though the landing screen completed a render before the display driver had fully loaded for the Clara, hence that 'write' never made it to the screen. 

Where should I try to add a brief sleep to delay rendering the System screen in the case of Clara HD model?

I note that in https://github.com/XCSoar/XCSoar/blob/719cedbf614b037c75081fa86f7be3735fb0daf4/src/Kobo/Model.cpp#L58-L66 there isn't an entry for the Clara HD model so it's impressive that it works out of the box except for this.

I haven't tried getting a shell on the Kobo to be able to e.g. interactively run operations like... https://github.com/XCSoar/XCSoar/blob/719cedbf614b037c75081fa86f7be3735fb0daf4/src/Kobo/Model.cpp#L83
...is that a sensible way forward to be able to investigate the filesystem for this model and line up the Clara HD specific values  and driver invocations correctly?

cefn:
Wondering if it's to do with https://github.com/XCSoar/XCSoar/blob/e7066d5a716ac8d19bc1b7dec3a58820260360f3/src/Screen/FB/TopCanvas.cpp#L314-L315 potentially as the HD models seem to need to wait for a buffer sync. So introducing explicit model detection support could resolve a few things for the Clara HD including getting the DPI right.

flightninja:
From a bit of testing it appears (at least for me) that even forcing frame_sync = true for Clara HD, and hence calling Wait(); still results in the same blank screen issue at start up.

I'm wondering (only a guess) if this could possibly be because the Wait(); function makes use of an older 'API call', and it looks like the newer kernel / chipset in the Clara may potentially require newer 'API call'?

[...]

PR #787, opened 2021-12-02
Fix startup for Clara

GeekGoneOld:
Brief summary of the changes
----------------------------
This adds a 0.1s delay in rcS file for Kobo to allow touch screen and display to initialize properly.  This was specifically tested for Clara HD (N249) but other models have had similar problems (blank screen on startup or touch not active) and those models may now work better.

#611
#317
#515 (maybe, not tested by me)

### DIFF
--- a/kobo/rcS
+++ b/kobo/rcS
@@ -80,6 +80,11 @@ echo "dc 0" > /sys/devices/platform/pmic_light.1/lit
 echo '/mnt/onboard/XCSoarData/crash/core.%p' >/proc/sys/kernel/core_pattern
 ulimit -c unlimited
 
+# workaround for Kobo Clara HD: the touch screen doesn't respond
+# and the initial screen render is blank if we are too fast
+# delay should not cause any problem and might help other models
+usleep 100000
+
 # launch user script
 if [ -f /mnt/onboard/XCSoarData/kobo/init.sh ]; then
     source /mnt/onboard/XCSoarData/kobo/init.sh


MaxKellermann:
Oh, is the Kobo hardware really so badly bugged that this simple "sleep" fixes all problems?

About the PR: I like that you added an extensive comment to the script. But the commit message is bad; this PR contains text which is missing in the commit message. I'd rather have an empty PR with only commits, if the commits have proper commit messages.
The commit message should also have "Closes" tags for these 3 issues, so they get closed automatically by GitHub as soon as I merge your PR.

I need to verify your code with the Kobo I received yesterday.

GeekGoneOld:
Yes, "sleep" seems to fix a few things in Kobo.  You can see at least one
other in rcS.

I really HATE using timers to solve problems.  I prefer to confirm
completion of a dependency.  Unfortunately, I don't know how to find the
dependency in this case.

In the future, I will use the commit for my extensive comments and the PR
for a quick comment.

Glad to hear you have a Clara HD now.  I will be submitting a few more PRs
for other functions that don't work (Wifi, Export USB, OTG, reboot and tty
naming).

Again, Max, many thanks for your patience with me.

MaxKellermann:
So... will you improve the commit message?

MaxKellermann:
Yes. My merge commit:
```
Author: Max Kellermann <max.kellermann@gmail.com>  2021-12-03 18:37:52
```
Your force-push:
```
Committer: Keith Laidlaw <laidlaw1546@gmail.com>  2021-12-03 18:37:53
```
... phew, that was close!

MaxKellermann:
3d1a59d17ee707e67c37dff694ec9ed319fed520

PR #788, opened 2021-12-03
Allign box nomenclature for Wind info boxes.

hjr:
Followed suggestions from glider buddies not understanding the "Wind V", or "V Wind" headline in the box. 

Renaming the info box long names so that all wind related boxes group together.

### DIFF
--- a/doc/manual/en/ch10_infobox_reference.tex
+++ b/doc/manual/en/ch10_infobox_reference.tex
@@ -159,14 +159,14 @@ lift distribution over each part of the circle.}
 with the connected InfoBox dialogue. Pressing the up/down cursor keys to cycle through 
 settings, adjust the values with left/right cursor keys.}
 {figures/infobox-dialog-wind1.png}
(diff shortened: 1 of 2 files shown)


MaxKellermann:
Please do not generate `po/*` unless there is a specific reason to do so.

hjr:
Ok, was not sure on our preference on this any more. I intentionally covered all translation of the changed stringes, to not loose the still correct translation.
Should I seperate this, or remove entirely, because the next generation anyways will preserve the meaningfull .. ?

MaxKellermann:
So you did edit those PO files? If so, then those changes were hidden deep inside hundreds of completely unrelated changes.

hjr:
Actually just checked if po/* files are generated meaningful. Amended ..

MaxKellermann:
I have now regenerated all PO files and synced with Weblate. If you have any improvements for these files, you can either use Weblate or submit a GH PR. Just don't mix regenerated PO files with manual PO changes in one commit, that'll be very confusing.

Issue #676, opened 2021-07-31
USB-OTG setting on Clara HD crashes and bricks the Kobo


Boulderpilot1:
XCSoar versions having and not having the problem
-------------------------------------------------

 7.10, I don't know if other versions work 

Expected behavior 
-----------------

 Device should be able to run data from XC tracer III GPS to Kobo Clara HD using USB-OTG connection, as described on XC Tracer website

Actual behavior
---------------

As soon as you select USB-OTG, a prompt occurs "USB-OTG not supported for Clara HD, use at own risk?" Select yes and then Kobo immediately crashes and bricks. I re-imaged the SD to original image and retried from the beginning. The same problem occurred twice. Kobo remains bricked.

Steps to reproduce the behavior
-------------------------------

 After new install of XCSoar on Clara HD, select system, then enable USB-OTG, when prompted select "Yes" - this leads to a crash/brick

Do you have any idea what may have caused this?
-----------------------------------------------

No ideas

Do you have an idea how to solve the issue?
-------------------------------------------

No ideas

MaxKellermann:
This is not a bug. The message clearly states that the Kobo Clara HD is not supported, and you installed an incompatible kernel despite the warning. Of course it does not work.
To support the Kobo Clara HD, we need a volunteer to develop this, and that volunteer needs to have a Kobo Clara HD.

sepi:
@Boulderpilot1 is the device recoverable? Usually the term "bricked" means that the device is broken beyond repair and is now as good as a brick.

GeekGoneOld:
This has been fixed in 7.21.  The Enable USB-OTG button is greyed out for Clara HD.  Check commit ae4a195 on src/Kobo/SystemDialog.cpp

@MaxKellermann or @lordfolken  perhaps can close this issue?

PR #791, opened 2021-12-09
Disable OTG for Clara HD and update NEWS

GeekGoneOld:
Disable OTG for Clara HD to avoid bricking the unit since the kernel is not compatible

Brief summary of the changes
----------------------------

### DIFF
--- a/src/Kobo/SystemDialog.cpp
+++ b/src/Kobo/SystemDialog.cpp
@@ -58,7 +58,9 @@ SystemWidget::Prepare(ContainerWindow &parent, const PixelRect &rc) noexcept
   AddButton("Reboot", [](){ KoboReboot(); });
   AddButton(IsKoboOTGKernel() ? "Disable USB-OTG" : "Enable USB-OTG",
             [this](){ SwitchKernel(); });
-
+#ifdef KOBO
+  SetRowEnabled(SWITCH_KERNEL, DetectKoboModel() != KoboModel::CLARA_HD);
+#endif
   AddButton("Export USB storage", [this](){ ExportUSBStorage(); });
   SetRowEnabled(USB_STORAGE, !IsKoboOTGKernel());
 }
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -4,7 +4,7 @@ Version 7.21 - not yet released
 * devices
   - fix inverted cruise/circling mode in borgelt and xcvario driver
 * Kobo
-  - fix Clara HD touch screen
+  - Add support for Clara HD
 * Android
   - re-enable background location and comply with Google Play Store policy
 


MaxKellermann:
I think this message is rather pointless. It would be better to disable the button for unsupported Kobo models.
Why did you translate the dialog caption but not the error message?
More nitpicks:
- please rebase on master before you submit a PR (and rebase each time you re-push)
- git commit messages should start with a brief one-liner mentioning the subsystem and a very short description; then an empty line, and then (if needed) a longer description. Your commit message is just a long description which doesn't fit a line, look at the PR title

GeekGoneOld:
I'm happy to change it to disable instead.  I chose to translate only the message because I was following the exact style of the existing programming which only translated the message.  Also I missed entering the PR title by accident.  I will fix.  Sorry about the omission of rebase.  Please give me a few minutes before pulling the changes because it could take me more than one try to get the rebase and commit right.  I will post when I have completed my attempt.

MaxKellermann:
Rebase, not merge. A PR should never contain a merge commit.

src/Kobo/SystemDialog.cpp
```diff
@@ -70,8 +73,7 @@ SystemWidget::SwitchKernel()
   KoboModel model = DetectKoboModel();
   if (model == KoboModel::CLARA_HD)
   {
-    ShowMessageBox(_T("This feature is not yet available for Clara HD."), _("Error"),
-                    MB_OK);
+    switch_kernel_button->SetEnabled(0);
```
MaxKellermann, line 76: This doesn't make sense. It disables the button *after* the user has successfully pressed it.

MaxKellermann:
Now this PR contains two commits, and the second commit undoes the code added by the first one.

src/Kobo/SystemDialog.cpp
GeekGoneOld: Stupid me.  I mistakenly didn't test it because I was in git helltrying to fix my pull vs rebase mistake.  A few more minutes...

MaxKellermann:
You might want to try stgit (https://stacked-git.github.io/) to fix "git hell". I can't live without stgit, seriously. I use it all day, and it helps so much, I don't even need any branches other than "master".

GeekGoneOld:
I will look into stacked-git.

I have another commit to add to this PR shortly (compiling, but long compile due to rebase).  Would you prefer to discard this PR altogether and I just submit a clean one with the proposed changes or are you happy adding another commit to this one?

MaxKellermann:
You never need to close a PR to resubmit a new one. You can always do something like `git reset --hard origin/master`, do new commits, and force-push to this PR branch. (That's where stgit will make your life much easier, but everything is possible with plain git.)
For merging, I'd prefer not merging known-bad commits together with fixup commits.

GeekGoneOld:
I will do the reset.  Please be patient...

GeekGoneOld:
I think I have it all right now.  Tested on Clara HD successfully.

src/Kobo/SystemDialog.cpp
```diff
@@ -40,6 +40,8 @@ class SystemWidget final
     USB_STORAGE,
   };
 
+  Button *switch_kernel_button;
```
MaxKellermann, line 43: This variable is only used inside `Prepare()`, so better make it a local variable.

src/Kobo/SystemDialog.cpp
```diff
@@ -56,11 +58,15 @@ void
...
-
+  switch_kernel_button = AddButton(IsKoboOTGKernel() ?
+                                   "Disable USB-OTG" : "Enable USB-OTG",
+                                   [this](){ SwitchKernel(); });
   AddButton("Export USB storage", [this](){ ExportUSBStorage(); });
   SetRowEnabled(USB_STORAGE, !IsKoboOTGKernel());
+#ifdef KOBO
+  if (DetectKoboModel() == KoboModel::CLARA_HD)
+    switch_kernel_button->SetEnabled(0);
+#endif
```
MaxKellermann, line 69: Too complicated. Look at the existing code line above that - you can simply call `SetRowEnabled()`, and then you don't even need the variable.
Also note that `0` is not a `bool`. It gets casted implicitly to `bool`, but it's the wrong parameter type.

flightninja:
I see this PR is to disable the OTG button on CLARA_HD - I assume to avoid issues like this one (https://github.com/XCSoar/XCSoar/issues/676)?

Interestingly, according to this comment https://github.com/XCSoar/XCSoar/issues/317#issuecomment-743950245, the `CONFIG_USB_OTG=y` is set on the default CLARA_HD kernel, but other drivers are not enabled. I'm not sure if this will ever make it easier or not to support OTG on CLARA_HD in the future, or if a new kernel would still need to be compiled with additional flags?

MaxKellermann (changes requested):

MaxKellermann:
This PR changes only affects the Kobo Clara HD, but what makes this model so special that we need to disallow the user to make this decision here:
https://github.com/XCSoar/XCSoar/blob/78aa74fc2f0bedb7119f72dbef19f494a10094b4/src/Kobo/SystemDialog.cpp#L76-L77
Either we believe that we can't trust the user's decision, then this should apply to all unsupported models. Or we trust the user, then this whole change doesn't make sense. In any case, this PR is not consistent.
By the way, I don't like the whole `SwitchKernel()` function, because there are two checks for the Kobo model: once for the `ShowMessageBox()` call, and another one for the kernel image selection (`switch/case`). I hate code duplication. Adding a third copy to `Prepare()` (which this commit does) makes this only worse. There should only be one copy of the model-checking code for OTG kernel. Maybe one function, or one table (i.e. C++ array of structs), like this:
```
struct KoboKernel {
  KoboModel model;
  const char *regular_kernel_path, *otg_kernel_path;
};
static constexpr KoboKernel kobo_kernels[] = {
  { KoboModel::AURA2, "/opt/xcsoar/lib/kernel/uImage.aura2", "/opt/xcsoar/lib/kernel/uImage.aura2.otg" },
  // ...
}
static constexpr const char *
GetOtgKernel(KoboModel m) noexcept
{
  for (const auto &i : kobo_kernels)
    if (i.model == m)
      return i.otg_kernel_path;
  return nullptr;
}
```
Then you can do:
```
   SetRowEnabled(SWITCH_KERNEL, GetOtgKernel(DetectKoboModel()) != nullptr);
```
That gives you only one source of truth for whether OTG is supported with the current model.

GeekGoneOld:
Should be MUCH cleaner now.  Thanks very much for keeping me, and therefore the codebase, to a very high standard.

MaxKellermann:
It's not good code if there's nothing left to add; it's good code if there's nothing left to *remove*. (Old saying.)

If the commit diff looks simple, it's a good commit. This one is.

GeekGoneOld:
You asked why we are treating Clara HD as special.  Good question, two answers.
1. It is special.  Clara HD has OTG in the kernel and no kernel switching is required to change modes.  You merely have to write "host" or "gadget to devfs ci_hdrc.0/role.  This code was a temporarily put in to start the special case of Clara_HD.  More to come, but I have to build some kernel modules first.
2. I didn't want to make any changes to other Kobo models as I don't have any to test.  I don't actually know which ones work with the default kernels so I don't even know which ones are actually not bricked by the default.

flightninja:
@GeekGoneOld
1. Really interesting, thank you for explaining.
2. Also interesting. I have not tried much kernel work, apart from pulling the CLARA_HD kernel sources from Kobolabs Github (https://github.com/kobolabs/Kobo-Reader/tree/master/hw) about a year ago, and changing some of the config flags and then try rebuild it. I never managed to try installing it as I got stuck with differences between it being a zImage v.s. a uImage and I didn't want to brick mine. I also recall having some trouble with the tool-chain, and I think I was only able to build the CLARA_HD kernel using the "gcc-linaro-4.9.4-2017.01-x86_64_arm-linux-gnueabihf" tool-chain - this is supplied by Kobolabs also on their Github (https://github.com/kobolabs/Kobo-Reader/tree/master/toolchain). It seems Kobolabs have updated with newer CLARA_HD kernel sources in last 6month too by looks of their repo.

PR #792, opened 2021-12-09
Screen/Layout: increase padding for text boxes

lordfolken:
The padding is quite compressed in the current version:
2021-12-09-202708_640x480_scrot

This change improves the layout:
2021-12-09-202602_640x480_scrot

### DIFF
--- a/src/Screen/Layout.cpp
+++ b/src/Screen/Layout.cpp
@@ -115,7 +115,7 @@ Layout::Initialize(PixelSize new_size, unsigned ui_scale, unsigned custom_dpi) n
        the viewing distance is usually smaller */
     font_scale = font_scale * 2 / 3;
 
-  text_padding = VptScale(2);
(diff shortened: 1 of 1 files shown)


MaxKellermann:
The screenshot looks good, and such a change sure is necessary - but tweaking such central constants is dangerous; does this look good on all screen resolutions and layouts? We need dynamic layouts that work well on everything.
This global constant is used for many different things, not just strictly text padding. If we find out it has been over-used and your change does not work well for other uses, we should split it up into different globals, before doing this change.

lordfolken:
I did some testing with -small. The main issue is that the rowwidgets are now taller with this change. There are some config panels that are no longer touch accessible in expert mode. This however was already the case with for example the tracking configuration panel.

lordfolken:
I'll revisit this later. There are several aspects and a lot of refactoring needed.

PR #793, opened 2021-12-09
Control backlight brightness of the Kobo Glo HD in the system dialog

schmidi:
Brief summary of the changes
----------------------------

Enable and control the backlight of a Kobo Glo HD from the system dialog.

### DIFF
--- a/src/Kobo/SystemDialog.cpp
+++ b/src/Kobo/SystemDialog.cpp
@@ -38,7 +38,12 @@ class SystemWidget final
     REBOOT,
     SWITCH_KERNEL,
     USB_STORAGE,
+    INCREASE_BACKLIGHT_BRIGHTNESS,
+    DECREASE_BACKLIGHT_BRIGHTNESS
   };
+
+  Button *increase_backlight_brightness;
+  Button *decrease_backlight_brightness;
 
 public:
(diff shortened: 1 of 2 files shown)


src/Kobo/SystemDialog.cpp
```diff
@@ -63,6 +71,13 @@ SystemWidget::Prepare(ContainerWindow &parent, const PixelRect &rc) noexcept
 #endif
   AddButton("Export USB storage", [this](){ ExportUSBStorage(); });
   SetRowEnabled(USB_STORAGE, !IsKoboOTGKernel());
+
+  if(KoboCanChangeBacklightBrightness()) {
+    increase_backlight_brightness = AddButton("Increase Backlight Brightness", [this]() { IncreaseBacklightBrightness(); });
+    decrease_backlight_brightness = AddButton("Decrease Backlight Brightness", [this]() { DecreaseBacklightBrightness(); });
```
MaxKellermann, line 77: These lines will be executed conditionally, but if `!KoboCanChangeBacklightBrightness()`, then those control indexes don't exit. That's not a problem now, but for the next guy who adds new controls to this form.
To avoid this, add `RowFormWidget::AddDummy()` calls to the `else` branch. This reserves index numbers, but doesn't actually add controls.

PR #797, opened 2021-12-15
added polars for DG-800S, the pure glider version

iltis42:
Brief summary of the changes
----------------------------

Two new Polars for the DG-800S 15m and 18m pure glider version that comes with a lower mass what needs to be reflected with a new polar based on the lower reference weight. Polar taken from manufacturer handbook pages of DG-800S sent by the owner (DG copyright). 

none yet.


hjr:
Hey Max, Iltis,
new polars in our PolarStore should be added to the very end of the list. The profile is keeping the Index to a polar and we would urge almost every buddy to fix their profile adding somewhere in the middle.
Please correct before releasing.

Helmut

MaxKellermann:
Thanks, you are right, there's even a code comment explaining this:
https://github.com/XCSoar/XCSoar/blob/16ed59aa4a6c1bf4533a0a8583374dd9ff369081/src/Polar/PolarStore.cpp#L61-L62
I'll revert the commit.

MaxKellermann:
Reverted: 74fdc09a7dd3022c05e55308cff723dd6b862743
@iltis42 please resubmit

hjr:
And by the way .. the "dry mass" setting is exactly meant to compensate for the correct individual glider mass in case the aerodynamics of the glider are just the same.

MaxKellermann:
Indeed, probably many items in our existing list are redundant already. That's bad and confusing.
There are too many myths and fairytales about glider performance already in people's minds, and XCSoar shouldn't add more confusion to that problem.

iltis42:
Okay, in XCV i have solved the list issue by a dedicated index per polar with some gap in between, maybe a future improvement here. For now i will push this to the end, and if the idea with an index gets common sense, i propose to migrate this at times.

        { 1400, "DG 800C/18",39.96,106.00,-0.620,171.75,-1.466,214.83,-2.346,150,11.81},
	{ 1410, "DG 1000S/20",28.00,80,-0.51,125,-0.93,175,-2.80,160,17.53},
	{ 1420, "Diamant 18",28,72,-0.52,125,-0.95,160,-1.75,50,14.28},
	{ 1425, "Diana2",32.10,75,-0.48,120,-0.82,180,-2.22,240,8.64},
	{ 1430, "Discus A",29.67,80,-0.56,125,-1.00,175,-2.55,200,10.58},
	{ 1440, "Discus 2B",51.7,100,-0.75,125,-0.80,175,-1.50,200,10.16},
	{ 1450, "DUO Discus",30.5,80,-0.52,130,-1.00,177,-2.50,198,16.4},

iltis42:
> Indeed, probably many items in our existing list are redundant already. That's bad and confusing. There are too many myths and fairytales about glider performance already in people's minds, and XCSoar shouldn't add more confusion to that problem.

The drymass is fine for that purpose, hence just mentioning this is not the whole story: For the pure glider version it defaults to the reference weight what is not correct for this glider. I am not sure if users are well aware of that, just now i have one on the line who fell into that trap, crazy things happen in case, e.g. water dump went crazy :)

And the non motorized version has its own type label, DG-800S, not DG-800, and has explicit measured Idaflieg Polars, so why not using it ?  The dry weight is correct then, and the polar is exact as taken from user manual for this type, and we remove confusion for DG-800S owners what polar to select (as it is not in up to now), and what exactly to adjust there and how. My 5 cent to this discussion.

hjr:
The mentioned water dump issue is unrelated. NMEA sentence passing a ballast percentage not having agreed on the unit (here the reference mass) does just not work well across system boundaries.

hjr:
Handling masses in XCS could be improved. LX even option for pilot, copilot, ballast and trim. In addition to the already present dry mass to fit the polar to a specific glider those other items would be a perfect addition to the XCS flight setup.

PR #798, opened 2021-12-15
added polars for DG-800S, the pure glider version, at the end of pola…

iltis42:
Brief summary of the changes

Two new Polars for the DG-800S 15m and 18m pure glider version that comes with a lower mass what needs to be reflected with a new polar based on the lower reference weight. Polar taken from manufacturer handbook pages of DG-800S sent by the owner (DG copyright), added at the end of list.

See https://github.com/XCSoar/XCSoar/pull/797

### DIFF
--- a/src/Polar/PolarStore.cpp
+++ b/src/Polar/PolarStore.cpp
@@ -305,6 +305,9 @@ static constexpr PolarStore::Item internal_polars[] =
   // flight manual http://www.issoire-aviation.fr/doc_avia_gen/MdV_WA26P_R2.pdf
   // Contest handicap reference: http://docplayer.fr/79733029-Handicaps-planeurs-ffvv.html
   { _T("WA 26 P Squale"),  330, 0, 80, -0.61, 152, -2, 174, -3.0, 12.6, 225, 86 },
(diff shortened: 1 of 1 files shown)


MaxKellermann:
Is any of this data redundant with the existing DG800 polar? Aerodynamics should be the same, so ....
Can we deduplicate this? Is there another way to avoid adding another polar for the same glider?
(https://github.com/XCSoar/XCSoar/pull/797#issuecomment-994780855)

iltis42:
> Is any of this data redundant with the existing DG800 polar? Aerodynamics should be the same, so .... Can we deduplicate this? Is there another way to avoid adding another polar for the same glider? ([#797 (comment)](https://github.com/XCSoar/XCSoar/pull/797#issuecomment-994780855))

Aerodynamics are not exactly same, DG-800S comes with the Elevator of the DG-600 and fuselage is a bit less in width.
Also the maximum water ballast (part of polar data) is different, as the non motorized version comes with 60kg less, max water ballast runs up to 150 liters vs the 120 liters in the motorized version, and of course the significant lower reference / dry weight, last thing might be a bit confusing to users how exactly to adjust.  
As of this in my opinion it makes sense to add this two polars.

PR #800, opened 2021-12-18
Amend FLARM device task declaration to use waypoint Short Name instead of index

davespragg:
Allows up to 10 turnpoints per task (which is the same as the limit was previously, and the maximum a FLARM device will accept).

### DIFF
--- a/src/Device/Driver/FLARM/TextProtocol.hpp
+++ b/src/Device/Driver/FLARM/TextProtocol.hpp
@@ -33,6 +33,6 @@ Copyright_License {
  * (worst-case).
  */
 char *
-CopyCleanFlarmString(char *gcc_restrict dest, const char *gcc_restrict src);
+CopyCleanFlarmString(char *gcc_restrict dest, const char *gcc_restrict src, int nChars = 0);
 
 #endif
--- a/src/Device/Driver/FLARM/TextProtocol.hpp
+++ b/src/Device/Driver/FLARM/TextProtocol.hpp
@@ -26,13 +26,17 @@ Copyright_License {
 
 #include "util/Compiler.h"
 
+#include <cstddef>
+
 /**
  * Copy a string, ignoring all characters that are illegal in a
  * setting value.  There is no buffer overflow check; the destination
  * buffer must be large enough to fit all of the source string
  * (worst-case).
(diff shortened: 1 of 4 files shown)


src/Device/Driver/FLARM/TextProtocol.hpp
```diff
@@ -33,6 +33,6 @@ Copyright_License {
  * (worst-case).
  */
 char *
-CopyCleanFlarmString(char *gcc_restrict dest, const char *gcc_restrict src);
+CopyCleanFlarmString(char *gcc_restrict dest, const char *gcc_restrict src, int nChars = 0);
```
MaxKellermann, line 36: What exactly does this new parameter mean, why is it a signed integer, what does 0 mean, and what does this have to do with this change?

davespragg: It optionally allows the string to be truncated. If nChars > 0, the string will be shortened to nChars if it was previously longer. Default value of 0 retains previous functionality (ie the string won't be truncated).
Required for this change because we need the ShortName parameter to be 6 chars or less to avoid exceeding the 192 byte limit for the full declaration. This will normally be the case anyway, but if someone configures a waypoint with a very long short name, it's better to truncate it to 6 characters than be unable to declare the task.
There's no need for this to be signed, happy to change it to unsigned.

MaxKellermann: Your explanation belongs in an API documentation comment.
But anyway, your explanation is insufficient. Is "nChars" the size of the buffer, or is "nChars" the maximum length of the string? (i.e. does it include the null terminator?) This is not nit-picking, the difference is extremely important.
And yes, the parameter should be unsigned - not only that, its type must be `std::size_t`.
The name "nChars" is also confusing. The name implies that the caller tells how many characters are in another parameter. But that's not true - it's not only not about the number of characters given (but a capacity), but also not about "characters". (What is a "character"? A single *byte* in a UTF-8 string (what is confusingly called `char` in C), a Unicode *code point*, or a glyph?)

davespragg: Thanks - updated per your suggestion; hopefully the comments are a bit clearer now.

PR #133, opened 2018-08-04
Further settings for the map element list and more map elements by an…

ThHanika:
… adjustable hit range

The trigger for this expansion was the use of openvario without touch and without mouse operation. It was very difficult to use the center crosshair on the screen to click on a waypoint or airfield to display the map item list. Even a subsequently installed resistive touchscreen has not brought benefits because of its inaccuracy.

With this extension, you can now expand the area around the crosshairs, even so far that all visible elements can be listed by clicking on the map. Even during the flight in turbulent air, this extension has proven itself.

The maximum number of the list is 32 items and is sorted by distance.

Since this is my first "pull request", please check some errors.

PR #801, opened 2021-12-20
PolarStore: Correction/refinement for DG800 polars triggered through X…

hjr:
…CVario development.

see  XCVario PR #133 for details.


MaxKellermann:
Where does this new polar data come from? I wish every entry had a code comment linking to the original source.

Does a glider model called "DG-800M" exist? On the manufacturer website, I can't find any trace of it. But DG-800 doesn't exist as well - there's "DG-800A", "DG-808C, "DG-808S" etc. but no "DG-800" and no "DG-800M". I understand that you're trying to reduce confusion, but by using nomenclature that doesn't officially exist, I fear it will only create more confusion.

hjr:
Hey, your comments are always appreciated ;) and helpful.

You are right, the M is the correct Model extension for the 500M (with engine), but not for the 800. That should be a 800B then. Own one of those.. should know (lol).
Iltis comment of where they are from is correct, I digitized the manual. His commit raced a discussion about meaning of proper default masses and max ballast setting, as well as my digitizing work. (Later is a disconnected story and related to the unfortunate definition of the CAI !g protocol).
Will change the "M" to an "B".
Mantissa change is related to have a 100% identical polynom out of this (njanja) three point polynom definition (polar #28,29).

MaxKellermann:
So... the new polar coefficients are taken from the same official (paper) manual, but from different speeds? It's effectively the same polar?
What's the point of this change then? The commit message is rather not-explanatory.

hjr:
Ok ...
DG80018_suggestion
When I digitize a polar from paperwork I use xyscan and digitize a pile of x,y tuples. I then chose visually a best polynomial grade 2 fit that matches the speed range 100 - 180kmh best. The DG800 hast the funky kink at 170kmh, so no way at all to be all realistic with the model. 
The three point output is then generated, speeds are chosen arbitrarily to V(least sink), 130, 170. Reference mass is set according choice, the whole thing scaled accordingly. This all hast not been properly done up to this point (I fear). If you request to chose different speeds for the three point polynomial definition I can do this with no change of the polar. 
I also offer to publish the python/numpy scripts I did to accomplish all this as XCSoar/tools e.g.

MaxKellermann:
I'm just trying to understand what you're doing - so far, this sounds like a good approach and a good change, but it's under-documented. Everybody should be able to understand this PR, without having to ask you, just from your commit message and code comments.
My point is: if I have to ask such questions, it means, that we don't have enough documentation.
The current polar list is, of course, just as badly documented. I wish we had better code comments explaining all the data sources, as I said. Let's do this better now, to set an example of how it should be.

After you've gone through all the work getting better polar coefficients, let's do this properly to the end: please add a code comment to the polars you modified, with enough explanation about where and how you got this data. Only with proper documentation can we ensure that your work was invested well; it allows others to verify your work easily. Writing this documentation now takes a bit of time, but will save a lot of time in the future.

MaxKellermann:
And yes, please, I'd welcome to have that script merged into this repository, with some documentation on how to use it. That will for sure help others do the same improvements for other polars.

iltis42:
There are different way's to choose the 3 points from a polar, and esp. for the DG-800 this is a tricky thing. @hjr has done a great job in order to automate the process to digitze polars by a sophisticated toolset. I consider this as an optimization, and can follow the argumentation from @hjr after a short demo, nothing else could do a better job that this way i have seen to fit real world data into a second order approximation. 

Regarding the naming convention, as a compromise the DG-800S truly exists in handbook, and there are motorized models A,B and maybe C that not be useful. The B is the most common type, about 70% of DG-800 are type B. Type A and C mostly differ in features like different fuel tanks, like Rögerhaken. All have nothing to do with a different aerodynamic. 
To complete confusion there is also DG-808*, also a feature-set model, same aerodynamics.

After several intensive discussion over a couple of day's, our conclusion was to support the DG-800S, the pure glider version in 15 and 18 meters, plus the motorized version in 15 and 18 meters. Both variants come with different default wingload and different water ballast. All types A,B,C and 808 are considered to be aerodynamically same.

We have also agreed to the same polar for glider and motorized version, the handbooks do not differ (even the aerodynamic slightly differ between glider and motorized version, e.g. fuselage width and elevator), hence there hasn't been any documented difference in polars, its just the same polar, for another weight.

If someone feels uncomfortable using DG-800B (the most common model), for the motorized type identifier, i propose "DG-800" _without_ and appendix as a wildcard. But not DG-800M, this is not a known type, hence for me the DG-800B makes also sense as most pilots will find themselves there immediately, and few owners of A or C or other types know that their model is aerodynamically same with that. If we don't carry any appendix, it might remain in question what exactly is this.

This is my 5 cent for this discussion, remaining in best regards.

hjr:
Thx for the comment Itlis.
To what Max is correctly requesting.
- Will complete documentation and suggested process to digitize/import polars. I guess best in the python script itself.
- Will also start a discussion(s) on this topic. First I see the option to go with fourth order approximations. Then I see the internal polar format with four points not adequate, because it is not a minimal definition and equality of them is obfuscated, comparison difficult.

But I would suggest to do this on further PRs. Don't see the option to improve this PR significant in adding words to Iltis comment line. The polar are extracted from the DG-800 manuals.

MaxKellermann:
Here's an official list of DG models: https://web.archive.org/web/20110823031225/http://www.dg-flugzeugbau.de/Data/dg-typen-total-d.pdf
Let's add some confusion: you said DG-808[ABC] are the same, but there's no such thing as a DG-808A. There is DG-800A, and there is DG-800B which was sold as DG-808B, but some older DG-800B were apparently sold as just DG-800B. The DG-808C was consistently named, there was never a DG-800C.
Now look at the empty weights (18m):
- DG-800A: 327kg
- DG-800B MidWest: 338kg
- DG-800B Solo: 340kg
- DG-808B: 344kg
- DG-808C Competition: 350kg
- DG-808C Classic: 340kg

So if XCSoar provides a polar (with mass), it must exactly specify which one it means, even if aerodynamics are the same. We can't just say "DG80X with engine" and "DG-80X without engine"; that would be inaccurate, because the mass for different iterations of the model is different.

You can argue that the mass difference is small, and the practical difference is miniscule; that would certainly be reasonable, but this must be a pilot's decision. XCSoar must be accurate, it must be trusted. Any deviation from our accurate data must be an explicit user decision.

MaxKellermann:
Thanks @hjr, looking forward to merge your script and more improvements. I merged this PR for now.

hjr:
Masses are a third thread to open. The XCS Plane/Polar dialog offers a dry mass to be specific on the mass of the specific glider you fly. 
Here is actually more mixed in, because you are expected to enter all mas you cannot dump by opening the ballast valve.
.. should change threads for this ..

iltis42:
The empty weight's above from the handbook mostly are not fully true for the concrete glider, gliders get heavier over time, not only because of cockpit instrumentation and gimmicks like bugwipers, solar panels, oxygen equipment, etc. As commonly known, composite materials soak water over time. The effect of gaining accuracy by using different models is some better default, and yes, i am a fan of having a polar for the explicit model, hence i agree as well, the benefit is small and the differences are low, and if i remember correctly it was also the idea here to clean up redundancy, hence this looks like cranking back in this question here, right ?
Having a new thread to cover the exact masses at takeoff, including pilot, including passenger, maybe even tail ballast i guess will help a lot to gain accuracy here. I wonder if any pilot had understood to change dry mass if pilot was lighter or heavier that the dry mass included, even me i didn't dig into there and just used the default for my glider what came with a considerable error as the dry carbon Nimbus 2C is considerably lighter (almost 60kg), that the glass type.

PR #803, opened 2021-12-20
PolarStore: Updated handicap factors for 2022, added specific wingspa…

hjr:
…n or name addition if needed to identify model.


hjr:
Dont know how to generate a new PR when the previous on the same branch is still open ..

so here is the polar helper script.

hjr:
Impressed by this python checker. Will be back here soon :)

MaxKellermann:
I'll merge the Python script, thanks for that - but please run "pylint" and check the warnings. For example this one looks legit:
```
tools/polar_import.py:297:3: C0121: Comparison 'args.xyscan != None' should be 'args.xyscan is not None' (singleton-comparison)
```

hjr:
Ok, cool. thx for the hint. Yes I am a rare times python/numpy coder, but it is fun. Hope it gets us more valuable polars.

PR #809, opened 2021-12-23
Fix crash in python module's flight.path()

tpatja:
Brief summary of the changes
----------------------------
Fixed a crash in the python modules flight.path() method that I recently found when trying it out. Also made the python tests run with python 3.

No related issues or discussions found. Seems not many people are using the python module.

### DIFF
--- a/python/src/PythonConverters.cpp
+++ b/python/src/PythonConverters.cpp
@@ -86,6 +86,17 @@ GeoPoint Python::ReadLonLat(PyObject *py_location) {
   return location;
 }
 
+TimeStamp Python::PyLongToTimeStamp(PyObject* ts) {
+  if(!PyLong_Check(ts)) {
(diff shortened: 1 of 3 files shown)


MaxKellermann:
This commit is confusing and hard to read because it mixes 4 things in one commit, 2 of which are described. Each commit should only contain one change.

tpatja:
Hi Max. The PR is now split into 3 commits.

MaxKellermann:
Thanks, much better. Your third commit changes the indentation of the Py_BuildValue call, which hides what this commit is really doing.

tpatja:
Indentation issue fixed.

MaxKellermann:
Thanks again!

tpatja:
Contributing this small patch was the least I could do, given the huge amount of value XCSoar has given me for free over the years on my cross country soaring flights. Thank you for your hard work on XCSoar and happy holidays!

Issue #812, opened 2021-12-24
B50 cruise / climb switch of displays now incorrect after fix


gldrpilot:
XCSoar versions having and not having the problem:
Having problem 7.21, not having problem 7.20
-------------------------------------------------

Expected behavior:
Using 'Flaps Forces Cruise", switching between cruise / climb from Borgelt B800 worked correctly in version 7.20 and earlier. 
-----------------

Actual behavior:
A recent 'fix' in version 7.21 has been introduced to invert this which has resulted in the circling page being displayed when cruising and cruising page being displayed when climbing.
---------------

Install version 7.21 (Android)
Connect to B800
Switch modes & observer resultant display.
(I've double checked using Mike's simulator program and changing between version 7.20 and 7.21. I've uses this feature for a number of years and it has been correct until version 7.21.)
-------------------------------

Do you have any idea what may have caused this?
A recent update to the software.

-----------------------------------------------

Do you have an idea how to solve the issue?
Remove the fix, or make an option to select a 0 or 1 as cruise or climb.

-------------------------------------------

lordfolken:
https://github.com/XCSoar/XCSoar/pull/774 this was the discussion and merge. @iltis42 can we verify?

gldrpilot:
I have double checked using the simulator, see attached file of results. This suggests the status is inverted to the published documentation as mentioned in #774 
v7.20 works in the correct sense when connected to either the B800 and the simulator.
[Cruise climb status to PBB50 string.pdf](https://github.com/XCSoar/XCSoar/files/7775149/Cruise.climb.status.to.PBB50.string.pdf)
.

gldrpilot:
Accidentally closed this...

iltis42:
Fully agree, öooks like the Borgelt documentation is wrong, exactly inverted. Hence the B50 or B800 gadgets are like this in the field, and we can't change, so we needed to revert in XCSoar in order to fix this:
https://github.com/XCSoar/XCSoar/pull/774
and i we agree this as solution, and keep the workaround of this bug in XCSoar, i will also change XCVario either to inverted cruise sense, so things will work with both gadgets in future.

iltis42:
Can we close the issue with the above reversion?
Mike Borgelt has answered my mail and confirmed by looking at the code, that documentation has twisted the sense.

gldrpilot:
Thank you, I'll close this.

gldrpilot:
Hi
[...]

PR #813, opened 2021-12-26
Revert "Fix inverted cruise/circling mode in Borgelt and XCVario protocol"

lordfolken:
Reverts XCSoar/XCSoar#774 according to a discussion in https://github.com/XCSoar/XCSoar/issues/812 the documentation is inverted to the actual behavior.

### DIFF
--- a/src/Device/Driver/BorgeltB50.cpp
+++ b/src/Device/Driver/BorgeltB50.cpp
@@ -104,11 +104,11 @@ PBB50(NMEAInputLine &line, NMEAInfo &info)
   // inclimb/incruise 1=cruise,0=climb, OAT
   switch (line.Read(-1)) {
   case 0:
-    info.switch_state.flight_mode = SwitchState::FlightMode::CIRCLING;
+    info.switch_state.flight_mode = SwitchState::FlightMode::CRUISE;
     break;
 
   case 1:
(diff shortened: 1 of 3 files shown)


iltis42:
Inverted now the bit in XCVario, 
https://github.com/iltis42/XCVario/commit/7a2d93483f5057ee3e254b750f2cb9c5e815a2c6
And changed in handbook:
  G = 1 in climb, 0 in cruise, Note: Original Borgelt docu shows vice versa
Remains an email to Mike Borgelt so it applicable, he may update his documentation too...

iltis42:
Hi @lordfolken,
somehow this went wrong here: Changeset #774  removed CRUISE and set CIRCLING when Cruise bit is 0 (zero), in the datatstream. Correct was the older state, change #774 was false as Borgelt docu turned out to be be wrong, so we should have restore the older state, hence looking at the source, i see still this change here:

https://github.com/XCSoar/XCSoar/blob/master/src/Device/Driver/XCVario.cpp#L105

so apparently this revert was never mergerd ?
Phillip, can you confirm and maybe merge this, just now my 7.23 XCSoar behaves inverted regarding Vario/Cruise Mode.





https://github.com/XCSoar/XCSoar/blob/master/src/Device/Driver/XCVario.cpp

PR #815, opened 2021-12-27
Add 3x6 InfoBox layout

davespragg:
Admittedly this is yet another hard-coded infobox layout that won't be useful to everyone, but this one is pretty useful on a 7" display (eg the 7" Openvarios, of which there are a lot around), particularly in portrait mode. Uses less screen real-estate than the existing 3x5 layout, and is easily readable on a 7" screen mounted in an instrument panel.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -5,6 +5,8 @@ Version 7.22 - not yet released
   - add DG-800S
 * devices
   - FLARM: send turn point names in task declaration
+* user interface
+  - add 3x6 InfoBoxes layout
 * Android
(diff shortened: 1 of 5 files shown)


MaxKellermann:
You can imagine I don't like adding yet more such layouts - but until I (or somebody else) implement dynamic layouts, I guess I should merge this. And one day replace all of this dirty code with a simple formula.

davespragg:
Thanks! Dynamic layouts would be a much better solution; I'll have a think about it (as a long-term project) and maybe float some ideas on the forum.

PR #823, opened 2022-01-02
Split IsKoboOTGKernel into two routines

GeekGoneOld:
With Clara HD, OTG host mode is no longer tied to custom kernel
so this splits IsKoboOTGKernel into IsKoboCustomKernel and
IsKoboOTGHostMode.  This doesn't change anything but sets the
stage for adding OTG to Clara HD (and any others that have OTG
in the factory kernel).

Brief summary of the changes
----------------------------

### DIFF
--- a/src/Kobo/Kernel.cpp
+++ b/src/Kobo/Kernel.cpp
@@ -22,6 +22,9 @@ Copyright_License {
 */
 
 #include "Kernel.hpp"
+#include "Model.hpp"
+
+#include "system/FileUtil.hpp"
 
 #include <stdexcept>
 
@@ -106,9 +109,13 @@ KoboInstallKernel(const char *uimage_path)
 #endif
 
 bool
-IsKoboOTGKernel()
+IsKoboCustomKernel()
 try {
 #ifdef KOBO
+  /* All Kobo except Clara HD have a factory kernel without OTG mode so
+     a custom kernel is installed for OTG. */
+  if (DetectKoboModel() == KoboModel::CLARA_HD) return false; 
+
   FileReader file(Path("/proc/config.gz"));
   GunzipReader gunzip(file);
   BufferedReader reader(gunzip);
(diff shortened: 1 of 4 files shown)


GeekGoneOld:
I have no idea why there is #ifdef KOBO entries in the source code.  The code is in the /src/Kobo folder so I'm guessing it is from days gone by and no longer needed.  I tried to follow the lead of the existing code but I really was just guessing.  Please check carefully.  Otherwise, this PR allows to handle new factory kernels that include OTG.  I will have an additional PR that provides Clara OTG support shortly.

src/Kobo/Kernel.cpp
```diff
@@ -123,3 +128,16 @@ try {
...
   return false;
 }
+
+bool
+IsKoboOTGHostMode()
+try {
+#ifdef KOBO
+  if (DetectKoboModel() != KoboModel::CLARA_HD) return IsKoboCustomKernel();
+  /* for Clara HD, just test for existence of host mode file */
+  FileReader file(Path("/mnt/onboard/XCSoarData/kobo/OTG_Host_Active"));
```
MaxKellermann, line 138: This is a clumsy way of testing a file's existence; better use `File::Exists()`.

MaxKellermann:
> I have no idea why there is #ifdef KOBO entries in the source code. The code is in the /src/Kobo folder so I'm guessing it is from days gone by and no longer needed.

Some sources from src/Kobo/ do get compiled on other targets, for example the KoboMenu program can be developed and debugged on regular Linux.

src/Kobo/SystemDialog.cpp
```diff
@@ -83,6 +84,15 @@ SystemWidget::Prepare(ContainerWindow &parent, const PixelRect &rc) noexcept
   }
 }
 
+inline void
+SystemWidget::SwitchOTGMode()
+{
+#ifdef KOBO
+  if (DetectKoboModel() != KoboModel::CLARA_HD) return SwitchKernel();
+  return;
```
MaxKellermann, line 92: What's the point of this `return`?

GeekGoneOld:
For the Clara OTG implementation, there is some code between 91 and 92.  One way or the other, this return is at the end of the function and is technically not needed as the function has no return value (void).  For many years, though, I have used return whether the function returned a value or not.  If you want it gone, I'm happy to remove it but it may get me into git hell again (I have since put in the OTG code for Clara).  I would ask if you would let me remove it on the next PR (2 days away as I now have OTG working but must test it throughly).

GeekGoneOld:
Yes, I would have to agree!  I will fix this (and the”return” you commented on) within 48 hours

GeekGoneOld:
I've made the updates (File::Exists() and delete return).

GeekGoneOld:
Another edit 'cause failed the compile test which I missed due to #ifdef KOBO.  Hopefully will pass...

PR #612, opened 2021-05-06
WeGlide Uploading IGC files

August2111:
Brief summary of the changes
----------------------------

Add some components to prepare the upload feature
  * simple date picker for date of birth selection
  * integer selection picker for large numbers
  * curl  UploadFile component
  * Setting the relevant data in the 'Config/Setting/WeGlide Cloud'-Panel and in the 'Plane'-Panel
  * Uploading the IGC file with 3 options:
    *  1st: Direct upload automatically after downloading file from logger (all necessary data have to be set and additional the flag 'automatic upload'
    * 2nd: Calling the IGC-FilePicker with a (self defined) lua gesture or keypress event
    * 3rd: Upload directly over a lua file
 
Related issues and discussions
------------------------------
The upload is following the proposal Issue #505 from Bomilkar.

The 'Downloading WeGlide Task'-Job will be coming in a next step based on this!

PR #829, opened 2022-01-07
Add json lookup

August2111:
adds json/lookup from CM4All, needed for communication with WeGlide


MaxKellermann:
This is pointless because there is no code using it. This should be part of a PR which actually uses it.

August2111:
Ok, I thought it is better to split the whole PR in small portions...
So I will prepare the whole PR #612...

PR #830, opened 2022-01-11
Update WaypointList.cpp

Blaubart:
Usually waypoint list is empty if you are using it first time. This PR filters waypoint to airports using waypoint list first time. So you can set goto an airfield very quickly in case like land there


lordfolken:
How is this different from the alternates feature? (Drawing an inverse L on the screen, or Nav->Alternates)

MaxKellermann:
Bad commit message, and I don't understand the change. Wondering the same as @lordfolken.

Issue #795, opened 2021-12-11
.cupx file format


hjr:
What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------

I'd like to see the naviter cupx waypoint file format supported as we do support the cup file format. The 'x' additions are columns for a description and a list of picture file names. jpg picture files are embedded in the cupx file like in our xcm tar files. 

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------

Drop in cupx files to the XCSoarData folder from the glider pilot community and just use them, incl. the pics feature for outlanding fields etc.
(just wrote a python script to create the Airfield-Details.txt glue file to make use from an cupx file)

MaxKellermann:
This has been several times before, and in the end, this was impossible because the file format is proprietary, there exists no public specification, and nobody was ever willing to reverse-engineer it. So unless you have a public and free specification, we need to close this feature request.

hjr:
I am willing to give a reverse engineered spec a trial.

Plantain:
It's nothing particularly complex, just two .zip files concatenated - it
was first reverse engineered in a prior XCSoar ticket (now deleted), but
you can see snippets of it still
https://tophatsoaring.atlassian.net/browse/TH-760

MaxKellermann:
Why... would anybody ever concatenate ZIP files?? It's already a container format.

lordfolken:
Landewiesen.at is available as cupx. They must have some kind of implementation.

hjr:
Will try to connect to them.

hjr:
Christian Hynek (chynek@streckenflug.at) answered, you guys seem to have contact as well. 
So the plan is to create an handsome and open format to distribute landewiesen.at from its proper origin and have XCSoar enabled to read those format too.
You may use this thread to post updates on this.

lordfolken:
So i got a mail from naviter, that they updated the documentation and also described cupx format:

Documentation for CUP and CUPX file formats is updated now.
You can download it here:
. for CUP: https://downloads.naviter.com/docs/SeeYou_CUP_file_format.pdf
. for CUPX: https://downloads.naviter.com/docs/SeeYou_CUPX_file_format.pdf
Have a nice 2022 intro,

MaxKellermann:
This format is so braindead, I will not waste time with writing code to support it.
[...]

PR #833, opened 2022-01-12
Support for new SeeYou Fields in CUP files (UserData, Pics)

lordfolken:
## Waypoint Details with Pictures from CUP (2022) file

2022-01-12-005732_640x480_scrot

## New Waypoint Types (&icons)

2022-01-17-202113_1276x704_scrot
Partly implements #795

Reference:
https://downloads.naviter.com/docs/SeeYou_CUP_file_format.pdf

What works:
- Adding pictures from seeyou.cup only (no need for waypoint_details)
- Adding UserData field as text to the Waypoints Dialog
- All Waypoint types in new Standard

What is missing:
- [x] Currently the pictures are searched realtive to the XCSoarData folder. Not the subdirectory ```pics``` as per seeyou spec.
- [ ] We need a container format like xcm (uncompressed zip archive) to easily distribute the
  * points.cup
  * pics/pictures.jpg

  in one file, so its easy to handle and download.

- [x] It has to be checked how this interferes with the waypoint_details functionality, as this uses the same functions to add text and pictures
- [ ] The dlgWaypointDetails dialog should get a page counter in the caption (2/4) for example, to make it clear that there are more pages
- [x] Prevent path traversal

Test line:
```csv
"Goesgen Kkw Kueh",,CH,4722.000N,00758.150E,377.0m,18,,,,"KKWKUEHL","Hello World This a long text text this text will almost never end","goesgen.jpeg;goesgen 2.jpeg"
```


src/Waypoint/WaypointReaderSeeYou.cpp
```diff
@@ -299,6 +301,22 @@ WaypointReaderSeeYou::ParseLine(const TCHAR* line, Waypoints &waypoints)
...
+    std::forward_list<tstring> pics;
+    tstring pic_list = params[iPics];
+    tstring::size_type pos = 0;
+    tstring::size_type prev_pos = 0;
+    while ((pos = pic_list.find(_T(';'), prev_pos)) != tstring::npos) {
+      pics.push_front(pic_list.substr(prev_pos, pos - prev_pos));
+      prev_pos = pos + 1;
+    }
+    pics.push_front(pic_list.substr(prev_pos));
+    new_waypoint.files_embed = pics;
```
MaxKellermann, line 318:
```
for (std::string_view i : IterableSplitString(params[iPics]))
  new_waypoint.files_embed.emplace_front(i);
```
Simpler, faster and less memory waste!

lordfolken:
@cirrus any idea why this fails in Windows?

MaxKellermann:
The usual `char` vs `TCHAR` problem. You need `TIterableSplitString` which works with `TCHAR`.

src/Waypoint/WaypointReaderSeeYou.hpp
```diff
@@ -41,6 +41,8 @@ class WaypointReaderSeeYou final : public WaypointReaderBase {
   /* field positions for typical SeeYou *.cup waypoint file */
   unsigned iFrequency = 9;
   unsigned iDescription = 10;
+  unsigned iUserData = 11;
+  unsigned iPics = 12;
```
MaxKellermann, line 45: That doesn't belong here

src/util/tstring.hpp
```diff
@@ -5,8 +5,10 @@
 
 #ifdef _UNICODE
 using tstring = std::wstring;
+using tstring_view = std::wstring_view;
```
MaxKellermann, line 8: I think this should be a new header, without the heavy dependency on `<string>`

src/Waypoint/WaypointReaderSeeYou.cpp
```diff
@@ -299,6 +303,17 @@ WaypointReaderSeeYou::ParseLine(const TCHAR* line, Waypoints &waypoints)
     new_waypoint.comment = params[iDescription];
   }
 
+  if (iUserData < n_params) {
+    new_waypoint.details = params[iUserData];
+  }
+
+  if (iPics < n_params) {
+    for (tstring_view i : TIterableSplitString(params[iPics], _T(';'))) {
+      tstring p = _T("pics/");
```
MaxKellermann, line 334: What's ... this?

src/Waypoint/WaypointReaderSeeYou.cpp
```diff
@@ -299,6 +303,17 @@ WaypointReaderSeeYou::ParseLine(const TCHAR* line, Waypoints &waypoints)
...
+      p += i;
```
MaxKellermann, line 335: This can force XCSoar to read arbitrary files from the filesystem, just by writing something like "../../../" etc. into a waypoint file. XCSoar should validate file names loaded from data files.

Data/icons/map_ndb.svg
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
MaxKellermann, line 1: What's the license?

src/Waypoint/WaypointReaderSeeYou.cpp
lordfolken: The problem is that params[iPics] comes as wide when on WIN64, and normal on UNIX. 
So this is the platform-neutral implementation, by using the tstring_view alias and _T for translation of the strings. 
The "pics/" directory is what is required by the CUP standard. As I didn't want to break files.embed for the waypoints_details.txt usecase i decided to add it here in the cup parser.

MaxKellermann: Not the CUP standard, but the CUPX standard, and the CUP parser should not hard-code assumptions from another standard, or else something else will break.

Data/icons/map_ndb.svg
lordfolken: They are GPL3 https://github.com/Akaflieg-Freiburg/enroute/blob/master/LICENSE

src/Waypoint/WaypointReaderSeeYou.cpp
lordfolken: i.c. your point. Will validate the path to prevent directory traversal.

Data/icons/map_ndb.svg
MaxKellermann: That has an important implication - it effectively makes all XCSoar builds GPLv3-only, removing the GPLv2 option.

src/util/tstring.hpp
lordfolken: ok will move to new header.

src/Waypoint/WaypointReaderSeeYou.cpp
lordfolken: https://downloads.naviter.com/docs/SeeYou_CUP_file_format.pdf

Its not clear to me here where cup standard ends and cupx begins. 
The field is described in the cup standard document:

"A string field containing the filename(s) of an JPG image stored in pics/ folder of a seperate pics.zip file.See SeeYou CUPX file format specification. Multiple filenames must be separated by commas and string mustbe embraced in double quotes."

Since we don't want to implement CUPx's insane zip concatenation:
 * we would have a single archive with  points.cup and pics/foo.jpg
 * search: the archive and the XCSoarData for pics/foo.jpg

lordfolken: Since this is an extension of the standard, with the column counter check, this shouldn't affect the parsing of the "legacy" cup files.

MaxKellermann: Naviter didn't know the boundary between CUP and CUPX, so they mixed it up in the documentation. This is obviously a documentation mistake. They were not careful enough to document the difference.

lordfolken: It looks like there isn't really a boundary. 
When i download cup files from soaringspot.com, all 14 fields are there. (although empty)
https://www.soaringspot.com/en_gb/nacional-clase-club-santa-rosa-2022/downloads

So the question is here, whether we try to read pictures when we have only the file and not an archive (cupx or otherwise.)

MaxKellermann: The CUP parser doesn't know (and shouldn't be aware) if it reads a CUP file from a CUPX/ZIP container.

src/Waypoint/WaypointReaderSeeYou.cpp
lordfolken: Ideally this should be in file_embed because, this code also gets used by waypoint_details.txt

src/Waypoint/WaypointReaderSeeYou.hpp
MaxKellermann: How is this resolved?

glidingFisi:
@lordfolken are you currently working on this PR? I reviewed yesterday the information out of this PR and the issue discussion. In parallel, I downloaded the landewiesen cupx version. 
It is a little bit strange, because there is a special cupx header before the zip files.
Did you already tried to extract this landewiesen cupx?

src/Waypoint/WaypointReaderSeeYou.cpp
lordfolken: See https://github.com/XCSoar/XCSoar/pull/833/commits/7c3fc9e653b595af70a5cf63e6dfdd441c148573

lordfolken:
@MaxKellermann  could we merge this? The remaining tasks are not a hard dependency. 
The container format needs to be invented, and the gui indicator is a generic usability problem.

MaxKellermann:
See https://github.com/XCSoar/XCSoar/pull/833/commits/46ec6671fd440be80b3a29ba19e6dd9ef31ac98d#r786255721

lordfolken:
Ah i was mainly talking about hte follow up commits such as
https://github.com/XCSoar/XCSoar/pull/833/commits/dba78f87acd096a7dc468c9de8a5f4d867825f34 for the new waypoint tiypes (according to seeyou 2022) and the new icons to represent them. Also the path traversal guard.

lordfolken:
I have restructured the commits and created a build up.
The first part of the PR adds the parsing of the new cup 2022 format. (new fields and interpretation)
The second part extends the waypoint engine for new waypoint types in cup 2022.
The third part mainly concerns itself with rendering of the icons on the map, and updating said icons.

MaxKellermann:
Cherry-picked with amendment, removing the "pics" prefix.

PR #841, opened 2022-01-18
Add OTG to Clara HD

### LINKED ISSUE #794: Add OTG serial support to Clara HD
XCSoar version
--------------

7.20+

Add OTG serial support to Clara HD

It should be possible to connect Clara HD to serial devices (Flarm, varios) using OTG.

GeekGoneOld:
Added OTG support including modules for serial, sound and
networking.  Note the modules are included in the release
separately (just as are the OTG kernels for other models).
If they are absent (e.g. a local build), no failure will occur.

NOTE!!! To make this complete, a tgz file must be put on the build server
in the SAME folder as the OTG kernels already present.  Failure to
do this will NOT cause a failure, just a loss of OTG functionality for 
Clara HD.  Let me know who to send the tgz file to.  I have tested it
on my build setup.

closes #794

Brief summary of the changes
----------------------------

### DIFF
--- a/kobo/rcS
+++ b/kobo/rcS
@@ -85,6 +85,35 @@ ulimit -c unlimited
 # delay should not cause any problem and might help other models
 usleep 100000
 
+# load kernel modules for ClaraHD as factory kernel is OTG
+# but missing modules.  Also put ClaraHD in the right mode
+model=`dd if=/dev/mmcblk0 bs=8 count=1 skip=64`
+if [ `expr substr $model 1 7` = SN-N249 ]; then
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/usb/serial/usbserial.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/usb/serial/aircable.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/usb/serial/ch341.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/usb/serial/cp210x.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/usb/serial/ftdi_sio.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/usb/serial/garmin_gps.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/usb/serial/mos7840.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/usb/serial/navman.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/usb/serial/pl2303.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/sound/core/snd-hwdep.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/sound/core/snd-rawmidi.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/sound/usb/snd-usbmidi-lib.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/sound/usb/snd-usb-audio.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/net/mii.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/net/usb/usbnet.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/net/usb/cdc_subset.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/net/usb/cdc_ether.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/net/usb/cdc_eem.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/net/usb/cdc_ncm.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/net/usb/rndis_host.ko"
+    mount -t debugfs none /sys/kernel/debug
+    if [ -f /mnt/onboard/XCSoarData/kobo/OTG_Host_Active ]; then
+        echo host >/sys/kernel/debug/ci_hdrc.0/role
+    fi
+fi
 # launch user script
 if [ -f /mnt/onboard/XCSoarData/kobo/init.sh ]; then
     source /mnt/onboard/XCSoarData/kobo/init.sh
--- a/src/Kobo/SystemDialog.cpp
+++ b/src/Kobo/SystemDialog.cpp
@@ -31,6 +31,10 @@ Copyright_License {
 #include "System.hpp"
 #include "Model.hpp"
 
+#include "system/FileUtil.hpp"
+
+#include <sys/stat.h>
+
 class SystemWidget final
   : public RowFormWidget {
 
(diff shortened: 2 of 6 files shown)


kobo/rcS
```diff
@@ -83,8 +83,38 @@ ulimit -c unlimited
 # workaround for Kobo Clara HD: the touch screen doesn't respond
 # and the initial screen render is blank if we are too fast
 # delay should not cause any problem and might help other models
-usleep 100000
-
+sleep 1
```
MaxKellermann, line 86: Why?

kobo/rcS
```diff
@@ -83,8 +83,38 @@ ulimit -c unlimited
...
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/net/usb/usbnet.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/net/usb/cdc_subset.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/net/usb/cdc_ether.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/net/usb/cdc_eem.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/net/usb/cdc_ncm.ko"
+    insmod "/opt/xcsoar/lib/modules/4.1.15/drivers/net/usb/rndis_host.ko"
+    if [ -f /mnt/onboard/XCSoarData/kobo/OTG_Host_Active ]; then
+        mount -t debugfs none /sys/kernel/debug
+        echo host >/sys/kernel/debug/ci_hdrc.0/role
+        umount /sys/kernel/debug
```
MaxKellermann, line 115: Why umount? If we need debugfs for a certain Kobo model, let's keep it mounted; this also avoids the fragile mount/umount code you added to C++.

src/Kobo/SystemDialog.cpp
```diff
@@ -87,9 +91,25 @@ SystemWidget::Prepare(ContainerWindow &parent, const PixelRect &rc) noexcept
 inline void
 SystemWidget::SwitchOTGMode()
 {
-#ifdef KOBO
```
MaxKellermann, line 90: Will this still compile without the `#ifdef`?

src/Kobo/SystemDialog.cpp
```diff
@@ -87,9 +91,25 @@ SystemWidget::Prepare(ContainerWindow &parent, const PixelRect &rc) noexcept
...
-  if (DetectKoboModel() != KoboModel::CLARA_HD) return SwitchKernel();
-#endif
+  if (DetectKoboModel() == KoboModel::CLARA_HD) {
+    mount("none", "/sys/kernel/debug", "debugfs", 0, "");
+    if (IsKoboOTGHostMode()) {
+      File::WriteExisting(Path("/sys/kernel/debug/ci_hdrc.0/role"), "gadget");
+      File::Delete(Path("/mnt/onboard/XCSoarData/kobo/OTG_Host_Active"));
+      switch_otg_mode->SetCaption("Enable USB-OTG");
```
MaxKellermann, line 99: There is on error handling; we shouldn't change the state of the UI just by guessing that this has worked.
First, there should be error handling. If an error occurs, the error should be reported to the user.
Second, the state of the UI should always reflect the state of the system. i.e. even if the operation seemed to have succeeded, check the system state again (call `IsKoboOTGHostMode()` again).

MaxKellermann (changes requested):

kobo/rcS
GeekGoneOld: With 100000us (0.1s) I had a few times when the Clara HD started with no touch screen.  I felt it was a harmless increase to make it 1s and I have not yet had a single failure.  Sad that the Kobo is that touchy!

MaxKellermann: So... this has nothing to do with OTG support?
Such a change needs to be a separate commit with an explanatory commit message.

kobo/rcS
GeekGoneOld: I'm more than happy to do this.  I can make the mount conditional only on being Clara HD and not consider OTG_Host_Active (except for the echo).  I will edit shortly.

src/Kobo/SystemDialog.cpp
GeekGoneOld: It compiles fine on my development machine.  I compile with TARGET=KOBO,

src/Kobo/SystemDialog.cpp
GeekGoneOld: Will change shortly...

kobo/rcS
GeekGoneOld: OK  will remove the change and (possibly) submit later.  Extra code in rcS may add enough delay.

src/Kobo/SystemDialog.cpp
GeekGoneOld: I will put back the ifdefs as they may be needed for UNIX... TARGETs.

kobo/rcS
GeekGoneOld: returned to original

kobo/rcS
GeekGoneOld: now mounted in rcS and never umounted

src/Kobo/SystemDialog.cpp
GeekGoneOld: ifdef is now back

MaxKellermann:
There's an executable called `-S` in your commit

src/Kobo/SystemDialog.cpp
GeekGoneOld: Added error handling on the mode change.  Note that optimisation in SwitchOTGMode() cached the result of IsKoboOTGHostMode() and caused an incorrect failure indication (as host mode changes in the function).  I used the __attribute__ to disable optimisation for SwitchOTGMode() and it works.

MaxKellermann: This is a bad solution; first, don't use `__attribute__`; it is a proprietary GCC extension. Even if we currently only use GCC, we should avoid using those extensions.
Second, disabling compiler optimization is a fragile kludge which steps away from really fixing the cause. This is caused by `IsKoboCustomKernel()` having a `const` attribute, which is wrong since you added code to the function which considers mutable runtime state (i.e. your change was buggy). Therefore, changing `const` to `pure` will fix the problem properly.

MaxKellermann: Also, please don't hard-code `/mnt/onboard/XCSoarData`; you should instead use the `LocalPath.hpp` library. Even if it's unlikely that this path ever changes.

GeekGoneOld:
Here is the ClaraHD_OTG_Drivers.tgz file zipped up (attachements can't be tgz) that must be in /opt/kobo/kernel/ on the build machine.  Note this is the very same directory that the otg kernels are in.  Unzip and put the .tgz file in that directory.  Don't gunzip (or tar) the tgz
[ClaraHD_OTG_Drivers.zip](https://github.com/XCSoar/XCSoar/files/7915785/ClaraHD_OTG_Drivers.zip)
.

MaxKellermann:
Please add instructions on how to reproduce your build to the developer manual

GeekGoneOld:
I will gladly add to the developer manual.  I'm assuming you are specifically referring to the kernel build and resulting modules.  Currently I don't see any information on kernel building in the manual so I suppose I will add an appendix.  Note that I have no knowledge of how to include the resulting kernel and/or modules in the official build.  I can only see where the build files look for these (/opt/kobo/kernel/) but I have no idea of how the files get there.  That is why I posted the file above for "someone" to put it where it must be on the "official" build machine.

src/Kobo/SystemDialog.cpp
GeekGoneOld: I just followed the lead of KoboExecNickel() in src/Kobo/System.cpp.  I assumed the mkdirs was because XCSoarData was not guaranteed to be there if XCSoar had not yet been started.  I am changing it to use the #defines but should we also change System.cpp in a separate PR?

PR #851, opened 2022-01-30
Update documentation for Kobo kernel build

GeekGoneOld:
Detailed description of how to build the Kobo kernels
with specific example for Clara HD.  Also the original
and new .config files are added to GitHub for reference
following the lead of older kernel work.

Brief summary of the changes
----------------------------


lordfolken:
can you --rebase?

GeekGoneOld:
Also, @MaxKellermann can you confirm that per the documentation the tgz file for Clara HD OTG has been copied into the /opt/kobo/kernel folder on every machine that creates release KoboRoot.tgz files.  I know you usually do the releases but others like lordfolken sometimes do the release (I think) so whatever machine they do the release on must have the Clara tgz file just as it is done with the six kernel files.

GeekGoneOld:
> can you --rebase?

I did just a short while ago.  I'm pretty sure I was 0 behind and 1 ahead.

GeekGoneOld:
OK, I give up.  Where is the developer manual source?  Is it still .tex?  What is sphynx and where is it?



> > can you --rebase?
> 
> I did just a short while ago. I'm pretty sure I was 0 behind and 1 ahead.

DanD222:
I would guess it's the .rst files in the doc folder which have the commit messages  "_doc: move the rest of the developer manual to Sphinx_", but also this one for example:  https://github.com/XCSoar/XCSoar/blob/master/doc/build.rst

GeekGoneOld:
> I would guess it's the .rst files in the doc folder which have the commit messages "_doc: move the rest of the developer manual to Sphinx_", but also this one for example: https://github.com/XCSoar/XCSoar/blob/master/doc/build.rst

The commits you mention are mostly deletion of files.  The developer manual you download on xcsoar.org/discover/manual.html is the one I'm looking for the source for.  It was the most recent and most comprehensive guide.  My guess is we have moved that source to some other service but we won't post it on xcsoar.org until the next release, so I'm stuck.

The file you mention seems to predate the developer manual and does not contain as much info.  I'm sure there is more somewhere else.

If it is right there in front of me, I apologise but I would love someone to point it out.  Also, if there is an easy conversion process, help would be appreciated.  I made an entire appendix (compiling kernel) for the old doc (.tex) and if I can easily make it an appendix (or chapter) of the new manual that would be great.  The compile kernel section of the file you mention is out of date and a bit thin.

lordfolken:
The problem was that the developer manual is written in latex. This is a great tool for creating anything from beautifully lay-outed books, thesis, manuals etc. The problem, is ts relatively arcane syntax. For the dev manual this isn't needed, but its much more important that docs be current. Hence the (wip) move to rst, a much more manageable format.

GeekGoneOld:
I will close this PR without including it and will submit a revised version for the new developer manual in a few days.

PR #852, opened 2022-01-30
-version switch, radio command buttons

lordfolken:


lordfolken:
The version switch still needs work. XCSoar.log shouldn't be created in that case. So there needs to be a check in XCSoar.cpp

src/CommandLine.cpp
```diff
@@ -68,6 +68,10 @@ CommandLine::Parse(Args &args)
     if (s[1] == '-')
       s++;
 
+    if (StringIsEqual(s, "-version", 8)) {
+      exit(EXIT_SUCCESS);
+    };
```
MaxKellermann, line 73: Extra semicolon

MaxKellermann (changes requested):

PR #859, opened 2022-02-03
WeGlide step2 add dialog form before and after the upload

August2111:
After the 1st PR #612 there a few tasks were still open:

Brief summary of the changes
----------------------------

- Before starting the upload is a new dialog available, where the user can check and change the necessary settings for uploading: IGC file, user and aircraft - or break, if he don't like the upload after the download of logger file.
- Message window at the end of dialog should have a better style (not an multiline MessageBox). This new dialog is prepared for a fast response to the WeGlide server about copilot, (short) message test and - maybe a deleting after a wrong upload.
- The error handling after known http codes (at http code 406 mostly the file is already uploaded, or at 400 a parameter is wrong - this should be displayed)
- Internal: Switch to structures similar and better aligned to the WeGlide objects 'User', 'Aircraft' and 'Flight'!

Continuation of work on PR #612 and Issue #505 (from December 2020, The main points are now solved! With some cherry picks from #866 issues from #612 solved in XCSoar 7.23.

Again some things are still open:
---------------------------------

- (Solved here: _Better pre-upload query with the ability to 'recalibrate' the pilot and aircraft_)
- Simple TaskDownload of any WeGlide task: Either your own task or that of another pilot (See #523)
- (Solved here: _A better selection of user and aircraft_)
- Creation of Lua access to this functionality for event-driven user guidance (keyboard or gestures)

### DIFF
--- a/src/Cloud/weglide/WeGlideObjects.hpp
+++ b/src/Cloud/weglide/WeGlideObjects.hpp
@@ -1,0 +1,65 @@
+#pragma once
+
+#include "util/StaticString.hxx"
+#include "time/BrokenDate.hpp"
+
+#include <cinttypes>
+#include <tchar.h>
+
+
+namespace WeGlide {
+
+struct User {
+  uint32_t id;
+  BrokenDate birthdate;
+  StaticString<0x80> name;
+
+  constexpr bool IsValid() const noexcept
+  {
+    return id > 0;
+  }
+
+  void Clear() noexcept
+  {
+    id = 0;
+    name.clear();
+    birthdate.Clear();
+  }
+};
+// because it is used in Settings and there it has to be 'trivial' type!
+  static_assert(std::is_trivial<User>::value, "type is not trivial");
+
+struct Aircraft {
+  uint32_t id = 0;
+  StaticString<0x40> name;
+  StaticString<4> kind; // 'MG' - motor aircraft, GL - Glider...
+  StaticString<10> sc_class;
+
+  constexpr bool IsValid() const noexcept
+  {
+    return id > 0;
+  }
+  void Clear() noexcept {
+    id = 0;
+    name.clear();
+    kind.clear();
+    sc_class.clear();
+  }
+};
+
+struct Flight {
+  uint64_t flight_id = 0;
+  User user;
+  Aircraft aircraft;
+  StaticString<0x40> igc_name;
+  StaticString<0x40> scoring_date;
+  StaticString<0x40> registration;
+  StaticString<0x40> competition_id;
+
+  constexpr bool IsValid() const noexcept
+  {
+    return flight_id > 0;
+  }
+};
+
+};
--- a/src/Cloud/weglide/WeGlideSettings.hpp
+++ b/src/Cloud/weglide/WeGlideSettings.hpp
@@ -24,7 +24,7 @@ Copyright_License {
 #ifndef XCSOAR_CLOUD_WEGLIDE_WEGLIDESETTINGS_HPP
 #define XCSOAR_CLOUD_WEGLIDE_WEGLIDESETTINGS_HPP
 
-#include "time/BrokenDate.hpp"
+#include "WeGlideObjects.hpp"
 
 #include <cstdint>
 
@@ -55,10 +55,11 @@ struct WeGlideSettings {
   static constexpr char gliderlist_uri[] = "https://raw.githubusercontent.com/"
     "weglide/GliderList/master/gliderlist.csv";
 
-  uint32_t pilot_id;
-  BrokenDate pilot_birthdate;
+  WeGlide::User pilot;
 
   void SetDefaults() noexcept;
 };
+
+static_assert(std::is_trivial<WeGlideSettings>::value, "type is not trivial");
 
 #endif  // XCSOAR_CLOUD_WEGLIDE_WEGLIDESETTINGS_HPP
--- a/test/src/RunWeGlideUploadFlight.cpp
+++ b/test/src/RunWeGlideUploadFlight.cpp
@@ -40,14 +40,14 @@ Copyright_License {
 struct Instance : CoInstance {
   const Net::ScopeInit net_init{GetEventLoop()};
 
-  boost::json::value value;
+  boost::json::value json;
 
-  Co::InvokeTask DoRun(const WeGlideSettings &settings,
-                       uint_least32_t glider_type,
+  Co::InvokeTask DoRun(const WeGlide::User &user,
+                       uint_least32_t aircraft_id,
                        Path igc_path,
                        ProgressListener &progress)
   {
-    value = co_await WeGlide::UploadFlight(*Net::curl, settings, glider_type,
+    json = co_await WeGlide::UploadFlight(*Net::curl, user, aircraft_id,
                                            igc_path, progress);
   }
 };
@@ -56,28 +56,28 @@ int
 main(int argc, char *argv[])
 try {
   Args args(argc, argv, "PILOT BIRTHDAY GLIDER IGCFILE");
-  const unsigned pilot = atoi(args.ExpectNext());
+  WeGlide::User pilot({ (uint32_t) atoi(args.ExpectNext()) });
   const char *birthday_s = args.ExpectNext();
   const unsigned glider = atoi(args.ExpectNext());
   const auto igc_path = args.ExpectNextPath();
+
   args.ExpectEnd();
 
   unsigned year, month, day;
   if (sscanf(birthday_s, "%04u-%02u-%02u", &year, &month, &day) != 3)
     throw "Failed to parse date";
 
-  WeGlideSettings settings;
-  settings.pilot_id = pilot;
-  settings.pilot_birthdate = {year, month, day};
+  pilot.birthdate = {year, month, day};
+
 
   Instance instance;
   ConsoleOperationEnvironment env;
 
-  instance.Run(instance.DoRun(settings, glider,
+  instance.Run(instance.DoRun(pilot, glider,
                               igc_path, env));
 
   StdioOutputStream _stdout(stdout);
-  Json::Serialize(_stdout, instance.value);
+  Json::Serialize(_stdout, instance.json);
   return EXIT_SUCCESS;
 } catch (...) {
   PrintException(std::current_exception());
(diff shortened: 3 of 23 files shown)


August2111:
Because some workflow/build errors rework started

August2111:
Unfortunately I don't know why Codacy has this problems with this 3 struct members:

- User::Id
- Flight_flight_id and
- HttpResponse::code

All 3 members are definitely used both in XCSoar and in the test RunWeGlideUploadFlight.cpp...
Please have a look to this...

MaxKellermann:
I'm confused, is this the same as https://github.com/XCSoar/XCSoar/pull/859?

August2111:
> I'm confused, is this the same as https://github.com/XCSoar/XCSoar/pull/859?

For this reason I had reset this PR to 'Draft' - the first three commits were the most important to me, mainly because of the bug fixes, so I submitted them as PR#866 - I wanted PR #859 again first open when I've fixed the 3 Codacy issues. But as I noticed from PR #862, commits with 'smaller' Codacy issues are still accepted? I'm struggling with this right now because I have no reason for this warning (unlike the #862 warnings)

August2111:
moved the 1st tree commits to PR #866

src/Cloud/weglide/WeGlideObjects.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include "util/StaticString.hxx"
+#include "time/BrokenDate.hpp"
+
+#include <cinttypes>
+#include <tchar.h>
+
+namespace WeGlide {
+
+  struct User {
```
MaxKellermann, line 11: Bad indent

src/Cloud/weglide/WeGlideSettings.hpp
```diff
@@ -24,7 +24,9 @@ Copyright_License {
 #ifndef XCSOAR_CLOUD_WEGLIDE_WEGLIDESETTINGS_HPP
 #define XCSOAR_CLOUD_WEGLIDE_WEGLIDESETTINGS_HPP
 
+#include "WeGlideObjects.hpp"
 #include "time/BrokenDate.hpp"
+#include "util/StaticString.hxx"
```
MaxKellermann, line 29: Why?

src/Profile/ProfileKeys.cpp
```diff
@@ -97,8 +97,8 @@ const char LanguageFile[] = "LanguageFile"; // pL
 const char InputFile[] = "InputFile"; // pL
 const char PilotName[] = "PilotName";
 const char WeGlideEnabled[] = "WeGlideEnabled";
-const char WeGlidePilotID[] = "WeGlidePilotID";
-const char WeGlidePilotBirthDate[] = "WeGlidePilotBirthDate";
+const char WeGlideUserID[] = "WeGlideUserID";
+const char WeGlideUserBirthDate[] = "WeGlideUserBirthDate";
```
MaxKellermann, line 101: Are you aware that such changes break existing configurations?

test/src/RunWeGlideUploadFlight.cpp
```diff
@@ -42,7 +42,7 @@ struct Instance : CoInstance {
 
   boost::json::value value;
 
-  Co::InvokeTask DoRun(const WeGlideSettings &settings,
+  Co::InvokeTask DoRun(const struct WeGlide::User &user,
```
MaxKellermann, line 45: No `struct`, again!

src/Cloud/weglide/WeGlideObjects.hpp
August2111: Sorry, I've overlooked that

src/Profile/ProfileKeys.cpp
August2111: Yes, I am. No one uses this code (and setting) outside of some developers, but since that's the nomenclature of objects on WeGlide, it's definitely a better name. It is better to straighten it out at the beginning than to work with different identifiers later!

test/src/RunWeGlideUploadFlight.cpp
August2111: Oops, overlooked

src/Profile/ProfileKeys.cpp
MaxKellermann: It's released in XCSoar 7.23, so people will start using it now.
If you break existing configurations, no matter if you think that's okay, that should at least be documented in the commit message. Not doing so is hiding it from reviewers. Don't hide stuff from me.

August2111: I hoped we could get that done before release 7.23 - but I can see that that was a bit 'sporty' on my part ;-), because it was definitely in the starting blocks.

August2111: By the way, I at least hinted at that in the commit message, but probably not clear enough that it directly affects the values ​​already set.
Should that also be mentioned in the NEWS.txt - or is it enough in the commit message?
A description in the manual is also still pending, but I didn't want to tackle that until everything was completely finished! At the moment there is still no dialog window before the upload about the confirmation and possible change of the set values ​​(user, aircraft, filename...). I've come a long way there, too, but I'll have to check it out more carefully before I submit it.

src/Cloud/weglide/WeGlideSettings.hpp
August2111: Not necessary after struct declaration

August2111:
I would like to make some small changes in the setup of objects.
Please wait a moment until I'm done with this.

src/Profile/ProfileKeys.cpp
August2111: Since version 7.23 has been released for some time - and I currently see no further progress on the way to a fast 7.24 - you are doing a lot to turn the OpenVario upside down! - I have decided to to leave the identifiers in the settings as they are already published.
In the end it's best that way: The WeGlide objects are called 'User', but they are used for 'Pilot' and 'Copilot' (and others)

August2111:
@MaxKellermann Now the 'small changes' should be done.
In the meantime I have checked the code several times, I hope that there are not many errors in it.
A rebase to the current 'master' has also been completed!
Unfortunately the old Codacy problem 'unused field' is still not clear...

src/Cloud/weglide/WeGlideObjects.hpp
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  void Clear() noexcept
+  {
+    id = 0;
+    name.clear();
+    birthdate.Clear();
+  }
+};
+// because it is used in Settings and there it has to be 'trivial' type!
```
MaxKellermann, line 29: What does that mean?

src/Cloud/weglide/WeGlideObjects.hpp
```diff
@@ -0,0 +1,65 @@
...
+  static_assert(std::is_trivial<User>::value, "type is not trivial");
```
MaxKellermann, line 30: Bad indent

src/Cloud/weglide/WeGlideObjects.hpp
```diff
@@ -0,0 +1,65 @@
+#pragma once
+
+#include "util/StaticString.hxx"
+#include "time/BrokenDate.hpp"
+
+#include <cinttypes>
```
MaxKellermann, line 6: Why this include?

src/Cloud/weglide/WeGlideObjects.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <tchar.h>
```
MaxKellermann, line 7: Why this include?

src/Cloud/weglide/WeGlideObjects.hpp
```diff
@@ -0,0 +1,65 @@
...
+  }
+  void Clear() noexcept {
+    id = 0;
+    name.clear();
+    kind.clear();
+    sc_class.clear();
+  }
+};
+
+struct Flight {
```
MaxKellermann, line 50: Unused struct

MaxKellermann:
The first commit doesn't even compile!

src/Cloud/weglide/WeGlideObjects.hpp
August2111: The structure is usable only if it is a 'trivial type' - if I don't pay attention to this, I get an error somewhere else, which in itself is not self-explanatory - at least not for me. With the static_assertion in the following line, such an incorrect member type or member function is immediately displayed correctly.

src/Cloud/weglide/WeGlideObjects.hpp
August2111: Ok, ;-(

src/Cloud/weglide/WeGlideObjects.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <cinttypes>
+#include <tchar.h>
+
+
+namespace WeGlide {
+
+struct User {
+  uint32_t id;
+  BrokenDate birthdate;
+  StaticString<0x80> name;
```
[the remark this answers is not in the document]
August2111: Yes, the new field 'name' is necessary for a later administration of the users - and in my opinion also for the structure itself not wrong and incomprehensible. Inserting it later inflates the history unnecessarily, but I'm trying to understand that such extensions are only made when they are needed... The now larger structure makes it clearer, in my opinion, why I outsourced it to a structure at all - well, that shouldn't play a role for an extension anyway...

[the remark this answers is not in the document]
MaxKellermann: Add it in the commit that actually uses it. Or add it right from the start. But don't add it hidden in a refactoring commit.

src/Cloud/weglide/WeGlideObjects.hpp
August2111: Ok, that is the same with the additional field in struct User...

src/Cloud/weglide/WeGlideObjects.hpp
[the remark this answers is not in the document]
MaxKellermann: It makes no sense to add a struct that nobody uses. And even less in a commit that is supposed to do refactoring. All this does is confuse me and waste my time.

PR #861, opened 2022-02-04
Small Changes for MSVC/_WIN32

August2111:
various necessary and optional changes for MSVC compilation

### DIFF
--- a/test/src/DumpVario.cpp
+++ b/test/src/DumpVario.cpp
@@ -28,7 +28,11 @@ Copyright_License {
 
 #include <stdio.h>
 #include <stdlib.h>
-#include <unistd.h>
+#if defined(_MSC_VER)
+# include <windows.h>
+#else
+# include <unistd.h>
+#endif
 
 int
 main(int argc, char **argv)


src/Task/ValidationErrorStrings.cpp
```diff
@@ -47,7 +47,7 @@ static_assert(ARRAY_SIZE(validation_error_strings) == unsigned(TaskValidationErr
               "Wrong array size");
 
 const TCHAR*
-getTaskValidationErrors(const TaskValidationErrorSet v)
+getTaskValidationErrors(const TaskValidationErrorSet &v)
```
MaxKellermann, line 50: ..... why oh why?
I don't understand the commit message.

August2111: From my understanding, passing a reference to a subfunction is better than copying the entire object. But of course that's only true if the caller can't destroy the object while it's still being used in the subfunction (as can happen in multithreaded applications though). Is that the case here? Then, of course, I immediately withdraw my amendment!

MaxKellermann: > From my understanding, passing a reference to a subfunction is better than copying the entire object.

But ... why is that your understanding? When I ask "why", an answer like "because that's my understanding" is never a satisfying answer. It doesn't explain anything at all.

I feared that you would reply with something like that, but when I ask "why" questions, I always want to first assume that you have good reasons, and that you can explain them. But in this case, again, I don't think you have good reasons, only some misguided (corporate?) C++ policy somebody taught you, but you never questioned it. Free yourself of this misguided policies, boot your brain and start thinking clearly.

Take a step back, and look what `TaskValidationErrorSet` really is. Did you look it up when you wrote that code? (`git grep` is your friend, it presents you the definition in a millisecond or so):
https://github.com/XCSoar/XCSoar/blob/299a01a61cb46c5743599d4ac0650ae1478ac9d6/src/Engine/Task/Factory/ValidationError.hpp#L57
And... what's this `EnumBitSet`?
https://github.com/XCSoar/XCSoar/blob/299a01a61cb46c5743599d4ac0650ae1478ac9d6/src/util/EnumBitSet.hxx#L37-L47

That's it - it's only a trivial wrapper for a 32 bit integer. In your imagination, passing a 64 bit pointer to a 32 bit integer is better than passing the 32 bit integer.
You didn't notice how poor of an idea that is because, I believe, you didn't know what you were doing, you didn't know it's just a 32 bit integer.

In the caller code, that 32 bit integer is probably already in a CPU register. After your patch, the compiler has to change all callers to allocate 4 bytes of stack space and copy the CPU register to the stack allocation, and assign the address of that stack allocation to that CPU register, to implement the new calling convention of the function you changed. Can you understand how messed up this overhead is you are introducing, for no reason?

Some objects should not be passed by value (if they are not consumed by the callee (since C++11)), but there's no unconditional rule or policy to do that for all classes. Use the grey matter in your head to decide that. References to trivial objects that can be stored in just one CPU register almost never make sense as a reference/pointer. This here is one, and it doesn't make sense.

Until 9 years ago, `TaskValidationErrorSet` was  a `std::vector`, and that object should be passed by reference. And indeed it was. It was my commit which changed this to a trivial integer which also removed the reference from the `getTaskValidationErrors()` parametr. Look: https://github.com/XCSoar/XCSoar/commit/21cc9ab4d80b4aaf788278c1ffa9e512b3b10665#diff-e5fdef1541b218e40e141c2d2bf831605f83ebb0be9c5a24da27075120f7b18a

Removing the reference was my explicit decision 9 years ago, and your commit now attempts to revert that decision.

test/src/DumpVario.cpp
```diff
@@ -28,7 +28,11 @@ Copyright_License {
 
 #include <stdio.h>
 #include <stdlib.h>
-#include <unistd.h>
+#if defined(_MSC_VER)
+# include <windows.h>
```
MaxKellermann, line 32: > the function declarations are in windows.h

MSDN documentation disagrees.

August2111:
Thanks for cherry-pick of the 1st two commits

test/src/DumpVario.cpp
August2111: Yes, in `windows.h` fom MinGW - but not in `Windows.h` from
`C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\um` - used from MSVC compiler
From my point of view MSDN is right about that.

MaxKellermann: I have **no idea** what your response means. I don't understand any of it.

August2111: With MSVC I have no unistd.h available, this is a part of the MinGW package only available for the GCC compiler. With a pure Windows system with VisualeStudio installed there is no header file unistd.h! 
All the Windows system headers you can find in the current Windows Kit folder.

MaxKellermann: I know you have no unistd.h, but ... what's your point?
My point was that your statement about windows.h is wrong. What do your replies have to do with that? You confuse me.

PR #866, opened 2022-02-06
Weglide step1 Small Update for bugfixing issues on PR #612

August2111:
With PR #612 there are 2 bugs at uploading IGC files from logger:

Brief summary of the changes
----------------------------

- Bugfix: After downloading a file from a logger the dialog question had the wrong logic!
- Bugfix:  An incorrect or faulty path type was previously used - that's why the IGC file was not found and no upload took place ;-( (f.e. on OpenVario,...)

### DIFF
--- a/src/Logger/ExternalLogger.cpp
+++ b/src/Logger/ExternalLogger.cpp
@@ -354,7 +354,7 @@ ExternalLogger::DownloadFlightFrom(DeviceDescriptor &device)
       weglide_settings.pilot_id > 0) {
       // ask whether this IGC should be uploaded to WeGlide
       if (ShowMessageBox(_("Do you want to upload this flight to WeGlide?"),
-        _("Upload flight"), MB_YESNO | MB_ICONQUESTION) != IDYES) {
+        _("Upload flight"), MB_YESNO | MB_ICONQUESTION) == IDYES) {
         WeGlide::UploadIGCFile(igc_path);
       }
     }
--- a/src/Logger/ExternalLogger.cpp
+++ b/src/Logger/ExternalLogger.cpp
@@ -339,8 +339,8 @@ ExternalLogger::DownloadFlightFrom(DeviceDescriptor &device)
     FormatIGCFilenameLong(name, date, header.manufacturer, header.id,
                           header.flight);
 
-    Path igc_path = AllocatedPath::Build(logs_path, name);
-    transaction.SetPath(igc_path);
+    auto igc_path = AllocatedPath::Build(logs_path, name);
+    transaction.SetPath((Path)igc_path);
     
     try {
       transaction.Commit();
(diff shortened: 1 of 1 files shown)


August2111:
Because I have some problems with the usage of Codacy Code Analysis I split the PR #859 in two parts - one mainly for the bugfixing - and the improvement part will be come later...

August2111:
This commit is critical for version 7.23 to have correct and expected behavior!

src/Logger/ExternalLogger.cpp
```diff
@@ -354,8 +354,9 @@ ExternalLogger::DownloadFlightFrom(DeviceDescriptor &device)
       weglide_settings.pilot_id > 0) {
       // ask whether this IGC should be uploaded to WeGlide
       if (ShowMessageBox(_("Do you want to upload this flight to WeGlide?"),
-        _("Upload flight"), MB_YESNO | MB_ICONQUESTION) == IDYES) {
-        WeGlide::UploadIGCFile(igc_path);
+                           _("Upload flight"),
+                           MB_YESNO | MB_ICONQUESTION) == IDYES) {
+          WeGlide::UploadIGCFile((Path)igc_path);
```
MaxKellermann, line 361: What UTF8-alignment-path bug does this change fix?

src/Logger/ExternalLogger.cpp
```diff
@@ -339,8 +339,8 @@ ExternalLogger::DownloadFlightFrom(DeviceDescriptor &device)
     FormatIGCFilenameLong(name, date, header.manufacturer, header.id,
                           header.flight);
 
-    Path igc_path = AllocatedPath::Build(logs_path, name);
-    transaction.SetPath(igc_path);
+    auto igc_path = AllocatedPath::Build(logs_path, name);
+    transaction.SetPath((Path)igc_path);
```
MaxKellermann, line 343: I don't understand this change, and your commit message doesn't explain it. It doesn't even explain what did not work. It explains nothing.

August2111: I don't understand and don't know why the casting in the line 342 
         `Path igc_path = AllocatedPath::Build(logs_path, name);`
was wrong- the path string in `igc_path` was in my trials with OpenVario only random mix of characters... Unfortunately I haven't a good setup for debugging in the OpenVario on my desk - I made some experiments with try and error to solve this issue - and with a variable declariation `AllocatedPath igc_path = ...`  (or `auto igc_path = ...`) the upload works as expected, but of course I have to cast the igc_path to the subfunctions.

If I have some free time tomorrow I'll dig a little deeper/debug to find out what's wrong in the previous code - but usually I just wanted to use the bottom layer (and not make it worse myself) as I often don't understand which functions are allowed for the 'XCSoar Framework' - and which are not!

August2111:
Rebased only...

src/Logger/ExternalLogger.cpp
MaxKellermann: Isn't ... it obvious why your code didn't work? It is so obvious that I wonder why I didn't see it when I read your code before merging it.
https://github.com/XCSoar/XCSoar/blob/9a559397c2b32236b5cd519415e41498a643b702/src/Logger/ExternalLogger.cpp#L342
So you have a function which returns an `AllocatedPath`, and you assign it to a `Path`.
What happens now with the `AllocatedPath` temporary? You know C++, do you? You know the answer.
Of course, it gets destructed right after the `Path igc_path` variable got initialised.
Where does `igc_path` now point to? To memory which was already freed by `~AllocatedPath()`.

I didn't understand your change, because it was a very obscure, and your commit showed that you still didn't understand what you were doing. You did random things until the compiler stopped complaining, and until XCSoar stopped crashing. That's "try'n'error" programming, but it's bad practice.

Please please never let me review code from you that was created with "try'n'error" programming. If you don't know what you're doing, at least state clearly so in the commit message. I can always explain, but please please don't leave me wondering what the hell you were doing, and don't let me ask all those "why??" questions when I could better already know that you don't have an answer, because you have no idea either.

August2111:
...taken over with cherry-pick, the last topic will be committed in #859

PR #865, opened 2022-02-06
Radio frequency card

davespragg:
Brief summary of the changes
----------------------------

Display a list of user-configurable radio channels, and allow user to update frequency on a connected radio
Invoked from Info menu #2 or pressing F9

Related issues and discussions
------------------------------

See https://forum.xcsoar.org/viewtopic.php?f=19&p=22935#p22935

PR #867, opened 2022-02-07
Radio Frequency Card - supercedes #865

davespragg:
Implement Frequency Card Dialog - displays a list of radio channels (name, frequency) and allows a connected radio to be tuned to the frequency selected.

Brief summary of the changes
----------------------------
Display a list of user-configurable radio channels, and allow user to update frequency on a connected radio
Invoked from Info menu https://github.com/XCSoar/XCSoar/pull/2 or pressing F9

Supercedes #865
See https://forum.xcsoar.org/viewtopic.php?f=19&p=22935#p22935

Tested in Windows (standalone) and Cubie/OpenVario (with ACD57 and KRT2 radio)


src/Dialogs/FrequencyDialog.cpp
```diff
@@ -0,0 +1,199 @@
...
+FrequencyDialogShowModal() noexcept
+{
+  static std::vector<FrequencyListRenderer::RadioChannel> channels;
+
+  const DialogLook &look = UIGlobals::GetDialogLook();
+
+  auto widget = std::make_unique<FrequencyListWidget>(look,&channels);
+
+  static bool first = true;
+  static bool haveRows = false;
```
MaxKellermann, line 177: No global variables, please.

src/Dialogs/FrequencyDialog.cpp
```diff
@@ -0,0 +1,199 @@
...
+FrequencyListWidget::CreateButtons(WidgetDialog &dialog)
+{
+	  standby_button = dialog.AddButton(_("Set Standby Frequency"), [this](){
+	    unsigned index = GetCursorIndex();
+	    assert(index < channels->size());
+	    const RadioChannel *channel = &channels->at(index);
+	    ActionInterface::SetStandbyFrequency(channel->radio_frequency,
+	            channel->name.c_str());
+	    cancel_button->Click();
+});
```
MaxKellermann, line 96: Bad indent

src/Dialogs/FrequencyDialog.cpp
```diff
@@ -0,0 +1,199 @@
...
+    return true;
+  }
+
+  void OnActivateItem(unsigned index) noexcept override;
+};
+
+void
+FrequencyListWidget::CreateButtons(WidgetDialog &dialog)
+{
+	  standby_button = dialog.AddButton(_("Set Standby Frequency"), [this](){
```
MaxKellermann, line 89: Bad indent

MaxKellermann:
I'm not yet convinced that we need another dialog. I still think this should be integrated in the Alternates dialog, because it's the dialog you open up when you need a plan B.

src/Dialogs/FrequencyDialog.cpp
```diff
@@ -0,0 +1,199 @@
...
+  void Prepare(ContainerWindow &parent, const PixelRect &rc) noexcept override;
+
+  /* virtual methods from class List::Handler */
+  void OnPaintItem(Canvas &canvas, const PixelRect rc,
+                   unsigned index) noexcept override {
+    assert(index < channels->size());
+
+    const RadioChannel &channel = channels->at(index);
+
+    FrequencyListRenderer::Draw(canvas, rc, channel,
```
MaxKellermann, line 75: Why is this in a separate library?

src/Dialogs/FrequencyDialog.cpp
```diff
@@ -0,0 +1,199 @@
...
+public:
+  /* virtual methods from class Widget */
+  void Prepare(ContainerWindow &parent, const PixelRect &rc) noexcept override;
+
+  /* virtual methods from class List::Handler */
+  void OnPaintItem(Canvas &canvas, const PixelRect rc,
+                   unsigned index) noexcept override {
+    assert(index < channels->size());
+
+    const RadioChannel &channel = channels->at(index);
```
MaxKellermann, line 73: Why use method `at()`?

src/Dialogs/FrequencyDialog.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+	if (StringIsEmpty(line))
+	  continue;
+
+	TCHAR ctemp[4096];
+	if (_tcslen(line) >= ARRAY_SIZE(ctemp))
+	  /* line too long for buffer */
+	  continue;
+
+	const TCHAR *params[20];
```
MaxKellermann, line 150: Why ... 20?

davespragg:
I don't think this belongs in the Alternates dialog specifically because it's not intended for a "Plan B" situation. Use cases would include switching to a Comp frequency, switching to an airfield frequency because you're going close to their airspace (or want to enter it, but not to land), switching to a cloud flying frequency because you want to thermal into cloud, switching to an XC training/discussion/situational awareness frequency.

src/Dialogs/FrequencyDialog.cpp
davespragg: Refactored

src/Dialogs/FrequencyDialog.cpp
davespragg: Fixed

src/Dialogs/FrequencyDialog.cpp
davespragg: Fixed

src/Dialogs/FrequencyDialog.cpp
davespragg: Refactored to use []

src/Dialogs/FrequencyDialog.cpp
davespragg: Misunderstanding of how ExtractParameters behaves when parsing too many parameters; now fixed.

src/Dialogs/FrequencyDialog.cpp
```diff
@@ -167,24 +179,18 @@ FrequencyListWidget::UpdateList() noexcept
 void
 FrequencyDialogShowModal() noexcept
 {
-  static std::vector<FrequencyListRenderer::RadioChannel> channels;
+  static std::vector<FrequencyListWidget::RadioChannel> channels;
```
MaxKellermann, line 182: Another global variable

src/Dialogs/FrequencyDialog.cpp
```diff
@@ -167,24 +179,18 @@ FrequencyListWidget::UpdateList() noexcept
...
 
   const DialogLook &look = UIGlobals::GetDialogLook();
 
   auto widget = std::make_unique<FrequencyListWidget>(look,&channels);
 
-  static bool first = true;
-  static bool haveRows = false;
-
-  if (first == true || FrequenciesFileChanged == true) {
+  if (channels.size() == 0 || FrequenciesFileChanged == true) {
```
MaxKellermann, line 188: Never check `size()==0`, try `empty()` instead, which is often faster.
What happens if the file is empty? Will it reload the file every time?

src/Dialogs/FrequencyDialog.cpp
```diff
@@ -37,14 +36,18 @@ Copyright_License {
...
   const DialogLook &dialog_look;
   TextRowRenderer row_renderer;
   Button *active_button, *standby_button, *cancel_button;
+
+public:
+  struct RadioChannel {
+	  tstring name;
+	  RadioFrequency radio_frequency;
+  };
   std::vector<RadioChannel> *channels;
```
MaxKellermann, line 51: Why is this a pointer? Why is it not `const`?

src/Dialogs/FrequencyDialog.cpp
```diff
@@ -86,22 +98,22 @@ class FrequencyListWidget final
...
-	    unsigned index = GetCursorIndex();
-	    assert(index < channels->size());
-	    const RadioChannel *channel = &channels->at(index);
-	    ActionInterface::SetStandbyFrequency(channel->radio_frequency,
-	            channel->name.c_str());
-	    cancel_button->Click();
+  standby_button = dialog.AddButton(_("Set Standby Frequency"), [this](){
+    unsigned index = GetCursorIndex();
+    assert(index < channels->size());
+    const RadioChannel *channel = &(*channels)[index];
```
MaxKellermann, line 104: Why is this a pointer?

src/Dialogs/FrequencyDialog.cpp
```diff
@@ -70,10 +73,19 @@ class FrequencyListWidget final
...
+    // Draw name and frequency
+    row_renderer.DrawTextRow(canvas, rc, channel.name.c_str());
+
+    if (channel.radio_frequency.IsDefined()) {
+    	StaticString<30> buffer;
+    	TCHAR radio[20];
+      channel.radio_frequency.Format(radio, ARRAY_SIZE(radio));
+      buffer.Format(_T("%s MHz"), radio);
+      row_renderer.DrawRightColumn(canvas, rc, buffer);
+      }
```
MaxKellermann, line 87: Bad indent

MaxKellermann:
I still don't think this is a good case for a new top-level dialog, but I'll agree to merge the feature, if you promise to keep maintaining the code.

The code needs some adjustments, and we'll probably need a few more rounds of fixups.

I'm not particularly happy with the file format. You used a simple file format with two columns, and you used the `ExtractParameters()` function which is horrible legacy code from 15 years ago or so.
Your code ignores parser errors completely, and just makes bad lines disappear.

How should users write those files? The file format is undocumented. Before we merge it, there must be documentation. I think it's okay to have the documentation in the developer manual.

What worries me most is how narrow this feature is, and that it's designed in a way that can hardly be extended later. I fear that this feature will be legacy cruft that we have to carry around forever, but we'll never be able to adapt it to new requirements, because it may break the feature for existing users.

The simple file format is a good feature, but it's also not extensible. For example, think of a frequency for the landing site; why can't I open up that waypoint's details from your new dialog? Or an airspace frequency - I can't open the airspace details or zoom the map on that airspace. Because with your file format, there cannot ever be an association with other objects.
These features do not exist now, but the narrow way you designed your code, these features cannot be added in the future.

Please rethink your design on how to make it extensible. Maybe you can come up with a file format that allows future extensions, but is still simple enough to write.

davespragg:
Thanks - that's really helpful feedback. I'll go away and have a think...

Only reason I haven't written any documentation yet is that I wanted to finalise the design first; I'll definitely do that when we're agreed how this should work.

src/Dialogs/FrequencyDialog.cpp
davespragg: This is a pattern I borrowed from elsewhere in the codebase (eg dlgChecklist), but as you've pointed out separately there's quite a bit of legacy code. What would be the preferred way of persisting an object (in this case the collection of channels, but my questions's more general) between invocations of the dialog?

MaxKellermann: The checklist dialog is a special example of bad and old code which I havn't rewritten, because I never cared about this feature.
You can't imagine how many global variables I have removed in the last 13 years. Hundreds? Or thousands? I can't remember. XCSoar used to be a VERY horrible mess of intertwined global state. It's still messy, but it used to be much worse. My task is to keep new bad code out, and try to reduce the amount of existing bad code.
There still isn't a good "one-fits-all" solution for global state; for example there is Components.hpp which is a huge unholy collection of global variables which havn't found a proper home somewhere else. But at least they get initialised and deinitialised properly. (More or less; if the XCSoar process gets recycled by the Android OS, and the BSS section doesn't get reinitialised, it gets dirty... sometimes, I wish I didn't know and care about all that, then I could sleep better. I could talk for days about how horrible real-life code is.)

In your case, I'd say: what's the point of persisting this object? You want to load it only once. But for now (because we don't have a proper solution to add global state for random one-off features), just load the file every time the dialog opens. It's a kilobyte or so, probably less. Don't bother caching it. For now.
Of course, in the long run, it should be loaded only once, and XCSoar should access the file system only when necessary. But we're not there yet.

MaxKellermann:
No progress for more than a year.

PR #889, opened 2022-03-14
Add support for Kobo Libra H2O

### LINKED ISSUE #882: Kobo Libra H2O touch rotated and partly unreachable
XCSoar 7.23
Kobo Libra H2O

Expected behavior
-----------------

Touch screen is usable.

Actual behavior
---------------

Touch input seems 90° rotated. Button shows animation for action despite touching the 90° rotated position.
Not the whole screen is reachable.

Install XCSoar 7.23 on Kobo Libra H2O

cmoetzing:
The device has similar hardware as Kobo Clara HD.

Closes #882

### DIFF
--- a/src/Kobo/Model.cpp
+++ b/src/Kobo/Model.cpp
@@ -68,6 +68,7 @@ static constexpr struct {
   { "SN-N249", KoboModel::CLARA_HD },
   { "SN-N306", KoboModel::NIA },
   { "SN-N418", KoboModel::LIBRA2 },
+  { "SN-N873", KoboModel::LIBRA_H2O },
 };
 
 static KoboModel
--- a/src/Kobo/Model.hpp
+++ b/src/Kobo/Model.hpp
@@ -39,6 +39,7 @@ enum class KoboModel {
   CLARA_HD,
   NIA,
   LIBRA2,
+  LIBRA_H2O,
 };
 
 [[gnu::const]]
--- a/src/Kobo/System.cpp
+++ b/src/Kobo/System.cpp
@@ -71,9 +71,11 @@ bool
 KoboReboot()
 {
 #ifdef KOBO
-  /* CLARA_HD requires the -f option to for reboot to work */
-  if  (DetectKoboModel() == KoboModel::CLARA_HD || DetectKoboModel() == KoboModel::LIBRA2)
+  KoboModel kobo_model = DetectKoboModel();
+  if  (kobo_model == KoboModel::CLARA_HD || kobo_model == KoboModel::LIBRA2
+      || kobo_model == KoboModel::LIBRA_H2O)
   {
+    /* some models require the -f option for reboot to work */
     return Run("/sbin/reboot", "-f");
   }
   return Run("/sbin/reboot");
@@ -153,6 +155,7 @@ KoboExportUSBStorage()
 
   case KoboModel::CLARA_HD:
   case KoboModel::LIBRA2:
+  case KoboModel::LIBRA_H2O:
     InsMod("/drivers/mx6sll-ntx/usb/gadget/configfs.ko");
     InsMod("/drivers/mx6sll-ntx/usb/gadget/libcomposite.ko");
     InsMod("/drivers/mx6sll-ntx/usb/gadget/usb_f_mass_storage.ko");
@@ -180,7 +183,9 @@ void
 KoboUnexportUSBStorage()
 {
 #ifdef KOBO
-  if(DetectKoboModel() == KoboModel::CLARA_HD || DetectKoboModel() == KoboModel::LIBRA2)
+  KoboModel kobo_model = DetectKoboModel();
+  if(kobo_model == KoboModel::CLARA_HD || kobo_model == KoboModel::LIBRA2
+      || kobo_model == KoboModel::LIBRA_H2O)
   {
     RmMod("g_file_storage");
     RmMod("usb_f_mass_storage");
@@ -236,6 +241,7 @@ KoboWifiOn()
     break;
 
   case KoboModel::CLARA_HD:
+  case KoboModel::LIBRA_H2O:
     InsMod("/drivers/mx6sll-ntx/wifi/sdio_wifi_pwr.ko");
     InsMod("/drivers/mx6sll-ntx/wifi/8189fs.ko");
     break;
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -4,6 +4,8 @@ Version 7.29 - not yet released
   - fix crash when arc airspace has no center
 * Android
   - get rid of not usable USB interfaces in the 'Device -> Port' list
+* Kobo
+  - Add support for Libra H2O
 
 Version 7.28 - 2022/10/29
 * data files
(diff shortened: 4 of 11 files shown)


src/Kobo/System.cpp
```diff
@@ -178,7 +179,7 @@ void
 KoboUnexportUSBStorage()
 {
 #ifdef KOBO
-  if(DetectKoboModel() == KoboModel::CLARA_HD)
+  if(DetectKoboModel() == KoboModel::CLARA_HD || DetectKoboModel() == KoboModel::LIBRA_H2O)
```
MaxKellermann, line 182: Line toooooooooooooooooooooooooo long

cmoetzing: add a line break

src/DisplayOrientation.hpp
```diff
@@ -47,9 +49,13 @@ static constexpr DisplayOrientation DEFAULT_DISPLAY_ORIENTATION =
 constexpr DisplayOrientation
 TranslateDefaultDisplayOrientation(DisplayOrientation orientation)
 {
-  return orientation == DisplayOrientation::DEFAULT
-    ? DEFAULT_DISPLAY_ORIENTATION
-    : orientation;
+  if(orientation == DisplayOrientation::DEFAULT) {
+    if(DetectKoboModel() == KoboModel::LIBRA_H2O) {
```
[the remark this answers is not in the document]
cmoetzing: First of all: I am not a C++ developer. :)
I can compile the code and it runs on my Kobo (with correct screen input).
I have no idead what `constexpr` means. Can you please elaborate why the function should not be called here and what a solution would be?

lordfolken:
hmm you removed our workflows. You can disable the workflows in your repository settings.

lordfolken (changes requested):
Checkout your branch before your changes and then do
```git cherry-pick <commitid>```
Then force push again. 
that way you can eliminate this change

**Comment by cmoetzing** [comment]
I don't know if you allow global variables. Overall this is a cleaner version (I think). Open for suggestions.

**Comment by MaxKellermann** [comment]
I'm confused by that commit diff. It moves chunks of code around, but it's documented to add a feature. I can't see what is really changed.

**Comment by cmoetzing** [comment]
> I'm confused by that commit diff. It moves chunks of code around, but it's documented to add a feature. I can't see what is really changed.

Split it into several commits.

File: src/DisplayOrientation.hpp [hunk 3]
```diff
@@ -34,25 +34,18 @@ enum class DisplayOrientation : uint8_t {
...
 };
 
-#ifdef KOBO
-/* Kobo defaults to portrait */
-static constexpr DisplayOrientation DEFAULT_DISPLAY_ORIENTATION =
-  DisplayOrientation::PORTRAIT;
-#else
-/* everything else defaults to landscape */
-static constexpr DisplayOrientation DEFAULT_DISPLAY_ORIENTATION =
+static DisplayOrientation default_display_orientation =
```
MaxKellermann: I don't understand this whole commit, and the commit message doesn't explain anything.

src/Kobo/Model.cpp
```diff
@@ -78,12 +82,22 @@ DetectKoboModel(const char *p)
   return KoboModel::UNKNOWN;
 }
 
-KoboModel
-DetectKoboModel()
-{
+void InitKoboModel() {
```
MaxKellermann, line 85: Bad coding style

src/Kobo/Model.hpp
```diff
@@ -37,10 +37,11 @@ enum class KoboModel {
...
   NIA,
+  LIBRA_H2O
 };
 
-gcc_const
-KoboModel
-DetectKoboModel();
+void InitKoboModel();
+
+KoboModel GetKoboModel();
```
MaxKellermann, line 45: This is an API change which has nothing to do with Libra H2O, does it? This change obfuscates the commitdiff and makes review very hard. I don't even know why you're doing this; your commit message doesn't explain anything.

src/Kobo/Model.cpp
```diff
@@ -78,12 +82,22 @@ DetectKoboModel(const char *p)
...
   char buffer[16];
-  if (!ReadFromFile("/dev/mmcblk0", 0x200, buffer, sizeof(buffer)))
-    return KoboModel::UNKNOWN;
+    if (!ReadFromFile("/dev/mmcblk0", 0x200, buffer, sizeof(buffer)))
```
MaxKellermann, line 87: Bad indent, changed for no obvious reason to 4 spaces.

src/Kobo/Model.cpp
```diff
@@ -78,12 +82,22 @@ DetectKoboModel(const char *p)
...
+      kobo_model = KoboModel::UNKNOWN;
+    else
+      kobo_model = DetectKoboModel(buffer);
+
+  switch(kobo_model) {
+    case KoboModel::LIBRA_H2O:
```
MaxKellermann, line 93: Bad indent.

src/Kobo/PowerOff.cpp
```diff
@@ -164,7 +168,7 @@ int main(int argc, char **argv)
 
   /* now we can power off the Kobo; the picture remains on the
      screen */
-  if (DetectKoboModel() == KoboModel::GLO_HD)
+  if (GetKoboModel() == KoboModel::GLO_HD)
```
MaxKellermann, line 171: Lots of changes unrelated to Libra H2O

MaxKellermann (changes requested):
I'm confused by this PR.

MaxKellermann:
> Input and display does not work properly

What does that mean, exactly? "Does not work" is the worst possible problem description, because all it does is state the presence of the problem, but says nothing else.

>  The Libra uses a different default display orientation.

What does "default display orientation" mean in this context? How is it part of the problem?

> The above is why there are also changes in places general Kobo stuff is handled, not only Kobo Libra stuff.

You can do any number of things in one PR, those different things should be separate commits, or else you make it very hard to review your changes. The second thing that made your PR very hard to review is the lack of a good commit message; there was no explanation for anything.

MaxKellermann:
> With "default display orientation" I mean the literal meaning of the words: the orientation of the drawn Image on the screen if you do a fresh install on the device an power it up the first time without changing any settings. The default drawing direction of XCSoar means I see a 90° rotated image on the device which runs outside the screen edges so not the whole image is visible and touch input does not trigger actions under you finger but approx. 90° rotated aswell.

Where does this come from? Is the touch screen input rotated as well? What is the fb rotation setting?

MaxKellermann:
> How do I check?

`/sys/class/graphics/fb0/rotate` and `/sys/class/graphics/fbcon/rotate`

cmoetzing:
```
~ # cat /sys/class/graphics/fb0/rotate
3
```
File `/sys/class/graphics/fbcon/rotate` does not exist.

ssosas:
Hi all, I have a similar problem on Aura H2O with xcsoar v7.25. All works fine but display is mirrored (180°) and you need to guess where real buttons are to interact with the touchscreen.

Have you any v7.xx beta which fix that?

Many thanks!

cmoetzing:
Base on the recent support for the Libra 2 I updated my PR. It is indeed a lot cleaner now.

PR #907, opened 2022-04-17
Weather/Rasp: Thermalmap, 15min Intervals and Rasp URLs in Repository

Developer66:
> Is there a way to do this on android or to generally introduce transparency to the RASP overlays?

I don't think there presently is. 
However Topography can have an alpha value and is then rendered transparently: src/Topography/TopographyFIleRenderer.cpp

### DIFF
--- a/src/Weather/Rasp/RaspCache.cpp
+++ b/src/Weather/Rasp/RaspCache.cpp
@@ -39,9 +39,9 @@ RaspCache::RaspCache(const RaspStore &_store, unsigned _parameter) noexcept
 RaspCache::~RaspCache() noexcept = default;
 
 static constexpr unsigned
-ToHalfHours(BrokenTime t)
+ToQuarterHours(BrokenTime t)
 {
-  return t.hour * 2u + t.minute / 30;
+  return t.hour * 4u + t.minute / 15;
 }
 
 const TCHAR *
@@ -62,7 +62,7 @@ RaspCache::GetMapLabel() const
 void
 RaspCache::SetTime(BrokenTime t)
 {
-  unsigned i = t.IsPlausible() ? ToHalfHours(t) : 0;
+  unsigned i = t.IsPlausible() ? ToQuarterHours(t) : 0;
   time = i;
 }
 
@@ -91,7 +91,7 @@ RaspCache::Reload(BrokenTime time_local, OperationEnvironment &operation)
          time */
       return;
 
-    effective_time = ToHalfHours(time_local);
+    effective_time = ToQuarterHours(time_local);
     assert(effective_time < RaspStore::MAX_WEATHER_TIMES);
   }
 
--- a/src/Weather/Rasp/RaspStore.cpp
+++ b/src/Weather/Rasp/RaspStore.cpp
@@ -104,7 +104,7 @@ RaspStore::MapItem::MapItem(const TCHAR *_name)
 BrokenTime
 RaspStore::IndexToTime(unsigned index)
 {
-  return BrokenTime(index / 2, index % 2 == 0 ? 0 : 30);
+  return BrokenTime(index / 4, (index % 4) * 15);
 }
 
 unsigned
--- a/src/Weather/Rasp/RaspStore.hpp
+++ b/src/Weather/Rasp/RaspStore.hpp
@@ -47,7 +47,7 @@ struct GeoPoint;
 class RaspStore {
 public:
   static constexpr unsigned MAX_WEATHER_MAP = 50; /**< Max number of items stored */
-  static constexpr unsigned MAX_WEATHER_TIMES = 48; /**< Max time segments of each item */
+  static constexpr unsigned MAX_WEATHER_TIMES = 96; /**< Max time segments of each item */
 
   struct MapInfo {
     const TCHAR *name;
--- a/src/Weather/Rasp/RaspStore.hpp
+++ b/src/Weather/Rasp/RaspStore.hpp
@@ -46,7 +46,7 @@ struct GeoPoint;
  */
 class RaspStore {
 public:
-  static constexpr unsigned MAX_WEATHER_MAP = 16; /**< Max number of items stored */
+  static constexpr unsigned MAX_WEATHER_MAP = 50; /**< Max number of items stored */
   static constexpr unsigned MAX_WEATHER_TIMES = 48; /**< Max time segments of each item */
 
   struct MapInfo {
--- a/src/Weather/Rasp/RaspStyle.cpp
+++ b/src/Weather/Rasp/RaspStyle.cpp
@@ -2,7 +2,7 @@
 Copyright_License {
 
   XCSoar Glide Computer - http://www.xcsoar.org/
-  Copyright (C) 2000-2021 The XCSoar Project
+  Copyright (C) 2000-2022 The XCSoar Project
   A detailed list of copyright holders can be found in the file "AUTHORS".
 
   This program is free software; you can redistribute it and/or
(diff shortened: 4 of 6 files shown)


lordfolken:
linking this with the repository here: https://github.com/XCSoar/xcsoar-data-content/issues/237

lordfolken:
Hey @Developer66 
I just merged: https://github.com/XCSoar/xcsoar-data-content/issues/237, so yours and other rasp sources should show up in this branches build.

Developer66:
Hey thanks.
I started some work locally to add the current and new providers to the data repository... some work to do

lordfolken:
can you squash the two last commits please? so we have a buildable version for each commit.

Developer66:
It should work now @lordfolken

src/Weather/Rasp/RaspCache.cpp
```diff
@@ -39,9 +39,9 @@ RaspCache::RaspCache(const RaspStore &_store, unsigned _parameter) noexcept
 RaspCache::~RaspCache() noexcept = default;
 
 static constexpr unsigned
-ToHalfHours(BrokenTime t)
+ToQuarterHours(BrokenTime t)
 {
-  return t.hour * 2u + t.minute / 30;
+  return t.hour * 4u + t.minute / 15;
 }
```
MaxKellermann, line 45: This is a rather intrusive and fragile change. This shouldn't be hidden inside a commit that pretends to just add a new RASP provider.

src/Weather/Rasp/RaspStore.cpp
```diff
@@ -104,7 +104,22 @@ RaspStore::MapItem::MapItem(const TCHAR *_name)
...
+    case 2 :
+      minute = 30;
+      break; // then prints "2"
+    case 3 :
+      minute = 45;
+      break; // then prints "2"
+    default:
+      minute = 0;
+      break; // then prints "2"
+  }
```
MaxKellermann, line 121: A very obscure and complicated way to write `(index % 4) * 15`

src/Weather/Rasp/RaspStore.hpp
```diff
@@ -48,8 +48,8 @@ struct GeoPoint;
  */
 class RaspStore {
 public:
-  static constexpr unsigned MAX_WEATHER_MAP = 16; /**< Max number of items stored */
-  static constexpr unsigned MAX_WEATHER_TIMES = 48; /**< Max time segments of each item */
+  static constexpr unsigned MAX_WEATHER_MAP = 50; /**< Max number of items stored */
+  static constexpr unsigned MAX_WEATHER_TIMES = 96; /**< Max time segments of each item */
```
MaxKellermann, line 52: The necessity for this change is not documented anywhere

src/Weather/Rasp/RaspStyle.cpp
```diff
@@ -116,8 +116,312 @@ static constexpr ColorRamp rasp_colors[6][NUM_COLOR_RAMP_LEVELS] = {
     { 300, { 255, 153, 0 }},
     {1000, { 255, 102, 0 }},
   },
+
+  // Thermalmap.info Color schemes
+
+  // Vertical speed (wave, Convergence)
+  // Offset 5000 cm/s
+  // Multiplyfactor 1000
+  // Blue to white to Purple
```
MaxKellermann, line 125: I have no idea what this multi-line comment expressed in single-line comment syntax means

src/Weather/Rasp/RaspStyle.cpp
```diff
@@ -25,7 +25,7 @@ Copyright_License {
 #include "ui/canvas/Ramp.hpp"
 #include "Terrain/RasterRenderer.hpp"
 
-static constexpr ColorRamp rasp_colors[6][NUM_COLOR_RAMP_LEVELS] = {
+static constexpr ColorRamp rasp_colors[21][NUM_COLOR_RAMP_LEVELS] = {
```
MaxKellermann, line 28: No more "magic numbers" please

src/Weather/Rasp/RaspStyle.cpp
```diff
@@ -116,8 +116,312 @@ static constexpr ColorRamp rasp_colors[6][NUM_COLOR_RAMP_LEVELS] = {
...
+      {667, {146, 146, 146}},
+      {750, {132, 132, 132}},
+      {833, {118, 118, 118}},
+      {917, {105, 105, 105}},
+      {1000, {91, 91, 91}},
+    },
 };
 
+
+
```
MaxKellermann, line 424: Why add two blank lines here?

src/Weather/Rasp/RaspStyle.cpp
```diff
@@ -116,8 +116,312 @@ static constexpr ColorRamp rasp_colors[6][NUM_COLOR_RAMP_LEVELS] = {
     { 300, { 255, 153, 0 }},
     {1000, { 255, 102, 0 }},
   },
+
+  // Thermalmap.info Color schemes
```
MaxKellermann, line 120: Instead of adding more array entries here, please create new variables with expressive names.

src/Weather/Rasp/RaspStyle.cpp
```diff
@@ -135,5 +439,49 @@ const RaspStyle rasp_styles[] = {
...
+  { _T("Wave_5000m"), rasp_colors[6], 5, false },
+  { _T("Wave_6000m"), rasp_colors[6], 5, false },
+  { _T("Wave_7000m"), rasp_colors[6], 5, false },
+  { _T("Temperature2m"), rasp_colors[9], 4, false },
+  { _T("Rain"), rasp_colors[10], 3, false },
+  { _T("PFD_Day_DuoDiscus"), rasp_colors[11], 3, false },
+  { _T("PFD_Day_LS4"), rasp_colors[11], 3, false },
+  { _T("PFD_Day_K8"), rasp_colors[11], 3, false },
+  { _T("Windspeed10m"), rasp_colors[12], 3, false },
+  { _T("Windspeed_500m"), rasp_colors[12], 3, false },
```
MaxKellermann, line 460: Why do all have an underscore, only "10m" not?

MaxKellermann (changes requested):

src/Dialogs/Weather/RASPDialog.cpp
```diff
@@ -47,8 +46,24 @@ Copyright_License {
...
+#include "net/http/DownloadManager.hpp"
+#include "ui/event/Notify.hpp"
+#include "thread/Mutex.hxx"
+#include "ui/event/PeriodicTimer.hpp"
+
+#include <vector>
+#endif
 #include <stdio.h>
 
+std::vector<AvailableFile> items;
```
MaxKellermann, line 65: Global variable!

src/Dialogs/Weather/RASPDialog.cpp
```diff
@@ -62,6 +77,27 @@ class RASPSettingsPanel final
 
   BrokenTime time;
 
+  #ifdef HAVE_DOWNLOAD_MANAGER
+    FileRepository repository;
+
+      /**
```
MaxKellermann, line 83: Bad indent

src/Dialogs/Weather/RASPDialog.cpp
```diff
@@ -62,6 +77,27 @@ class RASPSettingsPanel final
...
+     * This mutex protects the attributes "downloads" and
+     * "repository_modified".
```
MaxKellermann, line 85: ... what??

src/Dialogs/Weather/RASPDialog.cpp
```diff
@@ -154,7 +191,12 @@ class RaspProviderRenderer : public ListItemRenderer {
 
   void OnPaintItem(Canvas &canvas, const PixelRect rc,
                    unsigned i) noexcept override {
-    row_renderer.DrawTextRow(canvas, rc, rasp_providers[i].name);
+    std::string str = items.at(i).name;
```
MaxKellermann, line 194: Why does this copy the name to a new heap allocation?

src/Dialogs/Weather/RASPDialog.cpp
```diff
@@ -154,7 +191,12 @@ class RaspProviderRenderer : public ListItemRenderer {
...
+    TCHAR *name=new TCHAR[str.size()+1];
```
MaxKellermann, line 195: And now copying into a second heap allocation? Who is responsible for freeing this allocation? Does anybody? Is this a memory leak?

src/Dialogs/Weather/RASPDialog.cpp
```diff
@@ -154,7 +191,12 @@ class RaspProviderRenderer : public ListItemRenderer {
...
+    name[str.size()] = 0;
+    std::copy(str.begin(), str.end(), name);
```
MaxKellermann, line 197: Why copy this?

src/Dialogs/Weather/RASPDialog.cpp
```diff
@@ -170,19 +212,17 @@ DownloadRASP(const char *url, Path path, ProgressListener &progress)
 void
 RASPSettingsPanel::Download() noexcept
 {
-  unsigned n = 0;
-  for (auto i = rasp_providers; i->url != nullptr; ++i)
-    ++n;
+  unsigned n = items.size();
 
-  assert(n > 0);
+//  assert(n > 0);
```
MaxKellermann, line 217: What??

src/Dialogs/Weather/RASPDialog.cpp
```diff
@@ -170,19 +212,17 @@ DownloadRASP(const char *url, Path path, ProgressListener &progress)
...
 
   RaspProviderRenderer renderer;
   int i = ListPicker(_("Download"), n, 0, renderer.CalculateLayout(GetLook()),
                      renderer);
   if (i < 0)
     return;
 
-  const char *url = rasp_providers[i].url;
+  const char *url = items.at(i).GetURI();
```
MaxKellermann, line 225: Why `at()` and not `operator[]`?

MaxKellermann (changes requested):

build/main.mk
```diff
@@ -389,7 +389,6 @@ XCSOAR_SOURCES := \
 	$(SRC)/Weather/Rasp/RaspCache.cpp \
 	$(SRC)/Weather/Rasp/RaspRenderer.cpp \
 	$(SRC)/Weather/Rasp/RaspStyle.cpp \
-	$(SRC)/Weather/Rasp/Providers.cpp \
```
MaxKellermann, line 392: This should be done in the commit that removes the last user

Developer66:
Thanks for the review. I will have a look at all points and try to fix them.

---------------------

@MaxKellermann Do you have any idea why this is happening:

The display on Android is very different from that of Windows, for example. It seems that the map images under Android are downsampled from the resolution. This is certainly appropriate to a certain extent in view of the lower hardware performance, but it distorts the results strongly.
Unfortunately, we know too little about the render engine to make the necessary adjustments here.
The map images are generated by geojapser (https://github.com/fb/jasper-xcsoar) with tilesize 256. Changing the tilesize has little or no effect.

Windows:
163382651-1f9399eb-cfef-4818-b124-f64cc09394fd

Android:
163382687-5f3a2cac-e352-414d-b94c-91e0c9efd0bd

MaxKellermann:
That's off-topic. Write a new issue, but add proper screenshots. The two you just posted show different map areas, different zoom levels, different map settings.

lordfolken (changes requested):

src/Dialogs/Weather/RASPDialog.cpp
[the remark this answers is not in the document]
Developer66: The URL now definded in the repository file... when this can not be loaded properly or does not exists, `n` could be 0... the code does only render this items in a loop. So there will be no need to check the available items

Developer66:
I will give it at try in the next few days

src/Weather/Rasp/RaspStyle.cpp
lordfolken: Please see our style guide for comment formatting: 
https://xcsoar.readthedocs.io/en/latest/policy.html?highlight=comment#code-style

lordfolken:
This now includes:
 - the repository migration of the rasp sources (already merged) 
 - the rasp style definition 
 - the 15 minute interval. (This is WIP)

src/Weather/Rasp/RaspStyle.cpp
lordfolken: @Developer66 in your file that you provide, can you please correct the layer name?

src/Weather/Rasp/RaspStyle.cpp
lordfolken: I reformated them, but i'm also unsure about what exactly the lines mean. @Developer66 care to comment?

src/Weather/Rasp/RaspStyle.cpp
```diff
@@ -17,7 +17,7 @@ Copyright_License {
 
   You should have received a copy of the GNU General Public License
   along with this program; if not, write to the Free Software
-  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1306, USA.
```
MaxKellermann, line 20: .... ???

lordfolken: I'm just as puzzled. That shouldn't be there.

src/Weather/Rasp/RaspStore.hpp
lordfolken: i moved them now apart. The WEATHER_MAP is needed to support the large amount of layers in thermalmap's files. 
The WEATHER_TIMES is the theoretical value of entries that can now be reached with the 15 minute intervals.

lordfolken:
So i think with this the rework is pretty complete.
I would like @Developer66 point of view,  that everything works as he intended in the original pr.

Developer66:
> So i think with this the rework is pretty complete. I would like @Developer66 point of view, that everything works as he intended in the original pr.

Thank you very much for the work.
I just compiled it myself and it looks pretty good. 

The only thing is, if I set the bounding box of the JPEG2000 too big, a lot of details get lost.
So I will split the bounding box into the individual countries, similar to the terrain data. But this change will only be in the xcsoar-data-content repository. I am currently working on this and will then start a corresponding pull.

lordfolken:
Thanks for reviewing and pushing the boundaries. ;-) 

Did you see the comment on the Windspeed_10m layer? You would need to change the layer name inside the produced files. 
I second also the smaller sections for manageability reasons.

Developer66:
No problem.

Jep. Is already fixed in my code.
I think so too. Most of the devices should now be able to display bigger files but at least not all.
Tiling or transparency would be nice but something for the future you said.

PR #913, opened 2022-04-25
Rework Menu to fit action->newstate paradigm

### LINKED ISSUE #912: Make on-off buttons meaning more concistent
XCSoar version
--------------

Version 7.20-Android (commit e0ff401c6).

For now, some buttons says "On" or "Off" meaning the state of a feature is respectively "Enable**d**" or "Disable**d**", like the SnailTrail mode button ([code here](https://github.com/XCSoar/XCSoar/blob/ca4be3c26eda602ab0323b75bd98164f20195680/src/Menu/ExpandMacros.cpp#L315)).
Some other says "On" meaning "Enable": a press on the button will toggle the state of the feature to "On", and vice versa, like the Airspace toggle button ([code here](https://github.com/XCSoar/XCSoar/blob/ca4be3c26eda602ab0323b75bd98164f20195680/src/Menu/ExpandMacros.cpp#L331)).

This behavior is already inconsistent, but is worse in other languages. For instance, in French, "Off" is translated ["Desactivé"](https://github.com/XCSoar/XCSoar/blob/ca4be3c26eda602ab0323b75bd98164f20195680/po/fr.po#L892) (Disable**d**) instead of "Désactiver" (Disable), which is the opposite of what the airspace toggle button mean.

I see two solutions:

- Create two separate "On" and "Off" translation, one for the state, one of the action, like "Off_state" and "Off_action" which would be translated correctly in other languages.
- Make the behavior consistent by only having buttons giving the actual text of the feature.

I prefer the second option, but this will inverse the actual meaning of some buttons and might disturb peoples.

lordfolken:
Rework menus to make clear that an action is following a button press, that pressing the button will get you to the state on the label of the button. 

Note: the windows font change would not be required, but i still think we should change the font on windows to whats currently default on the platform. 

2022-04-27-223053_681x741_scrot
2022-04-27-222915_681x718_scrot
2022-04-27-222921_681x718_scrot

### DIFF
--- a/Data/Input/default.xci
+++ b/Data/Input/default.xci
@@ -479,7 +479,7 @@ mode=Display1
 type=key
 data=9
 event=ScreenModes cycle
-label=$(NextPageName)
+label=Page Show\n$(NextPageName)
 location=8
 
 mode=Display1
--- a/Data/Input/default.xci
+++ b/Data/Input/default.xci
@@ -430,7 +430,7 @@ event=Mode default
 event=StatusMessage Pilot event announced
 event=Logger note PEV
 event=PilotEvent
-label=Pilot\nEvent
+label=Pilot Event\nAnnounce
 location=7
 
 mode=Nav2
--- a/Data/Input/default.xci
+++ b/Data/Input/default.xci
@@ -361,7 +361,7 @@ type=key
 data=6
 event=Calculator
 event=Mode default
-label=Task
+label=Task Manager
 location=5
 
 mode=Nav1
@@ -420,7 +420,7 @@ event=Mode default
 event=StatusMessage Dropped marker
 event=Logger note Mark
 event=MarkLocation
-label=Mark\nDrop
+label=Marker Drop
 location=6
 
 mode=Nav2
@@ -610,7 +610,7 @@ type=key
 data=8
 event=WaypointEditor
 event=Mode default
-label=Waypoints Editor
+label=Waypoint Editor
 location=7
 
 mode=Config2
@@ -796,7 +796,7 @@ type=key
 data=9
 event=Checklist
 event=Mode default
-label=Check\nlist
+label=Checklist
 location=8
 
 mode=Info1
@@ -822,7 +822,7 @@ type=key
 data=6
 event=Status all
 event=Mode default
-label=Status
+label=Flight Status
 location=5
 
 mode=Info2
@@ -1062,7 +1062,7 @@ event=Mode default
 event=StatusMessage Dropped marker
 event=Logger note Mark
 event=MarkLocation
-label=Mark\nDrop
+label=Marker Drop
 location=10
 
 mode=RemoteStick
@@ -1202,7 +1202,7 @@ type=key
 data=0
 event=Checklist
 event=Mode default
-label=Check\nlist
+label=Checklist
 location=28
 
 mode=RemoteStick
@@ -1218,7 +1218,7 @@ type=key
 data=0
 event=Status all
 event=Mode default
-label=Status
+label=Flight Status
 location=30
 
 mode=RemoteStick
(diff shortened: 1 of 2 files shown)


Hugal31:
It looks great! Thant you for being so quick.

Hugal31:
I would have kept the current behavior for the tristate buttons tho.

lordfolken:
> I would have kept the current behavior for the tristate buttons tho.
The tristate buttons don't change. Its just the arrow is added, to make it clear that pushing it will set X to Y.

ulipo:
I like it, go for it!
The most common issue my friends report about XCSoar is the somehow misleading buttons behavior, this would be a big step forward!

Hugal31:
> The tristate button's behavior doesn't change. Its just the arrow is added, to make it clear that pushing it will set X to Y.

You're right. So maybe, to keep things coherent, display the current state for bot the toggle AND the tristate buttons.

lordfolken:
I think the way this was designed originally is akin to professional MFDs:
image

So an action is a departure from a current state to a new state. The Button's label shows the state that you will be heading to. 

But of course users of these where trained on them and not meant to intuitively understand what the UI needs to do.

Current paradigms in UI's always show you the current state and pressing a button, changes that state. So this is what is confusing users with the current implementation.

MaxKellermann:
First commit message doesn't explain why the fonts are changed

src/Menu/ExpandMacros.cpp
```diff
@@ -359,11 +359,11 @@ LookupMacro(const TCHAR *name, bool &invalid)
     return GetMapSettings().auto_zoom_enabled ? _("Manual") : _("Auto");
   } else if (StringIsEqual(name, _T("TopologyToggleActionName")) ||
              StringIsEqual(name, _T("TopographyToggleActionName"))) {
-    return GetMapSettings().topography_enabled ? _("Off") : _("On");
+    return GetMapSettings().topography_enabled ? _("☑") : _("☐");
```
MaxKellermann, line 362: Those don't need to be translated, do they?

lordfolken:
>  XCSoar tries to show the current state inside the button, but that's bad UI design.

So currently it says:
Airspace off
Topology off
Terrain off

with "off" actually being a state and not an action which confuses users rightly so. 

So it should be 
Hide Airspace 
Hide Topology 
Hide Terrain
And change this to Show when the switch is toggled.

The current commits reflect this now.

Data/Input/default.xci
```diff
@@ -594,23 +594,23 @@ mode=Config2
...
+label=Profile\nManager
 location=5
 
 mode=Config2
 type=key
 data=8
 event=WaypointEditor
 event=Mode default
-label=Waypoints Editor
+label=Waypoint\nEditor
```
MaxKellermann, line 613: I don't like having explicit layout decisions in translatable strings. We should separate content and layout as much as possible. If the button doesn't line-break automatically, we should improve that algorithm instead of adding layout specs here.

MaxKellermann:
The last commit is a mix of various changes to button labels; I'd prefer to have separate commits, because some changes are obviously good and others are questionable and have nothing to do with this PR. We should merge obviously-good quickly, and discuss the rest.

lordfolken:
> The last commit is a mix of various changes to button labels; I'd prefer to have separate commits, because some changes are obviously good and others are questionable and have nothing to do with this PR. We should merge obviously-good quickly, and discuss the rest.

I redid it and split it into different commits. 
The one i'm unsure about is https://github.com/XCSoar/XCSoar/pull/913/commits/c0e3297148865ffec9f97d86a9689cd9880f3f76

lordfolken:
So i had a look at  how the commericals do it: 


They also represent state in the labels.

lordfolken:
I'm closing this PR, as the initial motivation was now merged in 
https://github.com/XCSoar/XCSoar/commit/17ccc4bb38b13c1d8762629ddad0d4590b6d55a4

Some of the label corrections will be done in a new pr.

PR #915, opened 2022-04-27
Flarm bugfix reading flightlist

August2111:
Brief summary of the changes
----------------------------

With catching of exception(s) and extension of a timeout period it is possible to get all downloadable flights of a flarm device in a list - and not the wrong message 'flightlist is empty'

This changes should be solve the described problem (There was mostly a difference in the view of recorded flights between reading with FlarmTool and XCSoar)

This issue is not only linked to XCSoar itself: I observed this in the last year very often only with the PowerMouse! But with the new Flarm update 7.06 I observed this on all other devices (PowerFlarm, normal Flarm, ...) I think, Flarm changed something on the timing... ;-(

### DIFF
--- a/src/Device/Driver/FLARM/BinaryProtocol.cpp
+++ b/src/Device/Driver/FLARM/BinaryProtocol.cpp
@@ -2,7 +2,7 @@
 Copyright_License {
 
   XCSoar Glide Computer - http://www.xcsoar.org/
-  Copyright (C) 2000-2021 The XCSoar Project
+  Copyright (C) 2000-2022 The XCSoar Project
   A detailed list of copyright holders can be found in the file "AUTHORS".
 
   This program is free software; you can redistribute it and/or
--- a/src/Device/Driver/FLARM/BinaryProtocol.cpp
+++ b/src/Device/Driver/FLARM/BinaryProtocol.cpp
@@ -23,6 +23,7 @@ Copyright_License {
 
 #include "Device.hpp"
 #include "CRC16.hpp"
+#include "Device/Error.hpp"
 #include "Device/Port/Port.hpp"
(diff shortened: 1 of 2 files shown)


MaxKellermann:
First line of the commit message is waaaaay too long. Should be 80 chars or less. The "flarm-bugfix" prefix is useless noise. "Device/Driver/FLARM.cpp:" can be abbreviated to "Driver/FLARM:".

MaxKellermann:
> On my desk the BinaryPing throws an exception

Which exception? It's not a bare `std::exception`, is it?

MaxKellermann:
Then catch just `DeviceTimeout`. Why catch other exceptions here?

August2111:
> Why catch other exceptions here?

Because the flight list is empty in such a case, although flights were previously detected in the device - and nobody knows why...

MaxKellermann:
> Because the flight list is empty in such a case, although flights were previously detected in the device - and nobody knows why...

... contradicts with:

> All observed exceptions were DeviceTimeout exceptions

August2111:
I'm still working on this issue, probably next week I will present a better solution ;)

August2111:
Thanks for cherry-picking the first three commits 2 weeks ago, so the flight download with a PowerFlarm is possible again. With ClassicFlarm, switching to BinaryMode is still faulty because the loop is interrupted by the DeviceTimeout exception during BinaryPing ;-(
With the appropriate exception handler for BinaryPing (with a return value of 'false'), the switch then works perfectly!

August2111:
This is a rebase with master branch with c1a7213f

lordfolken:
So I tested this with a FLARM F5 2.0 running 7.06. With this i was able to download several flights, and the device returns after to normal operation.

PR #920, opened 2022-05-02
Engine/Contest/Solvers: Charron contest

timower:
Brief summary of the changes
----------------------------

Adds the Charron contest based on the rules at:
https://www.lvzc.be/index.php/secretariaat-2/downloads/charron-line/2013-reglement-charronbeker-2022/file

[Charron](https://www.lvzc.be/charronline/2022/overzicht.php) is an online contest from LVZC. The rules are:
 * Maximum 5 legs under 200km, 6 legs above.
 * Minimum 20km legs.
 * 4 points per km with handicap.

None

### DIFF
--- a/src/Engine/Contest/Solvers/Charron.cpp
+++ b/src/Engine/Contest/Solvers/Charron.cpp
@@ -1,0 +1,72 @@
+/*
+Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2021 The XCSoar Project
+  A detailed list of copyright holders can be found in the file "AUTHORS".
+
+  This program is free software; you can redistribute it and/or
+  modify it under the terms of the GNU General Public License
+  as published by the Free Software Foundation; either version 2
+  of the License, or (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
+
+#include "Charron.hpp"
+
+
+Charron::Charron(const Trace &_trace, bool _plus_200km) noexcept
+    : ContestDijkstra(_trace, true, _plus_200km ? 6 : 5, 1000, 20'000),
+      plus_200km(_plus_200km)
+{
+}
+
+SolverResult
+Charron::Solve(bool exhaustive) noexcept
+{
+  const auto result = ContestDijkstra::Solve(exhaustive);
+  if (result != SolverResult::VALID) {
+    return result;
+  }
+
+  const auto &solution = GetBestSolution();
+
+  auto distance = 0.0;
+  GeoPoint previous = solution[0].GetLocation();
+  for (unsigned i = 1; i < num_stages; ++i) {
+    const GeoPoint &current = solution[i].GetLocation();
+    distance += current.Distance(previous);
+    previous = current;
+  }
+
+  // Reject the solution if the distance was less than 200km.
+  // Also reset the best solution set by ContestDijkstra::Solve.
+  if (plus_200km && distance < 200'000) {
+    AbstractContest::Reset();
+    return SolverResult::FAILED;
+  }
+
+  return result;
+}
+
+ContestResult
+Charron::CalculateResult() const noexcept
+{
+  ContestResult result = ContestDijkstra::CalculateResult();
+
+  // 4 points per km.
+  const auto score_factor = 4 / 1000.0;
+  result.score = ApplyHandicap(result.distance * score_factor);
+
+  return result;
+
+}
--- a/src/Engine/Contest/Solvers/Charron.hpp
+++ b/src/Engine/Contest/Solvers/Charron.hpp
@@ -1,0 +1,60 @@
+/* Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2021 The XCSoar Project
+  A detailed list of copyright holders can be found in the file "AUTHORS".
+
+  This program is free software; you can redistribute it and/or
+  modify it under the terms of the GNU General Public License
+  as published by the Free Software Foundation; either version 2
+  of the License, or (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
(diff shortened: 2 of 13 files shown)


src/Engine/Contest/Solvers/Charron.cpp
```diff
@@ -0,0 +1,72 @@
...
+    return result;
+  }
+
+  const auto &solution = GetBestSolution();
+
+  float distance = 0.0f;
+  GeoPoint previous = solution[0].GetLocation();
+  for (unsigned i = 1; i < num_stages; ++i) {
+    const GeoPoint &current = solution[i].GetLocation();
+    distance += current.Distance(previous);
```
MaxKellermann, line 47: This implicitly casts the `double` return value to `float`. I don't think there's a performance advantage for using `float` here; please use `double` like the rest of XCSoar.

src/Engine/Contest/Solvers/Charron.cpp
```diff
@@ -0,0 +1,72 @@
...
+  return result;
+}
+
+ContestResult
+Charron::CalculateResult() const noexcept
+{
+  ContestResult result = ContestDijkstra::CalculateResult();
+
+  // 4 points per km.
+  const auto score_factor = 4 / 1000.0f;
```
MaxKellermann, line 67: Same here - no `float` please. It will be casted to `double` anyway, so all you do here is lose precision.

src/Engine/Contest/Solvers/Charron.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * Allows up to 5 legs for distances less than 200km,
+ * otherwise 6 legs are allowed. All legs must be longer than 20km.
+ *
+ * See:
+ * https://www.lvzc.be/charronline/2022/overzicht.php
+ * https://www.lvzc.be/index.php/secretariaat-2/downloads/charron-line/2013-reglement-charronbeker-2022/file
+ */
+class Charron : public ContestDijkstra {
+
+  bool plus_200km = false;
```
MaxKellermann, line 39: This default value is useless, because it will be assigned by the constructor; I'd rather remove the initialiser here and make the field `const`, so the compiler will bail out if a new constructor forgets to initialise it.

src/Engine/Contest/Solvers/ContestDijkstra.cpp
```diff
@@ -35,14 +35,17 @@ static constexpr unsigned CONTEST_QUEUE_SIZE = 5000;
...
   :AbstractContest(finish_alt_diff),
    NavDijkstra(n_legs + 1),
    TraceManager(_trace),
    continuous(_continuous),
-   incremental(false)
+   incremental(false),
+   min_distance(_min_distance)
 {
   assert(num_stages <= MAX_STAGES);
+  assert(min_distance >= 0.0f);
```
MaxKellermann, line 48: This field is an integer - don't compare it with a float!

src/Engine/Contest/Solvers/ContestDijkstra.cpp
```diff
@@ -237,26 +240,34 @@ ContestDijkstra::AddEdges(const ScanTaskPoint origin,
...
-      /* After excessive thinning, the exact TracePoint that matches
-         the required altitude difference may be gone, and the
-         calculated result becomes overly pessimistic.  This code path
-         makes it optimistic, by checking if the previous point
-         matches. */
-
-      /* TODO: interpolate the distance */
-      const unsigned d = weight * CalcEdgeDistance(origin, destination);
-      Link(destination, origin, d);
+    const unsigned distance = GetPoint(destination).DistanceTo(originGeo);
```
MaxKellermann, line 251: This integer cast loses precision for no reason. And it is calculated even if no `min_distance` is configured, adding overhead to all implementations.
Do we really need this field in the base class?

src/Engine/Contest/Solvers/ContestDijkstra.cpp
```diff
@@ -237,26 +240,34 @@ ContestDijkstra::AddEdges(const ScanTaskPoint origin,
...
+
+    // Check if the distance is withing the minimum distance.
+    // Also allows zero distance legs, because if a minimum distance is set not
+    // all solutions will use all legs.
```
MaxKellermann, line 255: This is rather obscure - can't we find a better, more reliable and easier-to-understand way to check if a leg is used at all?

src/Engine/Contest/Solvers/ContestDijkstra.hpp
```diff
@@ -87,7 +92,8 @@ class ContestDijkstra : public AbstractContest, protected NavDijkstra, public Tr
   ContestDijkstra(const Trace &_trace,
                   bool continuous,
                   const unsigned n_legs,
-                  const unsigned finish_alt_diff = 1000) noexcept;
+                  const unsigned finish_alt_diff = 1000,
+                  const unsigned _min_distance = 0.0f) noexcept;
```
MaxKellermann, line 96: An integer initialised with a float??

MaxKellermann (changes requested):

src/Engine/Contest/Solvers/ContestDijkstra.cpp
timower: I moved this to the base class, but has to make `AddEdges` virtual as a result.
Is that acceptable to you?

src/Engine/Contest/Solvers/ContestDijkstra.cpp
timower: It's the least intrusive way I found, otherwise the `Dijkstra.hpp` needed significant changes to handle varying amount of stages. IMO this is fine, 0 length legs don't show up in the interface and work quite nicely.

timower:
Thanks for the quick review and sorry about the integer types, changed float to unsigned last minute.

src/Engine/Contest/Solvers/ContestDijkstra.cpp
MaxKellermann: Your commit which makes it virtual adds a lot of duplicate code. That is all but acceptable.
I'm always cautious to make methods virtual, because the overhead this involves can sometimes be prohibitive (and sometimes doesn't matter), but in this case, the "virtual" is not the problem.
I suggest moving the "min_distance" feature to a separate commit (but leaving it on `class ContestDijkstra`), convert it to `double` to avoid implicit casts, and do the `DistanceTo()` call only if `min_distance>0`.

src/Engine/Contest/Solvers/ContestDijkstra.cpp
MaxKellermann: OK, agree, leave it like that for now. But (tiny nitpick) please use proper multi-line comment syntax for multi-line comments. `//` is a C++ extension explicitly designed for single-line comments.

src/Engine/Contest/Solvers/ContestDijkstra.cpp
timower: Ok, will do. By separate commit you mean separate pull request?

MaxKellermann: No, separate commit, same PR.
Check out [stgit](https://stacked-git.github.io/guides/tutorial/), it's a terrific tool which helps you manage PR branches in a proper and clean way, lets you move patches around, and helps avoid unclean fixup commits like the one you added to this PR.

Issue #892, opened 2022-03-21
docker android build


volkmarnissen:
XCSoar versions having and not having the problem
-------------------------------------------------
XCSoar main branch

Expected behavior
-----------------
I wanted to do a docker build for android and followed the documentation in https://github.com/XCSoar/XCSoar/tree/master/ide/docker
I was expecting to get a build without any configuration.

First approach:
```
docker run \
    --mount type=bind,source="$(pwd)",target=/opt/xcsoar \
    -it ghcr.io/xcsoar/xcsoar/xcsoar-build:latest xcsoar-compile ANDROID
```

Actual behavior
---------------
The build failed, because the android sdk and ndk were not present at all.

Second approach:
```
docker run \
    --mount type=bind,source="$(pwd)",target=/opt/xcsoar \
    -it ghcr.io/xcsoar/xcsoar/xcsoar-build:latest /bin/bash
```
I found out, that 
```
 /root/install-debian-packages.sh and
 /root/install-android-tools.sh
```
were not executed. Even if this is stated in https://github.com/XCSoar/XCSoar/blob/master/ide/docker/Dockerfile
Additionally, I had to call
```
git submodule init
git submodule update
```

Adding the following lines to the build/local-config.mk fixed the rest:
```
ANDROID_SDK := /root/opt/android-sdk-linux
ANDROID_NDK := /root/opt/android-ndk-r23b
ANDROID_BUILD_TOOLS_DIR = $(ANDROID_SDK)/build-tools/29.0.3
ANDROID_SDK_PLATFORM =android-30
```
Now, the packages were built nicely.
So, the docker build feature is great. It is also fast.:+1:
I have no experience with make or docker, so I don't know how to fix this issue on my own.

This issue can be used as a workaround for the problem.

Steps to reproduce the behavior
-------------------------------
Use ubuntu impish as docker host and proceeded as described in 
https://github.com/XCSoar/XCSoar/tree/master/ide/docker
to call an ANDROID build

Do you have any idea what may have caused this?
-----------------------------------------------
Yes. See above.

Do you have an idea how to solve the issue?
-------------------------------------------
No. :slightly_frowning_face:
But if someone  points me to the right direction, I can try to fix it.

lordfolken:
Please post error messages you are getting. 

This container is part of our build pipeline and this built an android package 2 days ago. See:
https://github.com/XCSoar/XCSoar/actions/runs/2008414338

[...]

PR #923, opened 2022-05-03
Added support for number pad keyboard, Enhanced lua with dialogs, introduced in dialog keyboard

### LINKED ISSUE #885: Developer Manual is not up-to-date Lua 5.4 and Ubuntu 21 are required
master branch:
The developer Manual doesn't tell anything about  which Ubuntu version to use
it refers to Lua 5.2 as a depenency. This won't work.
Another dependency is G++10.

I used developer Manual 7.0 to build XCSoar from github

Expected behavior
-----------------

Actual behavior
---------------

Try an older Ubuntu. You will fail because G++10 is required.
the other messages will follow during build process.

Isn't it better to list the known depenencies in the github start page(https://github.com/XCSoar/XCSoar)?
This is easier to maintain.

### LINKED ISSUE #902: Make Waypoint list dialog more end user friendly.
XCSoar version
--------------
XCSoar main branch

The waypoint list dialog (and others) require modal dialogs to enter text.
It is better to have a in dialog keyboard, which shows the allowed characters...
Advantage: 
- You can enter a character and you will see the updated list immediately.
   So don't need to guess whether you keyed in enough characters.
- You can select a distance and then add the name filter with only a few key board clicks
  The allowed characters will consider the distance filter.

Additionally, the waypoint list dialog (and others)  could support  [(number pads)](https://www.amazon.de/Ziffernblock-Bluetooth-Alcey-Nummernblock-Tastaturerweiterung/dp/B07TXNTPML)
They are cheap and can be used to call a lot of XCSoar functions without using the main menu.

I'd like to enable unexperienced users to be able to select waypoints quicker than now.

The number pad can be located on your leg. You can use it for all interactions with XCSoar.

This feature is implemented at [github](https://github.com/volkmarnissen/XCSoar/tree/NumKeyPad).
- It enhances the key codes to support all keys on the numpad.
- It uses lua to map the keys to the different dialogs and functions(zoom/pan)
- It is more convinient to enter texts in the in dialog keyboard.

It's tested on X11, ANDROID and PC.
It's compilable for KOBO. (I didn't test compiling for PI, because there is no docker image for it)
It's not tested on PI and KOBO.(I am not able to compile for PI and I have no KOBO hardware)

I'd like to integrate it into XCSoar main branch as soon as possible. That's why  the in dialog keyboard was added  to the waypoint list dialog only. Once it's integrated to the main branch, I'll integrate in dialog keyboard into more dialogs. 
It's designed to do this with a few code changes.

If you want to test it in a separate branch, please create it. Then I will push it into it.
Alternatively if someone volunteers to test it, it can also be done in my github branch.

volkmarnissen:
Brief summary of the changes
----------------------------
Added support for new keycodes for a numberpad
Enhance Lua functions with more dialogs
Add In Dialog to Waypoint list
Enable debugging for Android Build
Updated Developer documentation and NEWS.txt

Closes #885
Closes #902
In progress #892


src/ui/event/shared/Event.hpp
```diff
@@ -40,7 +40,7 @@ struct Event {
     TIMER,
 #endif
 
-    CALLBACK,
+    _CALLBACK, //CALLBACK is used by windows compiler
```
MaxKellermann, line 43: What the ... is this? Why is it hidden in this commit?

MaxKellermann:
This PR looks like a confusing mess of unrelated changes in each commit, some of which are undocumented. I can't review this.

volkmarnissen:
The commits deal with the topics stated in the Commit message.
I was not able to build android locally without the changes in android.mk, so I had to add the changes to test android build. The docker build was also not working for me even after restoring all mk files from master.
I know, that I have gaps in knowledge of XCSoar development. However, this is most likely true for all new contributors. I am sure, you are aware, that a lot of knowledge is not available to the public. That makes it hard to develop for a beginner.

I know also, that the PR request contains a lot of changes, however, it was not possible to test individually one by one on all platforms.
The lua changes are required to test the keyboard properly. The in dialog keyboard is in one commit, but it contains a lot of changes. This is due to the fact, that is is complex (but easy to use).

I am sure, there are a lot of pilots who would love to have a number pad keyboard and the in dialog keyboard. It is much more convienient than the touch screen approach.  Meanwhile, I tested it also during flights. It really helps.

So, if you don't want to review it, I will accept it. 
However,  in this case I will stop the effort contributing it to XCSoar  and use the code on my own device only.
Additionally, I will stop contributing here at all.

May be we can find another way to get it integrated. If you have any ideas, let me know. 
Otherwise just close the PR.

volkmarnissen:
It doesn't make sense to split the commit any further. May be one could move one or the other small change to a new commit, but that won't help much. The topic is complex. So, the commits are also complex.
I understand, that you can't review it as it is and I know, that splitting the commits doesn't really make sense.
So, I'll close the PR.

MaxKellermann:
> It doesn't make sense to split the commit any further. May be one could move one or the other small change to a new commit, but that won't help much. The topic is complex. So, the commits are also complex.

Every aspect of this is wrong.
I'm glad you'll stop contributing code to XCSoar, because the project doesn't need people who don't listen and don't want to learn, because that only wastes the time of people who do.

PR #927, opened 2022-05-15
Rework Menu Labels

### LINKED ISSUE #912: Make on-off buttons meaning more concistent

lordfolken:
this is a followup to #913 

Some labels are either
- incomplete to siginify function
- not correct english
- not conforming to <context> action

Each change is now a commit, as requested in #913

### DIFF
--- a/Data/Input/default.xci
+++ b/Data/Input/default.xci
@@ -303,7 +303,6 @@ type=gesture
 data=LDR
 event=Checklist
 
-
 mode=default
 type=gesture
 data=URDL
@@ -796,7 +795,7 @@ type=key
 data=9
 event=Checklist
 event=Mode default
-label=Check\nlist
+label=Checklist
 location=8
 
 mode=Info1
@@ -1203,7 +1202,7 @@ type=key
 data=0
 event=Checklist
 event=Mode default
-label=Check\nlist
+label=Checklist
 location=28
 
 mode=RemoteStick
--- a/Data/Input/default.xci
+++ b/Data/Input/default.xci
@@ -430,7 +430,7 @@ event=Mode default
 event=StatusMessage Pilot event announced
 event=Logger note PEV
 event=PilotEvent
-label=Pilot\nEvent
+label=Pilot Event\nAnnounce
 location=7
 
 mode=Nav2
--- a/Data/Input/default.xci
+++ b/Data/Input/default.xci
@@ -361,7 +361,7 @@ type=key
 data=6
 event=Calculator
 event=Mode default
-label=Task
+label=Task Manager
 location=5
 
 mode=Nav1
--- a/Data/Input/default.xci
+++ b/Data/Input/default.xci
@@ -419,7 +419,7 @@ event=Mode default
 event=StatusMessage Dropped marker
 event=Logger note Mark
 event=MarkLocation
-label=Mark\nDrop
+label=Marker\nDrop
 location=6
 
 mode=Nav2
@@ -1062,7 +1062,7 @@ event=Mode default
 event=StatusMessage Dropped marker
 event=Logger note Mark
 event=MarkLocation
-label=Mark\nDrop
+label=Marker\nDrop
 location=10
 
 mode=RemoteStick
--- a/Data/Input/default.xci
+++ b/Data/Input/default.xci
@@ -390,7 +390,7 @@ type=key
 data=0
 event=Setup Alternates
 event=Mode default
-label=Alternates$(CheckWaypointFile)
+label=Alternates\nList$(CheckWaypointFile)
 location=9
 
 # -------------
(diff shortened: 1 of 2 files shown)


src/Menu/ExpandMacros.cpp
```diff
@@ -153,12 +153,12 @@ ExpandTaskMacros(const TCHAR *name,
...
         if (has_optional_starts && !common_stats.active_has_previous) {
-          return _("Next Startpoint");
+          return _("Startpoint Next");
         } else {
           invalid |= !common_stats.active_has_previous;
           return previous_is_start
-            ? _("Start Turnpoint")
-            : _("Previous Turnpoint");
+            ? _("Turnpoint Start")
+            : _("Turnpoint Previous");
```
kedder, line 161: Hm, what's the motivation behind this? "Start turnpoint" sounds better than "Turnpoint Start" to me (I'm not a native English spearker though).

lordfolken: Thx for reviewing, and starting a discussion. 

The idea was to have a bigger to smaller kind of ordering. So first ```context``` and then an ```action```.
All other labels on the buttons are already in that format:
Zoom +, or Task Abort 
So the idea would be to look for Turnpoint first, and then the aciton. Its not necessarily meant to be "written" english.

kedder: I see, thanks for explanation!

kedder (approved):

Data/Input/default.xci
```diff
@@ -430,7 +430,7 @@ event=Mode default
 event=StatusMessage Pilot event announced
 event=Logger note PEV
 event=PilotEvent
-label=Pilot\nEvent
+label=Pilot Event\nAnnounce
```
MaxKellermann, line 432: I don't think the verb should be at the end; I mean, read it out loud and tell me it's good.

Data/Input/default.xci
```diff
@@ -390,7 +390,7 @@ type=key
 data=0
 event=Setup Alternates
 event=Mode default
-label=Alternates$(CheckWaypointFile)
+label=Alternates\nList$(CheckWaypointFile)
```
MaxKellermann, line 393: I don't think adding "List" improves anything here.
There's not even a verb - note I don't think that all buttons should have a verb - but it's not even consistent with the policy this PR attempts to meet (a policy which I don't like).

src/Menu/ExpandMacros.cpp
```diff
@@ -80,26 +80,26 @@ ExpandTaskMacros(const TCHAR *name,
     if (StringIsEqual(name, _T("WaypointNext")) ||
         StringIsEqual(name, _T("WaypointNextArm"))) {
       invalid = true;
-      return _("Next Turnpoint");
+      return _("Turnpoint Next");
```
MaxKellermann, line 83: This is awkward. "Turnpoint Next" - what does that mean? I don't like swapping those words, because the label doesn't make sense that way, it's confusing and thus wastes the user's attention.

MaxKellermann (commented):
I cherry-picked a few commits, but I'm not quite convinced that the remaining changes are really an improvement.

lordfolken:
I agree that these changes where quirky and didn't feel right. Hence also the PR. Thank you for merging the rest.

MikeBravo15:
Can you please export text to Weblate for translation? Thanx in advance

PR #929, opened 2022-05-19
Manual: Merged contributors into User Manual

Exadios:
Items from the AUTHORS file in the root directory of the source distribution
and items from the User Manual merged into the User Manual.

Brief summary of the changes
----------------------------

Update of contributors in the User Manual.

### DIFF
--- a/doc/manual/en/contributors_dev.tex
+++ b/doc/manual/en/contributors_dev.tex
@@ -27,4 +27,33 @@
 \item Paolo Ventafridda \url{coolwind@email.it}
 \item James Ward \url{jamesward22@gmail.com}
 \item John Wharington \url{jwharington@gmail.com}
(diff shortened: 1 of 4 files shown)


lordfolken (approved):
Thx. Maybe there is a way to parse AUTHORS.txt and include it here?

Exadios:
The reality is that there is only one reason to record the authors at all - their relation to the GPL. Otherwise, who cares? With that in mind I think that there's no reason to record the authors in the User Manual at all. The users are probably not interested.

Possibly the list should be replaced with a paragraph or two summarizing the history of XCSoar, or maybe nothing at all.

MaxKellermann:
Agree, this should have been removed instead of updated. This doesn't belong in the user manual.

PR #931, opened 2022-05-19
Windows: Fix Assert in WndForm

### LINKED ISSUE #901: WndForm runs into assertion in Windows when closing with ESC key (keyboard)
XCSoar main branch run on windows.

Expected behavior
-----------------
Open the checklist dialog
Press ESC **key** (The button on the screen works properly)
Expectation: The dialog should close (but nothing happens)
Press the key again
You get multiple assertions because hidden windows are going to be hidden

Actual behavior
---------------
The dialog doesn't close when pressing ESC key on the keyboard
Nothing happens
If you press ESC again, you get assertions.

See above. It is reproducible.

In the code "Form/Form.cpp", there is some #ifndef WIN_USER (line 396) part which might be part of the problem.
However, I am not so deep into it, that I can fix it.
My version of Form.cpp has a small modification, but I don't think, that it causes the issue, because it doesn't apply to the ESC key.
If you are not able to reproduce, just close the issue. I will look into it.

No.

lordfolken:
fixes #901

### DIFF
--- a/src/Form/Form.cpp
+++ b/src/Form/Form.cpp
@@ -392,12 +392,10 @@ WndForm::ShowModal()
         }
       }
 
-#ifndef USE_WINUSER
       if (event.GetKeyCode() == KEY_ESCAPE) {
         modal_result = mrCancel;
         continue;
       }
-#endif
 
 #ifdef KOBO
       if (event.GetKeyCode() == KEY_POWER) {


lordfolken:
@volkmarnissen can you please test this?

volkmarnissen:
Test is OK

PR #932, opened 2022-05-23
Critical Errors found by Codacy

lordfolken:
Two functions that throw runtime errors, that are encapsulated by other function calls, that have noexpect defined.

### DIFF
--- a/src/Airspace/AirspaceParser.cpp
+++ b/src/Airspace/AirspaceParser.cpp
@@ -217,7 +217,7 @@ struct TempAirspaceType
   }
 
   void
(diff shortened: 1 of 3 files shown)


MaxKellermann:
The `CurlGlobal` commit is bad. The codacy warning was correct, but this just moves the problem to the caller. If this ever throws, I think it's okay to crash XCSoar. It's an operation that fails only in the presence of code bugs or on a very critical condition which cannot be recovered anyway. Moving this one stack layer up doesn't solve anything.

lordfolken:
Thank you for the explanation. I will remove the commit.

PR #934, opened 2022-05-30
Windows port adjust

August2111:
Brief summary of the changes

  - use 6 devices in serial port list (like all other systems)
  - allow connection to COM port 10 to 255 (f.e. for BlueTooth user)
  - remove index based COM port setting (obsolet since XCSoar 5)

  - no

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -21,6 +21,8 @@ Version 7.24 - not yet released
 * Windows
   - new default fonts
   - use 6 devices in serial port list (like all other targets)
+  - allow connection to COM port 10 to 254 (f.e. for BlueTooth user)
+  - remove index based COM port setting (obsolete since XCSoar 5)
 * XCVario/Borgelt
   - revert cruise/circling mode change 
 * Kobo
--- a/src/Device/Port/ConfiguredPort.cpp
+++ b/src/Device/Port/ConfiguredPort.cpp
@@ -99,7 +99,19 @@ OpenPortInternal(EventLoop &event_loop, Cares::Channel &cares,
     if (config.path.empty())
       throw std::runtime_error("No port path configured");
 
+#ifdef _WIN32
+    StaticString<0x20> port_path;
+    // use the windows style of device names:
+    port_path.Format(_T("\\\\.\\%s"), config.path.c_str());
+    if (!port_path.empty() && port_path.back() == TCHAR(':')) {
+      /* On Windows strip the trailing colon if the port name setting
+      has the old style 'COMx:' instead of 'COMx' */
+      port_path.Truncate(port_path.length() - 1);
+    }
+    path = port_path.c_str();
+#else
     path = config.path.c_str();
+#endif
     break;
 
   case DeviceConfig::PortType::BLE_HM10:
--- a/src/Dialogs/Device/PortDataField.cpp
+++ b/src/Dialogs/Device/PortDataField.cpp
@@ -132,7 +132,7 @@ DosDeviceExists(const TCHAR *path) noexcept
 static void
 FillDefaultSerialPorts(DataFieldEnum &df) noexcept
 {
-  for (unsigned i = 1; i <= 10; ++i) {
+  for (unsigned i = 1; i <= 255; ++i) {
     TCHAR buffer[64];
     _stprintf(buffer, _T("COM%u"), i);
     if (DosDeviceExists(buffer))
--- a/src/Dialogs/Device/PortDataField.cpp
+++ b/src/Dialogs/Device/PortDataField.cpp
@@ -31,6 +31,10 @@
 #include "Device/Port/TTYEnumerator.hpp"
 #endif
 
+#ifdef _WIN32
+# include <fileapi.h>
+#endif
+
 #ifdef ANDROID
 #include "java/Global.hxx"
 #include "java/String.hxx"
@@ -117,13 +121,22 @@ DetectSerialPorts(DataFieldEnum &df) noexcept
 
 #if defined(_WIN32) && !defined(HAVE_POSIX)
 
+[[gnu::pure]]
+static bool
+DosDeviceExists(const TCHAR *path) noexcept
+{
+  TCHAR dummy[MAX_PATH];
+  return QueryDosDevice(path, dummy, std::size(dummy)) > 0;
+}
+
 static void
 FillDefaultSerialPorts(DataFieldEnum &df) noexcept
 {
   for (unsigned i = 1; i <= 10; ++i) {
     TCHAR buffer[64];
-    _stprintf(buffer, _T("COM%u:"), i);
-    AddPort(df, DeviceConfig::PortType::SERIAL, buffer);
+    _stprintf(buffer, _T("COM%u"), i);
+    if (DosDeviceExists(buffer))
+       AddPort(df, DeviceConfig::PortType::SERIAL, buffer);
   }
 }
 
--- a/src/Profile/DeviceConfig.cpp
+++ b/src/Profile/DeviceConfig.cpp
@@ -103,28 +103,6 @@ LoadPath(const ProfileMap &map, DeviceConfig &config, unsigned n)
   return map.Get(buffer, config.path);
 }
 
-static bool
-LoadPortIndex(const ProfileMap &map, DeviceConfig &config, unsigned n)
-{
-  char buffer[64];
-  MakeDeviceSettingName(buffer, "Port", n, "Index");
-
-  unsigned index;
-  if (!map.Get(buffer, index))
-    return false;
-
-  /* adjust the number, compatibility quirk for XCSoar 5 */
-  if (index < 10)
-    ++index;
-  else if (index == 10)
-    index = 0;
-
-  TCHAR path[64];
-  _stprintf(path, _T("COM%u:"), index);
-  config.path = path;
-  return true;
-}
-
 void
 Profile::GetDeviceConfig(const ProfileMap &map, unsigned n,
                          DeviceConfig &config)
@@ -151,30 +129,12 @@ Profile::GetDeviceConfig(const ProfileMap &map, unsigned n,
   if ((!have_port_type ||
        config.port_type == DeviceConfig::PortType::ANDROID_USB_SERIAL ||
        config.port_type == DeviceConfig::PortType::SERIAL) &&
-      !LoadPath(map, config, n) && LoadPortIndex(map, config, n))
+       !LoadPath(map, config, n))
     config.port_type = DeviceConfig::PortType::SERIAL;
 
   MakeDeviceSettingName(buffer, "Port", n, "BaudRate");
   if (!map.Get(buffer, config.baud_rate)) {
-    /* XCSoar before 6.2 used to store a "speed index", not the real
-       baud rate - try to import the old settings */
-
-    static constexpr unsigned speed_index_table[] = {
-      1200,
-      2400,
-      4800,
-      9600,
-      19200,
-      38400,
-      57600,
-      115200
-    };
-
-    MakeDeviceSettingName(buffer, "Speed", n, "Index");
-    unsigned speed_index;
-    if (map.Get(buffer, speed_index) &&
-        speed_index < ARRAY_SIZE(speed_index_table))
-      config.baud_rate = speed_index_table[speed_index];
+    config.baud_rate = 0;
   }
 
   MakeDeviceSettingName(buffer, "Port", n, "BulkBaudRate");


lordfolken:
So not a complete review, but some points here:
 Your commit messages:
 - Please use max 80 char for the first line
 - Typical convention if platform specific is to use the target name like: WIN64:  for example
 - I like your explanation and reasoning in the commit messages, but please remove the emotional component. :)

src/Device/Port/SerialPort.cpp
```diff
@@ -61,8 +62,11 @@ SerialPort::Open(const TCHAR *path, unsigned _baud_rate)
 
   DCB PortDCB;
 
+  StaticString<0x20> port;
+  port.Format(_T("\\\\.\\%s"), path);
+
   // Open the serial port.
-  hPort = CreateFile(path,
+  hPort = CreateFile(port.c_str(),
```
MaxKellermann, line 69: This violates this class's API, because it opens a different path than the one specified by the caller.

src/Profile/DeviceConfig.cpp
```diff
@@ -100,29 +100,21 @@ LoadPath(const ProfileMap &map, DeviceConfig &config, unsigned n)
...
-{
-  char buffer[64];
-  MakeDeviceSettingName(buffer, "Port", n, "Index");
-
-  unsigned index;
-  if (!map.Get(buffer, index))
+#ifdef _WIN32 
+  // on Windows don't use the form 'COM1:' anymore, use 'COM1'
+  if (map.Get(buffer, config.path)) {
+    if (config.path[config.path.length() - 1] == ':') {
```
MaxKellermann, line 106: Buffer overflow bug here; and a very clumsy way to access `config.path.back()`.

src/Profile/DeviceConfig.cpp
```diff
@@ -100,29 +100,21 @@ LoadPath(const ProfileMap &map, DeviceConfig &config, unsigned n)
...
+      config.path[config.path.length() - 1] = 0;
+      // map.Set(buffer, config.path);  // unfortunately map is a const object!
```
MaxKellermann, line 108: .... what????

src/Dialogs/Device/PortDataField.cpp
```diff
@@ -120,10 +124,13 @@ DetectSerialPorts(DataFieldEnum &df) noexcept
...
 {
-  for (unsigned i = 1; i <= 10; ++i) {
+  for (unsigned i = 1; i < 0xFF; ++i) {
     TCHAR buffer[64];
-    _stprintf(buffer, _T("COM%u:"), i);
-    AddPort(df, DeviceConfig::PortType::SERIAL, buffer);
+    constexpr uint32_t buffer_length = 0x100;
+    TCHAR portbuffer[buffer_length];
+    _stprintf(buffer, _T("COM%u"), i);
+    if (QueryDosDevice(buffer, portbuffer, buffer_length) != 0)
```
MaxKellermann, line 132: What does this do? This is obscure.

MaxKellermann:
Don't! Use! So! Many!! Exclamation!!! Marks!!!!

NEWS.txt
```diff
@@ -19,6 +19,9 @@ Version 7.24 - not yet released
   - new asw27 polar from idaflieg
 * Windows
   - new default fonts
+  - use 6 devices in serial port list (like all other systems)
+  - allow connection to COM port 10 to 255 (f.e. for BlueTooth user)
+  - remove index based COM port setting (obsolet since XCSoar 5)
```
MaxKellermann, line 24: typo - and even if it's obsolete, what's the justification for removing XCSoar 5 compatibility?

MaxKellermann:
> * all installed ports of the PC can be used
> * no uninstalled port is in the device list!
> * no more restriction of port numbering COM1 - COM9
> * remove function LoadPortIndex: obsolete thins XCSoar 5!

Rule of thumb: if your commit message contains an enumerated list, the commit is too crowded with unrelated changes.

src/Device/Port/SerialPort.cpp
August2111: Maybe, but this is a common way on Windows to select a serial port. I don't know any program, where I select the complete portname, every time the choice is 'COM3' or 'COM25' and so on. If you want to have the complete portname in the setting file (profile), you have to add this '\\\\.\\' after the selection in the port list, but this make not so much sense for me and makes the profile file more incomprehensible from my point of view!

MaxKellermann:
> But the reasoning behind keeping windows at 4 was shortsighted.

10 years is not so "short", for something that can be changed by removing 4 lines of code, for something nobody has ever needed (or will ever need?). I wanted to keep the UI as simple as possible for Windows CE users, and having 6 devices meant some contemporary screens suddenly needed a scrollbar.

Even today, I doubt anybody really uses (or wants to use) more than 4 ports on any Windows box with XCSoar 7. We dropped WinCE support 7 years ago; seriously, does anybody fly with a Windows PC?

Dropping the Windows restriction to 4 ports is okay, but it is only a theoretical improvement. It doesn't solve a practical problem, it doesn't add a feature anybody needs.

(This is just my personal assumption. Maybe there is really somebody who has a Windows PC in his glider, and has more than 4 Bluetooth devices, who knows. Crazy idea.)

August2111:
Maybe I'm the crazy guy, I manage a lot of BlueTooth devices via my Windows 10 laptop - and each BT connection creates its own port, which I then use to read out the devices. Of course, you can also do this with a smartphone - but I also like to do it with my notebook - and that's where I'm bothered by a - in my opinion pointless - limitation, especially since Windows CE support has been discontinued for ages, as you already say.
 In addition, the Windows version is also popular for playing and getting to know XCSoar, and the difference between the systems should be as small as possible!

lordfolken:
Well the ammount of devices controlled by XCSoar also have been steadily increasing. Radio, OGN, AirControl Display...

src/Profile/DeviceConfig.cpp
August2111: clumsy - okay. I am and will remain a newbie in your framework... And I see the problem with length() == 0 now!

src/Device/Port/SerialPort.cpp
August2111: I hoped to solve this problem on the root - without any API limitations. Maybe this is wrong. I will rebuild this part!

src/Profile/DeviceConfig.cpp
August2111: An overwriting of this profile setting isn't necessary - so I will remove this line. If the user make a new setting this don't have any colon!

NEWS.txt
August2111: To cleanup from old unused settings and debloating the system - or do you think, this is used somewhere?

src/Profile/DeviceConfig.cpp
MaxKellermann: That's not "my framework", that's the C++ standard library; are you a newbie to C++?
(See: https://en.cppreference.com/w/cpp/container/vector/back https://en.cppreference.com/w/cpp/container/array/back https://en.cppreference.com/w/cpp/container/list/back etc.)

src/Profile/DeviceConfig.cpp
MaxKellermann: I still have no idea what that means or what your code means or what that code comment means.

NEWS.txt
MaxKellermann: I don't know if it's used and it may (or may not) be reasonable to remove it - but you are removing this as an unrelated part of a cluttered bigger commit, and that's not reasonable.

src/Profile/DeviceConfig.cpp
August2111: I never worked with std::array up to now, so I'm a newbie with this!

NEWS.txt
August2111: The commit has to be split in two parts, you are right!

src/Device/Port/SerialPort.cpp
August2111: Building the device name moved to Device/Port/ConfiguredPort.cpp to fullfill the API

src/Dialogs/Device/PortDataField.cpp
August2111: Described in the new commit message

src/Device/Descriptor.cpp
```diff
@@ -483,7 +485,13 @@ try {
     msg.Format(_T("%s: %s (%s)"), _("Unable to open port"), name,
                (const TCHAR *)what);
 
-    env.SetErrorMessage(msg);
+    try {
+	    if (msg.length() > MAX_PATH-1)
+	      msg.Truncate(MAX_PATH-1);
```
MaxKellermann, line 490: > Because msg.Format could be bigger then the allowed length this has to truncated!

... why?

src/Device/Descriptor.cpp
```diff
@@ -483,7 +485,13 @@ try {
...
+      env.SetErrorMessage(msg);
+    } catch (const std::exception &e) {
+      return false;
```
MaxKellermann, line 493: > catch exception with wrong env variable

... why?

August2111: I'll postpone this troubleshooting until later, for this PR it doesn't matter for the time being. I once got a buffer overflow here that I would have liked to have cleared - but I guess I'll have to do a loop again...

MaxKellermann: What, a buffer overflow? Your commit messages does not say anything about a buffer overflow. And even if there were a buffer overflow - buffer overflows cannot be caught by a `catch` block. Nothing of this makes any sense.

August2111: I have to do something more in this topic, why the formatted string is longer then allowed - but I will do it later, not yet - and I have postponed the commit.

src/Profile/DeviceConfig.cpp
```diff
@@ -100,7 +100,20 @@ LoadPath(const ProfileMap &map, DeviceConfig &config, unsigned n)
 {
   char buffer[64];
   MakeDeviceSettingName(buffer, "Port", n, "Path");
+#ifdef _WIN32 
+  // on Windows don't use the form 'COM1:' anymore, use 'COM1'
```
MaxKellermann, line 104: That's wrong. A different part of this PR changes this aspect, but making this change here breaks the device configuration dialog.

MaxKellermann:
The commit message of "SerialPort remove function LoadPortIndex: obsolete since XCSoar 5! " does not make sense:
> The index based port baud rate does not fit to the baud rates of a real device:
> For example Flarm has a complete different baud rate list w/o 1200, 2400 and
> 115200 baud. So it should be never used in the future!

1. this code does not exist to be "used in the future"; it exists only to migrate old profiles
2. due to that, the limitation of available baud rates is not relevant for anything
3. therefore, nothing in this commit message explains why you want to remove this piece of code
4. the commit message only states things that are not relevant for what the commit is doing

src/Profile/DeviceConfig.cpp
August2111: Why? I only strip the colon from the profile setting, if the setting is with the old style. So in the dialog, in a new setting, in the opening of port  - overall I use the colonless portname. I don't see a break in the configuration dialog?

MaxKellermann: This code makes the profile settings inconsistent. Some parts of XCSoar still use a trailing colon, others do not. If you believe there should not be a colon, one commit should remove it everywhere, the commit should not do anything else, and the commit message should explain why this change should be done.
Also note that the colon change breaks all downgrades to an older XCSoar version. Such a compatibility breakage should be well justified. I don't see any justification at all.

August2111: And what is your proposal? To stay with the trailing colon itself for a downgradable setting? Otherwise by stripping the colon an older XCSoar has no connection to this port... But if a user set this to a port COM15 for example he also loses connection to that port. 
I prefer a cut at this point: From now on there should only be settings without a trailing colon - the workaround still takes the old setting into account. I'm not aware that there are other port names in Windows that have a colon after them.
But as you said before: Probably nobody uses Windows in a flight configuration.

MaxKellermann: At this point, I don't propose anything - you have proposed merging your code, but without explaining it. Once I have your explanation, I can review it, but I don't have it. I was just pointing out aspects that should be explained, that you should consider when writing the explanation. Your work is incomplete, so it's your turn.

src/Device/Port/ConfiguredPort.cpp
```diff
@@ -99,7 +99,19 @@ OpenPortInternal(EventLoop &event_loop, Cares::Channel &cares,
...
 
+#ifdef _WIN32
+    StaticString<0x20> port_path;
+    // use the windows style of device names:
+    port_path.Format(_T("\\\\.\\%s"), config.path.c_str());
+    if (!port_path.empty() && port_path.back() == TCHAR(':')) {
+      /* On Windows strip the trailing colon if the port name setting
+      has the old style 'COMx:' instead of 'COMx' */
+      port_path.Truncate(port_path.length() - 1);
+    }
```
MaxKellermann, line 110: What's the point of copying the trailing colon to the buffer if you're going to remove it later?

src/Device/Port/ConfiguredPort.cpp
```diff
@@ -163,35 +175,32 @@ OpenPortInternal(EventLoop &event_loop, Cares::Channel &cares,
     if (!ip_address.IsValid())
       throw std::runtime_error("No IP address configured");
 
-    return std::make_unique<TCPClientPort>(event_loop, cares,
-                                           ip_address, config.tcp_port,
-                                           listener, handler);
+    return std::make_unique<TCPClientPort>(event_loop, cares, ip_address,
+                                           config.tcp_port, listener, handler);
```
MaxKellermann, line 179: Why??

src/Device/Port/ConfiguredPort.cpp
```diff
@@ -163,35 +175,32 @@ OpenPortInternal(EventLoop &event_loop, Cares::Channel &cares,
...
   }
 
   case DeviceConfig::PortType::TCP_LISTENER:
-    return std::make_unique<TCPPort>(event_loop, config.tcp_port,
-                                     listener, handler);
+    return std::make_unique<TCPPort>(event_loop, config.tcp_port, listener,
+                                     handler);
```
MaxKellermann, line 184: Why??

MaxKellermann:
I'm unable to review this PR because there are so many coding style changes. Please avoid wasting my time with stuff like that, please fix your editor and please review your own commit diffs before you push them.

src/Device/Port/ConfiguredPort.cpp
```diff
@@ -83,11 +83,11 @@ WrapPort(const DeviceConfig &config, PortListener *listener,
...
-static std::unique_ptr<Port>
-OpenPortInternal(EventLoop &event_loop, Cares::Channel &cares,
-                 const DeviceConfig &config, PortListener *listener,
-                 DataHandler &handler)
-{
+static std::unique_ptr<Port> OpenPortInternal(EventLoop &event_loop,
+                                              Cares::Channel &cares,
+                                              const DeviceConfig &config,
+                                              PortListener *listener,
+                                              DataHandler &handler) {
```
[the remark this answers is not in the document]
August2111: Format mistake, also for your next 'Why?' too... ;-(

src/Device/Port/ConfiguredPort.cpp
```diff
@@ -99,7 +99,19 @@ OpenPortInternal(EventLoop &event_loop, Cares::Channel &cares,
     if (config.path.empty())
       throw std::runtime_error("No port path configured");
 
+#ifdef _WIN32
+    StaticString<0x20> port_path;
```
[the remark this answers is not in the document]
August2111: From my point of view yes: 4 Byte header `\\\\.\\`, 6 Byte portname (COM1 - COM255) -and maybe the trailing colon...

src/Device/Port/ConfiguredPort.cpp
August2111: I want to strip the trealing colon, if it is still present from a previous profile setting (not used anymore in the future) - without touching the config.path.

August2111:
> I'm unable to review this PR because there are so many coding style changes. Please avoid wasting my time with stuff like that, please fix your editor and please review your own commit diffs before you push them.

Please excuse the formatting errors in the ConfiguredPort.cpp file - I've always tried to avoid such errors by carrying out intensive checks, but unfortunately they keep slipping through...

lordfolken:
@MaxKellermann i would cherry pick this except the "serial port remove function" commit. Would you agree?

MaxKellermann:
Disagree. There are unanswered review questions.

MaxKellermann:
And there are unfixed bugs, which remain unfixed even after I pointed them out in review comments.

MaxKellermann:
> August2111 requested a review from MaxKellermann 29 days ago 

@August2111 this is really annoying. Why review again if you keep ignoring my previous reviews?

MaxKellermann:
No reply

lyusupov:
> > But the reasoning behind keeping windows at 4 was shortsighted.
> 
> 10 years is not so "short", for something that can be changed by removing 4 lines of code, for something nobody has ever needed (or will ever need?). I wanted to keep the UI as simple as possible for Windows CE users, and having 6 devices meant some contemporary screens suddenly needed a scrollbar.
> 
> Even today, I doubt anybody really uses (or wants to use) more than 4 ports on any Windows box with XCSoar 7. We dropped WinCE support 7 years ago; seriously, does anybody fly with a Windows PC?
> 
> Dropping the Windows restriction to 4 ports is okay, but it is only a theoretical improvement. It doesn't solve a practical problem, it doesn't add a feature anybody needs.
> 
> (This is just my personal assumption. Maybe there is really somebody who has a Windows PC in his glider, and has more than 4 Bluetooth devices, who knows. Crazy idea.)

Guys. I apologize for making a comment into a Closed ticket  - it is a bad practice, I agree- but it looks like this is the the only pace where you discuss current NUMDEV value...

This is an example where all the six XCSoar Device ports are in use:



there are 2 ADS-B protocols _(1090ES and UAT)_ one FLARM source, one OGN source, one FANET and one PilotAwarre. Accumulated air traffic from all these sources goes into XCSoar's Radar screen.

you can read this page https://github.com/lyusupov/SoftRF/wiki/Octave-Concept if you would like to see more details....

PR #936, opened 2022-05-31
InfoBoxes/Contenet/Weather: Update wind arrow according to heading variations

### LINKED ISSUE #914: Wind arrow infobox not updating according to heading variations
7.12 and newer (tested on Kobo Touch and WIN64).
Works as expected on 7.11 and older.

Expected behavior
-----------------
When the aircraft changes heading, the wind arrow infobox should update to show the relative wind direction (with aircraft nose pointing to the top of the screen)

Actual behavior
---------------
The arrow always stays in the same position, even if the numerical values may change; the arrow direction changes only when you touch/interact with the infobox.

Start a SIM flight, set some values for wind (I used 5 Kt and 90°), make the aircraft do some turning and see the wind arrow remaining the same.
Touch the infobox to see that the arrow direction updates correctly once.
Rinse and repeat.

I thought it could have something to do with this lines in the changelog:
 Kobo
    throttle sensor-triggered screen updates
    improve performance of graphical InfoBoxes

but it does the same also on windows version, so I have no clue...

No idea :-(
IANAC but I can try to do my best if someone point me in the right direction...

styblope:
Fixes #914


MaxKellermann:
I don't understand the fix, and the commit message doesn't explain it.

MaxKellermann:
... and this is necessary because ...
... and checking both track and heading is necessary because ....
(Explanations belong in the commit message, not in GitHub comments)

MaxKellermann:
The fix is still weird, because it checks `track_available`, but never actually uses the track. This depends on implementation details of `BasicComputer`, and InfoBox code shouldn't make any assumptions about implementation details of other parts of XCSoar. Instead, `ComputeHeading()` should update `heading_available` (copy `track_available`, or maybe `max(track_available, ground_speed_available)`).

PR #941, opened 2022-06-07
Input/InputEventsActions: end pattern list with \0

styblope:
Pointer gets out of bounds in `ScanMultiplePatterns` if pattern list isn't terminated by empty string. This results in picking up extra undefined patterns and misbehaviour.


MaxKellermann:
Good fix, but why did you omit your explanation in the commit message?

PR #943, opened 2022-06-10
Documentation: Removed surplus contributor list

Exadios:
The definitive contributor list is contained in the file "AUTHORS" in
the source root directory. It should not be replicated here.


Exadios:
The build-docs.yml fails presumably because the ch13_history.tex file still references the now deleted files. This is not the case at this end so I do not understand what is happening.

Exadios:
One more try at a pull request

MaxKellermann:
The CI log shows which file fails to build, but even easier is `git grep contributors_dev.tex`

lordfolken (approved):

PR #949, opened 2022-06-21
DownloadManager: Use AllocatedPath for downloads

### LINKED ISSUE #944: Unable to download Repository / Files - Files generated in CWD
The repository List isn't displayed and nothing is downloadable.

Affected Versions:
XCSoar current head 64b0fbe64742f269bbd325bd1c6fa976ef1265f

Version that are fine: 
v7.23

Target is UNIX

Expected behavior
-----------------
Opening the File Manager should display a list of the resources to download.
There shouldn't be strange files in the CWD directory. XCSoar should use a temporary
directory under ~/.xcsoar/tmp/ not the local working directory which might not be writeable.

Actual behavior
---------------
The List never shows when pressing the "add" button.
The current working directory is full of strangely named files.

```bash
rm  -rf ~/.xcsoar
mkdir downloadtest
cd downloadtest
~/git-xcsoar/XCSoar/output/UNIX/bin/xcsoar
```
1. start xcsoar in simulator mode (from terminal)
2. config-config->file manager
3. Press "add"

Nothing happens.

4. Exit XCSoar

```bash
 $  ls -lah downloadtest/
total 18K
drwxr-xr-x   2 folken folken   3 Jun 11 10:34  .
drwxrwx--- 120 folken folken 232 Jun 11 10:33  ..
-rw-r--r--   1 folken folken 40K Jun 11 10:34 '.Cٹ;V'
```
The file contains the download repository.

Potentially the changes to the io/filestream around the 13th of may. 
I will try to bisect.

lordfolken:
Fixes #944

### DIFF
--- a/src/net/http/DownloadManager.cpp
+++ b/src/net/http/DownloadManager.cpp
@@ -243,7 +243,7 @@ private:
 
 static Co::InvokeTask
 DownloadToFile(CurlGlobal &curl,
-               const char *url, const Path path,
+               const char *url, AllocatedPath path,
                std::array<std::byte, 32> *sha256,
                ProgressListener &progress)
 {


MaxKellermann:
The cause must be a compiler bug, because I see nothing wrong with the "bad" commit.
If this PR fixes the problem, it's only a workaround, and a bad one, because it forces overhead in the API.

lordfolken:
So i used the standard gcc that comes with debian
```
gcc --version
gcc (Debian 10.2.1-6) 10.2.1 20210110
```
For fun i reverted the commit on the current master, and i have the problem again.

PR #956, opened 2022-07-01
Move kernel binaries into repository

### LINKED ISSUE #880: Kobo prebuilt blobs not getting included in build
Kobo prebuilt blobs (kernels and modules) are not reliably integrated into the build system.  A more organized and managed system should be implemented to avoid releases that have old or missing blobs.

7.23 was released with a missing blob (Clara HD modules)
All prior releases were probably released OK but the problem is that the system of how to include the blobs is difficult and certainly error prone.

Expected behavior
-----------------

I expect that a developer can have a PR accepted and that is all that is required to have the blobs included.

Actual behavior
---------------

If a PR is submitted that involves blob changes (addition/deletion/modification), the blob must be handled outside the PR system.  There is no way I could see to ensure that the prebuilt blob is available to the build.

Probably historical.  Kernel build is an offline task due to complexity.  This means the kernels are prebuilt blobs.  Don't want to have prebuilt blobs in the repository and found a temporary shortcut for the small number that were needed (7 as of now I believe).  Worked OK for a while since kernel editing is VERY rare and only done by the most senior team members (who know how to access the build machine(s)).  When a very junior team member (me) submitted a PR, all I could do was to send emails and notes in the PR to say "please put this file in such and such a folder on the build machine".  The PR was accepted but the blob was never put in the required folder.

Yup.  A couple.
1. Allow blobs in the repository.
If you do this, the blobs can be part of the PR and when the PR is pulled, it is consistent.  Note these could be in a hidden folder 

2. If you really hate blobs in the repository, have another site (e.g. developer.xcsoar.org/blobs) or repository where files can be downloaded from.  If you do this, the download of the blobs can be done in the PR and the posting to the download site can be done by he who approves the PR.  Note this method is not as reliable as the first since it requires the PR approver to do an extra step (or the PR submitter to submit a PR for the "other" repository as well).

My preference is #1 as it is a single approval of a PR that leaves a consistent result.  Solution 2 requires a second step by the PR approver.  Both methods, however, mean that a build by anyone with access to the repository builds a complete release without having to manually put the blobs on his local machine.

I am VERY happy to do the work so I'm simply asking the powers that be (e.g. @MaxKellermann) for guidance so it will meet the quality standards of the project.  Note if option 2 is chosen, someone else will have to set up the site or repository but I can still do almost all the work.

GeekGoneOld:
The kernel binaries were not included or fetched from a known
source.  They were required to be put in a fixed directory
on every machine on which KoboRoot.tgz was built.  This made
it inaccessible to add/change/edit by PR.
Closes #880

Brief summary of the changes
----------------------------

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -61,6 +61,7 @@ The following people and organizations have contributed code to XCSoar:
  Nicolas Massart <nmassart@gmail.com>
  Uwe Augustin <info@flapsonline.de>
  Ronald Niederhagen <ronald_niederhagen@freenet.de>
+ Keith Laidlaw <laidlaw1546@gmail.com>
 
 Documentation:
 
--- a/build/kobo.mk
+++ b/build/kobo.mk
@@ -80,17 +80,77 @@ BITSTREAM_VERA_FILES = $(patsubst %,$(BITSTREAM_VERA_DIR)/%.ttf,$(BITSTREAM_VERA
 THIRDPARTY_TOOL_NAMES = simple_usbmodeswitch
 THIRDPARTY_TOOL_FILES = $(addprefix $(THIRDPARTY_LIBS_ROOT)/bin/,$(THIRDPARTY_TOOL_NAMES))
 
-KOBO_KERNEL_DIR = /opt/kobo/kernel
-
 CA_URL = https://curl.se/ca/cacert-2022-04-26.pem
 CA_ALTERNATIVE_URL = $(CA_URL)
 CA_MD5 = aa5ac583708ca35225ac2d230f4acb62
-
 CA_DOWNLOAD = $(DOWNLOAD_DIR)/$(notdir $(CA_URL))
 
 $(CA_DOWNLOAD): | $(DOWNLOAD_DIR)/dirstamp
 	@$(NQ)echo "  GET     $@"
 	$(Q)./build/download.py $(CA_URL) $(CA_ALTERNATIVE_URL) $(CA_MD5) $(DOWNLOAD_DIR)
+
+KOBO_KERNEL_URL = https://download.xcsoar.org/contrib/kobo/kernel/kobo-1.0.uImage
+KOBO_KERNEL_ALTERNATIVE_URL = $(KOBO_KERNEL_URL)
+KOBO_KERNEL_MD5 = 207f1b732330ac0bdc9c1fa1d16dd402
+KOBO_KERNEL_DOWNLOAD = $(DOWNLOAD_DIR)/$(notdir $(KOBO_KERNEL_URL))
+
+$(KOBO_KERNEL_DOWNLOAD): | $(DOWNLOAD_DIR)/dirstamp
+	@$(NQ)echo "  GET     $@"
+	$(Q)./build/download.py $(KOBO_KERNEL_URL) $(KOBO_KERNEL_ALTERNATIVE_URL) $(KOBO_KERNEL_MD5) $(DOWNLOAD_DIR)
+
+KOBOOTG_KERNEL_URL = https://download.xcsoar.org/contrib/kobo/kernel/kobootg-1.0.uImage
+KOBOOTG_KERNEL_ALTERNATIVE_URL = $(KOBOOTG_KERNEL_URL)
+KOBOOTG_KERNEL_MD5 = 7415607d93edb1426a10e278abc6919c
+KOBOOTG_KERNEL_DOWNLOAD = $(DOWNLOAD_DIR)/$(notdir $(KOBOOTG_KERNEL_URL))
+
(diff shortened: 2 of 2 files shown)


MaxKellermann:
Big fat **VETO**. A git repository should never have compiled binaries.

GeekGoneOld:
> Big fat **VETO**. A git repository should never have compiled binaries.

I don't have a problem with that.  You mentioned that about 5 months ago when I identified the issue.  I have been asking since then as to how to include compiled binaries in the build (as is already done with Kobo kernels) and created this solution to get the conversation going as my previous attempts (in issue #880 and earlier in PR #870 and PR #841) have not generated any answer.

PLEASE HELP

Currently there are at least 7 pre-built binaries (6 Kobo kernels and 1 tgz of drivers) involved in the build for KOBO and the method to include them in the build is get a copy of them and put them on each and every build machine in a specific directory (/opt/kobo/kernel).  The six kernels have already been done on the official build machines though contributors with forks would have to do it manually on their machines, copied from KoboRoot.tgz.  The seventh (drivers) is still not on the official build machines and can only be found in the conversation of PR #841!!!!

I haven't been able to find out who put them on the official build machines, even by PMing some key kernel contributors.  No contributor like me has the access required to do it so I need your help solving this issue.

So please tell me how you would like this done.  I have identified the problem and proposed two solutions.  Since this solution is unacceptable and I don't have access to complete the second solution, I'm asking once again for guidance on how you would like it done.

Failure to do this before 7.24 will mean that a second release will happen with OTG support for CLARA HD but no drivers unless the user goes to the conversation for PR841, downloads the tgz and either adds it to KoboRoot.tgz manually or puts them on the Kobo manually.

lordfolken:
What are the sources for these binaries? Can we reproduce them?
Is this normal kernel compilation? 
Such things could be achived in the build chain.
Binaries in git defeat the purpose of git and opensource.

lordfolken:
If its nit possible i can provide a download location on download.xcsoar.org and we can add a pull step.

GeekGoneOld:
> What are the sources for these binaries? Can we reproduce them? Is this normal kernel compilation? Such things could be achived in the build chain. Binaries in git defeat the purpose of git and opensource.

These are normal kernel cross compilation.  The six kernels were done some time ago by others who are no longer active.  The drivers for Clara HD were kernel cross compilation by me.  The kernel builds and driver builds were both challenging as they are a cross compile and only certain versions of toolchain work.  I have full documentation (step-by-step) of my build which generated the drivers.  I have no documenation (except the config.gz) on the six kernels.  Given enough time I could probably create them myself.  I'm not skilled enough to do it in the build, though.

> If its nit possible i can provide a download location on download.xcsoar.org and we can add a pull step.

This was my second (and preferred) solution.  I, of course, have no ability to post files on that site so I was unable to do that solution until I got your ear!  If it is an acceptable solution to you, I would be thrilled to do a PR providing you tell me the URL where I can find the files.

I can send you all 7 files or you can get the six from KoboRoot.tgz in ./opt/xcsoar/lib/kernel or from the official build machine at /opt/kobo/kernel.  The final one is in the conversation of PR #870 (most recent).

GeekGoneOld:
@lordfolken 
Here is the PR using download.xcsoar.org.  

Please let me know how to send you the binaries to put in downloads.xcsoar.org/content/KoboKernels (or whatever directory you see fit).  They are too big to attach to this comment.  I can PM them to you.  After you have put them there I will test the download portion before you pull the PR.

lordfolken:
https://download.xcsoar.org/contrib/kobo/kernel/ will be the path. (content is managed by xcsoar-data-content)
However for me this is temporary. This needs to be reproducible.

lordfolken:
The binaries are now available in http://download.xcsoar.org/contrib/kobo/kernel/
Please checksum the downloads in the solution you propose.

GeekGoneOld:
> The binaries are now available in http://download.xcsoar.org/contrib/kobo/kernel/

Thank you ever so much.  How would I go about requesting files be added/changed/deleted?

> Please checksum the downloads in the solution you propose.

I see you put the SHA hash files in the directory.  I checked the MD5 hash as is "normal" with the build download system.
> The binaries are now available in http://download.xcsoar.org/contrib/kobo/kernel/ Please checksum the downloads in the solution you propose.

GeekGoneOld:
> https://download.xcsoar.org/contrib/kobo/kernel/ will be the path. (content is managed by xcsoar-data-content) However for me this is temporary. 

Temporary?  What would be the permanent solution?

> This needs to be reproducible.

Reproducible?  I'm not sure what it is that must be reproducible.  It is as reproducible as it has been since custom kobo kernels were introduced but is now much more maintainable.

lordfolken (approved):

PR #964, opened 2022-07-07
SkyLines: Honor usage conditions of FileDescriptor

Exadios:
The documentation for FileDescriptor::Close() states that it should
only be called if File::Descriptor::IsDefined() is true. So do that.


MaxKellermann:
This PR is bad. `socket` is not a `FileDescriptor` but a `UniqueFileDescriptor`. Please revert, @lordfolken

MaxKellermann:
Wait. I was wrong. The commit message was misleading, because it's not a `FileDescriptor` at all, but a `UniqueSocketDescriptor`, and that class doesn't have a special `Close()` implementation like `UniqueFileDescriptor` has, which is a mistake in that API.

PR #965, opened 2022-07-07

Exadios:
Modify some, but not all,  headers to track them as files. This allows
header dependency graphs to be produced.


MaxKellermann:
I don't agree with cluttering those `@file` commands everywhere. I don't understand the commit message (@lordfolken, do you?). Therefore I don't agree with merging this PR.

Issue #921, opened 2022-05-02
Flarm Targets should be icons according to aircraft type


lordfolken:
Currently all Flarm targets are the same triangle. In order to differentiate between different aircraft, the targets should be iconifed:
I.e. powerplane/glider/paraglider/helicopter. 

The should:
 * be only an outline of an icon
 * retain the color for above/below
 * have some form of direction to indicate direction of travel

lordfolken:
@MikeBravo15 could you design some aircraft icons, as you did for the waypoint icons?

Basically top down and clearly indicating a direction of travel?

The list would be the following:

![Screenshot_20230426-075738_1](https://user-images.githubusercontent.com/407371/234482974-e7f265b2-f23c-4a6a-a085-acf229ac35f7.png)

lordfolken:
The idea is to display them on the map and the large traffic radar. I'll help you with the got stuff.

MikeBravo15:
I will try to draw something

MikeBravo15:
Do you want a dual set of icons? - ie "info" state and "warning" state (some red element)?

lordfolken:
I think we should color them at runtime. So please do them white or black on a transparent background.

MikeBravo15:
Work preview

![obrazek](https://user-images.githubusercontent.com/42399455/235749574-cddae428-e101-4d1b-a88f-d3268a64c758.png)
Is white border OK?

Sundown3867:
Enclosed please find collection of icons 
![obrazek](https://user-images.githubusercontent.com/127534937/236619726-8d630deb-212f-49dc-8c90-22b49fd05dfc.png)
[FlarmWarningIcons.zip](https://github.com/XCSoar/XCSoar/files/11412144/FlarmWarningIcons.zip)

fraktal-sb:
As long categories in Flarm and Transponder are not harmonized we will have problems to display the right aircraft symbol. Example:  Some UL-Transponder use in their setup B4 (Ultralight) which will be interpreted as category 7 in FLARM which is paraglider. The correct setup in the UL- transponder should be "Light" (less than 7000 kg) and this would be translated as category 8 in FLARM. So it would be good to have an option to bypass the symbol display and still have the triangle.

lordfolken:
I'm aware. 

The current traffic model is based on FLARM. I would like to have a general "traffic" object, that is not opinionated, and then have pluggable adapters that translate between the source and the traffic model. 

[...]

PR #972, opened 2022-07-20
Enhance Map Icons (WIP)

### LINKED ISSUE #1130: Some icons are worse than in v7.24
v7.8 has, v7.4 does not. I assume this change was made at v7.8

Expected behavior
-----------------
Whilst I accept that the icons needed to be reviewed, I think that some of the changes make XCSoar less usable. I think it's important to be able to take in information at a glance, and some of the new icons are larger and more complex, making that more difficult.

1. Mountain top (type 7) and Obstacle (type 8) have white backgrounds.

image

2. Normal waypoint (type 1) is now a faint crosshairs, making it much harder to pick out.

3. Obstacle (type 8) is now represented by what appears to me to be a transmitting tower. Previously it was just a simple A-frame tower, so suitable for other obstacles such as windmills.

4. Mountain top icon was previously a small black triangle, which made it ideal for point heights. Now it's a large icon with a snowcapped peak and a couple of subsidiary peaks, and visually similar to MtPass (type 6)
image

5. A couple of new icons (type 20 & 22) have been introduced. On close inspection, these appear to be hang gliders, but it isn't obvious at all... my first guess was satellite dishes. These are not documented. Oddly, there doesn't appear to be a type 21 icon.

6. Likewise, there are undocumented light blue icons for VOR & NDB (types 9 & 10) and type 19 - a light blue triangle in a circle.

As I said, I welcome the effort to review and improve the icons. However, I would have been very happy if *additions* had been made to the existing set, rather than replacing some of them. I believe that there needs to be some rigor in documenting the set of icons available to developers, so that they can be used by all, and so that a change doesn't have a detrimental effect on pre-existing maps. So I think the solution is not in fixing this particular gripe, but in formalising a process for amending icons and assigning numbers to them.

A better solution would be to move the icons from the executable to separate e.g. .PNG files, so that a map developer could package up appropriate icons in the .XCM along with the topography Shapefiles.

lordfolken:


MaxKellermann:
The color red should only be used to indicate imminent danger; it is way too prominent in the new map_tower icon. It's also too prominent in the existing map_obstacle. If we use red everywhere, it loses its significance.
The new map_mountain_top has odd geometry at the right corner of the white polygon.
The new map_pass has a very thin black outline where the white road leaves the black area. I don't think we need to invert the outline here, but if we do, the line width should be the same as the white outline on the rest of the shape; and there shouldn't be a transparent region between white road and black outline.
I think adding thin outlines to vector graphics doesn't work well with all output resolutions; I wonder if we'd better generate those outlines automatically after rasterizing, to ensure a proper outline width at all display resolutions.

hjr:
Think some of the symbols just got to small on current screen resolutions. We might need to compensate with proper rendering according to screen resolutions.

elgandoz:
Hello guys, I would like to propose some changes to the icon set.
I like the new concept although I noticed few things that could be improved:

- icons do not have a consistent visual weight (some appear prominent, some almost disappear in the background)
- some icons are rendered blurry due to the aliasing/scaling, this could be improved by uniforming and simplifying the design
- many of them have too many details, making them hard to recognise especially on a phone mounted on a bouncy cockpit.
- uniform the general style (some have drop shadow, thin/thick/no borders, tiny/big details), taking inspiration from topographic conventions and trying to maintain the current visuals.
- uniform the sizing: every icon should be contained in the same canvas in order to have consistent size
- improve contrast, especially for the coloured ones.
- Simplify the markup: SVG is an open standard that can be read and edit by application, but many icons contains lots of filters and vendor metatags that could create inconsistent rendering. This could be done by proper export and/or using cli tools like [SVGO](https://github.com/svg/svgo).

My intention is to provide a vector mockup, placed on a rendered map, which could be exported after having it discussed and refined.
I'm just checking if it's worth working on it, or if the ship has already sailed, since the current icons are now released.
I'm posting on this PR since I couldn't find a specific issue.

MaxKellermann:
@elgandoz no ship has sailed, and improvements to all aspects of XCSoar are welcome.
Always keep in mind that visuals are difficult to improve. There are a hundred thousand pilots flying with XCSoar, who have a hundred thousand and one opinions on how XCSoar should look like, and no matter what you do, one group or the other will be unhappy.
Sometimes, things that appear to be an improvement for some is a regression for others, and sometimes, there are aspects that flight computers should obey like keeping red reserved for indicating danger. I think some entity has published some guidelines somewhere on color use in flight instruments - many years ago, @jwharington did some research.

On the other hand, it's fairly easy to improve XCSoar, because much of XCSoar is not very consistent and badly designed. Many of the design aspects of XCSoar, if somebody would submit them to me today, I'd reject them because they are poor (they predate my involvement in XCSoar). But that's where our efforts start - we have a software that works for many pilots, but it's far from as good as it should be.

lordfolken:
@elgandoz please also be aware of #921 if you are creating icons. Flarm traffic could also use bitmaps that signify the aircraft type.

MikeBravo15:
Actually I am working on the icons overhaul also. My aim is to use 26x26 px, optimized (no bloat) svg with viewbox parameter as template for all XCSoar icons. I started with waypoint icons. As reference I use this document by Naviter, describing individual icon style - https://downloads.naviter.com/docs/SeeYou_CUP_file_format.pdf. Here you can see my progress so far

NewIcon_scrsht_2

Here you can see my PC testbuild XCSoar with icons in actio - right bottom icon is not changed yet.

NewIcon_scrsht

 I would like to make unifor graphic style of all icons of the sw with more uniform size.

lordfolken:
closing this in favour of #1155

PR #974, opened 2022-07-27

Exadios:
Modify some, but not all, headers to track them as files in Doxygen
produced documentation. This allows header dependency graphs to be
produced.


MaxKellermann:
This just adds a lot of noise just to make some random tool you're using happy. No, this doesn't belong in the XCSoar source code. Fix your tools instead. See db9fdd8b016745c672333eb00bb90b41e9fbbbb5

PR #978, opened 2022-08-05
Additional info to the "TC gain" info box adding up the below zero va…

### LINKED ISSUE #977: TC loss as additional info to TC gain
XCSoar version
--------------
7.23  as well as 7.24

I'd like to see the over all altitude loss while thermalling and trying to catch a thermal. Mathematically spoken the integral of all vario values below zero freo TC begon to end.

Comparing and optimizing thermalling strategies.

Writing this in the name of Michael Godoj mgodoj@gmx.net

hjr:
…rio as "loss" for the current thermal.

Closes #977


hjr:
Do we have an issue in our android build job definition?

hjr:
Had the chance to check and fly with the new feature. It does not reset properly on starting a thermal. Patch coming.

hjr:
Now the TC gain is properly set to zero on entering a climb. Was a sleeping bug.

MaxKellermann:
This PR branch is a confusing mess, can't review.
The commit message of "Properly clear OneClimbInfo before entering" does not explain anything, and I don't understand the code change.

lordfolken:
@hjr you need to rebase your code, and when you pull from master use --rebase.

hjr:
@lordfolken, thx, that was the hint I needed. Forgot about the rebase .. in a moment.

src/Computer/GlideComputerAirData.cpp
```diff
@@ -134,8 +134,10 @@ GlideComputerAirData::ProcessVertical(const MoreData &basic,
 
   if (calculated.circling)
     CurrentThermal(basic, calculated, calculated.current_thermal);
-  else
+  else {
     calculated.current_thermal = calculated.last_thermal;
+    calculated.current_thermal.Clear();
```
MaxKellermann, line 139: Why?

src/Computer/GlideComputerAirData.cpp
```diff
@@ -189,14 +191,14 @@ GlideComputerAirData::CurrentThermal(const MoreData &basic,
...
-  } else
-    current_thermal.Clear();
+  current_thermal.start_time = circling.climb_start_time;
+  current_thermal.end_time = basic.time;
+  double tmp_gain = basic.TE_altitude - circling.climb_start_altitude_te;
+  if ( tmp_gain < current_thermal.gain ) {
+    current_thermal.loss += tmp_gain - current_thermal.gain;
+  }
+  current_thermal.gain = tmp_gain;
+  current_thermal.CalculateAll();
```
MaxKellermann, line 201: I don't understand this change.

src/NMEA/ClimbInfo.hpp
```diff
@@ -46,6 +46,9 @@ struct OneClimbInfo
   /** Altitude gained while in the thermal [m]. May be negative. */
   double gain;
 
+  /** Altitude lost while in the thermal [m]. */
+  double loss;
```
MaxKellermann, line 50: Who writes this new field?

MaxKellermann (commented):
I still don't understand this code change. Either something's wrong, or I need more explanation.

src/Computer/GlideComputerAirData.cpp
hjr: CurrentThermal() was never ever called with undefined climb_start_time time stamp. This was a sleeping bug and the Clear never got called. Whereas the Clear was not necessary until the addition of the "loss". The loss gets calculated incremental, the gain just as simple difference, no need for variable initialization.

src/Computer/GlideComputerAirData.cpp
MaxKellermann: But ... why assign `current_thermal` and then clear everything of it? This way, the assignment has no effect. This doesn't appear to make sense.

src/Computer/GlideComputerAirData.cpp
MaxKellermann: If you believe this is a bug, it should be fixed (and documented) in a separate commit.
The way you did this is hide it in an unrelated commit. That's bad.

src/NMEA/ClimbInfo.hpp
MaxKellermann: Yes, I see that now. Too bad you made your diff so hard to read by doing two unrelated things in one commit, one of which required changing the indent of the code where you inserted new code. This makes your code change unreadable.

hjr:
IMG_2210
Generally spoken, the "loss" info gives a quality to what you achieved by thermalling. The more straight forward you kept the glider in climbing air the less "loss" will add up until you leave the spot. I skipped plotting this also for the last thermal info box.

src/Computer/GlideComputerAirData.cpp
hjr: It is at least my best guess. Have not been keen enough to create a PR based on a best guess and w/o fixing an effective failure mode. 
Call it now the needed change for the additional loss attribute.

src/Computer/GlideComputerAirData.cpp
hjr: Hm, true, it's not a reference.

hjr: Removed the assignment, completed the Clear().

src/Computer/GlideComputerAirData.cpp
MaxKellermann: If you're only guessing, then I'll refrain from merging this PR. I don't like merging code if the author is only guessing.

src/Computer/GlideComputerAirData.cpp
MaxKellermann: And why does it need to be cleared?

hjr: "loss" is a sum that is expected to start from beginning of circling. It needs to be cleared when switching to circling mode. The current implementation of this callback does allow to distiguish circling mode from none circling mode. Thus we can clear, whenever not in circling mode to have the sum zeroed from entering circling mode on.
I am open for more efficiant ways here. The choosen one should be bullet proof.

src/Computer/GlideComputerAirData.cpp
hjr: I "guess" I dont understand you. My statement was that I refuse to push a so called fix on vague guess. What do you want to see here. Pushing a fix based on a gues .. not acting on a guess? Are you just messing up my words for fun or for not having the time reading?
The current implementation needs a zeroed start value, the current TC gain implementation does not. The straight forward coded change you see here.
Please advice clearly.

MaxKellermann: I'm not messing with anything, I was just asking about a code change, and you said it fixed a bug, then you said it's just a guess. That confuses me.
If you think there's a bug, we should fix that bug first, before we discuss merging the feature this PR is about. But at least it should be a separate commit, before the commits adding this new feature, and the commit message should explain the bug being fixed. That's what I want to see (in the presence of a bug, one I don't know).

src/Computer/GlideComputerAirData.cpp
MaxKellermann: I don't understand your response. It begins with explaining something about "loss", but this part of the change is not about "loss", it's about clearing `calculated.current_thermal`, but none of what you just wrote convinces me that `calculated.current_thermal` should be cleared.
If not circling, `calculated.current_thermal` should contain information about the most recent thermal; it is explicitly implemented that way; clearing it would make that information unavailable.

src/Computer/GlideComputerAirData.cpp
hjr: So, I do not see a bug, nor a missbehavior of the software. I do see  a if - else, where the else never gets executed. For the intention of this I can only speculate. 
1) It could be that this "if" just was a safty measure in case CurrentThermal() would have been called with undefined TC start time stamp. all fine
2) It could have been that the intention/expectation was to really have the TC struct initialized when not in circling mode. Whereas, there is no need with the current feature set.
Useless to speculate, I easily can slip the change in a fix and a feature addition. Will force push the change and keep it in one PR. But we will have a commit commet that will be describing the change.

src/Computer/GlideComputerAirData.cpp
hjr: Ok, my understanding is that with setting the end of thermal time stamp and changing the circling mode the TC struct is copied over to the "last TC" variable and the TC is unused until you hit the next thermal. I'll look deeper into that.

src/Computer/GlideComputerAirData.cpp
MaxKellermann: If you don't know what this `if` is about and whether it's useful, then please don't submit a PR removing it. And, even more importantly: don't hide such a change in an unrelated commit.

Maybe it's really not useful, but maybe it is. We don't know. But let's not blindly hack the code without knowing what we're doing.

MaxKellermann:
The core of this PR is just a few lines; adding a new field, setting it, showing it in the UI; instead, this PR is a complicated beast doing several things, changing indent, fixing a comment, adding a new formatting feature, hiding an alleged bug fix. Had it been commited in the most simplest form, I wouldn't understand the use of this feature, but at least I could have merged it.

src/Computer/GlideComputerAirData.cpp
```diff
@@ -190,13 +189,15 @@ GlideComputerAirData::CurrentThermal(const MoreData &basic,
                                      OneClimbInfo &current_thermal)
 {
   if (circling.climb_start_time.IsDefined()) {
-    current_thermal.start_time = circling.climb_start_time;
+    if (current_thermal.start_time != circling.climb_start_time) {
```
MaxKellermann, line 192: Do not use `==` or `!=` to compare floating point values. (I don't understand what this change is trying to do.)

src/Computer/GlideComputerAirData.cpp
```diff
@@ -132,10 +132,9 @@ GlideComputerAirData::ProcessVertical(const MoreData &basic,
                         calculated);
   AverageClimbRate(basic, calculated);
 
-  if (calculated.circling)
+  if (calculated.circling) {
     CurrentThermal(basic, calculated, calculated.current_thermal);
-  else
-    calculated.current_thermal = calculated.last_thermal;
```
MaxKellermann, line 138: Why .. remove this? I don't get it, and I don't get your commit message.

src/Computer/GlideComputerAirData.cpp
```diff
@@ -132,10 +132,9 @@ GlideComputerAirData::ProcessVertical(const MoreData &basic,
                         calculated);
   AverageClimbRate(basic, calculated);
 
-  if (calculated.circling)
+  if (calculated.circling) {
```
MaxKellermann, line 135: Why add curly braces here? This obscures the commit diff.

src/Computer/GlideComputerAirData.cpp
```diff
@@ -360,39 +361,36 @@ GlideComputerAirData::LastThermalStats(const MoreData &basic,
...
-  else
-    calculated.last_thermal_average_smooth =
-        LowPassFilter(calculated.last_thermal_average_smooth,
-                      calculated.last_thermal.lift_rate, LOW_PASS_FILTER_THERMAL_AVERAGE_ALPHA);
-
-  ThermalSources(basic, calculated, calculated.thermal_locator);
+  // event condition: enter cruise mode, alias leaving climbe mode
+
+  if (calculated.climb_start_time.IsDefined())
+  {
```
MaxKellermann, line 373: I don't understand any of these changes. None of this appears to be about clearing TC data.

src/NMEA/ClimbInfo.hpp
```diff
@@ -45,6 +45,9 @@ struct OneClimbInfo
   /** Altitude gained while in the thermal [m]. May be negative. */
   double gain;
 
+  /** Altitude lost while in the thermal [m]. */
```
MaxKellermann, line 48: I think this needs a better description; the semantics of this field should really be well-defined. Even from your code, I have difficulties deriving the meaning of this field (and I still don't know the practical use, and I believe it is a very fragile and inaccurate value, because it depends so much about the vario's update rate and the volatility of the TE altitude).

src/Computer/GlideComputerAirData.cpp
hjr: It just triggers the very first call. Ther "enter climb" condition.

src/Computer/GlideComputerAirData.cpp
hjr: All logic what to do when leaving a climb transitioning to cruise, or being in cruise mode, is hosted in LastThermalStats() except this line. I joined to the rest of the logic to have it consistent. This assignment is only needed once when transitioning to cruise mode, not every time updating while in cruise mode.

src/Computer/GlideComputerAirData.cpp
hjr: I like curles.

src/Computer/GlideComputerAirData.cpp
hjr: I made the effort to understand the given lines. Here the outcome:
The couple return statements at the beginning of the method do prevent a proper action on all off them applying. That we do need here, it's the  above mentioned assignment of last_thermal, and the Clear() of the current_thermal.
In other words: `(calculated.circling || !last_circling) return;` or better `(!calculated.circling && last_circling) do {}` is just the important transition trigger climb to cruise. That is the routines purpose.
Then we have the condition of gain <=0 or thermal duration to small to trash the current_thermal information and not shift it into last_thermal.
That is what you clearly can read here now. The assignment of the last_thermal was a kludge. current thermal is at this moment well defined and last_thermal should just show exactly that.

src/NMEA/ClimbInfo.hpp
hjr: Sure I will add some more words here.

hjr: Did so

PR #979, opened 2022-08-05
Misc documentation

Exadios:


MaxKellermann:
No, no, and again no. https://github.com/XCSoar/XCSoar/pull/974

There are good bits in this PR, but they are obscured by bullshit changes in the same commit, i.e. I can't even cherry-pick the good bits. That's why it's important that you do separate commits for separate things.

PR #983, opened 2022-08-15
Eliminate duplicate waypoints in list

### LINKED ISSUE #816: Duplicate Waypoints on short waypoint files
The waypoints  apperar twice.
It is a normal *.cup - File. Exportet from Strepla für LX9000

Having:
XCSoar v7.19 (debian)
XCSoar7 v.20-Android
It's the same behavior an Windows 7.21-PC

Having not:
XCSoar 7.18 (debian)

Expected behavior
-----------------

A waypoint corresponds to a waypoint in the CUP-File

Actual behavior
---------------

2 identic waypoints corresponds to 1 waypoint in the CUP-File

* Load the attached file as the sole file
* Open the waypoint list via for example the guesture

No Idea

No. Sorry

### LINKED ISSUE #958: Waypoint list displays duplicate entries for .cup

7.18 is OK
7.19 has the problem
Possibly introduced in 9533b1221259257112d658a97474dddf9957cc03

Expected behavior
-----------------

When using guesture "L", I expect to see the consolidated list of waypoints with each one appearing only once.

Actual behavior
---------------

If a .cup waypoint has both a Title and Code (fields 1 and 2) non-blank, it appears in the waypoint list twice

Using any .cup file with a line that has both Title and Code (field 1 and 2), on the map display guesture "L" to call up the waypoint list and you will see that entry twice in the list (not necessarily consecutive).

Perhaps it was a small error when the second field support was introduced in 7.19 or it has to do with filtering on both fields.  If it is the latter, I would hope a solution can be found to support filtering and only show once.

Not really.  I'm unfamiliar with the code.  I'm happy to help if I can get guidance.

GeekGoneOld:
Retains filtering name and shortname but deletes duplicates
(e.g. when the shortname is the same as the name).  Also
eliminates duplicates when distance is the filter.  Bonus: this
should slightly speed up rendering of the waypoints on the map.
Closes #958
Closes #816

Brief summary of the changes
----------------------------

### DIFF
--- a/src/Waypoint/WaypointList.cpp
+++ b/src/Waypoint/WaypointList.cpp
@@ -62,3 +62,18 @@ WaypointList::SortByDistance(const GeoPoint &location) noexcept
 {
   std::sort(begin(), end(), WaypointDistanceCompare(location));
 }
+
+void
+WaypointList::SortByName() noexcept
+{
+  std::sort(begin(), end(), [](WaypointListItem &a, WaypointListItem &b)
+                            {return a.waypoint->name < b.waypoint->name;});
+}
+
+void
+WaypointList::MakeUnique() noexcept
+{
+  auto new_end = std::unique(begin(), end(), [](WaypointListItem &a, WaypointListItem &b)
+                                        {return a.waypoint->id == b.waypoint->id;});
+  erase(new_end, end());
+}
--- a/src/Dialogs/Waypoint/WaypointList.cpp
+++ b/src/Dialogs/Waypoint/WaypointList.cpp
@@ -282,6 +282,10 @@ FillList(WaypointList &list, const Waypoints &src,
 
   if (filter.distance > 0 || !filter.direction.IsNegative())
     list.SortByDistance(location);
+  else
+    list.SortByName();
+
+  list.MakeUnique();
 }
 
 static void
--- a/src/Waypoint/WaypointList.hpp
+++ b/src/Waypoint/WaypointList.hpp
@@ -54,7 +54,9 @@ public:
 class WaypointList: public std::vector<WaypointListItem>
 {
 public:
+  void SortByName() noexcept;
   void SortByDistance(const GeoPoint &location) noexcept;
+  void MakeUnique() noexcept;
 };
 
 #endif


GeekGoneOld:
I assume the failure of build-macos is not caused by the PR.  I didn't do anything with a jpeg library.

src/Waypoint/WaypointList.cpp
```diff
@@ -62,3 +62,30 @@ WaypointList::SortByDistance(const GeoPoint &location) noexcept
 {
   std::sort(begin(), end(), WaypointDistanceCompare(location));
 }
+
+class WaypointNameCompare
```
MaxKellermann, line 66: What's the point of this class? It has no fields; this should better be a standalone function with no `this` pointer overhead, or maybe even a lambda.

src/Waypoint/WaypointList.cpp
```diff
@@ -62,3 +62,30 @@ WaypointList::SortByDistance(const GeoPoint &location) noexcept
...
+WaypointList::SortByName() noexcept
+{
+  std::sort(begin(), end(), WaypointNameCompare());
+}
+
+void
+WaypointList::MakeUnique() noexcept
+{
+  auto new_end = std::unique(begin(), end(), [](WaypointListItem &a, WaypointListItem &b)
+                                        {return a.waypoint->name == b.waypoint->name;});
```
MaxKellermann, line 89: Why compare names? There could be two different waypoints with the same name (but different location, type etc.). That's not a good way to enforce unique waypoints, I think.

MaxKellermann (changes requested):

MaxKellermann:
Why sort by name again, anyway? The vector is already sorted.

src/Waypoint/WaypointList.cpp
GeekGoneOld: I have changed it to a lambda

src/Waypoint/WaypointList.cpp
GeekGoneOld: I'm happy to use the waypoint id.  That, of course, would be unique.  I used the name because I felt that it would be reasonable to expect the name to be unique (NOT the shortname) and I also felt it would be most reasonable to present the list in some order and the most reasonable order was by name.

If I use waypoint id as the uniqueness, I would have to sort twice (once by id then unique() then by name).

GeekGoneOld:
> 

I don't believe that is true.  I used my .cup files (attached) and removed the sort.  It resulted in duplicates in the list (at the very bottom).
[cupfile.zip](https://github.com/XCSoar/XCSoar/files/9384272/cupfile.zip)

src/Waypoint/WaypointList.cpp
GeekGoneOld: Rethinking it, if the name is unique, sorting by name would allow unique() to work as two with the same name would also have the same id and the second one would be eliminated even if we unique() on id!  As a bonus, we only sort once (by name).

Caution, though.  If, by chance, there were three with the same name in the list, two of which were the same id and the third a different waypoint, the list COULD contain a duplicate if the order was ID1, ID2, ID1.  I would consider this a very minor issue that would be acceptable as it is caused by non-unique names (which should be discouraged) and no data is lost to the user.

I have updated the PR to unique by id after sorting by name.

MaxKellermann:
> > Why sort by name again, anyway? The vector is already sorted.
> 
> I don't believe that is true.

You're right - the list is sorted, but by the name by which it was inserted (which may be the primary name or the short name), not by its primary name.

PR #989, opened 2022-08-16
Terrain: Set contrast/brightness according to modern screens

lordfolken:
I think we are giving XCSoar's Terrain Renderer not enough credit.
This change changes the default setting for the renderer (which is an expert setting)
to a high brightness and lower contrast value. 
The shading is still visible and the appearance is more detailed.

Before:
2022-08-17-010016_958x351_scrot
After:
2022-08-17-010037_958x351_scrot


MaxKellermann:
D'oh, the old screenshot looks awful.
0b306ea0ad6681069e72a49edebfe697021ca9c2 introduced those ugly old defaults, more than 16 years ago.

PR #990, opened 2022-08-16
gh: build-native only upload artifact on push

lordfolken:
Currently PR builds fail, as it tries to upload the android binary to XCSoar, which of course isn't permitted and the secrets aren't available.


hjr:
Hm, it does not seem to solve the issue. Checks failed again with the same staging error???

PR #996, opened 2022-08-18
github actions: build-native, fix indents, scope variables, prevent staging task on PR/Forks

lordfolken:


hjr:
Applause!

lordfolken:
Well i hope it works when pushed to master. The rules kind of prevent the task from running here, but doesn't mean it got it wrong and it does work in master.

.github/workflows/build-native.yml
```diff
@@ -83,12 +83,25 @@ jobs:
           make -j$(nproc) TARGET=${{env.TARGET }} DEBUG=${{ env.DEBUG }} USE_CCACHE=y V=2 everything check
 
       - name: upload artifact
-        if: github.event_name == 'push'
```
MaxKellermann, line 86: Why remove this now?

lordfolken: This is in the wrong place. This uploads the artefact to the github actions result page, where you can download the artefacts from for a (days) limited time. 
The staging one in the next step uploads to download.xcsoar.org/testing/ and should only happen on push to master and repo being xcsoar/xcoar.

MaxKellermann: But this doesn't disable the upload; it removes the condition, re-enabling the upload for non-push events. I'm confused.

lordfolken: There are 2 uploads here:
 - upload artifact
 - upload to staging server

The artifact is uploaded on a run to the github action's results page. Example here:
https://github.com/XCSoar/XCSoar/actions/runs/2885709841

This is not a problem for PRs and Forks, as there are namespaced per action run. So even if a fork uses the ci to uplaod an artiifact it will only be available on that partitcular artifact page.

The upload to staging server uploads the binaries to https://download.xcsoar.org/testing/
- this is available to users without a github account
- the downloads are available for unlimited ammount of time (gh's actions artefacts get erased at some point)
- represent current state of development
- (Future plan: to actually do releases on push of tags)

Of course this action cannot run from fork's or PRs as the secrets defined in the organization are missing, and they shouldn't be uploaded.

lordfolken: i split the commit now 2 different ones. One that renables the uploads for the artefact upload and another which implements the uploads for the staging server. Does that help?

MaxKellermann: Having separate commits clearly communicates that these are different issues being addressed. I get confused when two things which I both don't understand individually are getting mixed in one commit; that prevents me from understanding either, because it suggests that these belong together.

lordfolken: I totoally understand.

PR #1008, opened 2022-08-24
GlideComputerAirData.cpp: Ensure TC <-> TL logic on climb/cruise transitions

hjr:
…itching to circling mode.

Max, please help here. You added this code 2011, it's long ago, I know.

Current Thermal() only gets called with defined clime_start_time time stamp. I would see the moment to clear the current_thermal struct at the moment of entering the circling mode. That is when climb_start_time gets updated.
By triggering on this change I think a meaningful moment for clearing the current_thermal is found.

### DIFF
--- a/src/Computer/GlideComputerAirData.cpp
+++ b/src/Computer/GlideComputerAirData.cpp
@@ -132,10 +132,9 @@ GlideComputerAirData::ProcessVertical(const MoreData &basic,
                         calculated);
   AverageClimbRate(basic, calculated);
 
-  if (calculated.circling)
+  if (calculated.circling) {
     CurrentThermal(basic, calculated, calculated.current_thermal);
-  else
-    calculated.current_thermal = calculated.last_thermal;
+  }
 
   lift_database_computer.Compute(calculated.lift_database,
                                  calculated.trace_history.CirclingAverage,
@@ -190,13 +189,15 @@ GlideComputerAirData::CurrentThermal(const MoreData &basic,
                                      OneClimbInfo &current_thermal)
 {
   if (circling.climb_start_time.IsDefined()) {
-    current_thermal.start_time = circling.climb_start_time;
+    if (current_thermal.start_time != circling.climb_start_time) {
+      current_thermal.start_time = circling.climb_start_time;
+      current_thermal.Clear();
+    }
     current_thermal.end_time = basic.time;
     current_thermal.gain =
       basic.TE_altitude - circling.climb_start_altitude_te;
     current_thermal.CalculateAll();
-  } else
-    current_thermal.Clear();
+  }
 }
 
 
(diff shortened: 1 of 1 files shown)


MaxKellermann:
The commit message doesn't even explain why. How can I review something that you did not even try to explain?

hjr:
Hey Max, this is the question going to yourself. The code never clear() the current_thermal struct, even a line for it was provided. You requested the fix for it, even I explained there is no harm introduced with it.
You are the author of those code line not doing anything, never being called. Please give me a hint what commit message could be you would like to see here.

MaxKellermann:
I did not request a fix, because I don't know that a fix is necessary. You said there may be a bug, and a fix may be necessary, and I said if you believe a fix is necessary, submit a fix with a proper explanation. Since you submitted a fix, it's up to you to explain what was wrong and why you believe it was wrong.

hjr:
Last attempt, Max. What was your intend with line 199 in GlideComputerAirData.cpp? Your answer will give me a clue if there is a change needed or not. I.e. this proposed one.

MaxKellermann:
> What was your intend with line 199 in GlideComputerAirData.cpp?

This line is reached when already circling, but no climbing start time was recorded yet. Therefore, the statistics of the current thermal must be zero, because no data is available (not yet).

hjr:
Ok, thanks for the insight.
Still have the issue that I flew with this TC loss for about 10h and did debugging investigation replaying igc files and never witnessed current_thermal.clear() has been called, and thus tc_loss been reset.
I do understand the subtle difference you want to make in-between start circling and start of climb. Whereas I see in line 167ff CirclingComputer.cpp the circling flag set to true and the climb_start_time set the same moment. So as of my understanding this confirms my observation: There is no circling with climb_start_time undefined or unset. On top of this I miss to find the code that sets the climb_start_time back to 0 or undefined when i.e. entering cruise mode.

My conclusion here is I let this up to you to investigate further. We do not have an issue with it. The TC loss can be implemented not touching this status quo start_climb_time logic at all.

hjr:
My suggestion to have TC data cleared at the meaningful spots, also added comments.

For the moment TC data is collected but graded as attempt I did prefer to clear the TC data instead of copying the TL data over it. It does result in a TC display of "--", meaning not defined, instead of the same data as the TL gain info box.

Otherwise this change complies to your explanation and intent of the TC TL logic. But it does not change this mentioned gap between entering circling and climb phase. This does per line 167ff CirclingComputer.cpp not exist. As a consequence the TC data never got reinitialized on climbe/cruise transitions.

The change would ease the addition of the loss sum, because the Clear() can now be used to reset the loss sum properly.

igc replay tested so far.

hjr:
moved commit into #978

PR #1018, opened 2022-08-26
Ignore executable rights on tty device file descriptors

### LINKED ISSUE #1012: Native serial ports e.g. ttyS7, ttyS3 not seen under ANDROID target
Android 11 and XCSoar 7.24 and older versions tested down to 7.20 don't see those interfaces

Expected behavior
-----------------
Open Device manager dialog and all native tty devices for serial port's e.g. ttyS1, etc are seen

Actual behavior
---------------
Those ports do not list up at all in android, similiar if under linux the user running XCSoar is not in the group dialout

Launch XCSoar on an android board with native serial devices and check device dialog which interfaces exist and which do not. Further info and testcase also here: #1002 

Similar to linux when the user running XCSoar is not in the group 'dialout' where those interfaces aren't seen as well,
and commented by Max under #1002, "Android doesn't give XCSoar permissions to use those ports", its likely a permission problem. 

Not really, any idea's how to workaround, or fix this would be great. 
Older app's on those boards linked with older SDKs still can (see #1002), so the issue IMHO comes with XCSoar in conjunction with newer SDK.  
A similar issue we have with scoped storage where new SDK linked apps won't be able to access folders in the files system. On OS level a serial /dev/ttyS3 device is nothing more than a file descriptor, hence also older XCSoar versions where exactly this works (7.20) also won't see the tty's so this might be also different story.
I have a board like this on my table, and can build for android debug version and do tests to get more light into this.

Comments and other ideas welcome.

iltis42:
Brief summary of the changes
----------------------------
Ignore tty devices with executable rights on their device file descriptor to make them usable in XCSoar

Closes #1012

### DIFF
--- a/src/Device/Port/TTYEnumerator.cpp
+++ b/src/Device/Port/TTYEnumerator.cpp
@@ -66,7 +66,7 @@ TTYEnumerator::Next() noexcept
       /* truncated - ignore */
       continue;
 
-    if (access(path, R_OK|W_OK) == 0 && access(path, X_OK) < 0)
+    if (access(path, R_OK|W_OK) == 0 )
       return path;
   }
 


lordfolken:
Indeed that should only be rw no x.

MaxKellermann:
Your commit message doesn't explain why.
You don't even explain why you believe tty devices should be executable. What does it even mean when a character device is executable?

iltis42:
> Your commit message doesn't explain why. You don't even explain why you believe tty devices should be executable. What does it even mean when a character device is executable?
I haven't stated anywhere that tty devices should be executable, instead i have stated that executable rights are useless, see my comment in #1012, exactly "There is no use at all for executable rights here. Kernels allow to set this flag, but won't allow to execute device files (2) either, so this flag is mostly useless at all and for our application safely can be ignored."
and my proposal to fix the issue #1012 is to simply ignore the executable rights flag on the tty and look into the R/W flags only.

MaxKellermann:
I don't want to search GitHub for explanations. A commit message should contain all that is needed to understand why a change is necessary. Your commit message contains nothing.

MaxKellermann:
btw. I think this PR is bad, but maybe that's only because I don't have enough information (in this case: not enough = none at all).
This check is necessary to avoid listing directories (which are executable) as serial ports. Simply removing the check is a regression.
Merging a regression needs a good justification, but as I said, there is none currently.

iltis42:
I don't see directories under /dev that start with "tty", and at first there is a filter that masks everything that does not start with tty as the first thing:
https://github.com/XCSoar/XCSoar/blob/dfeecc7390225abbdc90720e28fa933e22e7b82c/src/Device/Port/TTYEnumerator.cpp#L36
So i put into question if there is a real need for that, as no directory will appear at that place where executable devices are masked.

MaxKellermann:
That's a very poor and unconvincing "explanation". (Actually, it's not an explanation at all, it's just the refusal to provide an explanation.)
"On my computer, the corner case your code attempts to protect against doesn't appear to exist, so I'm gonna remove the protection for everybody."
By that reasoning, we can remove 90% of XCSoar's code, because almost all of a program's code deals with corner case and error handling - why bother, errors shouldn't happen anyway ...

Seriously, we can't move forward until you can explain your suggested code change.

iltis42:
Not sure what else info exactly is missing here, maybe its about differences between OSes?
Yes UNIX based OSes don't have executable rights on tty, 100% agree, that might be the reason why you aren't faced with that situation on your computer.
Meanwhile most Android gadgets usually don't feature native serial devices, but there are meanwhile several android boards on the market may have, e.g. rockchip based designs with the chip featuring 10 ttys (!) internally and usually there are a few of them externally wired. Adding a good 7 inch display and we got a great device to run XCSoar.
And exactly those devices on this boards may have this executable bit set by the kernel (an i got meanwhile a second board, from a different manufacturer that behaves similar), hindering to get those devices into service with XCSoar, meanwhile other applications see them and we can play with them. 
The intent of this pull request is to remove this IMHO surplus check, to allow those boards to run native serial ports.
Does this explanation cover the gap you see @MaxKellermann or can you please advice, what else is missing ?

MaxKellermann:
> may have this executable bit set by the kernel

This is the important piece, and it's missing in your commit message (and is still missing).
However, I doubt this - I don't believe that the kernel sets the executable bit. This may be buggy userspace, but certainly not the kernel.
You can't just remove this check. Instead, if you really really believe the check breaks your use case, replace it with a better check, but don't remove it. (And, of course, explain what you did and why.)

iltis42:
Except XCSoar, and the SerialAPI demo app we haven't added anything in the userspace (and even before launching SerialAPI, exe rights are there), maybe the factory which flashes this board are doing this. Usually native tty under Android is a root thing only, needed for initial access for flashing/testing/configuring the device, means normally there are no rights for userspace, and in userspace this can't be modified (just the "chmod 666" won't change anything, and the su hack apparently is no more allowed by google policy). It is either like this: Those board's are providing the user access for the ttys somewhere in a rooted control flow, a control flow that is not in our hands, or if we alter this, next Android update may do it again.
Regarding a better check: Agree we may better check this tty file descriptors, e.g. a clean POSIX negative check if the devices file descriptor is not a directory should already do what you intend, or maybe a better is the idea is to check if the file is a character device instead (ab)using the exe bit as a filter criteria, what should work in any environment in my opinion, e.g. the way its described here:
https://en.cppreference.com/w/cpp/filesystem/is_character_file
Thoughts ?

The commit message will be adapted of course, sure we need there something new, but lets first converge our discussion here for a solution we all agree.

MaxKellermann:
No, the std::filesystem API must not be used in XCSoar. It is a bloated misdesigned mess.

lordfolken:
Can we use F_OK instead of X_OK?        

"The mode specifies the accessibility check(s) to be performed,
       and is either the value F_OK, or a mask consisting of the bitwise
       OR of one or more of R_OK, W_OK, and X_OK.  F_OK tests for the
       existence of the file."

According to the man page that checks if the destination is a file. 

if (access(path, F_OK|R_OK|W_OK) == 0

Trying to determine folders from files by the executable bit is ideal. Technically you can remove the executable bit of a directory.

lordfolken:
So at least in my basic tests this worked
chmod 775 ttyS0 -> File listed in Device List
mkdir ttyS33 && chmod 777 -> Directory not in Device List

MaxKellermann:
> According to the man page that checks if the destination is a file.

If by "file" you mean "regular file", then no. The manpage is confusing; it says "file" when it really means "an entity with the given path name", which can be a regular file, directory, char/block device, socket, pipe (and maybe more exotic stuff, like virtual kernel files).

lordfolken:
i was also looking for a more detailed explanation then whats in the manpage. But it seems that they really only mean a file and not a directory. According to my test above. 

As suspected, the current code will happily list directories in the device list if the executable bits are removed:
mkdir ttyS44
chmod 666 ttyS44 -> Device Tree lists directory ttyS44

MaxKellermann:
> i was also looking for a more detailed explanation then whats in the manpage. But it seems that they really only mean a file and not a directory. According to my test above.

That still doesn't make sense, because a character device is not a (regular) file. So what you're saying is that F_OK accepts a character device but not a directory?

lordfolken:
You can stat a file you have no access to:
```bash
touch foo
chmod 000 foo
stat foo
```

MaxKellermann:
> You can stat a file you have no access to:

That's because stat neither reads nor writes anything.

lordfolken:
So just checking for F_OK (no other checks)
will list files it has no access to. 

code:
      if (access(path, F_OK) == 0)

file:
crwxrwxr-x 1 root dialout 4, 64 Aug 29 15:44 ttyS0
crw-rw---- 1 root dialout 4, 65 Aug 29 15:44 ttyS1
c--------- 1 root dialout 4, 66 Aug 29 15:44 ttyS2
crw-rw---- 1 root dialout 4, 67 Aug 29 15:44 ttyS3
drwxrwxr-x 2 root root       40 Aug 30 17:12 ttyS44

2022-08-30-173334_958x351_scrot

but also directories. (ttyS44) 

So in the end F_OK isn't the filter i was looking for.

iltis42:
Something like this came into my mind:
```
struct stat st;
if (stat(name, &st) != -1){   // do we have access ?
        if (  S_ISCHR(st.st_mode) == 0){  //  if its not a character device ?
            continue;   // next device
        }
   }
}else{  // user doesn't have access
    continue;  // next device
}
```

lordfolken:
We also need to handle symlinks. 
Basically resolve the symlink and check the target.

iltis42:
According to the POSIX standard: https://pubs.opengroup.org/onlinepubs/009696799/functions/stat.html
symlinks shall be covered by stat, afaik by default this is the mode (can be switched of by flags), we need to check.
"If the named file is a symbolic link, the stat() function shall continue pathname resolution using the contents of the symbolic link, and shall return information pertaining to the resulting file if the file exists."

MaxKellermann:
> Something like this came into my mind:

You mean something like this?
https://github.com/XCSoar/XCSoar/blob/f525c2aea6f196b85470e7d79da206ae163bf4a0/src/Device/Port/TTYPort.cpp#L185-L191

Don't repeat yourself - if code you need already exists, write a commit to move it to a library to make it available to other parts of XCSoar, then write a commit which uses the function.

iltis42:
Sure, this is the right approach. We are converging. Will look into this as next step tomorrow.

iltis42:
#1034 as a precondition to progress here is off the ramp

MaxKellermann:
fyi, you can have multiple commits in one PR; you don't need to make two PRs for that.

iltis42:
replaced by new PR #1035

PR #1029, opened 2022-08-30
code cleanup

fberst:
- use of [[maybe_unused]]

### DIFF
--- a/src/Device/Driver/IMI/Protocol/Conversion.cpp
+++ b/src/Device/Driver/IMI/Protocol/Conversion.cpp
@@ -2,7 +2,7 @@
 Copyright_License {
 
   XCSoar Glide Computer - http://www.xcsoar.org/
-  Copyright (C) 2000-2021 The XCSoar Project
+  Copyright (C) 2000-2022 The XCSoar Project
   A detailed list of copyright holders can be found in the file "AUTHORS".
 
   This program is free software; you can redistribute it and/or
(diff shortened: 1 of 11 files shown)


src/Device/Driver/IMI/Protocol/Conversion.cpp
```diff
@@ -31,9 +31,9 @@ Copyright_License {
 #include <stringapiset.h>
 #endif
 
-#define IMI_SECONDS_IN_MINUTE       (60)
-#define IMI_SECONDS_IN_HOUR      (60*60)
-#define IMI_SECONDS_IN_DAY    (24*60*60)
+static constexpr int IMI_SECONDS_IN_MINUTE = 60;
+static constexpr int IMI_SECONDS_IN_HOUR = 60*60;
+static constexpr int IMI_SECONDS_IN_DAY = 24*60*60;
```
MaxKellermann, line 36: Wrong type. This shouldn't be signed.

MaxKellermann:
btw. next stop: finally enable `-Wunused` ;-)

fberst:
> btw. next stop: finally enable `-Wunused` ;-)

I will have a look at it :-)
Isn't this already enabled because it's included in -Wall?

MaxKellermann:
> I will have a look at it :-) Isn't this already enabled because it's included in -Wall?

Unfortunately not.
https://github.com/XCSoar/XCSoar/blob/dfeecc7390225abbdc90720e28fa933e22e7b82c/build/warnings.mk#L16-L17

src/Device/Driver/IMI/Protocol/Conversion.cpp
fberst: i changed the types to unsigned

groundhogxc:
I guess my comment would have been better put here:

Would it be possible to please combine all these basically identical (and fairly trivial) minor file changes into big combined commits with a common description? The commit list becomes extremely cluttered in this way, and it is cumbersome to find other things - and these particular commits and their comments are, while probably good and important, not extremely informative when reading the descriptions :-)

But thanks for your work either way!

MaxKellermann:
> Would it be possible to please combine all these basically identical (and fairly trivial) minor file changes into big combined commits with a common description?

Agree. Splitting into so many micro-commits is not useful.

fberst:
ok, i will combine the changes into bigger commits next time

PR #1033, opened 2022-08-31
finally enable -Wunused

fberst:
As suggested by max i worked on enabling the -Wunused compiler-option.

Not quite sure about this. Tell me if i should split it into more smaller commits.

### DIFF
--- a/build/warnings.mk
+++ b/build/warnings.mk
@@ -14,7 +14,6 @@ CXX_WARNINGS += -Wno-format-truncation
 endif
 
 # disable some warnings, we're not ready for them yet
-CXX_WARNINGS += -Wno-unused-parameter
 CXX_WARNINGS += -Wno-missing-field-initializers
 CXX_WARNINGS += -Wcast-align
 
--- a/src/Cloud/Main.cpp
+++ b/src/Cloud/Main.cpp
@@ -2,7 +2,7 @@
 Copyright_License {
 
   XCSoar Glide Computer - http://www.xcsoar.org/
-  Copyright (C) 2000-2021 The XCSoar Project
+  Copyright (C) 2000-2022 The XCSoar Project
   A detailed list of copyright holders can be found in the file "AUTHORS".
 
   This program is free software; you can redistribute it and/or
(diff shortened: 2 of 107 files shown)


MaxKellermann:
```
In file included from test/src/KeyCodeDumper.cpp:27:
test/src/Main.hpp:198:14: error: unused parameter 'argc' [-Werror,-Wunused-parameter]
int main(int argc, char **argv)
             ^
test/src/Main.hpp:198:27: error: unused parameter 'argv' [-Werror,-Wunused-parameter]
int main(int argc, char **argv)
                          ^
test/src/KeyCodeDumper.cpp:69:31: error: unused parameter 'p' [-Werror,-Wunused-parameter]
  bool OnMouseDown(PixelPoint p) override {
```
Lots of warnings remain.

fberst:
Build-Target and options? 
I compiled it for UNIX / ANDROID / PC / KOBO with DEBUG=y and it worked for me without errors

MaxKellermann:
Build target `everything` - I assume you omitted all those test/debug programs.

fberst:
> Build target `everything` - I assume you omitted all those test/debug programs.

Ok that makes sense, i didn't think about the test programs, i will go through these and fix this

MaxKellermann:
And: sometimes, it's less ugly to just remove the parameter name instead of adding `[[maybe_unused]]` (oh, how ugly is C++ ...)
That has a nice side effect; it enforces that the parameter is really not used (because there's no name it could be referred to).

fberst:
> And: sometimes, it's less ugly to just remove the parameter name instead of adding `[[maybe_unused]]` (oh, how ugly is C++ ...) That has a nice side effect; it enforces that the parameter is really not used (because there's no name it could be referred to).

that's the next step, after i identified and marked all those places in the code i can start removing some of these parameters

MaxKellermann:
No, not remove parameters; remove the parameter **name**.
```
void Foo(int unused_arg) {} // screams warning at you
void Foo([[maybe_unused]] int unused_arg) {} // ugly
void Foo(int) {} // not ugly, but why is this parameter here in the first place?
```
Of course, you can remove unused parameters; but sometimes, you cannot, e.g. for virtual overrides, or if the parameter is used in an `assert()`.

PR #1037, opened 2022-09-02
code cleanup

fberst:
- more pragma once 
- use of static constexpr instead of preprocessor macros
- remove of unused parameter

### DIFF
--- a/src/Engine/Airspace/AirspaceWarningManager.cpp
+++ b/src/Engine/Airspace/AirspaceWarningManager.cpp
@@ -1,7 +1,7 @@
 /* Copyright_License {
 
   XCSoar Glide Computer - http://www.xcsoar.org/
-  Copyright (C) 2000-2021 The XCSoar Project
+  Copyright (C) 2000-2022 The XCSoar Project
   A detailed list of copyright holders can be found in the file "AUTHORS".
 
   This program is free software; you can redistribute it and/or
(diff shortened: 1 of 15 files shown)


src/Atmosphere/CuSonde.cpp
```diff
@@ -38,10 +38,10 @@ Copyright_License {
  * @see http://en.wikipedia.org/wiki/Lapse_rate#Dry_adiabatic_lapse_rate
  * @see http://pds-atmospheres.nmsu.edu/education_and_outreach/encyclopedia/adiabatic_lapse_rate.htm
  */
-#define DALR -0.00974
+static constexpr float DALR = -0.00974;
```
MaxKellermann, line 41: Should both be `double`, not `float`

src/Engine/Airspace/AirspaceWarningManager.cpp
```diff
@@ -28,7 +28,7 @@
 #include "AirspaceAircraftPerformance.hpp"
 #include "Task/Stats/TaskStats.hpp"
 
-#define CRUISE_FILTER_FACT 0.5
+static constexpr float CRUISE_FILTER_FACT = 0.5;
```
MaxKellermann, line 31: Same here - `float` is wrong, `double` should be used. In this case, changing it to `double` has probably no effect, because 0.5 can be represented exactly with both types, but for other constants, it may be relevant if using `float` leads to truncation.

MaxKellermann:
Watch out for more `float` vs `double` problems - the advantage of macros is that the compiler can choose any type, while a `constexpr` variable may force the compiler to truncate the value. (That is probably the only "advantage" of macros. Good to get rid of them, but some care should be taken to get the types right.)

PR #1042, opened 2022-09-08
Water ballast synch between XCV and XCS in absolute ballast mode

### LINKED ISSUE #824: XCVario to XCSoar ballast synchronisation errors
Hardware 
XCvario firmware v21.1230-17
Huawei P30 lite Android v10.0.0.491 XCSoar v7.21
Bi-directional data exchange over BT using XCV drivers and protocol
Earlier versions of XCS exhibit problem.

Problem
This may relate to issue #807
Loaded into XCV, dry mass, pilot weight, and polar reference mass expressed as wing loading kg/m2
and in XCS Planes, Wing Area, Ref Mass and Dry mass (which is to include pilot according to manual).
With zero litres of ballast on both XCV and XCS when I open the flight setup page on XCS there is 42L that has been loaded,  Closing the flight setup page writes back 42l to XCV, opening the flight page again 81l is then loaded to XCS which is then written back to XCV and so this continues each time I open the flight setup page,116l then 148l and so on. Pressing the Dump button in XCS produces some very interesting results!!!
Changing ballast in XCS to zero or any other number, writes back that number to XCV but the addition process starts again when XCS Flight Setup is opened.
It is not possible to zero or synchronise the ballast between XCS and XCV when bi-directional data exchange is running. The only solution is disable the bi directional data exchange "Sync from device" with the loss of the QNH,MC and Bugs sync, 
Adjusting the dry mass and the pilot weight to equal the ref mass in XCV also overcomes the problem but at the expense of STF on the XCV.

My Observation for what its worth!
Looking at the Cambridge protocol which I think is used, in part, in the Xcvario protocol, calls for a ballast ratio made up of the total mass of the glider (dry mass, pilot, water) divided by the polar ref mass). In the case of my setup this is 1.27 with no ballast (water).
Monitoring the data stream from XCV I see the ballast variable is indeed  1.27. 
XCS seems to be treating any ratio over 1 as additional ballast. 
If I set the total mass equal to the ref mass in XCV, as suggested above, I get a ballast ratio of one on the data stream and the problem goes away, STF on XCV not withstanding.

iltis42:
**Brief summary of the changes**
----------------------------
Migrate fractional ballast exchange to absolute values in XCVario protocol

- Implement new fields for absolute values crew weight and empty mass for both directions and for ballast from device
- Migrate .xcp file extension appending into PlaneGlue to support resue of FileWrite() and avoid duplicate code
- Add backward compatible XCVario protocol version handling, now having a protocol version 2 running in absolute ballast mode
- Save plane file when as well when synchronized by external settings

**Related issues and discussions**
Closes #824 
#1017
https://github.com/iltis42/XCVario/issues/202


src/Dialogs/Plane/PlaneListDialog.cpp
```diff
@@ -234,9 +234,7 @@ PlaneListWidget::NewClicked()
       continue;
     }
 
-    StaticString<42> filename(plane.registration);
-    filename += _T(".xcp");
-
+    StaticString<42> filename( PlaneGlue::GetFileName(plane) );
```
MaxKellermann, line 237: Simpler:
```
  const auto filename = PlaneGlue::GetFileName(plane);
```
Using `auto` also allows changing the function's return type (e.g. larger buffer) without changing all the callers.

src/Dialogs/Plane/PlaneListDialog.cpp
```diff
@@ -279,8 +277,7 @@ PlaneListWidget::EditClicked()
       continue;
     }
 
-    StaticString<42> filename(plane.registration);
-    filename += _T(".xcp");
+    StaticString<42> filename( PlaneGlue::GetFileName(plane) );
```
MaxKellermann, line 280: `auto`

src/Plane/PlaneFileGlue.cpp
```diff
@@ -208,3 +210,21 @@ PlaneGlue::WriteFile(const Plane &plane, Path path)
   buffered.Flush();
   file.Commit();
 }
+
+StaticString<42>
+PlaneGlue::GetFileName(const Plane &plane)
```
MaxKellermann, line 215: Missing `noexcept`

src/Plane/PlaneFileGlue.cpp
```diff
@@ -208,3 +210,21 @@ PlaneGlue::WriteFile(const Plane &plane, Path path)
...
+{
+  StaticString<42> filename(plane.registration);
+  filename += _T(".xcp");
+  return(filename);
```
MaxKellermann, line 219: Superfluous parantheses

src/Plane/PlaneFileGlue.cpp
```diff
@@ -208,3 +210,21 @@ PlaneGlue::WriteFile(const Plane &plane, Path path)
...
+}
+
+// Simplified interface to write plane file including file extension handling
```
MaxKellermann, line 222: API documentation should be in doxygen-style comments in the header

src/Plane/PlaneFileGlue.cpp
```diff
@@ -208,3 +210,21 @@ PlaneGlue::WriteFile(const Plane &plane, Path path)
...
+void
+PlaneGlue::WriteFile(const Plane &plane)
+{
+  StaticString<42> filename( GetFileName(plane) );
+  const auto path = LocalPath( filename );
```
MaxKellermann, line 227: Simpler and more efficient:
```
  const auto path = LocalPath(GetFileName(plane));
```
This allows the compiler to generate code to free the stack space occupied by `GetFileName()` immediately after returning from `LocalPath()`.

src/Plane/PlaneFileGlue.cpp
```diff
@@ -208,3 +210,21 @@ PlaneGlue::WriteFile(const Plane &plane, Path path)
...
+  LogFormat(_T("WriteFile path:%s"), path.c_str() );
```
MaxKellermann, line 228: Useless logging call

src/Plane/PlaneFileGlue.hpp
```diff
@@ -37,5 +38,7 @@ namespace PlaneGlue
 /**
  * Throws exception on error.
  */
-void WriteFile(const Plane &plane, Path path);
+  void WriteFile(const Plane &plane, Path path);
```
MaxKellermann, line 41: Why change indentation?

src/Plane/PlaneFileGlue.hpp
```diff
@@ -37,5 +38,7 @@ namespace PlaneGlue
...
+  void WriteFile(const Plane &plane);
```
MaxKellermann, line 42: bad indent

src/Plane/PlaneFileGlue.hpp
```diff
@@ -37,5 +38,7 @@ namespace PlaneGlue
...
+  StaticString<42> GetFileName(const Plane &plane);
```
MaxKellermann, line 43: Add `[[gnu::pure]]` and `noexcept` to allow more compiler optimizations

src/Blackboard/DeviceBlackboard.cpp
```diff
@@ -257,6 +257,22 @@ DeviceBlackboard::Merge()
   }
 }
 
+void
+DeviceBlackboard::SetCrewWeight(double weight,
+                  OperationEnvironment &env)
```
MaxKellermann, line 262: Bad indent

src/Blackboard/DeviceBlackboard.cpp
```diff
@@ -257,6 +257,22 @@ DeviceBlackboard::Merge()
...
+{
+  if (devices != nullptr)
+     devices->PutCrewWeight(weight, env);
+}
+
+void
+DeviceBlackboard::SetEmptyMass(double weight,
+                   OperationEnvironment &env)
```
MaxKellermann, line 270: Bad indent

src/Blackboard/DeviceBlackboard.hpp
```diff
@@ -165,6 +165,10 @@ class DeviceBlackboard
 
   void SetBallast(double fraction, double overload,
                   OperationEnvironment &env);
+  void SetCrewWeight(double weight,
+                    OperationEnvironment &env);
```
MaxKellermann, line 169: Bad indent

src/Device/Descriptor.cpp
```diff
@@ -957,6 +957,59 @@ DeviceDescriptor::PutBallast(double fraction, double overload,
   return true;
 }
 
+bool
+DeviceDescriptor::PutCrewWeight(double weight,
+                             OperationEnvironment &env) noexcept
+{
+  assert(InMainThread());
+
+  if (device == nullptr || !config.sync_to_device )
```
MaxKellermann, line 966: Stray space before the `)`

src/Device/Driver.hpp
```diff
@@ -101,6 +101,24 @@ class Device {
   virtual bool PutBallast(double fraction, double overload,
                           OperationEnvironment &env) = 0;
 
+  /**
+    * Send the new crew weight value to the device.
+    *
+    * @param total weight in kg of all crew members
```
MaxKellermann, line 107: kilogram is not a weight unit, but a mass unit. The unit of weight would be Newton.

src/Device/Driver.hpp
```diff
@@ -101,6 +101,24 @@ class Device {
...
+    * @return true on success
+    */
+  virtual bool PutCrewWeight(double weight,
+                   OperationEnvironment &env) = 0;
```
MaxKellermann, line 111: Bad indent

src/Device/Driver.hpp
```diff
@@ -101,6 +101,24 @@ class Device {
...
+
+  /**
+      * Send the new empty mass value of the glider to the device.
+      *
+      * @param empty mass in kg of the glider
+      * @return true on success
+      */
```
MaxKellermann, line 118: Baaaaaad indent

src/Device/Driver.hpp
```diff
@@ -101,6 +101,24 @@ class Device {
...
+   virtual bool PutEmptyMass(double weight,
```
MaxKellermann, line 119: PutEmpty**Mass** but the parameter is called **weight**??

src/Device/Driver/XCVario.cpp
```diff
@@ -31,6 +31,7 @@ Copyright_License {
 #include "NMEA/InputLine.hpp"
 #include "util/Clamp.hpp"
 #include "Atmosphere/Pressure.hpp"
+#include "LogFile.hpp"
```
MaxKellermann, line 34: Please ... no more useless log messages.

src/Device/Driver/XCVario.cpp
```diff
@@ -144,6 +153,35 @@ PXCV(NMEAInputLine &line, NMEAInfo &info)
   return true;
 }
 
+bool
+XVCDevice::XCV(NMEAInputLine &line, NMEAInfo &info)
+{
+  double value;
+  char topic[16];
+  line.Read(topic, 16);
+  if( StringIsEqual( topic, "crew-weight") ){
```
MaxKellermann, line 188: Bad coding style

src/Device/Driver/XCVario.cpp
```diff
@@ -206,6 +244,48 @@ XVCDevice::PutBallast(double fraction, [[maybe_unused]] double overload,
   return true;
 }
 
+bool
+XVCDevice::SentNMEAItem(const char *which, int value, OperationEnvironment &env)
+{
+  /* the new xcs protocol for various int items e.g. with absolute weights or protocol version incluing the NMEA checksum */
+  static constexpr auto timeout = std::chrono::seconds(2);
+  char line[32];
+  sprintf(line,"xcs,%s,%d", which, value );
```
MaxKellermann, line 328: Bad coding style

src/Device/Driver/XCVario.cpp
```diff
@@ -206,6 +244,48 @@ XVCDevice::PutBallast(double fraction, [[maybe_unused]] double overload,
...
+  LogFormat("sendNMEAItem: !%s%s", line, checksum );
+  return true;
+}
+
+bool
+XVCDevice::PutCrewWeight(double weight,
+                         OperationEnvironment &env)
+{
+  if( protocol_version == XCV_VERSION_2 ) {
+    return( SentNMEAItem( "crew-weight", iround(weight), env ) );
```
MaxKellermann, line 343: Bad coding style

src/Device/Driver/XCVario.cpp
```diff
@@ -206,6 +244,48 @@ XVCDevice::PutBallast(double fraction, [[maybe_unused]] double overload,
...
+  }
+  return true;
+}
+
+bool
+XVCDevice::PutEmptyMass(double weight,
+                        OperationEnvironment &env)
+{
+  if( protocol_version == XCV_VERSION_2 ) {
+    return( SentNMEAItem( "empty-weight", iround(weight), env ) );
```
MaxKellermann, line 353: Bad coding style

src/Device/Driver/XCVario.cpp
```diff
@@ -206,6 +244,48 @@ XVCDevice::PutBallast(double fraction, [[maybe_unused]] double overload,
...
+  }
+  return true;
+}
+
+bool XVCDevice::PutProtocolVersion( int version,
+                         OperationEnvironment &env )
+{
+  return( SentNMEAItem( "version", version, env ) );
```
MaxKellermann, line 361: Bad coding style

src/Device/Driver/XCVario.cpp
```diff
@@ -206,6 +244,48 @@ XVCDevice::PutBallast(double fraction, [[maybe_unused]] double overload,
...
+XVCDevice::PutEmptyMass(double weight,
+                        OperationEnvironment &env)
+{
+  if( protocol_version == XCV_VERSION_2 ) {
+    return( SentNMEAItem( "empty-weight", iround(weight), env ) );
+  }
+  return true;
+}
+
+bool XVCDevice::PutProtocolVersion( int version,
```
MaxKellermann, line 358: Bad coding style

src/Device/Driver/XCVario.cpp
```diff
@@ -206,6 +244,48 @@ XVCDevice::PutBallast(double fraction, [[maybe_unused]] double overload,
...
+                         OperationEnvironment &env )
```
MaxKellermann, line 359: Bad indent

src/NMEA/ExternalSettings.cpp
```diff
@@ -22,6 +22,7 @@ Copyright_License {
 */
 
 #include "ExternalSettings.hpp"
+#include "LogFile.hpp"
```
MaxKellermann, line 25: No no, don't add useless log messages everywhere!

src/Device/Driver/XCVario.cpp
```diff
@@ -177,42 +203,91 @@ XVCDevice::XCV(NMEAInputLine &line, NMEAInfo &info)
...
+      protocol_version = std::min( int(value), XCV_VERSION_2 );  // switch protocol version to the minimum version both can do
+    }
+  }
   else{
     LogFormat("Unknown XCV message element received: %s", topic );
   }
   return true;
 }
+
+void XVCDevice::LinkTimeout()
```
MaxKellermann, line 218: Bad coding style

src/Device/Driver/XCVario.cpp
```diff
@@ -177,42 +203,91 @@ XVCDevice::XCV(NMEAInputLine &line, NMEAInfo &info)
...
+{
+  protocol_version = XCV_VERSION_UNKNOWN;
+  link_timer = 5;
+  nmea_channel_up = false;
+  nmea_enable = false;
+  LogFormat("LinkTimeout, set protocol version to %d", XCV_VERSION_UNKNOWN);
+}
+
+
+bool XVCDevice::EnableNMEA([[maybe_unused]] OperationEnvironment &env)
```
MaxKellermann, line 228: Bad coding style

src/Device/Driver/XCVario.cpp
```diff
@@ -32,8 +32,13 @@ Copyright_License {
 #include "util/Clamp.hpp"
 #include "Atmosphere/Pressure.hpp"
 #include "LogFile.hpp"
+#include "Operation/Operation.hpp"
```
MaxKellermann, line 35: Why include this header?

src/Device/Driver/XCVario.cpp
```diff
@@ -42,28 +47,42 @@ class XVCDevice : public AbstractDevice {
...
+  void LinkTimeout() override;
+
   bool PutCrewWeight(double weight, OperationEnvironment &env) override;
   bool PutEmptyMass(double weight, OperationEnvironment &env) override;
   bool SentNMEAItem( const char *which, int value, OperationEnvironment &env );
+
+
+private:
+  int protocol_version = XCV_VERSION_UNKNOWN;
+  int link_timer = 5;
```
MaxKellermann, line 65: This looks like you implemented a timeout detection inside this driver, but there is already `OnLinkTimeout()`, which you also use. Don't add unnecessary complexity to the driver.

src/Device/Driver/XCVario.cpp
```diff
@@ -42,28 +47,42 @@ class XVCDevice : public AbstractDevice {
...
+  bool nmea_enable = false;
```
MaxKellermann, line 66: Fields should be documented

src/ApplyExternalSettings.cpp
```diff
@@ -83,11 +82,11 @@ EmptyMassProcessTimer()
   static Validity last_empty_mass_validity;
   const NMEAInfo &basic = CommonInterface::Basic();
   PolarSettings &polar(CommonInterface::SetComputerSettings().polar);
-  Plane &plane( CommonInterface::SetComputerSettings().plane );
   if( basic.settings.empty_mass_available.Modified( last_empty_mass_validity ) ){
+    Plane &plane( CommonInterface::SetComputerSettings().plane );
     polar.glide_polar_task.SetEmptyMass(basic.settings.empty_mass);
     plane.empty_mass = basic.settings.empty_mass;
-    LogFormat("External Settings new empty mass %f", basic.settings.empty_mass );
+    PlaneGlue::WriteFile( plane );
```
MaxKellermann, line 89: Should we really rewrite the whole plane file every time the vario decides to send a new value? What if a different vario repeats the (unmodified) value every second?

MaxKellermann (changes requested):
Lots of bad coding style.
Your commit messages are long enough, but should start with a single line a la "Subsystem: brief summary", followed by an empty line, and then your long commit message. The list of commits looks very confusing if you start it with a half-sentence.

MaxKellermann:
btw. this PR does much more than just water ballast sync. For example, the whole plane mass feature - which triggers a rewrite of the plane file - requires some more thought.
It would have been simpler to merge this if you had just done the water ballast first, but now it's already an entangled big PR with everything.
Next time, don't do everything at once. It makes everything harder for everybody, including you.

iltis42:
> btw. this PR does much more than just water ballast sync. For example, the whole plane mass feature - which triggers a rewrite of the plane file - requires some more thought. It would have been simpler to merge this if you had just done the water ballast first, but now it's already an entangled big PR with everything. Next time, don't do everything at once. It makes everything harder for everybody, including you.

Thanks for the comprehensive comments! 
The coding and documentation issues are all accepted, will work in everything in the next iteration, except the "timeout" story that is actually not a link timeout, actually LinkTimeout() is implemented either, see the override here: 
https://github.com/XCSoar/XCSoar/blob/689791df51bb50ee47ea12be96f81a452cc3b0c5/src/Device/Driver/XCVario.cpp#L218 , but a timeout on application layer to detect explicitly the link is taking $PXCV protocol, as devices might have this NMEA sentences sending (and taking XCVario protocol) stopped for some reason e.g. igc file upload or by explicit by setup (to another protocol), so we need this supervision mechanism on top of the link state to start protocol negotiation, or decide this is legacy protocol version 1. 

And yes, you are right, while altering the plane and crew masses will trigger the issue from #824, alone the exchange of ballast as an absolute value will fix the trouble. So i plan in the next iteration to drop all masses except ballast from this PR. 

Then in a second PR, we may cover crew/plane masses, as the most volatile part of the plane thinking of altering crew's or a FES glider flying with or without battery, so empty mass that should not be mixed up with ballast as this cannot be dumped. Mostly wing area and polar data can be considered as static, but only for the same wingspan, extension's and winglets may or may not get attached to the same plane, so exchange is also desired.

So we are thinking about a third PR to cover exchange of all data from the plane, such as polar data, wing area, name of plane, etc, and in a first iteration this might just go down to the device as it makes more sense, when XCSoar and the device is connected, to the current plane selected in the device. Reason: XCSoar offers a large library and and good overview and the ease of entering lots of data on a touch-screen compared to enter this with just rotary and a push button on the device, users might only want to manage this in XCSoar, and not in the device. If we decide in that XCSoar is the master, altering data in the current plane file of XCSoar might be not be desired.

src/Dialogs/Plane/PlaneDetailsDialog.cpp
```diff
@@ -165,6 +169,10 @@ PlaneEditWidget::Save(bool &_changed) noexcept
   changed |= SaveValue(MAX_SPEED, UnitGroup::HORIZONTAL_SPEED,
                        plane.max_speed);
 
+  if( device_blackboard != NULL ){
+       MessageOperationEnvironment env;
+       device_blackboard->SetEmptyMass(plane.empty_mass, env);
```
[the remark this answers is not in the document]
iltis42: Yup, we need to sent this on Activate either.

src/NMEA/ExternalSettings.cpp
iltis42: All logs removed, except one when a wrong protocol element is received.

src/Device/Driver/XCVario.cpp
```diff
@@ -42,28 +47,42 @@ class XVCDevice : public AbstractDevice {
...
+  bool nmea_channel_up = false;
```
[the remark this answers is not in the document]
iltis42: moved

src/Device/Driver/XCVario.cpp
iltis42: not clear about the issue here

src/Device/Driver/XCVario.cpp
iltis42: The Timer callback needs a null environment, in order to declare this.

src/Device/Driver/XCVario.cpp
iltis42: its actually an application layer XCV protocol supervision and on top of the LinkTimeout()

src/Device/Driver/XCVario.cpp
iltis42: done

src/ApplyExternalSettings.cpp
iltis42: to be discussed in the next PR, plane masses are removed for now

src/Dialogs/Plane/PlaneDetailsDialog.cpp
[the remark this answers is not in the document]
iltis42: For now plane masses are removed from this PR

src/Device/Driver/XCVario.cpp
iltis42: fixed

src/Device/Driver/XCVario.cpp
iltis42: stray space removed after (

src/Device/Driver/XCVario.cpp
iltis42: stray spaces removed

src/Device/Driver/XCVario.cpp
iltis42: stray spaces removed

src/Device/Driver/XCVario.cpp
iltis42: stray spaces removed

src/Device/Driver/XCVario.cpp
iltis42: stray space removed

src/Device/Driver/XCVario.cpp
iltis42: stray spaces removed and repetition of constant

src/Device/Driver/XCVario.cpp
iltis42: just only one log in an error condition

src/Device/Driver/XCVario.cpp
```diff
@@ -177,42 +203,91 @@ XVCDevice::XCV(NMEAInputLine &line, NMEAInfo &info)
       info.settings.ProvideBallastLitres( value, info.clock );
     }
   }
+  else if( StringIsEqual( topic, "version") ){
```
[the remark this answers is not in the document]
iltis42: stray spaces removed

src/Device/Driver/XCVario.cpp
iltis42: stray newline removed

src/Device/Driver.hpp
iltis42: will be renamed in plane masses PR

src/Device/Driver.hpp
iltis42: yup, to be fixed in next PR for plane masses

src/Device/Driver.hpp
iltis42: fixed

src/Device/Descriptor.cpp
iltis42: will be fixed in plane masses PR

src/Plane/PlaneFileGlue.hpp
iltis42: aligned to the scope of this function with one indent

src/Dialogs/Plane/PlaneListDialog.cpp
iltis42: to be done in PR for plane masses

src/Dialogs/Plane/PlaneListDialog.cpp
iltis42: to be done in PR for plane masses

src/Plane/PlaneFileGlue.cpp
iltis42: to be done in PR for plane masses

src/Plane/PlaneFileGlue.cpp
iltis42: will be removed in PR for plane masses

src/Plane/PlaneFileGlue.cpp
iltis42: to be added in PR for plane masses

src/Plane/PlaneFileGlue.cpp
iltis42: Sure, will be changed in PR for plane masses

src/Plane/PlaneFileGlue.cpp
iltis42: Used during test, will be removed in PR for plane masses

src/Plane/PlaneFileGlue.hpp
iltis42: Sure, to be aligned in PR for plane masses

src/Blackboard/DeviceBlackboard.cpp
iltis42: to be aligned in PR for plane masses

src/Blackboard/DeviceBlackboard.hpp
iltis42: to be aligned in PR for plane masses

src/Device/Driver.hpp
iltis42: weight needs to be replaced by mass in PR for plane masses

src/Blackboard/DeviceBlackboard.cpp
iltis42: to be aligned in PR for plane masses

src/Device/Driver/XCVario.cpp
MaxKellermann: This is bad, because it means that `OnSysTicker()` can block, but it must not. By abusing a `NullOperationEnvironment`, you escape the protection against this mistake I built into the API.

src/Device/Driver/XCVario.cpp
MaxKellermann: I don't understand. This needs further explanation.

iltis42: Once the link comes up, XCVario protocol needs further negotiation on application layer what protocol to run. To be backward compatible with older device SW, there is protocol number negotiation between XCS and the XCV device, and until this hasn't been finished, we can't clearly go into full operation, so there is a second decision on top of the link status below on application layer to decide about this. Older devices might not answer the protocol version query, so after the link is up and if within a timeout of 5 seconds there is no answer, XCS will run in legacy mode with fractional ballasts. And until then, as long as the protocol version is unknown, any sending of NMEA except protocol version negotiation, must not take place.

PR #1043, opened 2022-09-09
 Water ballast synch between XCV and XCS in absolute ballast mode

iltis42:
**Brief summary of the changes**

Migrate fractional ballast exchange to absolute values in XCVario protocol

- Implement new fields for ballast exchange in litres in both directions
- Add backward compatible XCVario protocol version 2 enabling new ballast handling once discovered

**Related issues and discussions**

Closes https://github.com/XCSoar/XCSoar/issues/824
https://github.com/XCSoar/XCSoar/discussions/1017
https://github.com/iltis42/XCVario/issues/202
Continues ballast aspect from #1042 that got automatically closed after fork rebased and synced to master.


MaxKellermann:
Commit messages are still badly formatted.
Did you test this code? I have a feeling this PR cannot work.

iltis42:
> Commit messages are still badly formatted. 
What is exactly the bad format in there, can you please advise improvements to be done?
Short summary too long?
image

> Did you test this code? I have a feeling this PR cannot work.
Yes, i have tested the change on Android and also on linux, and the sync works in both directions: The ballast setting in XCV appears in XCS dialog and vice versa and i see now as well the exact values (as now the absolute ballast water value in litres get exchanged), on both sides. Water dumping runs well. Below is a trace of water dump from the XCV device (there is no logging in XCS with my 7.27 version with this commit). If you want i can add more evidence of tests.
What is exactly your concern here ?

1) Trace of XCVario protocol TX, we see protocol version negotiation and then ballast messages created by XCV after change applied from XCS. Water dump now runs down consistently to zero.
I (00:00:22.085) Protocols.cpp: Got xcv protocol version: 2
I (00:00:22.086) Protocols.cpp: sendNMEACmd: !xcv,version,2*25
I (00:00:23.349) Protocols.cpp: sendNMEACmd: !xcv,bal-water,53*5C
I (00:00:24.857) Protocols.cpp: sendNMEACmd: !xcv,bal-water,51*5E
I (00:00:26.350) Protocols.cpp: sendNMEACmd: !xcv,bal-water,49*57
I (00:00:27.851) Protocols.cpp: sendNMEACmd: !xcv,bal-water,47*59
I (00:00:29.351) Protocols.cpp: sendNMEACmd: !xcv,bal-water,45*5B
I (00:00:30.867) Protocols.cpp: sendNMEACmd: !xcv,bal-water,43*5D
I (00:00:32.881) Protocols.cpp: sendNMEACmd: !xcv,bal-water,41*5F
I (00:00:34.352) Protocols.cpp: sendNMEACmd: !xcv,bal-water,39*50
I (00:00:35.853) Protocols.cpp: sendNMEACmd: !xcv,bal-water,37*5E
I (00:00:37.353) Protocols.cpp: sendNMEACmd: !xcv,bal-water,35*5C
I (00:00:38.855) Protocols.cpp: sendNMEACmd: !xcv,bal-water,33*5A
I (00:00:40.354) Protocols.cpp: sendNMEACmd: !xcv,bal-water,31*58
I (00:00:41.855) Protocols.cpp: sendNMEACmd: !xcv,bal-water,29*51
I (00:00:43.355) Protocols.cpp: sendNMEACmd: !xcv,bal-water,27*5F
I (00:00:44.861) Protocols.cpp: sendNMEACmd: !xcv,bal-water,25*5D
I (00:00:46.356) Protocols.cpp: sendNMEACmd: !xcv,bal-water,23*5B
I (00:00:47.857) Protocols.cpp: sendNMEACmd: !xcv,bal-water,21*59
I (00:00:49.371) Protocols.cpp: sendNMEACmd: !xcv,bal-water,19*52
I (00:00:51.373) Protocols.cpp: sendNMEACmd: !xcv,bal-water,17*5C
I (00:00:52.882) Protocols.cpp: sendNMEACmd: !xcv,bal-water,15*5E
I (00:00:54.859) Protocols.cpp: sendNMEACmd: !xcv,bal-water,13*58
I (00:00:56.359) Protocols.cpp: sendNMEACmd: !xcv,bal-water,11*5A
I (00:00:57.864) Protocols.cpp: sendNMEACmd: !xcv,bal-water,9*63
I (00:00:59.876) Protocols.cpp: sendNMEACmd: !xcv,bal-water,7*6D
I (00:01:01.361) Protocols.cpp: sendNMEACmd: !xcv,bal-water,5*6F
I (00:01:02.862) Protocols.cpp: sendNMEACmd: !xcv,bal-water,3*69
I (00:01:04.366) Protocols.cpp: sendNMEACmd: !xcv,bal-water,1*6B

MaxKellermann:
> > Commit messages are still badly formatted.
> 
> What is exactly the bad format in there, can you please advise improvements to be done? Short summary too long? image

See https://github.com/XCSoar/XCSoar/pull/1042#pullrequestreview-1101873465: "Your commit messages are long enough, but should start with a single line a la "Subsystem: brief summary", followed by an empty line, and then your long commit message. The list of commits looks very confusing if you start it with a half-sentence."

> What is exactly your concern here ?

Your new fields in `struct ExternalSettings` are never initialised and never copied by `Complement()`. I cannot imagine this really works, and if it does, it does only by coincidence or luck.

iltis42:
Okay, accepted for the commit, first line is too long or might have contained a newline. Will fix that.

Regarding:
> Your new fields in `struct ExternalSettings` are never initialised 

I have added just one field, the ballast_litres and yes, the field is not initalizes as other primitive types there, hence the ballast_litres_validity is cleared what is consistent with the handling of other primitive types in this subsystem. 
https://github.com/XCSoar/XCSoar/blob/b34e3caa1b2c805e256bfe58f52cce161078c718/src/NMEA/ExternalSettings.cpp#L27
Yes i saw as well some some gap in here, but if we go for that, IMHO we should as well cover the other fields in here as well with an initializer <field> = 0.0, right ?

Regarding:
> and never copied by `Complement()`.

In Complement() i see it copied, don't get the point here:
https://github.com/XCSoar/XCSoar/blob/b34e3caa1b2c805e256bfe58f52cce161078c718/src/NMEA/ExternalSettings.cpp#L76

MaxKellermann:
> In Complement() i see it copied, don't get the point here after L76:

You're right, I don't know what I looked at, but the "missing" code really exists in this PR.

src/Device/Driver/XCVario.cpp
```diff
@@ -31,33 +31,54 @@ Copyright_License {
 #include "NMEA/InputLine.hpp"
 #include "util/Clamp.hpp"
 #include "Atmosphere/Pressure.hpp"
+#include "Operation/Operation.hpp"
 #include <math.h>
 
+#define XCV_VERSION_UNKNOWN 0  // as long device has not yet answered and no timeout
+#define XCV_VERSION_1  1       // exchange legacy format with fractional ballast in $PXCV plus CAI302 !g command
+#define XCV_VERSION_2  2       // exchange absolute ballast in liters by !xc[s|v] messages
```
MaxKellermann, line 39: Don't use macros. Either use `constexpr` variables, or in this case, an non-strictly-typed `enum` is the appropriate choice. And it shouldn't be global, but be declared inside `class XVCDevice` to avoid namespace clashes.

src/Device/Driver/XCVario.cpp
```diff
@@ -31,33 +31,54 @@ Copyright_License {
...
+
 class XVCDevice : public AbstractDevice {
   Port &port;
 
+private:
+  int protocol_version = XCV_VERSION_UNKNOWN;  // controls the
```
MaxKellermann, line 45: Why is this a signed integer? Can it be negative?

src/Device/Driver/XCVario.cpp
```diff
@@ -31,33 +31,54 @@ Copyright_License {
...
+  int link_timer = 5;                          // timer to supervise $PXCV protocol talking peer on application level
```
MaxKellermann, line 46: Why is this a signed integer? Can it be negative?
The need for this and how it exactly works needs further documentation.

src/Device/Driver/XCVario.cpp
```diff
@@ -93,13 +120,15 @@ PXCV(NMEAInputLine &line, NMEAInfo &info)
   if (line.ReadChecked(value))
     info.settings.ProvideBugs(1 - Clamp(value, 0., 30.) / 100.,
                               info.clock);
+  if(protocol_version == XCV_VERSION_1){
+    double ballast_overload;
+    if (line.ReadChecked(ballast_overload))   // Parse ballast overload value
+      info.settings.ProvideBallastOverload(ballast_overload, info.clock);
+  }else{
```
MaxKellermann, line 127: Bad coding style

MaxKellermann (changes requested):

src/Device/Driver/XCVario.cpp
iltis42: Will put this into an enum inside class XCVDevice, sure.

src/Device/Driver/XCVario.cpp
iltis42: will be the corresponding enun type later.

src/Device/Driver/XCVario.cpp
```diff
@@ -138,43 +167,98 @@ PXCV(NMEAInputLine &line, NMEAInfo &info)
...
+  if (line.ReadChecked(x) && line.ReadChecked(y) && line.ReadChecked(z))
     info.acceleration.ProvideGLoad(SpaceDiagonal(x, y, z));
 
   return true;
 }
 
+bool
+XVCDevice::XCV(NMEAInputLine &line, NMEAInfo &info)
+{
+  double value;
```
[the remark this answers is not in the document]
iltis42: 'value' will be moved into corresponding scope where used.

src/Device/Driver/XCVario.cpp
```diff
@@ -138,43 +167,98 @@ PXCV(NMEAInputLine &line, NMEAInfo &info)
...
+// Scedule sending of new protocol version 2; peer shall answer with the same message with its version
+// Upon reception of XCV's version, the protocol will be switched, atfer timeout we assume legacy version 1
+void
+XVCDevice::OnSysTicker()
+{
+  // one all us up, repeat sending protocol version every second until timeout ( 5 seconds )
+  if(link_timer && nmea_enable && protocol_version == XCV_VERSION_UNKNOWN && nmea_channel_up){
+    link_timer--;
+    if(protocol_version == XCV_VERSION_UNKNOWN){ // sent our version until negotiation is finished
+      NullOperationEnvironment env;
```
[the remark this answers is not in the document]
iltis42: We can fix this. Am i good using a static PopupOperationEnvironment like devRestart() in Device/device.cpp ?

static PopupOperationEnvironment env;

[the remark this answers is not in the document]
MaxKellermann: No, that's just a different class for the same bad trick.
`OnSysTicker()` must not block! And as long as you're using `OperationEnvironment`, no matter which implementation, you have blocking code.
`devRestart()` runs in a different context; it creates an `OperationEnvironment` for a different thread.

src/Device/Driver/XCVario.cpp
```diff
@@ -138,43 +167,98 @@ PXCV(NMEAInputLine &line, NMEAInfo &info)
...
   return true;
 }
 
+bool
+XVCDevice::XCV(NMEAInputLine &line, NMEAInfo &info)
+{
+  double value;
+  char topic[16];
+  line.Read(topic, sizeof(topic));
+  if(StringIsEqual(topic, "bal-water")){
```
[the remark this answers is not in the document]
iltis42: Will add the _T("") around the string constant, can add an if() around line.Read() to check the return before going into StringIsEqual(), something else?

src/Device/Driver/XCVario.cpp
iltis42: will add spaces around else

src/Device/Driver/XCVario.cpp
```diff
@@ -93,13 +120,15 @@ PXCV(NMEAInputLine &line, NMEAInfo &info)
   if (line.ReadChecked(value))
     info.settings.ProvideBugs(1 - Clamp(value, 0., 30.) / 100.,
                               info.clock);
+  if(protocol_version == XCV_VERSION_1){
```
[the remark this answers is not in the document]
iltis42: add space after if in L123

src/Device/Driver/XCVario.cpp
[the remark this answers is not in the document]
iltis42: First, i see that any methods to sent NMEA strings need their OperationEnvironment, e.g. port.FullWriteString(checksum, env, timeout), e.g. the ActionInterface class is using e.g. to sent MC the "MessageOperationEnvironment env; devices->PutMacCready(mc, env);", would this be the correct way to address this? 

And second, the OnSysTicker(), should not be used as risk of blocking with any Operations env, and see no other useful method from base class, so to implement a timer to sent this message, am i going in the right direction with the PeriodicTimer class, to and scedule my own OnTimer() callback e.g. like below ? 
PeriodicTimer timer{[this]{ OnTimer(); }};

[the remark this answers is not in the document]
MaxKellermann: That's just as wrong - timers must not block as well. Blocking would block the entire thread and all other events registered there, and may freeze XCSoar.
Rule of thumb: if you don't already have an `OperationEnvironment`, you must not block.
All functions which take an `OperationEnvironment` can block for an arbitrary amount of time.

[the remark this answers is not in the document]
iltis42: Okay i see, great. Taking all this lessons learnt, i could imagine to initiate the sending of protocol version in the context of EnableNMEA(OperationEnvironment &env) ? There is an OperationalEnvironment,  i am allowed to block maybe as long until the first NMEA message from device comes in (?), and until timeout i may resent (even in a loop) the protocol version to the device ?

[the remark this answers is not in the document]
MaxKellermann: You can use `Port::Write()` but it may be a partial write and you are not allowed to retry if the write was really only partial, as that will result in unbounded wait times.
You may block in `EnableNMEA()`, but I would not advise to wait until the device just happens to reply with a version number; that would not freeze XCSoar, but it would delay each connection attempt to the device.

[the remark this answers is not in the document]
MaxKellermann: And: obtaining the version number has nothing to do with enabling NMEA ...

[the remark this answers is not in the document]
iltis42: Guess there is no need to wait in the exceptional condition of a partial write. Like an UDP message on TCP, the protocol version polling can be considered as a fire and forget, no blocking and further supervision needed in lower layer, as supervision we do in the application layer by detecting the peers answer. I think it is correct to user just Write() here. 
Look into that asap, and will repeat all test on my both targets and push this, along with any other fixes for your comments next week.

src/Device/Driver/XCVario.cpp
iltis42: unsigned now here:
https://github.com/XCSoar/XCSoar/blob/2b7ae12843d31dfd27a118d97ee66919a375fbe1/src/Device/Driver/XCVario.cpp#L45
and initialized in constructor:
https://github.com/XCSoar/XCSoar/blob/2b7ae12843d31dfd27a118d97ee66919a375fbe1/src/Device/Driver/XCVario.cpp#L55

src/Device/Driver/XCVario.cpp
iltis42: https://github.com/XCSoar/XCSoar/blob/2b7ae12843d31dfd27a118d97ee66919a375fbe1/src/Device/Driver/XCVario.cpp#L45

src/Device/Driver/XCVario.cpp
iltis42: https://github.com/XCSoar/XCSoar/blob/2b7ae12843d31dfd27a118d97ee66919a375fbe1/src/Device/Driver/XCVario.cpp#L132

iltis42: else spaces also corrected

src/Device/Driver/XCVario.cpp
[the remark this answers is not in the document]
iltis42: Changed consistently to Port::Write() in protocol version resend code.

src/Device/Driver/XCVario.cpp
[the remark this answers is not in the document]
MaxKellermann: No _T as this isn't TCHAR; but a space was missing between `if` and `(`.

src/Device/Driver/XCVario.cpp
```diff
@@ -31,33 +31,59 @@ Copyright_License {
...
+private:
+
+  bool nmea_enable = false;                     // EnableNMEA() even has been received
+  unsigned int protocol_negotiation_timer = 5;  // timer to supervise $PXCV protocol talking peer on application level
+  bool xcvario_protocol_up = false;             // application level XCVario protocol $PXCV msg received
 
 public:
-  XVCDevice(Port &_port):port(_port) {}
+  XVCDevice(Port &_port):port(_port) {
+    protocol_version = XCV_VERSION_UNKNOWN;
```
MaxKellermann, line 55: C++11 style initialiser is easier to read

src/Device/Driver/XCVario.cpp
```diff
@@ -31,33 +31,59 @@ Copyright_License {
...
   bool ParseNMEA(const char *line, struct NMEAInfo &info) override;
-
   bool PutMacCready(double mc, OperationEnvironment &env) override;
   bool PutBugs(double bugs, OperationEnvironment &env) override;
   bool PutBallast(double fraction, double overload, OperationEnvironment &env) override;
   bool PutQNH(const AtmosphericPressure &pres, OperationEnvironment &env) override;
+  bool EnableNMEA(OperationEnvironment &env) override;
+  void OnSysTicker() override;
+  void LinkTimeout() override;
+  bool SentNMEAItem(const char *which, unsigned int value);
```
MaxKellermann, line 67: Typo

src/Device/Driver/XCVario.cpp
```diff
@@ -138,43 +174,98 @@ PXCV(NMEAInputLine &line, NMEAInfo &info)
...
 // For documentation refer to chapter 10.1.3 Device Driver/XCVario in mulilingual handbook: https://xcvario.de/handbuch
+// Schedule sending of new protocol version 2; peer shall answer with the same message with its version
+// Upon reception of XCV's version, the protocol will be switched, atfer timeout we assume legacy version 1
+void
+XVCDevice::OnSysTicker()
+{
+  // repeat sending protocol version every second until timeout ( 5 seconds )
+  if (protocol_negotiation_timer && nmea_enable && protocol_version == XCV_VERSION_UNKNOWN && xcvario_protocol_up) {
+    protocol_negotiation_timer--;
+    if (protocol_version == XCV_VERSION_UNKNOWN) { // sent our version until negotiation is finished
```
MaxKellermann, line 247: Another typo: `sent` -> `send` ("sent" is past tense)

src/Device/Driver/XCVario.cpp
iltis42: Fixed typo by SendNMEAItem()

src/Device/Driver/XCVario.cpp
iltis42: fixed

src/Device/Driver/XCVario.cpp
```diff
@@ -31,33 +31,59 @@ Copyright_License {
...
+    XCV_VERSION_1,        // exchange legacy format with fractional ballast in $PXCV plus CAI302 !g command
+    XCV_VERSION_2         // exchange absolute ballast in liters by !xc[s|v] messages
+  };
+  unsigned int protocol_version;  // the protocol_version controls the mode for ballast exchange with XCVario device
+
+private:
+
+  bool nmea_enable = false;                     // EnableNMEA() even has been received
+  unsigned int protocol_negotiation_timer = 5;  // timer to supervise $PXCV protocol talking peer on application level
+  bool xcvario_protocol_up = false;             // application level XCVario protocol $PXCV msg received
```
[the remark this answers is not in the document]
iltis42: removed second private: and grouped bool at the end

iltis42:
> The first commit was already cherry-picked last week, but it's still in this PR.
Had overlooked this, now rebased to master in my second forced push.

MaxKellermann:
> Had overlooked this, now rebased to master in my second forced push.

That's merged, not rebased.

MaxKellermann:
Cherry-picked without the merge commit.

iltis42:
> That's merged, not rebased.

Yes, correct have seen you cherry picked now the changeset, so IMHO the only thing i have to to is to make a clean rebase next time.

PR #1046, opened 2022-09-11
Add polar for Silene E78/E78B

bhudson225:
Brief summary of the changes
----------------------------

Added the Issoire Aviation Silene E78/E78B polar from the flight manual

None


lordfolken (changes requested):
Please do a correct commit messageLike polarstore: added... 
Updated x is not helpful when searching in tge history for a change

MaxKellermann:
No reply

Issue #992, opened 2022-08-18
XCSoarData dir not accessible due to scoped storage policy of Android 11


henrik1g:
While testing I came across not being able to access the XCSoarData directory any more after completely uninstalling and reinstalling XCSoar binary on Android.

It seems to be due to the changed storage policy by Android 11.
See here: 
https://developer.android.com/training/data-storage/use-cases

The xcsoar code mentions in LocalPath.cpp
https://github.com/XCSoar/XCSoar/blob/970470f3ab2529e27200a19306d767348a712c34/src/LocalPath.cpp#L249
>         /* the "legacy" external storage directory is writable (either
>            because this is Android 10 or older, or because the
>            "preserveLegacyExternalStorage" is still in effect) - we
>            can use it */

However that only seems to be true if there is still an old installation of XCSoar that is merely updated. For a complete new install or reinstall. XCSoarData is no longer available.

> Note: If you set preserveLegacyExternalStorage to true, the legacy storage model remains in effect only until the user uninstalls your app. If the user installs or reinstalls your app on a device that runs Android 11, then your app cannot opt out the scoped storage model, regardless of the value of preserveLegacyExternalStorage.

Is there anyway around that? Personally I hope there is, since I would hate losing my whole data directory just by accidentally uninstalling XCSoar.
If there is no way around that and scoped storage is the new way eventually, this should be made more visible, to avoid confused users.

henrik1g:
A guess that means. There is no way around it.

So to offer some help for people coming across this post copy the contents of your XCSoarData folder (not the folder itself) to the new location:

/Android/data/org.xcsoar/files/

This seems to work for me so far. However be careful since I "think" uninstalling the application will momentarily erase the whole data directory with it. (->Backup)

fronzee:
I did downgrade to 7.20 version from 7.23 due to the same issue. It somehow
would be accessable until first restart of the device, and then after I
would need to reinstall xcsoar to access the folder easily. Otherwise some
other file browser is needed and access to very deep folder structure to
put the airspace and wp files to… very annoying :(

henrik1g:
I used this App to move the files. The first time it requested some special permission to access the /Android/data/ folder, but that was self explanatory.
https://play.google.com/store/apps/details?id=com.alphainventor.filemanager&hl=de&gl=US

MaxKellermann:
> However be careful since I "think" uninstalling the application will momentarily erase the whole data directory with it.

[...]

PR #995, opened 2022-08-18
Make UsbSerial work with multi-interface devices. (e.g. FTDI4232)

henrik1g:
Brief summary of the changes
----------------------------
UsbSerial fully supports multi interface devices. This change makes UsbSerialHelper.java capable providing this functionality to the main program. Some changes were necessary since the code of UsbSerialHelper assumed a one to one relationship of device and port, which is not the case for multiport devices.
Although quite some lines needed changing, the behaviour for single Port devices should not have changed at all.
For multiport devices however entries for each port (=interface) will be available and managed with this change.

Related issues and discussions
------------------------------
https://github.com/XCSoar/XCSoar/issues/968
Closes #968

PR #1048, opened 2022-09-14
Add Android Documents Provider

brunotl:
allow access to "/Android/data/org.xcsoar/files" from android file explorer

fix issue #992

Brief summary of the changes
----------------------------

### DIFF
--- a/android/src/AllFilesDocumentsProvider.java
+++ b/android/src/AllFilesDocumentsProvider.java
@@ -1,0 +1,338 @@
+/* Copyright (c) 2018-2020, Bruno de Lacheisserie
+ * All rights reserved.
+
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+
+ * - Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ * - Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the distribution.
+ * - Neither the name of nor the names of its contributors may be used to
+ *   endorse or promote products derived from this software without specific
+ *   prior written permission.
+
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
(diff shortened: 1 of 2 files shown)


lordfolken:
hey bruno, 
i tried as per your suggestion, but was failing at the androidx dependency. But i see that you resolved that now.

brunotl:
> hey bruno, i tried as per your suggestion, but was failing at the androidx dependency. But i see that you resolved that now.

androidx dependency was only for few java annotation, I've just removed the annotations ...

MaxKellermann:
But ... this doesn't fix #992, does it? #992 is about `XCSoarData` not being accessible anymore to the XCSoar app, but your code does not make it accessible, therefore it's not a solution to #992.
Quite contrary, all it does is add a shortcut to reach into the new-style data folder from other apps.

And, correct me if I'm wrong (I don't know these APIs): it effectively removes all protection from XCSoar's data folders, allowing arbitrary untrusted apps to manipulate any of XCSoar's data files. If this is true, then this opens a rather critical vulnerability, and allows arbitrary untrusted apps to invade the user's privacy.

brunotl:
> But ... this doesn't fix #992, does it? #992 is about `XCSoarData` not being accessible anymore to the XCSoar app, but your code does not make it accessible, therefore it's not a solution to #992. Quite contrary, all it does is add a shortcut to reach into the new-style data folder from other apps.

right...

> 
> And, correct me if I'm wrong (I don't know these APIs): it effectively removes all protection from XCSoar's data folders, allowing arbitrary untrusted apps to manipulate any of XCSoar's data files. If this is true, then this opens a rather critical vulnerability, and allows arbitrary untrusted apps to invade the user's privacy.

it does not remove any protection, it only allow user to manage the file using the Android UI. trusted (or untrusted) app can't use this provider without asking to the user to select file using the native android file picker...

lordfolken:
Its true that its not a "technical solution" in the sense that you are able to reuse the XCSoarData on the "Sdcard". But you already ruled that out, as it would require to use google's high level apis that don't allow many of the operations allowed by current posix standard.

But the big problem right now is, that users cannot access their flight logs and that they cannot add custom files to xcsoar's data dir. The File manager per default blocks all direct access to this directory.

Its not "arbituary untrusted apps". They need to have the storage "manage all files" permission in order to access it. Otherwise they are locked into their application directory just like xcsoar is  locked into its. And if threy would, they would have to allow explicit access via File Manager.

android/src/AllFilesDocumentsProvider.java
```diff
@@ -0,0 +1,338 @@
...
+
+        if (Document.MIME_TYPE_DIR.equals(mimeType)) {
+            if (!file.mkdirs()) {
+                throw new FileNotFoundException("Failed to create directory with name " +
+                        displayName + " and documentId " + parentDocumentId);
+            }
+        } else {
+            try {
+                file.createNewFile();
+                file.setWritable(true);
```
lordfolken, line 310: Should this be:
public boolean setWritable(boolean writable, boolean ownerOnly) 
so therefore true,true ? 
https://www.demo2s.com/android/android-file-setwritable-boolean-writable-boolean-owneronly.html

brunotl: it's : https://developer.android.com/reference/java/io/File#setWritable(boolean)

MaxKellermann:
Amended to update the testing/AndroidManifest.xml and removed the "fixes" tag (because this PR does not fix #995 

1e45a795f0c762fd6e4194abda0cc2b9569d8553

PR #1055, opened 2022-09-21
code cleanup

fberst:
- use of pragma once
- use of [[gnu::pure]]
- use of [[gnu::const]]
- use of static constexpr

### DIFF
--- a/src/Input/InputQueue.cpp
+++ b/src/Input/InputQueue.cpp
@@ -33,9 +33,9 @@ Copyright_License {
 
 static Mutex mutexEventQueue;
 
-#define MAX_GCE_QUEUE 10
+static constexpr std::size_t MAX_GCE_QUEUE = 10;
 static int GCE_Queue[MAX_GCE_QUEUE];
-#define MAX_NMEA_QUEUE 10
+static constexpr std::size_t MAX_NMEA_QUEUE = 10;
 static int NMEA_Queue[MAX_NMEA_QUEUE];
 
 void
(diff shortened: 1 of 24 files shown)


src/ui/canvas/fb/mxcfb.h
```diff
@@ -18,8 +18,8 @@
  *
  * @ingroup Framebuffer
  */
-#ifndef __ASM_ARCH_MXCFB_H__
-#define __ASM_ARCH_MXCFB_H__
+ 
+#pragma once
```
MaxKellermann, line 22: Don't edit third-party sources. This will only increase the diff if we update it from upstream.

src/Input/InputQueue.cpp
```diff
@@ -33,9 +33,9 @@ Copyright_License {
 
 static Mutex mutexEventQueue;
 
-#define MAX_GCE_QUEUE 10
+static constexpr int MAX_GCE_QUEUE = 10;
 static int GCE_Queue[MAX_GCE_QUEUE];
-#define MAX_NMEA_QUEUE 10
+static constexpr int MAX_NMEA_QUEUE = 10;
```
MaxKellermann, line 38: Wrong type. should be `std::size_t`, not a signed integer.

test/src/harness_task.hpp
```diff
@@ -24,7 +24,7 @@
 
 #include "Task/TaskManager.hpp"
 
-#define NUM_TASKS 5
+static constexpr int NUM_TASKS = 5;
```
MaxKellermann, line 27: More bad signed integers

test/src/harness_wind.hpp
```diff
@@ -26,7 +26,7 @@
 
 #include <stdio.h>
 
-#define NUM_WIND 9
+static constexpr int NUM_WIND = 9;
```
MaxKellermann, line 29: This one is still signed. Of course, the existing code is crap and that makes your life harder than necessary; once you migrate from macros to constexpr, some things are suddenly explicit and you can't get away with the wrong types.

Issue #509, opened 2021-01-01
Add Waypoint on Map


eidolonFIRE:
XCSoar version
--------------
v6.8 - Android

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------
The ability to add waypoint / turnpoint from arbitrary location selected on map. (GPS point, not just airports)
This could be a simple `from map` button in the `waypoints editor`.

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------
Quickly add waypoints either in flight or in the field. 

Use Case: "Hey, lets flight 10mi out to that stretch of beach we can land on. Lets set a waypoint there."

Adding waypoint from raw lat/long numbers in the `waypoints editor` is completely impractical.

elgandoz:
I would love this too, it's a very common use case in paragliding, where all you want from the flight computer is to know if you can reach that hill/trigger point/landing (not on your waypoint list) or how much climb you need before committing to the jump.

Meanwhile I sorted it by using custom events:
Create a file called `custom-events.xci` in the xcsoar files folder, with the following:
```
## Change What's Here to GoTo Marker
# @see https://groups.google.com/g/xcsoar-user/c/z3_zckhSh1E
mode=pan
type=key
data=APP4
event=NearestMapItems
event=MarkLocation
event=Mode default
label=GoTo\nMarker
location=4
```
The load the file from "Config > System > Look > Events" (Expert mode must be enabled)

This will swap the "What's here" button with "GoTo Marker" when in pan mode:
- Pan when you want to go, hit "GoTo Marker"
- A dialog will open with the closest waypoint and  the first should be your marker, already selected
- Hit GoTo

It works fairly well but the downside is that requires few taps and it will fill your waypoint list with markers, which you may edit later, but they may not necessary be useful.
|1|2|3|
|-|-|-|
|![Screenshot_20221008-144058_2](https://user-images.githubusercontent.com/5859926/194686695-9acd9529-ee23-49e6-9062-01a1fd8f2166.png)|![Screenshot_20221008-144110](https://user-images.githubusercontent.com/5859926/194686846-90060062-4f66-4510-8623-0af0de7d7a06.png)|![Screenshot_20221008-144234](https://user-images.githubusercontent.com/5859926/194686724-208b5154-7956-439d-99e0-573bac193886.png)|

Does anyone know if is there a way to achieve the same by using the `AddWaypoint` event, similarly to what the default takeoff event is doing?

PR #1065, opened 2022-10-10
Pan & goto with custom events

elgandoz:
Brief summary of the changes
----------------------------

- New parameter `goto` for the `AddWaypoint` event: this creates a temporary waypoint "(goto)" ~~and sets it as navigation target.~~
- Argument for `GotoLookup` event: parameter will be looked up within the waypoints names, if provided.

```
event=GotoLookup (goto)
event=AddWaypoint goto
```

Further informations
----------------------------

The goal is to provide an easy way for the user to configure an event in order to "Pan & Go" to any point on the map, particularly useful for "improvised" XC, very common in paragliding.

https://user-images.githubusercontent.com/5859926/194804173-c9122814-68bf-4fcf-9148-afff134048cc.mp4

Sample of the `custom-events.xci` used in the above demo:
```
mode=pan
type=key
data=APP4
# event=Pan off
event=GotoLookup (goto)
event=AddWaypoint goto
event=StatusMessage Goto point on the map
label=Goto\nPoint
location=4
```
Download here: [custom-events.xci.txt](https://github.com/XCSoar/XCSoar/files/12714777/custom-events.xci.txt)
_(Remove the .txt at the end)_

This functionality is a common pattern in other freeflying computers.

A possible enhancement of this feature would be enabling the "Goto" button (and maybe the first page of Details) in the "Map element at this location" dialog.

|SeeYou Navigator|XCTrack|XCSoar _(proposed)_|
|-|-|-|
|  |  |  |

- https://groups.google.com/g/xcsoar-user/c/z3_zckhSh1E
- https://github.com/XCSoar/XCSoar/issues/509

### DIFF
--- a/src/Engine/Waypoint/Waypoints.cpp
+++ b/src/Engine/Waypoint/Waypoints.cpp
@@ -342,30 +342,34 @@ Waypoints::CheckExistsOrAppend(WaypointPtr waypoint) noexcept
 }
 
 Waypoint
-Waypoints::GenerateTakeoffPoint(const GeoPoint& location,
-                                const double terrain_alt) const noexcept
+Waypoints::GenerateTempPoint(const GeoPoint& location, const double terrain_alt,
+                             const TCHAR *misc) const noexcept
 {
   // fallback: create a takeoff point
   Waypoint to_point(location);
   to_point.elevation = terrain_alt;
   to_point.has_elevation = true;
-  to_point.name = _T("(takeoff)");
-  to_point.type = Waypoint::Type::OUTLANDING;
+  to_point.name = misc;
+  if (StringIsEqual(misc, _T("(takeoff)")))
+    to_point.type = Waypoint::Type::OUTLANDING;
+  else
+    to_point.type = Waypoint::Type::NORMAL;
   return to_point;
 }
 
 void
-Waypoints::AddTakeoffPoint(const GeoPoint& location,
-                           const double terrain_alt) noexcept
+Waypoints::AddTempPoint(const GeoPoint& location, const double terrain_alt,
+                        const TCHAR *misc) noexcept
 {
   // remove old one first
-  WaypointPtr old_takeoff_point = LookupName(_T("(takeoff)"));
-  if (old_takeoff_point != nullptr)
-    Erase(std::move(old_takeoff_point));
+  WaypointPtr old_point = LookupName(misc);
+  if (old_point != nullptr)
+    Erase(std::move(old_point));
 
-  if (!GetNearestLandable(location, 5000)) {
+  if ((StringIsEqual(misc, _T("(takeoff)")) && !GetNearestLandable(location, 5000))
+      || (!StringIsEqual(misc, _T("(takeoff)")))) {
     // now add new and update database
-    Waypoint new_waypoint = GenerateTakeoffPoint(location, terrain_alt);
+    Waypoint new_waypoint = GenerateTempPoint(location, terrain_alt, misc);
     Append(std::move(new_waypoint));
   }
 
--- a/src/Engine/Waypoint/Waypoints.hpp
+++ b/src/Engine/Waypoint/Waypoints.hpp
@@ -186,19 +186,19 @@ public:
   }
 
   /**
-   * Generate takeoff waypoint
+   * Generate a temorary waypoint with a given name.
    *
    * @return waypoint copy
    */
-  Waypoint GenerateTakeoffPoint(const GeoPoint &location,
-                                double terrain_alt) const noexcept;
+  Waypoint GenerateTempPoint(const GeoPoint &location, double terrain_alt,
+                             const TCHAR *misc) const noexcept;
 
   /**
-   * Create a takeoff point or replaces previous.
+   * Create a temporary point with a given name and replaces the previous one.
    * This modifies the waypoint database.
    */
-  void AddTakeoffPoint(const GeoPoint& location,
-                       double terrain_alt) noexcept;
+  void AddTempPoint(const GeoPoint& location, double terrain_alt,
+                    const TCHAR *misc) noexcept;
 
   /**
    * Return the current home waypoint.  May be nullptr if none is
--- a/src/Input/InputEventsTask.cpp
+++ b/src/Input/InputEventsTask.cpp
@@ -101,7 +101,12 @@ InputEvents::eventGotoLookup([[maybe_unused]] const TCHAR *misc)
   if (!backend_components->protected_task_manager)
     return;
 
-  auto wp = ShowWaypointListDialog(*data_components->waypoints, basic.location);
+  WaypointPtr wp;
+  if (StringIsEmpty(misc)) {
+    wp = ShowWaypointListDialog(*data_components->waypoints, basic.location);
+  } else {
+    wp = data_components->waypoints->LookupName(misc);
+  }
   if (wp != NULL) {
     backend_components->protected_task_manager->DoGoto(std::move(wp));
     trigger_redraw();
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -4,6 +4,7 @@ Version 7.40 - not yet released
   - FLARM: Add 100m zoom option.
   - hide mouse cursor after 10 seconds of inactivity
   - Move glider icon on thermal assist according to 30s average.
+  - add events for navigating to a temporary Goto waypoint
 * Linux
   - fix the /etc/xcsoar data directory
 * Android
--- a/src/Engine/Task/Unordered/GotoTask.cpp
+++ b/src/Engine/Task/Unordered/GotoTask.cpp
@@ -90,8 +90,9 @@ GotoTask::TakeoffAutotask(const GeoPoint& location, const double terrain_alt)
 
   auto wp = waypoints.GetNearestLandable(location, 5000);
   if (!wp)
-    wp = std::make_unique<Waypoint>(waypoints.GenerateTakeoffPoint(location,
-                                                                   terrain_alt));
+    wp = std::make_unique<Waypoint>(waypoints.GenerateTempPoint(location,
+                                                                terrain_alt,
+                                                              _T("(takeoff)")));
 
   return DoGoto(std::move(wp));
 }
--- a/src/Input/InputEventsActions.cpp
+++ b/src/Input/InputEventsActions.cpp
@@ -70,6 +70,7 @@ doc/html/advanced/input/ALL  http://xcsoar.sourceforge.net/advanced/input/
 #include "Components.hpp"
 #include "BackendComponents.hpp"
 #include "DataComponents.hpp"
+#include "Terrain/RasterTerrain.hpp"
 
 #include <cassert>
 #include <tchar.h>
@@ -656,9 +657,26 @@ InputEvents::eventAddWaypoint(const TCHAR *misc)
   if (StringIsEqual(misc, _T("takeoff"))) {
     if (basic.location_available && calculated.terrain_valid) {
       ScopeSuspendAllThreads suspend;
-      way_points.AddTakeoffPoint(basic.location, calculated.terrain_altitude);
+      way_points.AddTempPoint(basic.location, calculated.terrain_altitude,
+                              _T("(takeoff)"));
       way_points.Optimise();
     }
+  } else if (StringIsEqual(misc, _T("goto"))) {
+    const auto location = GetVisibleLocation();
+    if (!location.IsValid())
+      return;
+
+    double elevation = 0;
+    if (data_components->terrain != nullptr) {
+      const auto h = data_components->terrain->GetTerrainHeight(location);
+      if (!h.IsSpecial()) {
+        elevation = h.GetValue();
+      }
+    }
+
+    ScopeSuspendAllThreads suspend;
+    way_points.AddTempPoint(location, elevation, _T("(goto)"));
+    way_points.Optimise();
   } else {
     Waypoint edit_waypoint = way_points.Create(basic.location);
     edit_waypoint.elevation = calculated.terrain_altitude;


src/Engine/Waypoint/Waypoints.cpp
```diff
@@ -420,4 +420,6 @@ Waypoints::AddGotoPoint(const GeoPoint& location,
   Append(std::move(new_goto_point));
 
   Optimise();
+  WaypointPtr new_goto_point_ptr = LookupName(_T("(goto)"));
```
elgandoz, line 423: @lordfolken How can I pass the `new_goto_point` without performing another lookup?
Sorry about the question, but this is my very first effort in C++

src/Engine/Waypoint/Waypoints.cpp
```diff
@@ -395,3 +395,31 @@ Waypoints::AddTakeoffPoint(const GeoPoint& location,
...
   Optimise();
 }
+
+Waypoint
+Waypoints::GenerateGotoPoint(const GeoPoint& location,
+                                const double terrain_alt) const
+{
+  Waypoint goto_point(location);
+  goto_point.elevation = terrain_alt;
+  goto_point.name = _T("(goto)");
```
elgandoz, line 405: The name can be adjusted, I was after something short.

elgandoz:
Instead of binding the DoGoto within the `addWaypoint` event, I extended the `GotoLookup` event in order to lookup the argument, the waypoint name.
EG: `event=GotoLookup Airfield`
In this way we can still bind the creation and navigation of the temporary waypoint in a custom event file.

elgandoz:
Is there an event to exit Pan mode?

src/Input/InputEventsTask.cpp
```diff
@@ -119,7 +119,13 @@ InputEvents::eventGotoLookup([[maybe_unused]] const TCHAR *misc)
   if (protected_task_manager == NULL)
     return;
 
-  auto wp = ShowWaypointListDialog(basic.location);
+  WaypointPtr wp;
+  if (StringIsEmpty(misc)) {
+    wp = ShowWaypointListDialog(basic.location);
+  } else {
+    auto waypoints = &way_points;
+    wp = waypoints->LookupName(_T(misc));
```
elgandoz, line 127: This part currently break the Windows build, can anyone help me out here?
```
src/Input/InputEventsTask.cpp: In function ‘void InputEvents::eventGotoLookup(const TCHAR*)’:
src/Input/InputEventsTask.cpp:127:32: error: ‘Lmisc’ was not declared in this scope; did you mean ‘misc’?
  127 |     wp = waypoints->LookupName(_T(misc));
      |                                ^~
make: *** [build/compile.mk:108: output/WIN64/dbg/src/Input/InputEventsTask.o] Error 1
```

elgandoz: Also, does anyone know why the macos build doesn't compile?
(I thought MacOS was no longer supported by XCSoar, despite I would love to see it running).

src/Engine/Waypoint/Waypoints.cpp
elgandoz: I used `(goto)` since it was the same pattern used by `(takeoff)`, but it has 2 downside:
- the final bracket gets cut off when using 5 letters display `(goto` (same issue as `(take`)
- the arguments for the 2 related events slightly differ, which could potentially lead to confusion. 
  It would be much nicer allowing:
  ```
  event=GotoLookup goto
  event=AddWaypoint goto
  ```

lordfolken:
The osx ios pipline is a sanity check. The build failure is unrelated.

To the pr. In general the waypoint creation is imo ok. 
Although i dislike the string match, and would have used the flags like the home waypoint for selection and deletion. But given takeoff is already implemented this way...

The tie in via inputEvents feels wrong. 

Also replacing the whats here button isnt friendly towards joystick users.

I would have tied it in via the

MapItemListWidget::OnGotoClicked() function.

lordfolken:
@MaxKellermann do you agree?

elgandoz:
Hey @lordfolken, thank you for the feedback!

> Although i dislike the string match, and would have used the flags like the home waypoint for selection and deletion. But given takeoff is already implemented this way...

I agree to that, but I wanted to start by little steps, without modifying too much, by using the existing "takeoff" pattern. Maybe I could try to refactor both later by using flags (like "home")?

> The tie in via inputEvents feels wrong.

I used the input event for few reasons: 
- it was the pattern I was already using but with markers (using custom events)
- it is the (takeoff) pattern for the `gce=TAKEOFF`
- it doesn't change any existing behaviour
- it can be used and customised as the user please (for example I added the `event=Pan off` in order to switch immediately back to the map)
- when set-up like the example, it allows the minimum number of taps/clicks to achieve the feature.
- events can be overridden (for example, removed the auto-marking of the takeoff waypoint since while hang/paragliding often you don't want to land back on top of the mountain where you launched)

> Also replacing the whats here button isn't friendly towards joystick users.

It is not, but that's part of the customisation the custom events allow. The change is not enforced by this PR, the sequence posted is just an example on how to use it.
On a paraglider we have both hands occupied on the controls and I wanted to minimise the number of tap/clicks: replacing the "What's here" was my preference since I can get to the same window by tapping directly on a point. To each their own.


> I would have tied it in via the `MapItemListWidget::OnGotoClicked()` function.

That would be an extension of this ticket, as I sort of mocked up in the screenshot (although not really connected to this very PR).
I was waiting to hear opinions on this one before starting. That would be the "proper" way, although not customisable as this one. I personally would like to have them both.

lordfolken:
Hey @elgandoz this was good work. Are you ready for review?

elgandoz:
Hey @lordfolken, sorry for the long absence.

The PR was ready for review all along.

I now fixed the refactoring conflicts by using `data_components`.
Could please check if these lines follow the right approach?
-  https://github.com/XCSoar/XCSoar/pull/1065/files#diff-db7439a8b995f3f0a1251a530565dee3741b42a6413d0b264d5073a735a1b076R669-R670
- https://github.com/XCSoar/XCSoar/pull/1065/files#diff-82e8497c828b489d7468e692741536dee5c9b0da61ca3e5e4f2a02cf923ee76dR108

Unfortunately Android doesn't build, but the issue doesn't seem related with this PR but with some build system issue.

MaxKellermann:
Your first commit consists almost completely of code you copied. Adding redundant code is not acceptable.

elgandoz:
> Your first commit consists almost completely of code you copied. Adding redundant code is not acceptable.
@MaxKellermann it's obviously the same pattern as the Home waypoint. Do you have at least any suggestion on how this could be done?

elgandoz:
@MaxKellermann @lordfolken, I rebased from master and adjusted the code in order to not have duplicated code.
Now it generates and adds the "takeoff" and "goto" waypoints from the same function.
Could you please review?

Android and Windows do not build but it doesn't seem related with this PR.

lordfolken:
This now looks concise. 
At some point we are going to have to find a better internal repesentation of waypoint properties, that are not text strings. But this isnt what this pr wants.

@MaxKellermann can wevconsider merging this?

src/Engine/Waypoint/Waypoints.hpp
```diff
@@ -186,19 +186,19 @@ class Waypoints {
...
   /**
-   * Generate takeoff waypoint
+   * Generate a temorary waypoint with a given name.
    *
    * @return waypoint copy
    */
-  Waypoint GenerateTakeoffPoint(const GeoPoint &location,
-                                double terrain_alt) const noexcept;
+  Waypoint GenerateTempPoint(const GeoPoint &location, double terrain_alt,
+                             const TCHAR *misc) const noexcept;
```
MaxKellermann, line 194: What does the name "misc" mean?

MaxKellermann:
> @MaxKellermann can we consider merging this?

No, because the overall concept of this implementation is confusing and too complicated.
Look at the example configuration:
```
event=GotoLookup (goto)
event=AddWaypoint goto
```
Now explain to me what this does, and why it is written that way. Who, except for the submitter, will ever be able to come up with this configuration?

So .. what this does is (1) add a new waypoint at the current location. How do we know it's the current location? Why does it say "goto" when it's not about "goto" at all? Why doesn't it say "current location" instead? Confusing!

Then (2) it does "goto" to this new waypoint. Uhhh... which waypoint? Not the "goto" waypoint, but an arbitrary waypoint whose name just happens to be "(goto)" (yes, name with parentheses). But ... why that exact name? Because it's an undocumented implementation detail of the current code, which can change any time, for example when somebody decides to translate the name. And what happens if this feature was used twice? Then we have two waypoints with that name. Which one will be used? I have no idea.

This feature is too complicated, confusing, fragile and undocumented.

src/Engine/Waypoint/Waypoints.hpp
elgandoz: I used the the same var name from [InputEvents::eventAddWaypoint(const TCHAR *misc)](https://github.com/XCSoar/XCSoar/pull/1065/files#diff-db7439a8b995f3f0a1251a530565dee3741b42a6413d0b264d5073a735a1b076R651) for consistency, but we can easily change it to something more meaningful (like "name")

MaxKellermann: That's a generic API signature, and the name has no special meaning there, other than "it's some string that the user specified in the file". The function you designed is not a generic API, but a very specific function. It's a different context. Saying you just copied the name is a rather lame excuse for not coming up with a meaningful name (or with API documentation).

elgandoz:
@MaxKellermann 
> Now explain to me what this does, and why it is written that way. Who, except for the submitter, will ever be able to come up with this configuration?

I don't think it's that different from what currently already happens in the default.xci [here](https://github.com/XCSoar/XCSoar/blob/ae1fa16ba31e5b78c3d27b25780716fdfa388392/Data/Input/default.xci#L47) and [here](https://github.com/XCSoar/XCSoar/blob/ae1fa16ba31e5b78c3d27b25780716fdfa388392/Data/Input/default.xci#L1054-L1056) 


> So .. what this does is (1) add a new waypoint at the current location. How do we know it's the current location? Why does it say "goto" when it's not about "goto" at all? Why doesn't it say "current location" instead? Confusing!

I am a bit confused by this statement... it creates a waypoint under the sight cross, where the user would like to "go to" (like the [event/label name](https://github.com/XCSoar/XCSoar/blob/ae1fa16ba31e5b78c3d27b25780716fdfa388392/Data/Input/default.xci#L1054-L1056)).
I'm not sure what you refer to with "current location", but that would make me think at the current location of the glider.


> Then (2) it does "goto" to this new waypoint. Uhhh... which waypoint? Not the "goto" waypoint, but an arbitrary waypoint whose name just happens to be "(goto)" (yes, name with parentheses). But ... why that exact name?

As I already explained in [a previous comment](https://github.com/XCSoar/XCSoar/pull/1065#issuecomment-1277022033), I used that parenthesis pattern only because the same arbitrary name is used for "takeoff" (displayed `(takeoff)`, or better `(take` using default settings), and I wanted to maintain a consistent behaviour with what's already happening.
I would rather use `goto` without parenthesis, it's a very easy change, I'm all for it.

> Because it's an undocumented implementation detail of the current code, which can change any time, for example when somebody decides to translate the name.

But isn't it the same for `(takeoff)`? If translated that in their own .xci is not going to work.
Those should be considered special waypoint labels.

> And what happens if this feature was used twice? Then we have two waypoints with that name. Which one will be used? I have no idea.

If used twice it removes the previously created waypoint with the same name, just like it happens for `(takeoff)`.
This would happen only for the labels `takeoff` and `goto`. If the user uses another arbitrary name a regular waypoint would be created (more than once if it already exist), but I believe that is the current behaviour already.

About the documentation, this is lacking and it took me a moment to understand what the current code is doing (see #509).
The goal of this PR is to improve the Event, by modifying as little as possible of the current behaviour. I would be happy to document the feature, including currently undocumented events, if this is getting accepted.

I previously [asked for recommendation](https://github.com/XCSoar/XCSoar/pull/1065#issuecomment-1772913291) on what can be done better but I received none.

src/Engine/Waypoint/Waypoints.hpp
elgandoz: sure a _lame excuse_... any other preferred name, or `name` is fine?

elgandoz:
Hey @lordfolken, I saw you marked this for the next release.

I was wondering if you still want me to:
- rename `goto` to something better
- as suggested by Max, change the `misc` parameter name for `GenerateTempPoint()`. Would `name` be better?
- uniform the name of the `goto` waypoint when added by `AddTempPoint()`, without the parenthesis?
   - I used the paranthesis because that was what  was happening for `(takeoff)` but maybe that's not necessary in this case, simplifying the call in .xci: `event=GotoLookup goto, event=AddWaypoint goto`
   
Also, as separate PR, I can have a look into using `MapItemListWidget::OnGotoClicked()`, the idea for this one was to use the custom action, like someone pointed out in that [Google forum post](https://groups.google.com/g/xcsoar-user/c/z3_zckhSh1E)

lordfolken:
> Hey @lordfolken, I saw you marked this for the next release.
Technically its the release after that. The strategy is to stabilize and fix all the current bugs in 7.42. for the moment. I also like the feature's functionality and i think it should be in XCSoar.

> I was wondering if you still want me to:
> 
>     * rename `goto` to something better

>     * uniform the name of the `goto` waypoint when added by `AddTempPoint()`, without the parenthesis?

The string as such is fine with me to be "goto". I'm also fine with the parentheses, to indicate to the user that this waypoint is generated.

* Its my plan to make the default waypoint name on the map the Shortname, and if not available the 5 first letters of the full name. So (goto) would be fully displayed as long as you write it into the shortname string of the waypoint. 

The main problem i see with extending the AddTempPoint()  function, is that we currently have no way of differentiating between a user provided waypoint and a generated one. A string match isn't enough to isolate this from the user playload. (this was already a problem with the current implemenation...) 

There should be a flag/enum in Waypoint.hpp to indicate that this waypoint is programmatically generated. 
Potentially the flag would be saved for each waypoint and, it could be for example an enum. This one should differentiate between files that have been loaded, the user.cup waypoints, takeoff and goto waypoints. This way we can handle all the use cases without string matches. 

>     * as suggested by Max, change the `misc` parameter name for `GenerateTempPoint()`. Would `name` be better?
i'd go for name as this is also in the waypoint struct. 

    
>       * I used the parenthesis because that was what  was happening for `(takeoff)` but maybe that's not necessary in this case, simplifying the call in .xci: `event=GotoLookup goto, event=AddWaypoint goto`

> Also, as separate PR, I can have a look into using `MapItemListWidget::OnGotoClicked()`, the idea for this one was to use the custom action, like someone pointed out in that [Google forum post](https://groups.google.com/g/xcsoar-user/c/z3_zckhSh1E)

Pan ultimate that should be the goal.

elgandoz:
Rebased and renamed variable

lordfolken:
Hey, there, are you still working on this? I'd love it for v7.45

lordfolken:
@coderabbitai full review

PR #1066, opened 2022-10-13
Waypoint Editor dialog: load saved waypoint type

elgandoz:
Brief summary of the changes
----------------------------
On the editing dialog for an individual waypoint, the _unchanged_ value for _Type_  it can be only  "Airport", "Landable" or "Turnpoint" (catch-all).

This PR aims to fix this by displaying the correct value for the "Type" waypoint.

### DIFF
--- a/src/Dialogs/Waypoint/dlgWaypointEdit.cpp
+++ b/src/Dialogs/Waypoint/dlgWaypointEdit.cpp
@@ -104,7 +104,7 @@ WaypointEditWidget::Prepare(ContainerWindow &, const PixelRect &) noexcept
            0, 30000, 5, false,
            UnitGroup::ALTITUDE, value.elevation);
   AddEnum(_("Type"), nullptr, waypoint_types,
-          value.IsAirport() ? 1u : (value.IsLandable() ? 2u : 0u),
+          (unsigned)value.type,
           this);
 }
 


src/Dialogs/Waypoint/dlgWaypointEdit.cpp
```diff
@@ -104,7 +104,7 @@ WaypointEditWidget::Prepare(ContainerWindow &, const PixelRect &) noexcept
            0, 30000, 5, false,
            UnitGroup::ALTITUDE, value.elevation);
   AddEnum(_("Type"), nullptr, waypoint_types,
-          value.IsAirport() ? 1u : (value.IsLandable() ? 2u : 0u),
+          (unsigned)value.type,
```
elgandoz, line 107: The value type has some offset, can anyone help me out here?
| Saving | Loading |
|:-|:-|
| Turnpoint | Turnpoint |
| Airport | Airport |
| Landable | Landable |
| Mountain pass | Mountain top |
| Mountain top | Mountain pass |
| Transmitter mast | Power Plant |
| Tower | Tunnel |
| Tunnel | Tower |
| Bridge | Transmitter mast |
| Power plant | Bridge |
| Vor | Dam |
| NDB | Castle |
| Dam | Intersection |
| Castle | Marker |
| Intersection | Control point |
| Marker | NDB |
| Control Point | PG Take off |
| PG Take off | PG LZ |
| PG LZ | Turnpoint |

MaxKellermann: The whole enum list is broken, always has been, see commit 4275c46bc6531448bca403e595efe4f2e16f58d2 @lordfolken!
Why the ... does it use hard-coded integers instead of `Waypoint::Type` values?

lordfolken: Argh. This is indeed crap. There should only be one waypoint defintion and both load and save should refeer to it.

PR #1069, opened 2022-10-20
Add support for Kobo Libra 2

ubx:
Brief summary of the changes
----------------------------

This PR adds support for the Kobo Libra 2. No Bluetooth support yet.

### DIFF
--- a/src/Kobo/System.cpp
+++ b/src/Kobo/System.cpp
@@ -72,7 +72,7 @@ KoboReboot()
 {
 #ifdef KOBO
   /* CLARA_HD requires the -f option to for reboot to work */
-  if  (DetectKoboModel() == KoboModel::CLARA_HD)
+  if  (DetectKoboModel() == KoboModel::CLARA_HD || DetectKoboModel() == KoboModel::LIBRA2)
   {
     return Run("/sbin/reboot", "-f");
   }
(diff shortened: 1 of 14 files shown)


MaxKellermann:
A PR branch should not contain merge commit - please clean up the branch.
There are linker failures in the CI.

src/Kobo/System.cpp
```diff
@@ -267,9 +278,11 @@ bool
 KoboWifiOff()
 {
 #ifdef KOBO
+  const char *interface =  (DetectKoboModel() == KoboModel::LIBRA2) ? "wlan0" : "eth0";
```
MaxKellermann, line 281: There are lots of copies of this one, but we shouldn't have redundant code.

ubx:
Build failure on Linux builds:
```
src/Kobo/Model.hpp:27:2: error: #error This header is only for Kobo builds
```

I'm confused, either we do not include the Kobo Menu in the Linux build, or we allow all necessary files, including Model.hpp.

MaxKellermann:
> Build failure on Linux builds:

7b5fe227e40e1324ca3e86fc71132a9f8cf88d3d turned the linker failure into a compiler failure, so you see the mistake earlier.

> I'm confused, either we do not include the Kobo Menu in the Linux build, or we allow all necessary files, including Model.hpp.

Neither. (Much of) KoboMenu is being developed and debugged on Linux, but obviously the really Kobo-specific things cannot be done, and `Kobo/Model.hpp` must not be used. If your PR adds code which uses more of `Kobo/Model.hpp`, it must be protected with `#ifdef KOBO`.

MaxKellermann:
7bc9dd41ecc61ca16ea7e7c34d2f2666ab895f1c

ubx:
Max, thank you for helping to merge the PR.

Issue #1070, opened 2022-10-22
Polar list


silverstate64:
XCSoar versions having and not having the problem
-------------------------------------------------

unknown - haven't had to create a new plane/polar in a while

Expected behavior
-----------------

be able to select an Arcus M polar

Actual behavior
---------------

no polar

Steps to reproduce the behavior
-------------------------------

Do you have any idea what may have caused this?
-----------------------------------------------

Do you have an idea how to solve the issue?
-------------------------------------------

fraktal-sb:
use "Config Plane New/Edit" and "Polar Para EN B/DHV12" and adapt it to your needs

fberst:
In the polar list there is already a polar for the Arcus, the Arcus M should be the same with a higher dry mass (which the user can easily adjust)

When i remember right there isn't any polar published by the manufacturer of these planes. If you have access to a distinct polar for the Arcus M please share it, than we can add it to the list. Otherwise there is nothing we can do about this.....

silverstate64:
Thanks for the responses.
I do have some Polar numbers for the Arcus M but i'm unsure of their origin. Below are the numbers for selecting the Arcus Polar in XCSoar and the Polar numbers I have. 
I've been using my numbers for 3 years.

Arcus select in XCSoar
Polar V     35kt              58kt              89kt
Polar W   -0.91kt          -1.3kt           -3.89kt

Reference Mass 761 lb

what i have:
Polar V       64kt             86kt             113kt
Polar W    -1.30kt         -2.06kt         -3.89kt

Reference Mass 1653 lb

fberst:
> Arcus select in XCSoar 
> Polar V 35kt 58kt 89kt >
> Polar W -0.91kt -1.3kt -3.89kt
> 
> Reference Mass 761 lb

That's obviously completely wrong....
The values in the xcsoar PolarStore.cpp:
` _T("Arcus"), 700, 185, 110, -0.64, 140, -0.88, 180, -1.47, 15.59, 50.0, 120, 430 }` (values in kg, km/h and m/s)

When i load it i get the exakt same wrong results as silverstate64. (only on android)
So it seems we have some bug when loading the values from the PolarStore.
And it seems we already have an reported issue for that: #1062

lordfolken:
So this has been imlemented in #1026

PR #1075, opened 2022-11-04
add Arcus M and Silence E78 polars

fberst:
- add Arcus M polar, data provided by [silverstate64](https://github.com/silverstate64) in issue #1070
- add Silence E78 polar, duplicate to pull request #1046, because original author didn't respond to fix his commits


MaxKellermann:
Do we really need separate Arcus and Arcus M polars?

MaxKellermann:
If it's only about usability, maybe yes, but then we shouldn't have two polars in the source code, but instead derive one from the other (at compile time). Redundant data is always a maintenance burden.

mlep:
Indeed, not being a programmer, my comment was only about usability.

hjr:
Your provided Arcus M polar is slightly different to what I got from Bernd Weber. Where do you have the polar from? Otherwise we introduced the empty weight attribute to the plane configuration. There is now merely no good reason to have two polars in the store, just because of mass differences. Eg we removed all two seater duplicates because of that.
ArcusPolars

hjr:
You might want to split the pull request of both polars to get the Silence in straight forward. (Or strip the Arcus M commit)

fberst:
I agree, we should merge the silence polar and close this pull request.

PR #1078, opened 2022-11-15
GaugeFLARM: Show only when traffic is within 4Km

### LINKED ISSUE #481: FLARM radar display only when traffic is within 2km
XCSoar version
--------------

6.8.16

Currently if enabled, the 2km ranged FLARM radar overlay displays all traffic targets coming in via NMEA 0183 $PFLAA sentences, regardless of distance. For situational awareness of close targets this is an excellent display. If the traffic source is sending targets greater than 2km distance, the FLARM radar overlay still shows targets, just bunched up around the edge.

The FLARM radar page should pop up when traffic is present at less than 2km. Other targets that are greater than 2km do not need to be decluttered from the map, just the map itself should appear and disappear based on the closest target being within 2km.

When multiple traffic targets are present the FLARM map becomes cluttered and less useable. It's fantastic for close-in awareness. There is value in seeing targets that are more than 2km away for situational awareness or tracking purposes. Looking for gliders ahead that are circling (and climbing), or watching arrivals/departures at a possible landout airport.
However, the FLARM map becomes cluttered as it populates every single NMEA 0183 $PFLAA target on the outside edge of the map. See: https://i.imgur.com/0NbtWO8.jpg or https://i.imgur.com/rWMa3PV.jpg (ignore the orange alert level 1, I set that manually)

Logic:
If any traffic in your traffic list relativeEast or relativeNorth is within 2,000m (positive or negative), then display the FLARM map. If not, hide it.

For users who don't want this functionality it can be enabled or disabled by a config at Gauges > FLARM, Other > FLARM radar.
Currently FLARM radar has **on** or **off**. It can be modified to include **on**, **on** **within 2km**, **off**.

lordfolken:
A proposal to implement #481
Presently when an ADS-B capable receiver is hooked up, such as PowerFLARM, the GaugeFLARM is always on display, as ADS-B displays traffic from 200km away.

This change limits that to a 4Km range.

**Note:** This only affects the FLARM Gauge. All traffic is still displayed on the map as before with the distance configured in the FLARM device.

Some Questions:
 * Do we need to make this distance configurable? 4Km is the typical Flarm Range
 * The gauge seems to show 2k max.
 * The implementation recalculates the traffic targets on every GPS Update. Is this efficient enough?

This can be simulated without a FLARM, by changing the range to for example 2000m, and setting a tcp-client to 4353 localhost.
```
./output/UNIX/bin/FeedNMEA tcp 4353 < ./test/data/driver/FLARM/pflaf01.nmea
```
(recording with 2000M filtering)

https://user-images.githubusercontent.com/407371/201847606-ffabe54e-fd4d-49e5-b77a-5dfef9f230ee.mp4

### DIFF
--- a/src/FLARM/List.cpp
+++ b/src/FLARM/List.cpp
@@ -40,3 +40,10 @@ TrafficList::FindMaximumAlert() const noexcept
 
   return alert;
 }
+
+bool
+TrafficList::InCloseRange() const noexcept
+{
+  return std::any_of(list.begin(), list.end(), [](const auto &traffic)
+    { return traffic.distance < (RoughDistance)4000; });
+}
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -14,6 +14,8 @@ Version 7.29 - not yet released
 * Polars
   - Add LS-5 polar
 * fix exchange frequencies crash when no frequency was set
+* user interface
+  - show FLARMGauge only when traffic is within 4Km
 
 Version 7.28 - 2022/10/29
 * data files
--- a/src/FLARM/List.hpp
+++ b/src/FLARM/List.hpp
@@ -199,6 +199,11 @@ struct TrafficList {
   constexpr unsigned TrafficIndex(const FlarmTraffic *t) const noexcept {
     return t - list.begin();
   }
+
+  /**
+   * Is set if traffic is present and closer than 4Km.
+   */
+  bool InCloseRange() const noexcept;
 };
 
 static_assert(std::is_trivial<TrafficList>::value, "type is not trivial");
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1079,6 +1079,9 @@ MainWindow::UpdateTrafficGaugeVisibility() noexcept
     if (HasDialog())
       return;
 
+    if (!flarm.traffic.InCloseRange())
+      return;
+
     if (!traffic_gauge.IsDefined())
       traffic_gauge.Set(new GaugeFLARM(CommonInterface::GetLiveBlackboard(),
                                        GetLook().flarm_gauge));


lyusupov:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[**FLARM CONFIGURATION SPECIFICATION**](https://flarm.com/wp-content/uploads/man/FTD-014-FLARM-Configuration-Specification.pdf)


image

DanD222:
> [**FLARM CONFIGURATION SPECIFICATION**](https://flarm.com/wp-content/uploads/man/FTD-014-FLARM-Configuration-Specification.pdf)
> 
> image

I’d prefer to not limit the viewing options at FLARM level, if that’s what you’re inferring - i.e. I’d still like to see distant aircraft on the map, but I’m OK with them appearing on the black FLARM gauge only within 4km.

lordfolken:
> I’d prefer to not limit the viewing options at FLARM level, if that’s what you’re inferring - i.e. I’d still like to see distant aircraft on the map, but I’m OK with them appearing on the black FLARM gauge only within 4km.

That's this. There is no configuration change in the Flarm. The map targets are unaffected and displayed at whatever distance is configured. This change only looks at all targets and displays the GaugeFLARM the superimposed small radar, whenever there is a target closer than 4km.

The reason for this is to make the pilot aware, that there is something in his vicinity. This was the original purpose of the design. But this has now been upset by ads-b targets and also powerflarm's increased reception at much greater distances.

DanD222:
> 
> That's this. There is no configuration change in the Flarm. The map targets are unaffected and displayed at whatever distance is configured. This change only looks at all targets and displays the GaugeFLARM the superimposed small radar, whenever there is a target closer than 4km.
> 
> The reason for this is to make the pilot aware, that there is something in his vicinity. This was the original purpose of the design. But this has now been upset by ads-b targets and also powerflarm's increased reception at much greater distances.

I understand - I was replying to lyusupov’s post with the FLARM RANGE parameter screenshot; I probably should have made that more clear.

src/FLARM/List.cpp
```diff
@@ -40,3 +40,16 @@ TrafficList::FindMaximumAlert() const noexcept
...
+TrafficList::InCloseRange() const noexcept
+{
+  const FlarmTraffic *inrange = NULL;
+
+  for (const auto &traffic : list) {
+    if ((int)traffic.distance < 4000)
+      inrange = &traffic;
+  }
+
+  return inrange;
```
MaxKellermann, line 54: Why does this return a pointer? Your API docs suggest it should be a `bool`, and its return value is used as a boolean. Imagine how much simpler this function body could be:
```
  return std::any_of(list.begin(), list.end(), [](const auto &traffic){ return traffic.distance < 4000; });
```

lordfolken: Applied as you suggested. Needed a cast for RoughDistance.

lordfolken:
@MaxKellermann can we merge this?

PR #1093, opened 2023-01-02
docker: Reorder permission change command of install scripts

psa-avat:
Brief summary of the changes
----------------------------
fix access rights on install scripts 

### DIFF
--- a/ide/docker/Dockerfile
+++ b/ide/docker/Dockerfile
@@ -5,10 +5,9 @@ ENV DEBIAN_FRONTEND=noninteractive
 # Install build dependencies
 RUN apt-get update && apt-get install -y python3 vim ccache python3-pkgconfig gdb && apt-get clean
 COPY ./provisioning/install-debian-packages.sh /root/install-debian-packages.sh
-RUN /root/install-debian-packages.sh && apt-get clean
+RUN chmod 0755 /root/install-debian-packages.sh && /root/install-debian-packages.sh && apt-get clean
 COPY ./provisioning/install-android-tools.sh /root/install-android-tools.sh
-RUN chmod 0755 /root/install-android-tools.sh /root/install-debian-packages.sh
-RUN /root/install-android-tools.sh && apt-get clean
+RUN chmod 0755 /root/install-android-tools.sh && /root/install-android-tools.sh && apt-get clean
 
 COPY ./docker/bin/* /usr/local/bin/
 WORKDIR /usr/local/bin


MaxKellermann:
I don't understand what problem this solves because the commit message doesn't explain anything.

ide/docker/Dockerfile
```diff
@@ -5,9 +5,10 @@ ENV DEBIAN_FRONTEND=noninteractive
 # Install build dependencies
 RUN apt-get update && apt-get install -y python3 vim ccache python3-pkgconfig gdb && apt-get clean
 COPY ./provisioning/install-debian-packages.sh /root/install-debian-packages.sh
-RUN /root/install-debian-packages.sh && apt-get clean
 COPY ./provisioning/install-android-tools.sh /root/install-android-tools.sh
-RUN chmod 0755 /root/install-android-tools.sh /root/install-debian-packages.sh
+RUN chmod 0755 /root/install-debian-packages.sh /root/install-android-tools.sh
```
MaxKellermann, line 9: How does changing the order of file names solve anything?

psa-avat:
You just need to chmod before executing otherwise you got an access deny

Le lun. 2 janv. 2023 à 22:16, Max Kellermann ***@***.***> a
écrit :

MaxKellermann:
So this doesn't actually "fix acces rights on install scripts", because they are already being fixed - your commit changes nothing about access rights. It just changes the order of commands. But your commit message is bad and misleading, and your diff is obscured by an unnecessary change.

psa-avat:
My commit fixes an access right error message that you got while executing
the original version

Le lun. 2 janv. 2023, 22:36, Max Kellermann ***@***.***> a
écrit :

MaxKellermann:
Did you understand that your commit is bad for 2 reasons, and those 2 reasons prevent it from getting merged?
I doubt you understood that, because your last reply makes little sense in the context of my last comment.

psa-avat:
Hi

 My commit makes the Dockerfile working for building a dockerimage, your's does not !
 Your dockerfile fails with an access denied error while trying to execute "/root/install-debian-packages.sh". This is due to the fact that the chmod is placed after the RUN. 

If you don't want my **fix**, it wont change my life, but it can save time to someone else !
It's not really rocketscience but avoid spending 1h to understand why a provided dockerfile is not properly working.

So it's up to you to merge or not this minor change. 

I'am preparing a branch with the integration of the new kobo model Clara HD 2e. I have to fully test it before pushing it. 
I'll make a single commit.

If you don't want this model supported, tell me now, I'll will save time and keep it for myself only.

Patrick

DanD222:
Patrick,

My guess is that what Max is trying to say is that you should change your commit message to say something like "Reorder permission change command of install scripts", and then also state the reason why this is necessary in the commit message (that is, not in the pull request comments).

I think Max takes issue with the fact that strictly speaking, access rights are being **_fixed_** by the chmod command, whereas what you're doing is changing the **_order_** the permission change (i.e. fix) is being applied in.

I once worked with a developer who was very particular about the way a question was posed to him and wouldn't give you a good answer unless you posed a good question, but, while a little frustrating at times, I found the experienced pretty educational and had a lot of respect for him :)

ide/docker/Dockerfile
psa-avat: You just need to chmod before executing otherwise you got an access deny

psa-avat:
Dan

  According to your advice I changed my comment. 
  But strictly speaking I think that I fixed a non working Dockerfile by reordering some instructions. 
  I'm not going to loose energy on this point anymore.

  Anyway, thanks for your comment.
Patrick

lordfolken:
The docker image is in use by the pipelines so its functional.
Can it be that the files in ide/provisioning/ are not marked as executable in your checkout, and you are rebuilding the image?
Anyway, my update to your commit should prevent that situation from arising again.

MaxKellermann:
Why is this chmod even necessary at all?
And if it is, why not `sh /root/install-debian-packages.sh` without chmod?
I don't get it at all. Is Docker really so complicated? (I've never used Docker.)

lordfolken:
> Why is this chmod even necessary at all? And if it is, why not `sh /root/install-debian-packages.sh` without chmod? I don't get it at all. Is Docker really so complicated? (I've never used Docker.)

The chmod isn't necessary. The scripts are already marked as executable in git. That is why it worked in most of our use-cacses. 
It may be that this person is building on a  filesystem that doesn't support unix-style permissions.

I thought about leaving the chmod away, and to do it as you suggested, why i didn't:
 * one is a shell script the other is a bash script according to the shebang. So it would have to be two different calls in the docker file. If anything there changes it would have to be changed in 2 different places (script and dockerfile). So i'd rather prevent the breakage by making sure that the script is in charge of picking its interpreter. Hence the chmod.

concerning docker//container images. They are completely brain dead. Its literally tech from the 80s.
Building a docker image is downloading a chroot, executing each RUN or COPY statement in it and making an incremental tar after each operation. That's it.  You can unpack container images with tar -xf. 

Running a container image is basically mounting all the layers on top of it each other and doing a chroot to it. Some instructions are there for bind mounting stuff in and out of the image, or creating a port forward to the image.

MaxKellermann:
> It may be that this person is building on a filesystem that doesn't support unix-style permissions.

Not a supported use case, I'd say. Away with those confusing complicated kludges. Don't let software design be led by broken systems.

MaxKellermann:
> Don't let software design be led by broken systems.

And IF we ever need to support a broken system, there MUST be a comment explaining the workaround. No obscure unexplained code!

psa-avat:
The issue appeared when I launched a Dockerbuild with this dockerfile from
a linux console running on my Synology NAS.
At first I thought I had not enough privileges to execute the build. I
restarted with a full root account and got the same result.
Reading the dockerfile, I saw that a chmod was made after the execution of
the file.
I put the chmod before to fix my issue and it worked fine.

In conclusion for my point of view, If the chmod is necessary, it should be
run before the exec. If not it should be removed.

My modification has not been made by magic ! It's driven by the access
denied error returned.





Le jeu. 5 janv. 2023 à 18:24, Max Kellermann ***@***.***> a
écrit :

MaxKellermann:
This PR is bad, it adds a bad kludge to work around issues on the submitter's computer which were not caused by XCSoar.

PR #1095, opened 2023-01-05
Windows: Tracking - the SkyLines loop doesn't never finished

August2111:
Brief summary of the changes
----------------------------

In case of the Windows-Target (WIN64 or PC)  their seems to be a blocking call to the skylines server, it is enabled and connected.
On closing XCSoar it hangs in this loop and never came back
The only way to stop XCSoar is to destroy it...

For Windows there is a NonBlocking property needed in the socket Open function (all other OS are doing this with a flag in the callíng function 'Read' or 'Write' itself).
In my opinion, this should solve the problem without any side effects - at least XCSoar behaves properly again under Windows.

This issue also has side effects on enabling and disabling the NMEA ports, which is also fixed with this commit.

### DIFF
--- a/src/Tracking/SkyLines/Client.cpp
+++ b/src/Tracking/SkyLines/Client.cpp
@@ -75,6 +75,10 @@ SkyLinesTracking::Client::Open(SocketAddress _address)
     handler->OnSkyLinesReady();
   }
 
+#ifdef _WIN32
+  GetSocket().SetNonBlocking();
+#endif
+
   return true;
 }
 


MaxKellermann:
I don't get it. This is a datagram socket which cannot ever block.
Also I don't get why you want to include `Protection.hpp`.

August2111:
The bug is easy to reproduce:
* Start on Windows the XCSoar-Win64-Target (for example)
* Set the complete SkyLines  options (f.e. interval = 5s, friends On, near traffic Off ) - with a valid key
* Set the feater On
With this setting you cannot close the program: You have to kill it (and of course after every restart too!)

The (wrong) header 'Protection.hpp' came from a erlier solution, excuse me, but  it was a hard thing to find the root cause for this...
Linux have the Flag SOCK_NONBLOCK, but on Windows you have to find an alternative...

August2111:
The 2nd commit should prevent an errors in the check list of the pull request only, has nothing to do with the actual intention of the PR

The next clang error (build-macos) related to 'sprintf'  is to solve in an extra additional PR because there are a lot of deprecated sprintf in the code...

MaxKellermann:
Right now, I don't want to reproduce this. I want an explanation. You explained nothing, you only posted a workaround for something you don't understand.

August2111:
> Right now, I don't want to reproduce this. I want an explanation. You explained nothing, you only posted a workaround for something you don't understand.

Ok, I will close this PR - and will open an new issue, maybe someone who knows more about it than me will take care of it ;-)

PR #1101, opened 2023-01-29
o not truncate wifi list if a scan result line …

ubx:
Brief summary of the changes
----------------------------

Fix the problem if a wifi entry has an empty ssid. The list is not any more truncated.


MaxKellermann:
This `break` is here because XCSoar assumes there's only unparseable garbage left. This was an explicit decision. Instead of removing the `break`, we should better check why parsing fails - and fix the problem.

MaxKellermann:
c0772da895a8e662ce75f87ab9d1688d4a42812c
Merged with two changes: 1. using `clear()` instead of copying an empty string which is faster; 2. using `!empty()` instead of `length()>0` which is faster

PR #1104, opened 2023-01-29
Show Waypoint Type in WaypointInfoWidget

lordfolken:
2023-01-29-230850_958x258_scrot

### DIFF
--- a/src/Dialogs/Waypoint/WaypointInfoWidget.cpp
+++ b/src/Dialogs/Waypoint/WaypointInfoWidget.cpp
@@ -108,6 +108,8 @@ WaypointInfoWidget::Prepare(ContainerWindow &parent,
   if (!waypoint->comment.empty())
     AddMultiLine(waypoint->comment.c_str());
 
+  AddReadOnly(_("Waypoint type"), nullptr, Waypoint::TypeMap[waypoint->type].c_str());
+
   if (waypoint->radio_frequency.Format(buffer.buffer(),
                                       buffer.capacity()) != nullptr) {
     buffer += _T(" MHz");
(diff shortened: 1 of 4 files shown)


lordfolken:
Two questions:
 * i had to add language.hpp in Engine/Waypoints.cpp I'm not sure whether drawing that in there is not violating layering?
 * Potentially the ShortName and Type field should be exchanged for better readability.

lordfolken:
Converting to draft. Windows doesn't compile. I have a proper implementation for that, but it ends up with an assertion: 
src/Language/Language.cpp Line:91 
Epxression language_allowed

lordfolken:
I saw that there was another similar construct in dlgWaypointEdit.cpp will try to refactor.

src/Engine/Waypoint/Waypoint.cpp
```diff
@@ -44,3 +45,26 @@ Waypoint::Project(const FlatProjection &projection) noexcept
   flat_location_initialised = true;
 #endif
 }
+
+std::map<Waypoint::Type, tstring> Waypoint::TypeMap = {
```
MaxKellermann, line 49: What could be a simple plain array, initialised at compile-time into `.rodata`, is now a runtime-initialised `std::map` with lots of heap allocations, with a complicated lookup procedure.

src/Dialogs/Waypoint/WaypointInfoWidget.cpp
```diff
@@ -108,6 +108,8 @@ WaypointInfoWidget::Prepare(ContainerWindow &parent,
   if (!waypoint->comment.empty())
     AddMultiLine(waypoint->comment.c_str());
 
+  AddReadOnly(_("Waypoint type"), nullptr, Waypoint::TypeMap[waypoint->type].c_str());
```
MaxKellermann, line 111: This will crash if the type happens not to be in that `std::map`.

Issue #1112, opened 2023-02-11
Add NMEA GPS information to ACD driver


XCNav:
XCSoar version
--------------

All versions

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------
Presently the ACD driver only supports Altitude and Com data exchange between XCSoar and ACD

Add GPS Position information

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------
The ADC supports an autotune function to dial in the frequency of the nearest alternate airport as standby freq. Therefore it needs a gps position. As devices running  XCSoar naturally provide GPS information it would be usefull to pass this as well to the ACD so no second data input from a gps source is needed.
 See above

hjr:
I like this idea. Need to be configurable, because people might use other gps sources connected to ACD.

DanD222:
It would also be great to extend the functionality of say the VT-01 transponder to ADS-B using GPS positional data, so an option to send GPS position information could probably be useful for the Generic driver as well? (or is the better way to do this using the xci system?)

Scumi:
> I like this idea. Need to be configurable, because people might use other gps sources connected to ACD.

You can disable processing of GPS data on each port in the ACD config, so there would be no need to do that on XCSoar side.

I restarted work on sending GPS data to the ACD from XCSoar based on my old PR #170. Not sure if I'm more successful this time, but refactoring goes quite well for now, so lets see...

XCNav:
Is there still a device needed. I can organize one by AirAvionics.
Just contact me via info@xcnav.de

Scumi:
For what? I have opened a PR, which I tested on my device. PR is linked right above your comment. It just needs review and merge.

Scumi:
@XCNav you can close this now, PR is merged.

lordfolken:
Closing as implemented.

XCNav:
Thanks to all!

PR #1117, opened 2023-02-15
Provide GPS data to AirControlDisplay

Scumi:
Brief summary of the changes
----------------------------

~~- added a short format for the DDMM.MMM location format~~
- added NMEAFormatter to format standard NMEA sentences
- updated the AirControlDisplay driver to forward GPS data via NMEA sentences to the device
- updated the CAILNAV driver to also use the new NMEAFormatter

#170 (previous attempt)
#1112 (recent request)

### DIFF
--- a/build/test.mk
+++ b/build/test.mk
@@ -109,6 +109,7 @@ TEST_NAMES = \
 	TestByteSizeFormatter \
 	TestTimeFormatter \
 	TestIGCFilenameFormatter \
+	TestNMEAFormatter \
 	TestLXNToIGC \
 	TestLeastSquares \
 	TestHexString \
@@ -476,6 +477,21 @@ TEST_IGC_FILENAME_FORMATTER_SOURCES = \
 	$(TEST_SRC_DIR)/TestIGCFilenameFormatter.cpp
 TEST_IGC_FILENAME_FORMATTER_DEPENDS = MATH UTIL TIME
 $(eval $(call link-program,TestIGCFilenameFormatter,TEST_IGC_FILENAME_FORMATTER))
+
+TEST_NMEA_FORMATTER_SOURCES = \
+	$(SRC)/Units/Descriptor.cpp \
+	$(SRC)/Units/System.cpp \
+	$(SRC)/Atmosphere/AirDensity.cpp \
+	$(SRC)/Device/Parser.cpp \
+	$(SRC)/Device/Driver/FLARM/StaticParser.cpp \
+	$(SRC)/FLARM/Traffic.cpp \
+	$(SRC)/FLARM/FlarmId.cpp \
+	$(SRC)/Formatter/NMEAFormatter.cpp \
+	$(TEST_SRC_DIR)/tap.c \
+	$(TEST_SRC_DIR)/FakeGeoid.cpp \
+	$(TEST_SRC_DIR)/TestNMEAFormatter.cpp
+TEST_NMEA_FORMATTER_DEPENDS = LIBNMEA GEO MATH IO UTIL TIME
+$(eval $(call link-program,TestNMEAFormatter,TEST_NMEA_FORMATTER))
 
 TEST_STRINGS_SOURCES = \
 	$(TEST_SRC_DIR)/tap.c \
--- a/build/test.mk
+++ b/build/test.mk
@@ -670,6 +670,7 @@ TEST_DRIVER_SOURCES = \
 	$(SRC)/Atmosphere/Pressure.cpp \
 	$(SRC)/Atmosphere/AirDensity.cpp \
 	$(SRC)/TransponderCode.cpp \
+	$(SRC)/Formatter/NMEAFormatter.cpp \
 	$(ENGINE_SRC_DIR)/Waypoint/Waypoint.cpp \
 	$(TEST_SRC_DIR)/tap.c \
 	$(TEST_SRC_DIR)/FakeMessage.cpp \
@@ -856,6 +857,7 @@ DEBUG_REPLAY_SOURCES = \
 	$(SRC)/IGC/Generator.cpp \
 	$(SRC)/Units/Descriptor.cpp \
 	$(SRC)/Units/System.cpp \
+	$(SRC)/Formatter/NMEAFormatter.cpp \
 	$(ENGINE_SRC_DIR)/Airspace/AirspaceWarningConfig.cpp \
 	$(ENGINE_SRC_DIR)/ThermalBand/ThermalBand.cpp \
 	$(ENGINE_SRC_DIR)/ThermalBand/ThermalSlice.cpp \
@@ -1366,8 +1368,11 @@ RUN_VEGA_SETTINGS_SOURCES = \
 	$(SRC)/Device/Util/NMEAWriter.cpp \
 	$(SRC)/Device/Util/NMEAReader.cpp \
 	$(SRC)/Device/Port/ConfiguredPort.cpp \
+	$(SRC)/Units/Descriptor.cpp \
+	$(SRC)/Units/System.cpp \
 	$(SRC)/Device/Config.cpp \
 	$(SRC)/Operation/ConsoleOperationEnvironment.cpp \
+	$(SRC)/Formatter/NMEAFormatter.cpp \
 	$(TEST_SRC_DIR)/FakeMessage.cpp \
 	$(TEST_SRC_DIR)/FakeLogFile.cpp \
 	$(TEST_SRC_DIR)/FakeLanguage.cpp \
@@ -2454,11 +2459,14 @@ EMULATE_DEVICE_SOURCES = \
 	$(SRC)/Device/Util/LineSplitter.cpp \
 	$(SRC)/Device/Util/NMEAWriter.cpp \
 	$(SRC)/Device/Util/NMEAReader.cpp \
+	$(SRC)/Units/Descriptor.cpp \
+	$(SRC)/Units/System.cpp \
 	$(SRC)/Device/Driver/FLARM/BinaryProtocol.cpp \
 	$(SRC)/Device/Driver/FLARM/CRC16.cpp \
 	$(SRC)/Device/Config.cpp \
 	$(SRC)/io/CSVLine.cpp \
 	$(SRC)/Operation/ConsoleOperationEnvironment.cpp \
+	$(SRC)/Formatter/NMEAFormatter.cpp \
 	$(TEST_SRC_DIR)/FakeLogFile.cpp \
 	$(TEST_SRC_DIR)/FakeLanguage.cpp \
 	$(TEST_SRC_DIR)/DebugPort.cpp \
--- a/src/Device/Driver/AirControlDisplay.cpp
+++ b/src/Device/Driver/AirControlDisplay.cpp
@@ -14,6 +14,10 @@
 #include "Math/Util.hpp"
 #include "util/StaticString.hxx"
 #include "util/Macros.hpp"
+#include "Formatter/NMEAFormatter.hpp"
+#include "NMEA/MoreData.hpp"
+#include "Operation/Operation.hpp"
+#include "time/PeriodClock.hpp"
 
 using std::string_view_literals::operator""sv;
 
@@ -105,6 +109,7 @@ ParsePAAVS(NMEAInputLine &line, NMEAInfo &info)
 
 class ACDDevice : public AbstractDevice {
   Port &port;
+  PeriodClock status_clock;
 
 public:
   ACDDevice(Port &_port):port(_port) {}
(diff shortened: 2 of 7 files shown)


src/Formatter/NMEAFormatter.hpp
```diff
@@ -0,0 +1,100 @@
...
+ * <7>  Speed over ground, 0.0 to 999.9 knots
+ * <8>  Course over ground 000.0 to 359.9 degrees, true
+ *      (leading zeros will be transmitted)
+ * <9>  UTC date of position fix, ddmmyy format
+ * <10> Magnetic variation, 000.0 to 180.0 degrees
+ *      (leading zeros will be transmitted)
+ * <11> Magnetic variation direction, E or W
+ *      (westerly variation adds to true course)
+ */
+bool FormatGPRMC(TCHAR *buffer, size_t size, const NMEAInfo &info);
```
lordfolken, line 45: Code-style issue.

Scumi: Ok, will check and update (and also fix the build errors for the tests).

src/Form/DigitEntry.cpp
```diff
@@ -747,6 +753,8 @@ DigitEntry::GetGeoAngle(CoordinateFormat format) const
           columns[9].value        ) / 60000.;
     break;
 
+  case CoordinateFormat::DDMM_MMM_SHORT:
+
```
lordfolken, line 757: code formatting

src/Formatter/NMEAFormatter.cpp
```diff
@@ -0,0 +1,131 @@
+/*
+Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2021 The XCSoar Project
```
lordfolken, line 5: copyright to current year

test/src/TestNMEAFormatter.cpp
```diff
@@ -0,0 +1,61 @@
+/* Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2022 The XCSoar Project
```
lordfolken, line 4: copyright year

build/test.mk
```diff
@@ -486,6 +487,22 @@ TEST_IGC_FILENAME_FORMATTER_SOURCES = \
...
 $(eval $(call link-program,TestIGCFilenameFormatter,TEST_IGC_FILENAME_FORMATTER))
 
+TEST_NMEA_FORMATTER_SOURCES = \
+	$(SRC)/Units/Descriptor.cpp \
+    $(SRC)/Units/System.cpp \
+    $(SRC)/Atmosphere/AirDensity.cpp \
+    $(SRC)/Device/Parser.cpp \
+    $(SRC)/Device/Driver/FLARM/StaticParser.cpp \
+    $(SRC)/FLARM/Traffic.cpp \
+    $(SRC)/FLARM/FlarmId.cpp \
```
lordfolken, line 497: Indents consistentcy

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -157,6 +163,25 @@ ACDDevice::ParseNMEA(const char *_line, NMEAInfo &info)
     return false;
 }
 
+void
+ACDDevice::OnSensorUpdate(const MoreData &basic)
+{
+  NullOperationEnvironment env;
+  char buffer[100];
```
lordfolken, line 170: codacy thinks that this can go up one bracket

src/Formatter/NMEAFormatter.hpp
```diff
@@ -0,0 +1,100 @@
+/*
+Copyright_License {
+
+  XCSoar Glide Computer - http://www.xcsoar.org/
+  Copyright (C) 2000-2022 The XCSoar Project
```
lordfolken, line 5: copyright year

Scumi:
Fixed the code style issues and improved on the logic.

I'm not happy with the amount of additions to test.mk, but I'm not sure what a good way around could be. I could create a new library (e.g. "FORMATTERS/libformatters.mk") and just add that as a dependency, which would reduce the clutter a bit.

Scumi:
Simplified the logic a bit further and fixed the windows build and general tests.

However, I cannot figure out, why the windows tests fail. If someone could help me out there or give me a hint, that would be well appreciated.

MaxKellermann:
Not ready for review as long as unit tests fail

Scumi:
> Not ready for review as long as unit tests fail

Build and tests are good now, you can review any time.

Scumi:
rebased

src/Formatter/NMEAFormatter.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+#include "Units/Units.hpp"
+#include "NMEAFormatter.hpp"
+
+void
+FormatGPRMC(char *buffer, size_t size, const NMEAInfo &info) noexcept
+{
+  const GeoPoint location = info.location_available
+    ? info.location
+    : GeoPoint::Zero();
```
MaxKellermann, line 32: `GeoPoint::Zero()` is a valid location that exists on earth. I don't like using "magic" values like that. Can't we just keep the columns empty if we have no location, instead of printing an arbitrary bogus location?

src/Formatter/NMEAFormatter.cpp
```diff
@@ -0,0 +1,117 @@
...
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+}
+*/
```
MaxKellermann, line 22: Please update to SPDX headers

src/Formatter/NMEAFormatter.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+#include "Units/Units.hpp"
+#include "NMEAFormatter.hpp"
```
MaxKellermann, line 25: A file should include its own header first

src/Formatter/NMEAFormatter.cpp
```diff
@@ -0,0 +1,117 @@
...
+FormatGPRMC(char *buffer, size_t size, const NMEAInfo &info) noexcept
+{
+  const GeoPoint location = info.location_available
+    ? info.location
+    : GeoPoint::Zero();
+
+  const BrokenDateTime now = info.time_available &&
+    info.date_time_utc.IsDatePlausible()
+    ? info.date_time_utc
+    : BrokenDateTime::NowUTC();
```
MaxKellermann, line 37: The fallback to the current time stamp seems surprising and wrong to me, it's not documented, and the system clock may be wrong. Think of a Kobo or a Raspi that has no battery-buffered real-time clock.
I say this library should not mandate any such fallback, it should only take values from its parameters and never do any system calls.

src/Formatter/NMEAFormatter.cpp
```diff
@@ -0,0 +1,117 @@
...
+               location.latitude.IsNegative() ? 'S' : 'N',
+               (int)location.longitude.AbsoluteDegrees(),
+               location.longitude.ToDMM().minutes,
+               location.longitude.ToDMM().decimal_minutes,
+               location.longitude.IsNegative() ? 'W' : 'E',
+               (double)Units::ToUserUnit(info.ground_speed, Unit::KNOTS),
+               (double)info.track.Degrees(),
+               now.day, now.month, now.year % 100,
+               info.variation.AbsoluteDegrees(),
+               info.variation.IsNegative() ? 'W' : 'E');
```
MaxKellermann, line 55: Code like this already exists:
https://github.com/XCSoar/XCSoar/blob/3fe70253b7d880b289427307625058aad12da07d/test/src/IGC2NMEA.cpp#L28-L36
Yours is more generic, which is a good thing, but we shouldn't have two implementations.

src/Formatter/NMEAFormatter.cpp
```diff
@@ -0,0 +1,117 @@
...
+  const BrokenDateTime now = info.time_available &&
+    info.date_time_utc.IsDatePlausible()
+    ? info.date_time_utc
+    : BrokenDateTime::NowUTC();
+
+  StringFormat(buffer, size,
+               "GPRMC,%02u%02u%02u,%c,%02u%02u.%03u,%c,%03u%02u.%03u,%c,%05.1f,%05.1f,%02u%02u%02u,%05.1f,%c",
+               now.hour, now.minute, now.second,
+               info.location_available ? 'A' : 'V',
+               (int)location.latitude.AbsoluteDegrees(),
```
MaxKellermann, line 43: Does this suffer from rounding errors? Your way of implementing this looks suspicious - it uses different methods to extract degrees and the fractional part; to guarantee that it's free from rounding errors, only one method should be used. The IGC2NMEA code uses `Angle::ToDMS()` to extract all components.

MaxKellermann (changes requested):

MaxKellermann:
I just saw that the code I complained about was not actually written by you, but you copied it from the CaiLNav driver. Copying code isn't good! Copying bad code is even worse!
Taking advantage of existing code is good, but instead of pretending you wrote it, the first commit should have just moved the code out of the CaiLNav driver as-is, as bad as it is, with all the mistakes (it is clear then that it's not your fault the code is bad, and on the other hand you don't claim being the author); and then follow-up commits can add more features or fix mistakes.

src/Formatter/NMEAFormatter.cpp
Scumi: done

src/Formatter/NMEAFormatter.cpp
Scumi: While I agree on your remark of "magic" values, I don't know how devices react when the latitude and longitude fields are left empty, together with the N/S/W/E identifiers. Most specifications of the NMEA0183 GPRMC sentence that I found don't mention that case, while a few just set it to 0000.0000. This is of course a valid location, but devices should discard the sentence based on field #2, which sets the position status to valid or invalid.

src/Formatter/NMEAFormatter.cpp
Scumi: done

Scumi:
Okay, I restructure the commits to reflect that.

src/Formatter/NMEAFormatter.cpp
Scumi: Will use `Angle::ToDMM()` as it has all necessary components ready to use.

src/Formatter/NMEAFormatter.cpp
Scumi: Agreed, updated the code using the new NMEAFormatter.

src/Formatter/NMEAFormatter.cpp
```diff
@@ -23,14 +23,16 @@ FormatGPRMC(char *buffer, size_t buffer_size, const NMEAInfo &info) noexcept
...
+               "GPRMC,%02u%02u%02u,%c,%s,%s,%05.1f,%05.1f,%02u%02u%02u,%05.1f,%c",
                now.hour, now.minute, now.second,
                info.location.IsValid() ? 'A' : 'V',
                lat_buffer,
                long_buffer,
                (double)Units::ToUserUnit(info.ground_speed, Unit::KNOTS),
                (double)info.track.Degrees(),
-               now.day, now.month, now.year % 100);
+               now.day, now.month, now.year % 100,
+               info.variation.AbsoluteDegrees(),
```
MaxKellermann: Must not use this field without checking its validity

MaxKellermann:
I don't understand the commit "Formatter/NMEAFormatter: add $ and Checksum directly in the formatter" - this adds complexity, but the commit message doesn't justify that additional complexity. And it changes the timeout to a different value.

Scumi:
> I don't understand the commit "Formatter/NMEAFormatter: add $ and Checksum directly in the formatter" - this adds complexity, but the commit message doesn't justify that additional complexity. And it changes the timeout to a different value.

Yes, the timeout should be the same as for `PortWriteNMEA()`. I will fix that.

I think the formatter should spit out a "ready-to-use" NMEA sentence and that also includes the starting $ and ends with a checksum. If you disagree I can remove that part, but will mention in the code comments that this might need to be added by anyone calling the formatter.

MaxKellermann:
I'm not sure what's the best API; it was left out because we have the PortWriteNMEA() function for that, so it was natural to omit the checksum in NMEA line generators, and add it only while sending; only the serial port transfer needs to be protected. There was no explanation at all in your commit message, and I still don't know why we should change that and not use PortWriteNMEA(). Certainly, PortWriteNMEA() isn't the best code design, but if we change it, we must be confident that the new code is better, and I'm not confident here.

Scumi:
Ok, fair enough. Will remove this part then and update the other commits. The IGC2NMEA-code looked so nice and clean with this.

MaxKellermann:
I like how simple the final patch to add the actual feature has become!

PR #1124, opened 2023-02-20
ActionInterface: only exchange frequencies when both are defined

Scumi:
Previously XCSoar would crash if a frequency swap was attempted without either of the frequencies being set.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -13,6 +13,7 @@ Version 7.29 - not yet released
   - Fix OTG mode for newer models
 * Polars
   - Add LS-5 polar
+* fix exchange frequencies crash when no frequency was set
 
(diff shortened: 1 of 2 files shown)


MaxKellermann:
How does it crash?

Scumi:
If I try to exchange the frequencies (without either or both being set in XCSoar before doing so), the following error is posted to the log:
```
xcsoar: ./src/RadioFrequency.hpp:90: constexpr unsigned int RadioFrequency::GetKiloHertz() const: Assertion `IsDefined()' failed.
Aborted (core dumped)
```

MaxKellermann:
This belongs in the commit message. Please add it and fix the coding style issue.

PR #1125, opened 2023-02-21
Fix backlight and add colour temperature to certain Kobo models

GeekGoneOld:
Brief summary of the changes
----------------------------

Added CLARA HD backlight, CLARAs HD and 2E and LIBRA 2 colour temp

### DIFF
--- a/src/Kobo/System.cpp
+++ b/src/Kobo/System.cpp
@@ -362,6 +362,7 @@ KoboCanChangeBacklightBrightness()
   case KoboModel::GLO_HD:
   case KoboModel::LIBRA2:
   case KoboModel::CLARA_2E:
+  case KoboModel::CLARA_HD:
     return true;
 
   default:
@@ -387,6 +388,7 @@ KoboGetBacklightBrightness()
 
   case KoboModel::LIBRA2:
   case KoboModel::CLARA_2E:
+  case KoboModel::CLARA_HD:
     if (File::ReadString(Path("/sys/class/backlight/mxc_msp430.0/brightness"), line, sizeof(line))) {
       result = atoi(line);
     }
@@ -417,6 +419,7 @@ KoboSetBacklightBrightness([[maybe_unused]] int percent)
 
   case KoboModel::LIBRA2:
   case KoboModel::CLARA_2E:
+  case KoboModel::CLARA_HD:
     File::WriteExisting(Path("/sys/class/backlight/mxc_msp430.0/brightness"), std::to_string(percent).c_str());
     break;
 
@@ -424,5 +427,110 @@ KoboSetBacklightBrightness([[maybe_unused]] int percent)
     // nothing to do here...
     break;
   }
+#endif
+}
+
+/**
+ * Gets the file that contains the background colour
+ * as text of integer 0-10
+ *
+ * Some models have more than one possibility and each
+ * possibility must be checked in a certain order
+ *
+ * @return the file that contains the background colour
+ * code or nullptr if background colour not supported
+ */
+const char *
+KoboGetBacklightColourFile() noexcept
+{
+#ifdef KOBO
+  constexpr const char * colour_files[3] = {
+    "/sys/class/leds/aw99703-bl_FL1/color",
+    "/sys/class/backlight/lm3630a_led/color",
+    "/sys/class/backlight/tlc5947_bl/color"
+  };
+  bool files_to_check[3] = {
+    false,
+    false,
+    false
+  };
+
+  switch (DetectKoboModel()) {
+  case KoboModel::CLARA_2E:
+    files_to_check[0] = true;
+    break;
+
+  case KoboModel::CLARA_HD:
+    files_to_check[1] = true;
+    break;
+
+  case KoboModel::LIBRA2:
+    files_to_check[0] = true;
+    files_to_check[1] = true;
+    files_to_check[2] = true;
+    break;
+
+  default:
+    return nullptr;
+  }
+  if (files_to_check[0] && File::Exists(Path(colour_files[0])))
+    return colour_files[0]; 
+  if (files_to_check[1] && File::Exists(Path(colour_files[1])))
+    return colour_files[1]; 
+  if (files_to_check[2] && File::Exists(Path(colour_files[2])))
+    return colour_files[2]; 
+#endif
+  return nullptr;
+}
+
+bool
+KoboCanChangeBacklightColour() noexcept
+{
+#ifdef KOBO
+  return KoboGetBacklightColourFile() != nullptr;
+#endif
+  return false;
+}
+
+/**
+ * Returns true if successful in fetching the current colour
+ * or false if not
+ *
+ * @param (by ref) colour is set to the current colour
+ * where 0 = cold (yellow) and 10 is hot (white)
+ */
+bool
+KoboGetBacklightColour([[maybe_unused]] unsigned int &colour) noexcept
+{
+#ifdef KOBO
+  char line[4];
+  if (File::ReadString(Path(KoboGetBacklightColourFile()),
+      line,sizeof(line))) {
+    int raw_value = atoi(line);
+    if (raw_value < 0)
+      colour = 0;
+    else
+      colour = raw_value;
+    return true;
+  }
+#endif
+  return false;
+}
+
+/**
+ * Changes the colour of the background lighting
+ *
+ * @param colour desired colour where 0 = cold (yellow)
+ * and 10 is hot (white). Other values are silently ignored
+ */
+void
+KoboSetBacklightColour([[maybe_unused]] int colour) noexcept
+{
+#ifdef KOBO
+
+  if(colour < 0) { colour = 0; }
+  if(colour > 10) { colour = 10; }
+  File::WriteExisting(Path(KoboGetBacklightColourFile()),
+                      std::to_string(colour).c_str());
 #endif
 }
--- a/src/Kobo/System.hpp
+++ b/src/Kobo/System.hpp
@@ -71,3 +71,15 @@ KoboGetBacklightBrightness();
 
 void
 KoboSetBacklightBrightness(int percent);
+
+const char *
+KoboGetBacklightColourFile() noexcept;
+
+bool
+KoboCanChangeBacklightColour() noexcept;
+
+bool
+KoboGetBacklightColour(unsigned int &colour) noexcept;
+
+void
+KoboSetBacklightColour(int colour) noexcept;
--- a/src/Kobo/SystemDialog.cpp
+++ b/src/Kobo/SystemDialog.cpp
@@ -47,12 +47,16 @@ class SystemWidget final
     USB_STORAGE,
     INCREASE_BACKLIGHT_BRIGHTNESS,
     DECREASE_BACKLIGHT_BRIGHTNESS,
+    INCREASE_BACKLIGHT_COLOUR_TEMPERATURE,
+    DECREASE_BACKLIGHT_COLOUR_TEMPERATURE,
   };
 
   Button *switch_otg_mode;
   Button *usb_storage;
   Button *increase_backlight_brightness;
   Button *decrease_backlight_brightness;
+  Button *increase_backlight_colour_temperature;
+  Button *decrease_backlight_colour_temperature;
 
 public:
   SystemWidget(const DialogLook &look):RowFormWidget(look) {}
@@ -64,6 +68,9 @@ private:
   void IncreaseBacklightBrightness();
   void DecreaseBacklightBrightness();
   void UpdateBacklightBrightnessButtons(int percent);
+  void IncreaseBacklightColourTemperature() noexcept;
+  void DecreaseBacklightColourTemperature() noexcept;
+  void UpdateBacklightColourButtons(unsigned int colour);
 
   /* virtual methods from class Widget */
   void Prepare(ContainerWindow &parent,
@@ -84,7 +91,21 @@ SystemWidget::Prepare([[maybe_unused]] ContainerWindow &parent, [[maybe_unused]]
     decrease_backlight_brightness = AddButton("Decrease Backlight Brightness", [this]() { DecreaseBacklightBrightness(); });
     int current_brightness = KoboGetBacklightBrightness();
     UpdateBacklightBrightnessButtons(current_brightness);
+    if(KoboCanChangeBacklightColour()) {
+      increase_backlight_colour_temperature = AddButton("Increase Backlight Colour Temperature",
+                                                        [this]() { IncreaseBacklightColourTemperature(); });
+      decrease_backlight_colour_temperature = AddButton("Decrease Backlight Colour Temperature",
+                                                        [this]() { DecreaseBacklightColourTemperature(); });
+      unsigned int current_colour = 0;
+      KoboGetBacklightColour(current_colour);
+      UpdateBacklightColourButtons(current_colour);
+    } else {
+      AddDummy();
+      AddDummy();
+    }
   } else {
+    AddDummy();
+    AddDummy();
     AddDummy();
     AddDummy();
   }
@@ -221,6 +242,38 @@ SystemWidget::UpdateBacklightBrightnessButtons(int percent)
   }
   if(increase_backlight_brightness != nullptr) {
     increase_backlight_brightness->SetEnabled(percent < 100);
+  }
+}
+
+inline void
+SystemWidget::IncreaseBacklightColourTemperature() noexcept
+{
+  unsigned int current_colour;
+  if (KoboGetBacklightColour(current_colour)) {
+    KoboSetBacklightColour(current_colour + 2);
+    UpdateBacklightColourButtons(current_colour + 2);
+  }
+}
+
+inline void
+SystemWidget::DecreaseBacklightColourTemperature() noexcept
+{
+  unsigned int current_colour;
+  if (KoboGetBacklightColour(current_colour)) {
+    if (current_colour < 2) current_colour = 2;
+    KoboSetBacklightColour(current_colour - 2);
+    UpdateBacklightColourButtons(current_colour - 2);
+  }
+}
+
+inline void
+SystemWidget::UpdateBacklightColourButtons(unsigned int colour)
+{
+  if(decrease_backlight_colour_temperature != nullptr) {
+    decrease_backlight_colour_temperature->SetEnabled(colour > 0);
+  }
+  if(increase_backlight_colour_temperature != nullptr) {
+    increase_backlight_colour_temperature->SetEnabled(colour < 10);
   }
 }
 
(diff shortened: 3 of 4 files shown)


lordfolken (changes requested):
While your change here is minimally invasive, it also adds to the unecessary complexity.

Can you 
* rewrite the GetInfo function as a case statement matching the model similar as in system.cpp?
* eliminate the clones by creating a function passing the path

The commit messages would be good if you can add kobo: to the beginning.

ubx:
It does not work on my Libra 2:
```
~ # ls /sys/class/backlight/lm3630a_led/color
ls: /sys/class/backlight/lm3630a_led/color: No such file or directory
```
but it would (I have tried it):
```
~ # ls /sys/class/leds/aw99703-bl_FL1/color 
/sys/class/leds/aw99703-bl_FL1/color
```
there are probably several HW revisions.
IMHO you should check if the path exists (maybe there are different possibilities)  before display the color menu.

Here my Libra 2 config:
``` 
~ # /bin/ntx_hwconfig -s /dev/mmcblk0
HW CONFIG v3.5 @1024 secno, size=74 bytes : 
[0] PCB='E70K10'
[1] KeyPad='PGUP+PGDN'
[2] AudioCodec='No'
[3] AudioAmp='No'
[4] Wifi='RTL8723DS'
[5] BT='RTL8723DS'
[6] Mobile='No'
[7] TouchCtrl='ekth6'
[8] TouchType='C-Type'
[9] DisplayCtrl='MX6SLL+SY7636'
[10] DisplayPanel='7" Right EPD'
[11] RSensor='KX122'
[12] MicroP='No'
[13] Customer='9'
[14] Battery='PR-284983N-1500mA'
[15] Led='W'
[16] RamSize='512MB'
[17] IFlash='Micro SD'
[18] ExternalMem='No'
[19] RootFsType='Ext4'
[20] SysPartType='TYPE3'
[21] ProgressXHiByte=0x01
[22] ProgressXLoByte=0x68
[23] ProgressYHiByte=0x02
[24] ProgressYLoByte=0xE4
[25] ProgressCnts=0x00
[26] ContentType=0x00
[27] CPU='mx6sll'
[28] UIStyle='Customer UI'
[29] RAMType='LPDDR2'
[30] UIConfig='Normal'
[31] DisplayResolution='1680x1264'
[32] FrontLight='TABLE0+'
[33] CPUFreq='NC'
[34] HallSensor='TLE4913'
[35] DisplayBusWidth='16Bits'
[36] FrontLight_Flags=BootON:OFF,TABLE1X:OFF,EN_INV:ON,(null):OFF,(null):OFF,(null):OFF,(null):OFF,(null):OFF
[37] PCB_Flags=NO_KeyMatrix:ON,FPC_Touch:OFF,LOGO_LED:OFF,RD_MODE:OFF,EPD_LV:OFF,EPDVDD_STDALONE:OFF,HP_DETECT:OFF,ISD_1V8:OFF
[38] FrontLight_LEDrv='LM3630'
[39] VCOM_10mV_HiByte=0xFF
[40] VCOM_10mV_LoByte=0x6A
[41] PCB_REV=0x00
[42] PCB_LVL='E'
[43] HOME_LED_PWM='No'
[44] PMIC='BD71828'
[45] FL_PWM='AW99703x2'
[46] RTC='BD71828'
[47] BootOpt=ESD:OFF
[48] Touch2Ctrl='No'
[49] Touch2Type='No'
[50] GPS='No'
[51] FM='No'
[52] RSensor2='No'
[53] LightSensor='No'
[54] TPFWIDByte0=0x00
[55] TPFWIDByte1=0x00
[56] TPFWIDByte2=0x00
[57] TPFWIDByte3=0x00
[58] TPFWIDByte4=0x00
[59] TPFWIDByte5=0x00
[60] TPFWIDByte6=0x00
[61] TPFWIDByte7=0x00
[62] GPU='NC'
[63] PCB_Flags2=eMMC@SD1:ON,WiFi@SD2:OFF,eMMC@SD2:OFF,USB_TYPEC:ON,Cilix:OFF,(null):OFF,(null):OFF,(null):OFF
[64] EPD_Flags=NO_VDD_CTRL:OFF,LP_PANEL:OFF,ACEP:OFF,CFA:OFF,XON:OFF,(null):OFF,(null):OFF,(null):OFF
[65] LAN='No'
[66] MobileIF='NC'
[67] PIR='No'
[68] PanelLaminationSrc='NC'
[69] FL_CHNS=0x04
[70] TTS='No'
[71] Test_Flags=0:OFF,1:OFF,2:OFF,3:OFF,4:OFF,5:OFF,6:OFF,7:OFF
[72] EPD_CFA_REV='Gen1'
[73] PanelModel='NC'
```

src/Kobo/System.cpp
```diff
@@ -424,3 +424,65 @@ KoboSetBacklightBrightness([[maybe_unused]] int percent)
...
 }
+
+bool
+KoboCanChangeBacklightColour()
+{
+#ifdef KOBO
+  switch (DetectKoboModel()) {
+    case KoboModel::LIBRA2:
+    case KoboModel::CLARA_2E:
+    case KoboModel::CLARA_HD:
```
MaxKellermann, line 435: Bad indent

src/Kobo/System.hpp
```diff
@@ -71,3 +71,12 @@ KoboGetBacklightBrightness();
 
 void
 KoboSetBacklightBrightness(int percent);
+
+bool
+KoboCanChangeBacklightColour();
+
+int
+KoboGetBacklightColour();
```
MaxKellermann, line 79: What does the return value mean? Why is it signed? Can it throw exceptions?

src/Kobo/System.hpp
```diff
@@ -71,3 +71,12 @@ KoboGetBacklightBrightness();
 
 void
 KoboSetBacklightBrightness(int percent);
+
+bool
+KoboCanChangeBacklightColour();
```
MaxKellermann, line 76: Can this method throw exceptions?

src/Kobo/SystemDialog.cpp
```diff
@@ -46,13 +46,17 @@ class SystemWidget final
     SWITCH_OTG_MODE,
     USB_STORAGE,
     INCREASE_BACKLIGHT_BRIGHTNESS,
-    DECREASE_BACKLIGHT_BRIGHTNESS
+    DECREASE_BACKLIGHT_BRIGHTNESS,
+    INCREASE_BACKLIGHT_COLOUR_TEMPERATURE,
+    DECREASE_BACKLIGHT_COLOUR_TEMPERATURE
```
MaxKellermann, line 51: Trailing comma missing

src/Kobo/SystemDialog.cpp
```diff
@@ -224,6 +244,33 @@ SystemWidget::UpdateBacklightButtons(int percent)
...
+  int current_colour = KoboGetBacklightColour();
+  KoboSetBacklightColour(current_colour + 2);
+  UpdateBacklightColourButtons(current_colour + 2);
+}
+
+inline void
+SystemWidget::DecreaseBacklightColourTemperature()
+{
+  int current_colour = KoboGetBacklightColour();
+  KoboSetBacklightColour(current_colour - 2);
```
MaxKellermann, line 259: What if the getter has failed? Why call the setter if it has failed?

src/Kobo/SystemDialog.cpp
```diff
@@ -224,6 +244,33 @@ SystemWidget::UpdateBacklightButtons(int percent)
...
+  UpdateBacklightColourButtons(current_colour - 2);
+}
+
+inline void
+SystemWidget::UpdateBacklightColourButtons(int colour)
+{
+  if(decrease_backlight_colour_temperature != nullptr) {
+    decrease_backlight_colour_temperature->SetEnabled(colour != 0);
```
MaxKellermann, line 267: This enables the button if the color is negative

src/Kobo/System.hpp
GeekGoneOld: The return value is an integer from 0 to 10 where 0 is lowest colour temperature (yellowish) and 10 is highest (white/blueish).  The actual number is not any known scale (like degrees K or percent).

The return value is only signed because it gets its value from atoi() which returns int.  Do you have a preferred designation (uint8_t?).  If you wish me to change it, would you like me to change the Brightness code (similar but 0-100)  to your preferred type as well?

I don't see how it could throw an exception.  It reads a string from the driver (which is noexcept) and calls atoi() which returns 0 in the case of invalid.

src/Kobo/System.hpp
GeekGoneOld: I don't see how it could throw an exception. It reads a string from the driver (which is noexcept) and calls atoi() which returns 0 in the case of invalid.

src/Kobo/System.hpp
```diff
@@ -71,3 +71,12 @@ KoboGetBacklightBrightness();
...
+
+int
+KoboGetBacklightColour();
+
+void
+KoboSetBacklightColour(int colour);
```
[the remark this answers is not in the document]
GeekGoneOld: If we are out of memory, it could throw an exception on std::to_string().  Would you like me to handle that in some specific way?  noexcept?

src/Kobo/SystemDialog.cpp
GeekGoneOld: Are you sure?  Line 49 (previously last in list) got a comma added.  Line 51 (new last in list) shouldn't have one.

src/Kobo/System.hpp
MaxKellermann: The explanation belongs in a doxygen-style code comment.

GeekGoneOld:
In reply to @lordfolken (way up near the top of the issue)
>     * rewrite the GetInfo function as a case statement matching the model similar as in system.cpp?
> 
done
>     * eliminate the clones by creating a function passing the path
> 
done (for colour)
> The commit messages would be good if you can add kobo: to the beginning.
>
done

GeekGoneOld:
In response to @ubx near the top of this issue (just below @lordfolken ):
> It does not work on my Libra 2:
> 

> There are probably several HW revisions, [see here](https://github.com/NiLuJe/koreader/blob/3841c7b1b18f7a0ec348d1fe3dfdf53a53641c5a/frontend/device/kobo/device.lua).
>
> IMHO you should check if the path exists (maybe there are different possibilities) before display the color menu.

Looked at your link and handled the "silent" model changes much better.  Should work now.

src/Kobo/System.cpp
GeekGoneOld: Fixed

src/Kobo/SystemDialog.cpp
GeekGoneOld: I don't see how the getter could fail.  The getter is now tested for presence of the file prior to enabling the buttons.

src/Kobo/SystemDialog.cpp
GeekGoneOld: Yes.  Good point.  This has been changed to be colour > 0

src/Kobo/System.hpp
GeekGoneOld: Will gladly do this.  Can you point me to a preferred doxygen sample as I would like to conform to this repo's standard?  Do you want the entire function def documented or just the colour return value (get) / call parameter (set)?

Also, would you like me to document both functions (set and get) as they both use colour?

src/Kobo/System.hpp
GeekGoneOld: I added noexcept

src/Kobo/System.hpp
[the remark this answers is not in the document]
MaxKellermann: No. OOM is never to be handled. All is lost on OOM anyway.
The fact that you didn't mark this function `noexcept` indicates that this function can throw exceptions, but it's not documented.

src/Kobo/SystemDialog.cpp
MaxKellermann: Yes, I'm sure. The comma was missing in commit 79b80960035 but this wasn't noticed in code review.

src/Kobo/SystemDialog.cpp
MaxKellermann: No. You made a mistake: you made the colour a *signed* integer. Why did you choose to make it signed, why did you choose to allow negative numbers?

src/Kobo/System.hpp
GeekGoneOld: I went ahead and did the code comment in the same style as some other (src/Engine/GlideSolvers/GlidePolar.cpp).  I commented both KoboGetBacklightColour and KoboSetBacklightColour and KoboGetBacklightColourFile.  I also added noexcept to all the ...BacklightColour... functions as none of them can cause an exception other than OOM.

src/Kobo/System.hpp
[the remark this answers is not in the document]
GeekGoneOld: changed to noexcept

src/Kobo/SystemDialog.cpp
GeekGoneOld: Added the comma

src/Kobo/SystemDialog.cpp
GeekGoneOld: OK.  I've fixed it to silently return if the getter fails.

GeekGoneOld:
Small change to allow build-ubuntu and build-sanitizer to pass.  Took a couple of trys as I introduced a codacity error that I also fixed.  build-macos still fails as it seems to with every recent PR.

src/Hardware/Battery.cpp
```diff
@@ -43,11 +43,13 @@ GetInfo() noexcept
...
   auto &external = info.external;
-  KoboModel kobo_model = DetectKoboModel();
 
   char line[256];
 
-  if (kobo_model == KoboModel::GLO_HD) {
+  switch (DetectKoboModel())
+  {
+  case KoboModel::GLO_HD:
+  case KoboModel::CLARA_HD:
```
MaxKellermann, line 52: Don't do code refactoring in the middle of a patch that is supposed to do something else. This causes a confusing commit diff which is hard to review. You can refactor code, but do it in a separate commit. In this diff, I can't see what really belongs to the backlight feature.

src/Kobo/System.cpp
```diff
@@ -359,12 +359,14 @@ KoboCanChangeBacklightBrightness()
...
-    case KoboModel::GLO_HD:
-    case KoboModel::LIBRA2:
-    case KoboModel::CLARA_2E:
-      return true;
-    default:
-      return false;
+  case KoboModel::GLO_HD:
+  case KoboModel::LIBRA2:
+  case KoboModel::CLARA_2E:
+  case KoboModel::CLARA_HD:
```
MaxKellermann, line 365: This indent change has nothing to do with backlight, does it?

src/Kobo/System.cpp
```diff
@@ -378,20 +380,23 @@ KoboGetBacklightBrightness()
...
+    if (File::ReadString(Path("/sys/class/backlight/mxc_msp430.0/brightness"), line, sizeof(line))) {
+      result = atoi(line);
+    }
+    break;
+
+  default:
+    // nothing to do here...
+    break;
   }
```
MaxKellermann, line 400: More indent changes

src/Kobo/System.cpp
```diff
@@ -408,16 +413,124 @@ KoboSetBacklightBrightness([[maybe_unused]] int percent)
...
-      break;
+  case KoboModel::GLO_HD:
+    File::WriteExisting(Path("/sys/class/backlight/mxc_msp430_fl.0/brightness"), std::to_string(percent).c_str());
+    break;
+
+  case KoboModel::LIBRA2:
+  case KoboModel::CLARA_2E:
+  case KoboModel::CLARA_HD:
+    File::WriteExisting(Path("/sys/class/backlight/mxc_msp430.0/brightness"), std::to_string(percent).c_str());
+    break;
```
MaxKellermann, line 424: Yet more confusing indent changes

src/Kobo/SystemDialog.cpp
```diff
@@ -63,7 +67,10 @@ class SystemWidget final
   void ExportUSBStorage();
   void IncreaseBacklightBrightness();
   void DecreaseBacklightBrightness();
-  void UpdateBacklightButtons(int percent);
+  void UpdateBacklightBrightnessButtons(int percent);
```
MaxKellermann, line 70: What does the parameter mean? Prior to the rename, it was clear, but it's not any longer.

src/Kobo/SystemDialog.cpp
```diff
@@ -63,7 +67,10 @@ class SystemWidget final
...
+  void IncreaseBacklightColourTemperature();
+  void DecreaseBacklightColourTemperature();
```
MaxKellermann, line 72: What exceptions do these methods throw? If not at all, add `noexcept`!

MaxKellermann:
I could not review any actual code changes this time because this PR had become a confusing mess of refactoring and indentation changes.

GeekGoneOld:
Split into 3, organized commits.  Should be far more readable, and reviewable now.

src/Hardware/Battery.cpp
GeekGoneOld: Yes, it grew into confusion.  I have split the PR into 3 well defined commits (the first being refactoring).

src/Kobo/System.cpp
GeekGoneOld: It does have to do with the backlight but backlight brightness, not backlight colour.  I have done all the refactoring in a separate commit that preceeds the additions so it should be easy to follow.

src/Kobo/System.cpp
GeekGoneOld: Also moved to the refactor commit

src/Kobo/SystemDialog.cpp
GeekGoneOld: The backlight has two controls: brightness and colour.  The existing code ignored colour and referred to the backlight brightness as "...Backlight..." in function names  The change to "...BacklightBrightness..." makes it easier to add other backlight controls (like colour) and actually makes the original parameter meaning clearer!

src/Kobo/SystemDialog.cpp
GeekGoneOld: I've added noexcept.

src/Kobo/System.cpp
```diff
@@ -426,3 +429,108 @@ KoboSetBacklightBrightness([[maybe_unused]] int percent)
...
+ * possibility must be checked in a certain order
+ *
+ * @return the file that contains the background colour
+ * code or nullptr if background colour not supported
+ */
+const char *
+KoboGetBacklightColourFile() noexcept
+{
+#ifdef KOBO
+  const char * colour_files[3] = {
```
MaxKellermann, line 447: This array should be `const` (or better: `constexpr`)

GeekGoneOld: Yup.  Squeeze all you can into compile!  Changed to constexpr const char * colour_files[3] = {

MaxKellermann: It was initialized at compile time already, but the difference is that it can now be in the `.rodata` section where it will be write-protected.

PR #1129, opened 2023-02-26
PolarStore: Add Std Austria S polar.

hjr:
As of a user request. The polar is taken from a Schmepp-Hirth sales ad.


lordfolken (changes requested):
Can you add a NEWS entry?

PR #1135, opened 2023-03-01
Code cleanup

lordfolken:


src/Dialogs/dlgCredits.cpp
```diff
@@ -68,7 +68,6 @@ LogoPageWindow::OnPaint(Canvas &canvas) noexcept
   unsigned int middle = width / 2;
 
   // Draw XCSoar Logo
-  x = middle - (logo_size.width / 2);
   canvas.Copy({x, y}, logo_size, logo, {0, 0});
```
MaxKellermann, line 71: This breaks the logo position, doesn't it?

src/Profile/PageProfile.cpp
```diff
@@ -41,7 +41,7 @@ Load(const ProfileMap &map, PageLayout &_pl, const unsigned page)
 {
   char profileKey[32];
   unsigned prefixLen = sprintf(profileKey, "Page%u", page);
-  if (prefixLen <= 0)
+  if (prefixLen == 0)
```
MaxKellermann, line 44: While this change is not incorrect, it reduces the robustness of the code (just in case somebody does something stupid and makes the variable signed); it is extremely unlikely that this robustness ever has an effect, but on the other hand, you did not describe any advantage of dropping it, so I must figure that this change is completely pointless.

lordfolken: While I think that your argument about robustness is more likely to hide bad code...:

I think the main mistake here is that we check for a bad value and if untrue we execute, 
When we should be checking whether we have a good value and then proceed with execution.

Will make a proposal.

lordfolken: Actutally the current state is a bug. 

Sprintf will return a negative value on failure. Which is currently converted to a unsigned value and execution will proceed...

prefixLen should be therefore signed and checked for a positive value.

MaxKellermann: True, the bad code was added by commit 6faa4f7008b
Related question: does anybody know how `sprintf()` can ever fail?

src/Profile/PageProfile.cpp
```diff
@@ -40,47 +40,47 @@ static void
...
-  map.Get(profileKey, temp);
-  switch (temp) {
-  case tlEmpty:
-    pl.valid = false;
-    break;
-
-  case tlMap:
-    pl.infobox_config.enabled = false;
-    break;
+  int prefixLen = snprintf(profileKey, 32, "Page%u", page);
```
MaxKellermann, line 43: Why `snprintf()` instead of `sprintf()`? What's the point?
And why the hard-coded magic value?

src/Profile/PageProfile.cpp
```diff
@@ -40,47 +40,47 @@ static void
...
+  if (prefixLen > 0 && prefixLen <= 32)
```
MaxKellermann, line 44: Off-by-one buffer overflow bug (forgot to consider the null terminator) and yet another hard-coded magic value

src/Profile/PageProfile.cpp
```diff
@@ -40,47 +40,47 @@ static void
...
+  {
+    PageLayout pl = PageLayout::Default();
+    strcpy(profileKey + prefixLen, "InfoBoxMode");
```
MaxKellermann, line 47: Since you felt it was important to check the `prefixLen` for overflowing the buffer - what happens in this line if `prefixLen==31`? The `strcpy()` would overflow the buffer.
I don't feel like this overflow check has a point, but in the context of your decision that it does have a point, the rest of this commit doesn't make sense.

src/Profile/PageProfile.cpp
```diff
@@ -40,47 +40,47 @@ static void
...
-    pl.valid = false;
-    break;
-
-  case tlMap:
-    pl.infobox_config.enabled = false;
-    break;
+  int prefixLen = snprintf(profileKey, 32, "Page%u", page);
+  if (prefixLen > 0 && prefixLen <= 32)
+  {
+    PageLayout pl = PageLayout::Default();
```
MaxKellermann, line 46: I hate how this commit introduces an indent level when the old version carefully avoided that. This obscures the commit diff for no reason, because this commit does two unrelated things.

MaxKellermann (changes requested):

August2111:
Excuse my objection, but do you really want to review someone else's code? In my opinion, this ZZip component should be used either as a submodule or as a 3rd party component - possibly with a patch if necessary - but actually I would just leave this foreign code as it is - if it doesn't cause any problems...

lordfolken:
I'm keeping my branch and will return to this later. As the most important parts have been merged, i will close this for now.

PR #1140, opened 2023-03-03
Plane details, take over handicap factor from list selection. 

### LINKED ISSUE #1132: Plane Details, list selected data items not all copied
Head rev of today. Linux desktop, sim mode.

Expected behavior
-----------------
Choose a glider type from the internal polar list should copy all data into the plane profile.

Actual behavior
---------------
The selection misses to copy the handicap factor as well as the max. cruise speed. Because the "New" button shows the form prefilled with the current active data, this might slip through the users eye and results in a wrong plane data mix.

Config -> Plane easy and always to reproduce.

The latest change/augmentation of the stored polar data in terms of masses might have missed out something.

Yes

hjr:
Brief summary of the changes
----------------------------
Copy one more item from the polar list selection.

Closes #1132

### DIFF
--- a/src/Dialogs/Plane/PlaneDetailsDialog.cpp
+++ b/src/Dialogs/Plane/PlaneDetailsDialog.cpp
@@ -185,6 +185,7 @@ PlaneEditWidget::PolarButtonClicked() noexcept
     LoadValue(TYPE, plane.polar_name.c_str());
 
   /* reload attributes that may have been modified */
+  LoadValue(HANDICAP, plane.handicap);
   LoadValue(WING_AREA, plane.wing_area);
   LoadValue(EMPTY_MASS, plane.empty_mass, UnitGroup::MASS);
   LoadValue(MAX_BALLAST, plane.max_ballast);


hjr:
Ok, dont know why this Austria polar slipped in here. Stupid git.

src/Dialogs/Plane/PlaneDetailsDialog.cpp
```diff
@@ -185,6 +185,7 @@ PlaneEditWidget::PolarButtonClicked() noexcept
     LoadValue(TYPE, plane.polar_name.c_str());
 
   /* reload attributes that may have been modified */
+  LoadValue(HANDICAP, static_cast<int>(plane.handicap));
```
MaxKellermann, line 188: Let's not do this cast, this is fragile code clutter. Instead, `RowFormWidget()` should have an (inline) overload for `unsigned` which does this cast.

hjr: agreed

MaxKellermann: I've added the overload, you can now amend this commit to remove the cast.

hjr: Thx, did so.

Issue #1133, opened 2023-02-27
our github CI chain seems broken 


hjr:
XCSoar versions having and not having the problem
-------------------------------------------------
none, but the CI build env. My local build is fine.

Expected behavior
-----------------
build-macros should run through smooth and convert graphics etc. pp.

Actual behavior
---------------
The build-macros seem stop after this log line:
output/data/icons/winpilot_marginal_rgb.png output/data/icons/alt_landable_field_rgb.png output/data/icons/map_vor_160_rgb.png output/data/icons/settings_160_rgb.png

Steps to reproduce the behavior
-------------------------------
Trigger build of xcs head on github.

Do you have any idea what may have caused this?
-----------------------------------------------
No

Do you have an idea how to solve the issue?
-------------------------------------------
No

Scumi:
Its "build-macos" for Mac (not macros).

Issue is this:
`error: unknown warning option '-Wno-deprecated-experimental-coroutine' [-Werror,-Wunknown-warning-option]
`

lordfolken:
Max wrote the following:
https://github.com/XCSoar/XCSoar/blob/master/build/flags.mk#L7-L16

So someone should remove the flags try to compile with clang=y and fix any breakage.

lordfolken:
Found this commit at LLVM:
https://github.com/llvm/llvm-project/commit/55900a0d207670cb7295b98b31bb9ed50613cd42

Probably Max is using a git checkout of LLVM to compile. 

The clang supplied with github's osx machines is probably dated in comparison. 
The readme lists it as 14.0.0 and 15.0.4 available via brew:
https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md

hjr:
Thx guys!

lordfolken:
@MaxKellermann what is the minimal version of clang required to build XCSoar? 
Debian seems to come with stone-age 11. Which we seem to be installing with the provision-scripts. 13 would be available from repositories.

fberst:
according to the developer manual, clang/LLVM 10 is the minimum version required,
but that doesn' work at the moment, i tried it in Debian with clang11 and even after disabling that option mentioned above i get a lot of errors. (concept / template related)....

It seems that clang11 doens't know std::invocable and std::predicate, i assume this libc++ version is too old.
No clue at the moment what to do about this......

[...]

PR #1141, opened 2023-03-04
bugfixes clang build

fberst:
this is related to #1133, but also adresses some issues when building on debian with older clang version (11)

- Driver KRT2: fix "error: case value not in enumerated type 'std::byte'" when building with clang on debian
- Concepts.hxx: check clang version before using std::invocable and std::predicate (because of build failure on debian with clang11)
- install-debian-packages: add libc++abi-dev, because of linker failure
- flags.mk: disable -Wno-deprecated-experimental-coroutine, we can't use this option if the minimum clang version required is 10 (according to developer manual).

### DIFF
--- a/src/util/Concepts.hxx
+++ b/src/util/Concepts.hxx
@@ -30,13 +30,12 @@
 #pragma once
 
 #include <concepts>
-#include <util/Compiler.h>
 
 /**
  * Compatibility wrapper for std::invocable which is unavailable in
  * the Android NDK r25b and Apple Xcode.
  */
-#if !defined(ANDROID) && !defined(__APPLE__) && (CLANG_VERSION >= GCC_MAKE_VERSION(14,0,0))
+#if !defined(ANDROID) && !defined(__APPLE__) && (!defined __clang__ || __clang_major__ >=14)
 template<typename F, typename... Args>
 concept Invocable = std::invocable<F, Args...>;
 #else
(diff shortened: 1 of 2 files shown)


MaxKellermann:
Please move explanations to the commit message. PR texts are lost in GH's proprietary database, only commit messages remain forever.

fberst:
I had some hope that this might fix the macos build failure, but to no avail.
Now it fails because sprintf is deprecated and should be replaced by snprintf, looks like another task for the future....

MaxKellermann:
> Now it fails because sprintf is deprecated and should be replaced by snprintf, looks like another task for the future....

We should ignore dumb warnings like that. `sprintf()` is fine (in some rare situations).

build/flags.mk
```diff
@@ -12,7 +12,7 @@ CXX_FEATURES += -fcoroutines-ts
 # suppress clang 16 warning "the '-fcoroutines-ts' flag is deprecated
 # and it will be removed in Clang 17"; TODO: remove -fcoroutines-ts
 # eventually
-CXX_FEATURES += -Wno-deprecated-experimental-coroutine
+#CXX_FEATURES += -Wno-deprecated-experimental-coroutine
```
MaxKellermann, line 15: Removing this breaks my clang 14 build.

src/util/Concepts.hxx
```diff
@@ -30,12 +30,13 @@
 #pragma once
 
 #include <concepts>
+#include <util/Compiler.h>
```
MaxKellermann, line 33: Please don't introduce this header dependency. If all you need to check is a major version, `__clang_major__` is simple enough.

PR #1145, opened 2023-03-06
PolarStore: update handicap factors for 2023

fberst:
I updated the handicapt factors for 2023

The official list can be found here: [https://docs.weglide.org/pdf/handicaps.pdf](url)


MaxKellermann:
The PR contains text that is missing in the commit message

PR #1146, opened 2023-03-07
Add onFlarmTraffic for Android BLE application

ygaillard:
Brief summary of the changes
----------------------------
Additional function used by NMEA parser and transferring the information over Bluetooth Low Energy(BLE).
Devices which are able to parse the PFLAA message from Flarm devices can use this function to show the aircraft on the map. The function consistently keeps the structure defined by Flarm. The variable types are selected in accordance to Flarm serial data protocoll: flarm.com/wp-content/uploads/man/FTD-012-Data-Port-Interface-Control-Document-ICD.pdf 

### DIFF
--- a/android/src/SensorListener.java
+++ b/android/src/SensorListener.java
@@ -61,6 +61,9 @@ public interface SensorListener {
                            double gspeed, double vspeed,
                            double bearing);
 
+  void onFlarmTraffic(int AlarmLevel,int RelativeNorth,int RelativeEast,int RelativeVertical,
+                      String ID,int Track,double TurnRate,int GroundSpeed,double ClimbRate,
+                      int AcftType,boolean Stealth);
   void onTemperature(double temperature_kelvin);
 
   void onBatteryPercent(double battery_percent);
--- a/src/Android/NativeSensorListener.cpp
+++ b/src/Android/NativeSensorListener.cpp
@@ -285,6 +285,30 @@ Java_org_xcsoar_NativeSensorListener_setGliderLinkInfo(JNIEnv *env,
 
 gcc_visibility_default
 JNIEXPORT void JNICALL
+Java_org_xcsoar_NativeSensorListener_onFlarmTraffic(JNIEnv *env,
+                                                    jobject obj,
+                                                    jint AlarmLevel,
+                                                    jint RelativeNorth,
+                                                    jint RelativeEast,
+                                                    jint RelativeVertical,
+                                                    jstring ID,
+                                                    jint Track,
+                                                    jdouble TurnRate,
+                                                    jint GroundSpeed,
+                                                    jdouble ClimbRate,
+                                                    jint AcftType,
+                                                    jboolean Stealth)
+{
+  jlong ptr = env->GetLongField(obj, NativeSensorListener::ptr_field);
+  if (ptr == 0)
+    return;
+
+  auto &listener = *(SensorListener *)ptr;
+  listener.OnFlarmTraffic(AlarmLevel,RelativeNorth,RelativeEast,RelativeVertical,Java::String::GetUTFChars(env, ID).c_str(),Track,TurnRate,GroundSpeed,ClimbRate,AcftType,Stealth);
+}
+
+gcc_visibility_default
+JNIEXPORT void JNICALL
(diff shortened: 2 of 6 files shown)


android/src/SensorListener.java
```diff
@@ -61,6 +61,9 @@ void onGliderLinkTraffic(long gid, String callsign,
                            double gspeed, double vspeed,
                            double bearing);
 
+  void onFlarmTraffic(int AlarmLevel,int RelativeNorth,int RelativeEast,int RelativeVertical,
+                      String ID,int Track,double TurnRate,int GroundSpeed,double ClimbRate,
+                      int AcftType,boolean Stealth);
```
MaxKellermann, line 66: Parameter names shouldn't use CamelCase
Why is `Track` and `GroundSpeed` an integer?

src/Android/NativeSensorListener.cpp
```diff
@@ -283,6 +283,30 @@ Java_org_xcsoar_NativeSensorListener_setGliderLinkInfo(JNIEnv *env,
...
+                                                    jdouble ClimbRate,
+                                                    jint AcftType,
+                                                    jboolean Stealth)
+{
+  jlong ptr = env->GetLongField(obj, NativeSensorListener::ptr_field);
+  if (ptr == 0)
+    return;
+
+  auto &listener = *(SensorListener *)ptr;
+  listener.OnFlarmTraffic(AlarmLevel,RelativeNorth,RelativeEast,RelativeVertical,Java::String::GetUTFChars(env, ID).c_str(),Track,TurnRate,GroundSpeed,ClimbRate,AcftType,Stealth);
```
MaxKellermann, line 307: Bad coding style: missing spaces after comma, line too long

src/Device/AndroidSensors.cpp
```diff
@@ -447,6 +447,63 @@ DeviceDescriptor::OnGliderLinkTraffic(GliderLinkId id, const char *callsign,
...
+     traffic->Clear();
+    traffic->id = id;
+
+    traffic_list.new_traffic.Update(basic.clock);
+  }
+
+  traffic->alarm_level = (FlarmTraffic::AlarmType)AlarmLevel;
+
+  traffic->relative_north = RelativeNorth;
+
```
MaxKellermann, line 481: Too many pointless empty lines - they have no meaning if an empty line is between *all* actual code lines; there's no grouping implied by them.

src/Device/AndroidSensors.cpp
```diff
@@ -447,6 +447,63 @@ DeviceDescriptor::OnGliderLinkTraffic(GliderLinkId id, const char *callsign,
...
+  traffic->speed = GroundSpeed;
+
+  traffic->climb_rate = ClimbRate;
+
+  traffic->type = (FlarmTraffic::AircraftType) AcftType;
+
+  traffic->stealth = Stealth;
+
+    // set time of fix to current time
```
MaxKellermann, line 500: This comment is wrong. You copied it from OnGliderLinkTraffic() but it's still wrong. And badly indented.

MaxKellermann (changes requested):
Bad commit message.

I don't quite get the point of your suggested change.
The PR text (which doesn't exist in the commit message) says "Devices which are able to parse the PFLAA message from Flarm devices can use this function to show the aircraft on the map" but what kind of device would that be? From where will that device get the PFLAA message? And how will it benefit from this change?

ygaillard:
Thanks for the reply. Actually, I didn't know you could also draft pull request. I know it is drafty.
Currently only my custom developped board is able to fulfill to parse PFLAA and transmit over BLE. I didn't published it yet, since I'm still working on it...

ygaillard:
The benefit is that all the parsing and data acquisition is relocated on a low energy powered MCU. Then XCSoar only needs to read the conditioned information.

ygaillard:
May I delete this request until I completely published my custom board?

MaxKellermann:
> The benefit is that all the parsing and data acquisition is relocated on a low energy powered MCU. Then XCSoar only needs to read the conditioned information.

I see no benefit here. You add lots of complex code to XCSoar so you can add complex hardware with complex firmware. Adding complexity is rarely a good idea.

PR #1148, opened 2023-03-09
Bug fix and extension of Windows serial ports

August2111:
Brief summary of the changes
----------------------------

* Fix bug:  remove unnecessary trailing colon in the search for the active serial ports of Windows.
* Allow connection to COM port 10 to 255 (f.e. for BlueTooth or VSPE user)

* Repetition or revision of PR #934 - all requirements of Max's code review incorporated.

### DIFF
--- a/src/Device/Port/ConfiguredPort.cpp
+++ b/src/Device/Port/ConfiguredPort.cpp
@@ -78,8 +78,16 @@ OpenPortInternal(EventLoop &event_loop, Cares::Channel &cares,
   case DeviceConfig::PortType::SERIAL:
     if (config.path.empty())
       throw std::runtime_error("No port path configured");
-
+#ifdef _WIN32 
+    {
+      // the usual windows style of device names:
+      auto *endp = CopyString(buffer, MAX_PATH, _T("\\\\.\\"));
+      CopyString(endp, MAX_PATH - (endp-buffer), config.path.c_str());
+      path = buffer;  // set path to the begin of the buffer
+    }
+#else
     path = config.path.c_str();
+#endif
     break;
 
   case DeviceConfig::PortType::BLE_HM10:
--- a/src/Dialogs/Device/PortDataField.cpp
+++ b/src/Dialogs/Device/PortDataField.cpp
@@ -115,10 +115,8 @@ FillDefaultSerialPorts(DataFieldEnum &df) noexcept
   for (unsigned i = 1; i <= 10; ++i) {
     TCHAR buffer[64];
     _stprintf(buffer, _T("COM%u"), i);
-    if (DosDeviceExists(buffer)) {
-      _tcscat(buffer, _T(":"));  // add the colon to the path string
+    if (DosDeviceExists(buffer))
       AddPort(df, DeviceConfig::PortType::SERIAL, buffer);
-    }
   }
 }
 
--- a/src/Dialogs/Device/PortDataField.cpp
+++ b/src/Dialogs/Device/PortDataField.cpp
@@ -114,9 +114,11 @@ FillDefaultSerialPorts(DataFieldEnum &df) noexcept
 {
   for (unsigned i = 1; i <= 10; ++i) {
     TCHAR buffer[64];
-    _stprintf(buffer, _T("COM%u:"), i);
-    if (DosDeviceExists(buffer))
-       AddPort(df, DeviceConfig::PortType::SERIAL, buffer);
+    _stprintf(buffer, _T("COM%u"), i);
+    if (DosDeviceExists(buffer)) {
+      _tcscat(buffer, _T(":"));  // add the colon to the path string
+      AddPort(df, DeviceConfig::PortType::SERIAL, buffer);
+    }
   }
 }
 
--- a/src/Dialogs/Device/PortDataField.cpp
+++ b/src/Dialogs/Device/PortDataField.cpp
@@ -112,7 +112,7 @@ DosDeviceExists(const TCHAR *path) noexcept
 static void
 FillDefaultSerialPorts(DataFieldEnum &df) noexcept
 {
-  for (unsigned i = 1; i <= 10; ++i) {
+  for (unsigned i = 1; i <= 255; ++i) {
     TCHAR buffer[64];
     _stprintf(buffer, _T("COM%u"), i);
     if (DosDeviceExists(buffer))
--- a/src/Profile/DeviceConfig.cpp
+++ b/src/Profile/DeviceConfig.cpp
@@ -80,7 +80,16 @@ LoadPath(const ProfileMap &map, DeviceConfig &config, unsigned n)
 {
   char buffer[64];
   MakeDeviceSettingName(buffer, "Port", n, "Path");
-  return map.Get(buffer, config.path);
+  bool retvalue = map.Get(buffer, config.path);
+#ifdef _WIN32
+  // the usual windows port names have no colon at the end
+  if (retvalue && (config.path.back() == TCHAR(':')) &&
+      config.path.StartsWith(_T("COM"))) {
+    // replace colon with '\0', so the path name has the correct length
+    config.path[config.path.length() - 1] = TCHAR('\0');
+  }
+#endif
+  return retvalue;
 }
 
 static bool
--- a/src/SystemSettings.cpp
+++ b/src/SystemSettings.cpp
@@ -15,7 +15,7 @@ SystemSettings::SetDefaults()
   } else {
     devices[0].port_type = DeviceConfig::PortType::SERIAL;
 #ifdef _WIN32
-    devices[0].path = _T("COM1:");
+    devices[0].path = _T("\\\\.\\COM1");
 #else
     devices[0].path = _T("/dev/tty0");
 #endif
--- a/src/util/StaticString.hxx
+++ b/src/util/StaticString.hxx
@@ -15,7 +15,7 @@
 #include <string_view>
 
 #ifdef _UNICODE
-#include <wchar.h>
+#include "util/WStringCompare.hxx"
 #endif
 
 bool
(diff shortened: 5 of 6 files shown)


August2111:
I don't understand the Codacy issue: The variable endp is used in line 106 in file src/Device/Port/ConfiguredPort.cpp as new destination...

August2111:
Together with PR #1153 the CI checks should be green and the commits should all conform to the XCSoar standard hopefully.

src/Profile/DeviceConfig.cpp
```diff
@@ -80,7 +80,16 @@ LoadPath(const ProfileMap &map, DeviceConfig &config, unsigned n)
...
   char buffer[64];
   MakeDeviceSettingName(buffer, "Port", n, "Path");
-  return map.Get(buffer, config.path);
+  bool retvalue = map.Get(buffer, config.path);
+#ifdef _WIN32
+  // the usual windows port names has no colon at the end
+  if (retvalue && (config.path.back() == TCHAR(':')) &&
+      config.path.StartsWith(_T("COM"))) {
+    // replace colon with 0, so the path name has the correct length
+    config.path[config.path.length() - 1] = 0;
```
lordfolken, line 89: Maybe not 0.. because that would result in COM0 which could be mistaken as valid.

src/Profile/DeviceConfig.cpp
```diff
@@ -80,7 +80,16 @@ LoadPath(const ProfileMap &map, DeviceConfig &config, unsigned n)
 {
   char buffer[64];
   MakeDeviceSettingName(buffer, "Port", n, "Path");
-  return map.Get(buffer, config.path);
+  bool retvalue = map.Get(buffer, config.path);
+#ifdef _WIN32
+  // the usual windows port names has no colon at the end
```
lordfolken, line 85: // the usual windows port names have no colon at the end

lordfolken (changes requested):
Your commit messages are not short and to the point.
While i value the additional verbosity and reasoning, it should be short and to the point. 
So  For example:
Dialogs/Device/PortDataField - remove unnecessary trailing colon

Fix regression from XXXXX ....

August2111:
> Your commit messages are not short and to the point.
> While i value the additional verbosity

Ok, that should explain the error of the previous commit - but it's certainly not necessary

src/Profile/DeviceConfig.cpp
August2111: Done

src/Profile/DeviceConfig.cpp
August2111: I changed the character to TCHAR('\0') - maybe that's clearer (and better in UNICODE) - but the result is NOT 'COM0'! e.g. the result of this operation on 'COM1:' is 'COM1' - and that's correct!

August2111:
I don't know if I'm following the exact procedure of the 'PR - Review- Change request - Change - Answer' procedure, I obviously did something wrong with the 1st PR ;-(

lordfolken (approved):

lordfolken:
@August2111 what about com ports not named "com"? 
image

August2111:
... but I didn't found a port named CNCxx up to now on my PC. From my point of view all devices, ports and others in windows could be open with '\\\\.\\[device name]', so I cannot see problems with other names...
But as I said - I didn't see such a com port...

August2111:
... and this problem has nothing to do with the (similar?) UNIX issue!

src/Dialogs/Device/PortDataField.cpp
```diff
@@ -114,9 +114,11 @@ FillDefaultSerialPorts(DataFieldEnum &df) noexcept
 {
   for (unsigned i = 1; i <= 10; ++i) {
     TCHAR buffer[64];
-    _stprintf(buffer, _T("COM%u:"), i);
-    if (DosDeviceExists(buffer))
-       AddPort(df, DeviceConfig::PortType::SERIAL, buffer);
+    _stprintf(buffer, _T("COM%u"), i);
+    if (DosDeviceExists(buffer)) {
+      _tcscat(buffer, _T(":"));  // add the colon to the path string
```
MaxKellermann: If this colon is not necessary, what's the point of adding overhead here to omit it and then add it? What regression? Your commit message is inadequate.

src/Device/Port/ConfiguredPort.cpp
```diff
@@ -78,8 +78,16 @@ OpenPortInternal(EventLoop &event_loop, Cares::Channel &cares,
   case DeviceConfig::PortType::SERIAL:
     if (config.path.empty())
       throw std::runtime_error("No port path configured");
-
+#ifdef _WIN32 
+    {
+      // the usual windows style of device names:
+      auto *endp = CopyString(buffer, MAX_PATH, _T("\\\\.\\"));
+      CopyString(endp, MAX_PATH - (endp-buffer), config.path.c_str());
+      path = buffer;  // set path to the begin of the buffer
```
MaxKellermann, line 86: This is obscure code and unnecessary overhead for using a temporary buffer.

src/Device/Port/ConfiguredPort.cpp
```diff
@@ -78,8 +78,16 @@ OpenPortInternal(EventLoop &event_loop, Cares::Channel &cares,
   case DeviceConfig::PortType::SERIAL:
     if (config.path.empty())
       throw std::runtime_error("No port path configured");
-
+#ifdef _WIN32 
+    {
+      // the usual windows style of device names:
```
MaxKellermann, line 83: What does that comment mean?

src/Dialogs/Device/PortDataField.cpp
```diff
@@ -115,10 +115,8 @@ FillDefaultSerialPorts(DataFieldEnum &df) noexcept
   for (unsigned i = 1; i <= 10; ++i) {
     TCHAR buffer[64];
     _stprintf(buffer, _T("COM%u"), i);
-    if (DosDeviceExists(buffer)) {
-      _tcscat(buffer, _T(":"));  // add the colon to the path string
```
MaxKellermann: I don't understand why this gets removed now, when you just added this line of code in the preceding commit.

src/Profile/DeviceConfig.cpp
```diff
@@ -80,7 +80,16 @@ LoadPath(const ProfileMap &map, DeviceConfig &config, unsigned n)
 {
   char buffer[64];
   MakeDeviceSettingName(buffer, "Port", n, "Path");
-  return map.Get(buffer, config.path);
+  bool retvalue = map.Get(buffer, config.path);
+#ifdef _WIN32
+  // the usual windows port names have no colon at the end
```
MaxKellermann, line 85: I have no idea what that means.

src/Profile/DeviceConfig.cpp
```diff
@@ -80,7 +80,16 @@ LoadPath(const ProfileMap &map, DeviceConfig &config, unsigned n)
...
+  if (retvalue && (config.path.back() == TCHAR(':')) &&
```
MaxKellermann, line 86: Buffer overflow bug here.

src/Profile/DeviceConfig.cpp
```diff
@@ -80,7 +80,16 @@ LoadPath(const ProfileMap &map, DeviceConfig &config, unsigned n)
...
+      config.path.StartsWith(_T("COM"))) {
+    // replace colon with '\0', so the path name has the correct length
```
MaxKellermann, line 88: Ehh, what?? What is the correct length and what does the colon have to do with that?

src/util/StaticString.hxx
```diff
@@ -15,7 +15,7 @@
 #include <string_view>
 
 #ifdef _UNICODE
-#include <wchar.h>
+#include "util/WStringCompare.hxx"
```
MaxKellermann, line 18: This is wrong and even if it were correct, this doesn't belong in this commit.

src/SystemSettings.cpp
```diff
@@ -15,7 +15,7 @@ SystemSettings::SetDefaults()
   } else {
     devices[0].port_type = DeviceConfig::PortType::SERIAL;
 #ifdef _WIN32
-    devices[0].path = _T("COM1:");
+    devices[0].path = _T("\\\\.\\COM1");
```
MaxKellermann, line 18: Your code is now inconsistent regarding what `path` means. Is this suposed to be a device name? Device name with colon? UNC path? You should document that, have proper migration for old profile values, and make your code consistent.

MaxKellermann (changes requested):
As always, your PRs are a confusing mess.

src/Dialogs/Device/PortDataField.cpp
August2111: I just can't believe what I'm reading there... In Windows, there can't be a colon in the search string in the port search function '`QueryDosDevice`', but in the port string the one for calling `CreateFile(path,...)` is in the old one Style ('COM1:') still urgently needed! So that the commit first fixes the bug that XCSoar introduced since Windows version 7.24, I removed the colon in the top line, but added it back to AddPort - knowing full well that I won't need this character later. But so that the commit also works as a standalone CherryPick and really fixes the error, that's how I did it! Incidentally, Max removed the '`DosDeviceExists(...)`' function, which he had requested at the time to make the search clearer. Ok, I can live with that - but the fact that it still does NOT find a serial device in the search function under Windows now is very bad! It would be better to show all 9 possible ports like before - regardless of whether they are available or not!But this also makes an extension to port numbers greater than 10 unnecessary, because you can't just add 255 ports to the list...

And apparently Max doesn't even seem to try out the code he's putting in, at least as far as the Windows part is concerned: that was the case more than a year ago, when he added the colon 
to commit [86bc99c](https://github.com/XCSoar/XCSoar/commit/86bc99c847ef31bf237a49f1d7c1a7508ed5ca56) - and it still is still the same today, where the code doesn't work with the current changes either! Max is doing such a great job but even with his limited time budget he has to accept that there are areas that others may have a better grasp of! Especially since this Windows code has no relevance at all for the main functionality of XCSoar on Android, Kobo and OpenVario!

PR #1151, opened 2023-03-11
Adding Logbook and the system info script to the new x-menu for OpenVario

Blaubart:
This PR is adding the menu button "Logbook" to the new x-menu for OpenVario to use the logbook script and it includes the new system info script.
It also prepares the menu button "System Settings" for the new settings option at the menu.


MaxKellermann:
This just replicates a feature that already exists in XCSoar (for the Kobo shutdown screen) in a badly written shell script. See https://github.com/Openvario/meta-openvario/pull/340#issuecomment-1468465808

PR #1153, opened 2023-03-13
Bugfixes clang build macos

August2111:
Brief summary of the changes
----------------------------

* change the make file flags.mk to avoid the build error on CI with MacOS
  * disable the warning 'deprecated-experimental-coroutine' on MacOS
  * Disable the warning 'deprecated declarations' on MacOS too
    * The check is carried out for all other targets, so that it can be omitted here from my point of view.

see closed PR #1141 from fberst


August2111:
In the 2nd commit I described how I would see a better and more sustainable solution: converting the sprintf function either to snprintf if the buffer length is available in the calling function, otherwise converting to StringFormatUnsafe and to allow internal calling of sprintf there only. This is marking the function as a potentially unsafe operation.
But that would be a lot of work, and I would only tackle it and do it if there was a chance of success of a merge into the master.

lordfolken:
@fberst you had a look at this in #1133. Would you see this as an acceptable solution?

August2111:
@lordfolken of course I can, but I think my 1st commit helps for both: Max could be use his clang version - and the build here on CI is without error.

lordfolken:
Well we are going to move to Debian 12 when its out. That comes with clang14.  But AFAIK only Darwin seems to complain at the sprintf.

August2111:
> But AFAIK only Darwin seems to complain at the sprintf.

Yes, I think so too.
I haven't been able to reproduce the problem with any other build system.
Unfortunately, this topic was 'hidden' by the other build bug on the Mac ;-( which has been around for quite a while. 
If there are similar problems when upgrading the versions (-> Debian 12), I (or we) have to try again, but I would be reluctant to give it a blanket release unless there is a need, this check for 'deprecated functions' should be done on as many systems as possible (I think anyway)

August2111:
> Well we are going to move to Debian 12 when its out. That comes with clang14

The MacOS has clang 14 too ;-)

PR #1154, opened 2023-03-13
Waypoint, Settings: Store home waypoint name instead id in profile.

### LINKED ISSUE #1122: Changing number of markers changes Home field
tested in v7.28

Expected behavior
-----------------

One can use the marker function to add and remove markers without it to cause the home field to change. 

Actual behavior
---------------

When a marker is added or existing markers are removed, basicly when number of markers changes. Next time you start up XCSoar the home field is different.

Start up XCSoar either in fly or sim mode --> add a marker --> close XCSoar --> open XCSoar fly or sim mode and check home field it will be different airfield. (Checking in sim mode is easiest as plane starts at home field).

From my testing it looks like home field is stored as some kind of list index and markers are added at the top of the list changing all the indices without changing what index is used for home thus changing home field.

Add markers at bottom of list or different list or store home field by name or coordinate.

ubx:
Fixes changing number of markers changes home field #1122.

Brief summary of the changes
----------------------------
This PR resolves the issue where changing the number of markers caused the home field to change.

Closes #1122

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -19,6 +19,8 @@ Version 7.29 - not yet released
   - Add Silene E78 polar
   - updated handicap factors for 2023
 * fix exchange frequencies crash when no frequency was set
+* configuration
+  - Fixes changing number of markers changes home field (#1122)
 
 Version 7.28 - 2022/10/29
 * data files
--- a/src/Computer/Settings.cpp
+++ b/src/Computer/Settings.cpp
@@ -38,14 +38,14 @@ PolarSettings::SetDefaults()
 void
 PlacesOfInterestSettings::ClearHome()
 {
-  home_waypoint = -1;
+  home_waypoint_name.clear();
   home_location_available = false;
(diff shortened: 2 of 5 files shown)


August2111:
I don't know if I'm wrong, but in my opinion the name of a waypoint doesn't have to be unique, there can be several waypoints with the same name, but they designate different places (e.g. train station, church, airfield,...) You need here necessarily a field with a unique identifier, so the ID is not wrong. The problem here is of course that the ID can change ;-(

ubx:
You are right, I was probably too hasty in providing this solution. Using a unique home waypoint ID would be the proper solution.

groundhogxc:
How about storing the name, altitude and the coordinates together as an ID string (or a hash of these). That should be unique, and if it is not, at least it will point to the right position and, importantly, to the right altitude.

ubx:
Obsolete, fixed in  #1287.

Issue #1116, opened 2023-02-14
High DPI icons


lordfolken:
XCSoar version
--------------
v7.28

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------
Current mobile screens have resolutions of up to 650dpi. 
XCSoar's icons currently scale to maximum of 160 DPI.

Ideally we can use the SVGs to generate higher DPI icons.

August2111:
Before we do this bmp/png to svg conversion in the program code, it is in my opinion faster to adjust the critical (too small) bitmaps. In my opinion, this includes, for example, the 'mode' icons cruise, climb and finalglide. Or do you think that all icons are generally displayed too small?

lordfolken:
It concerns a lot of icons displayed on the map, the flight mode icons you mentioned, but also the icons in the task manager, flarm status, gps, terrain collsion icons, metar icons etc. 

Unfortunately its a little more complex than just upscaling the icon:
 In general all bitmaps are generated during the build process from svgs here (in several different dpi versions)
https://github.com/XCSoar/XCSoar/blob/8f1cbf368cf591e03a83bf5d90cb4e26421bb1c3/build/resource.mk#L36-L68

The source svgs are here https://github.com/XCSoar/XCSoar/tree/master/Data/icons 

Additionally they are assigned a resource id in src/Resources.hpp (currently in 2 resolutions standard and "hd"). 

Then each resource is loaded at the appropriate place for example:
https://github.com/XCSoar/XCSoar/blob/8f1cbf368cf591e03a83bf5d90cb4e26421bb1c3/src/Look/MapLook.cpp#L84-L86

The load resource function seems to be defined here, and seems to have some ability to stretch icons. Maybe its artificially limited, or another issue:
https://github.com/XCSoar/XCSoar/blob/8f1cbf368cf591e03a83bf5d90cb4e26421bb1c3/src/ui/canvas/Icon.cpp#L61

August2111:
[...]

Issue #982, opened 2022-08-13
Waypoints replaced by blue dots with smaller zoom values


MikeBravo15:
XCSoar versions having and not having the problem
-------------------------------------------------

7.24

Actual/Expected behavior
-----------------

Since intruduction of the 7.24 version of XCS, waypoints got a new symbol, which I find nice. Hovever, when you zoom out, symbols are replaced with big blue dots. I find the dots far too big for my screen - they clutter the view with smaller zoom and also overlap landables. Blue dots are OK, just please make them smaller or size selectable. Thank you

Steps to reproduce the behavior
-------------------------------

Zoom out on map view

lordfolken:
The problem is that we lack any sort of scaling for images on screens. So too large for you might be too small for someone else.

MikeBravo15:
As at lower zoom values there is just blue circle displayed, is it possible to set it in the same manner as Landable size?

MikeBravo15:
Well, I would call it a regression than enhancement as at smaller zoom I usually do not look for waypoints not important for the current task and far away from my current position. I would even kill airspace information boxes at extreme small zoom to enhance clarity. What was the idea behind the blue dots?

MikeBravo15:
![Screenshot_20220823-210148](https://user-images.githubusercontent.com/42399455/186242063-a6ded0e6-5057-4954-aa65-e082609c5b0a.png)

MaxKellermann:
This... looks awful!

MaxKellermann:
I can't reproduce this look. When I zoom out, I don't see those circles.

MikeBravo15:
Shall I post both .cup files/other settins? I run XCS on Pixel 4a and Nokia 8 - same behavior

lordfolken:
https://github.com/XCSoar/XCSoar/blob/master/build/resource.mk#L42 is the line that resizes all the map icons.
we may need different ones for different sizes. But as long as we don't  have any sort of bitmap/dpi support this is a stop gap solution.

lordfolken:
-dpi=400 and i can get something like that on my laptop.

MaxKellermann:
I can't reproduce with -dpi=400, and I have no idea where this purple circle comes from.

Scumi:
Depends if you have a waypoint file selected with certain waypoint types.

MaxKellermann:
Always post files and settings. Not doing so wastes the time of developers for asking questions and waiting for answers, and it delays the bug fix a lot.

[...]

Issue #1026, opened 2022-08-28
Icons in task administration no more readable on KOBO


fraktal-sb:
XCSoar versions having and not having the problem
-------------------------------------------------

OK 7.23
NOK 7.24

Expected behavior
-----------------

the icons in task administation are no more radable on KOBO (Glo HD, 300 PPI)

Actual behavior
---------------

on this screenshot of 7.24 icons are to small

![2022-08-28 20_59_42-Clipboard](https://user-images.githubusercontent.com/69430537/187090513-b6e0262c-22f2-426a-ac7a-27906b08a940.png)

on this screenshot of 7.23 icons are OK

![2022-08-28 20_59_01-Clipboard](https://user-images.githubusercontent.com/69430537/187090560-28947308-dfee-4ce6-85fc-716e5bc49150.png)

Steps to reproduce the behavior
-------------------------------

check 7.24 against 7.23

Do you have any idea what may have caused this?
-----------------------------------------------

no

Do you have an idea how to solve the issue?
-------------------------------------------

no

MikeBravo15:
The same for Status icons

lordfolken:
@fraktal-sb can you retest with latest update?

fraktal-sb:
tested with current head on KOBO GLO (not HD)
icons still too small (see comparision 6.8.17/7.29)
![Icons](https://user-images.githubusercontent.com/69430537/227168347-c6aa5b53-6085-4ec2-88af-c7635e208c65.png)

ubx:
> @ubx since you have current master running. Do the icons now look good?

Screenshot from Libra 2
![task-manager](https://user-images.githubusercontent.com/516680/227286807-8b6b3eb7-6cb8-4604-8238-04348f38b981.jpg)

For my taste, still a little too small

MaxKellermann:
I bisected this and found the first bad commit: ff406bf664842298cf1c710cd9b99dd13b9e8298
Obviously, this is a bad change, because it hard-codes the output sizes of the icons at certain pixel sizes, completely ignoring the physical sizes.
@lordfolken!

MaxKellermann:
If it didn't, I'd have to revert the whole commit, because it had "side effects". That commit was obviously bad - not because it was buggy - but because this change was weird and undocumented.

[...]

PR #1155, opened 2023-03-15
Icons update

MikeBravo15:
Unification of icon size to 32x32px for Menu/Mode icons and to 22x22 for Map icons (diff. size for potential future development). Optimization of graphics, removed not visible geometry, added ViewBox for correct sizing. Waypoint icons are referenced via this Naviter document - https://downloads.naviter.com/docs/SeeYou_CUP_file_format.pdf .Added icon template file. Can contribute to #1116, #1130, ViewBox importance see in #982


lordfolken:
@XCSoar/kobo-team could someone test this on their kobos? 
i tested an android build and things looked good. 
Trying openvario now...

lordfolken:
The main issue is the flight status icon in the lower left. Its just too small to make out.

ubx:
I tested the Libra 2, it looked good, but I noticed that on Kobos, the flight status icon in the lower right is too small. This problem was already in v7.28. I think we should create a new issue to address it.

lordfolken:
Danke.
@ubx can you check for #1026  on the libra2?

ubx:
> Danke. @ubx can you check for #1026 on the libra2?
No icons at all, just text labels.

lordfolken:
> > Danke. @ubx can you check for #1026 on the libra2?
> > No icons at all, just text labels. I can't see any icons on Linux, Kobo, or Android.

Oh sorry its  a configuration option:
Look->Screen Layout (expert)
Tab dialo style set to icons

MikeBravo15:
I consider 22x22 px size for icons too small for high DPI screens. Can you please also check my test build mentioned at #1116 for bigger menu/mode icons and report back? Thank you very much

ubx:
Another option I missed. The icons are actually visible, although they might be a bit small.
task-manager

lordfolken:
So the discussion about icon sizing should go here: https://github.com/XCSoar/XCSoar/issues/1116

I like about this PR that all the icons are correctly sized, that the viewport is fixed etc. This gives us a good basis for introducing changes.

PR #1157, opened 2023-03-19
Disable the Werror flag in warnings.mk for deprecated-declarations at…

lordfolken:


lordfolken:
I think this fix by @August2111 is minimal as it gets. This fixes the current ci issues and allows XCSoar to build. 

We current have code that isn't building on clang versions smaller than 14 on Linux platforms. As we move to debian 12 when its relased, that should be an intermittent problem.

build/flags.mk
```diff
@@ -12,7 +12,15 @@ CXX_FEATURES += -fcoroutines-ts
 # suppress clang 16 warning "the '-fcoroutines-ts' flag is deprecated
 # and it will be removed in Clang 17"; TODO: remove -fcoroutines-ts
 # eventually
-CXX_FEATURES += -Wno-deprecated-experimental-coroutine
+ifneq ($(TARGET_IS_DARWIN),y)
+  EXE := $(findstring .exe,$(MAKE))
+  CLANG_MAIN_VERSION := $(shell expr `$(LLVM_PREFIX)clang++$(LLVM_SUFFIX)$(EXE) -dumpversion | cut -f1 -d.`)
+  CLANG_14_OR_MORE := $(shell expr $(CLANG_MAIN_VERSION) \>= 14)
```
lordfolken, line 18: @August2111 should we make the EXE call on L15 windows only?

August2111: The '.exe' is only found on Windows - and correctly remains empty on Linux or others.
This is about the host system, not the (cross compiled) target!
Of course you can also do it differently ;-)

lordfolken: I found this stackoverflow article (famous last words i know...) : https://stackoverflow.com/questions/1617877/how-to-detect-llvm-and-its-version-through-define-directives

But according to this you there should be __clang_major__ which should give you the version of clang. 

There are some examples in src/util/Compiler.h of using such directives. Care to have a go?

August2111: I also thought about this possibility - and rejected it again, because not all header files include this Compiler.h include file... You have to make sure that either the Compiler.h is already included before each include of stdio.h - or add this macro with __clang_major__ ;-(
I would prefer a sustainable solution like the one I described in the commit: Using the equivalent function 'StringFormatUnsafe(..)' - and inserting this macro in front of it, because that's the only place where sprintf is called..
But before I - or whoever - tackles such a conversion, I first suggested the quick solution that should be solved in the first step in the .mk files - but unfortunately there is no predefined __clang_major__ ;- (, you have to 'build' it first - by the way, also a stackoverflow suggestion :)

Issue #769, opened 2021-10-27
Font of Buttons, Forms, and Dialogs is not resized, when window is dragged


lordfolken:
Target = UNIX
Version = v7.20 

What I see:
When one starts with a very small resolution, and then changes the window by dragging, or resizing in another way, the fonts stay at the size when xcsoar started.
![2021-10-28-002203_640x480_scrot](https://user-images.githubusercontent.com/407371/139156092-66f93ff4-7d1f-4e9c-a2e6-5bc151bd2873.png)

What i expect:
On Resize of the window the Font size is being recalculated and applied.

PR #1158, opened 2023-03-22
Look/InfoBoxLook: Calc AutoFont to realistic no. Chars

lordfolken:
Autofont was only using 4 chars to calculate the font size. This gave a bad look on small and large displays.

Removing also the upper Limit with the fontscale introduced in 663c7825aeb as this breaks the autoscaling when resizing a window.

Brief summary of the changes
----------------------------


lordfolken:
#769 is related, but not yet solved.

lordfolken:
Merged in 7bf77a3

PR #1163, opened 2023-03-26
Expandquickmenu

lordfolken:


lordfolken:
@linuxianer99 does this feel ok?

August2111:
Hi @lordfolken , from my point of view, the extensions of MAX_ITEMS in `src/Form/GridView.hpp` and in `src/Input/InputConfig.hpp` are still missing before you can really use more than 32 items in the QuickMenu. I had increased this to 40 internally, but initially discarded it because increasing the number of buttons would not make things clearer ;-( Personally, I would have preferred the Quick Menu to be separated into a flight mode and a ground mode (of course also with the option of to switch manually) - and actually I'd rather have a TreeView at this point that's very easy to navigate with arrow keys, which is why I discarded expanding the number of bottoms...

August2111:
I was late for `src/Form/GridView.hpp`... ;-(

lordfolken:
From a Feature point of view, i agree that extending it is not the long term answer. 

Currently parts of the app are not accessible if you use a remote stick only. Which makes it hard to use, in keyboard/joystick only scenarios such as the openvario. 
So for example Pilot Event Announce is missing, something needed in competitions inflight.

So with the current resources in the project, i op to do this and put a long term redesign plan somewhere next off-season.

lordfolken:
THx for the feedback and tip about intpuconfig.hpp. @August2111

August2111:
> From a Feature point of view, i agree that extending it is not the long term answer.
> 
> Currently parts of the app are not accessible if you use a remote stick only. Which makes it hard to use, in keyboard/joystick only scenarios such as the openvario. So for example Pilot Event Announce is missing, something needed in competitions inflight.
> 
> So with the current resources in the project, i op to do this and put a long term redesign plan somewhere next off-season.

This sounds good for me! We (you?) have to bring the system in a state which is functional for this saison asap. For this enough things to do, maybe we should publish an intermediate state, which is much better then the 7.28 from october last year - and not wait for the final touch... If I can help you with a few simple things, please let me know - f.e. the release procedure for 7.29, if Max doesn't have time for it.
In my opinion it would be very good if we could discuss 'face to face', perhaps by e-mail: Uwe Augustin <Info@FlapsOnline.de>?

linuxianer99:
@lordfolken Fine for me :-)

MaxKellermann:
Sigh. I wish we would throw away this trash that we call menu, which is a 15 years old misdesign that was built specifically around the Altair hardware buttons, and instead write a proper menu (with dynamic screen layout) that works well everywhere.
The dlgQuickMenu was already a bad kludge which I accepted to allow OpenVario people to have a solution quickly, because the full rewrite wasn't going to happen soon enough. And this PR just adds more of that. Of course, it's necessary, because we still don't have the rewrite ...

PR #1167, opened 2023-04-01
Transponder squawk class, readout and setting for devices

Scumi:
Brief summary of the changes
----------------------------
- TransponderSquawk: added for storing transponder code
- Update transponder squawk through ActionInterface
- Device/Driver/AirControlDisplay: send and receive of transponder squawk
- Introduce InfoBoxes for transponder squawk
- Dialogs/Device/DeviceListDialog: show if transponder is available

#867
https://forum.xcsoar.org/viewtopic.php?f=30&t=3424&start=75

### DIFF
--- a/src/Device/Driver/AirControlDisplay.cpp
+++ b/src/Device/Driver/AirControlDisplay.cpp
@@ -9,8 +9,11 @@
 #include "NMEA/Checksum.hpp"
 #include "Atmosphere/Pressure.hpp"
 #include "RadioFrequency.hpp"
+#include "TransponderCode.hpp"
 #include "Units/System.hpp"
 #include "Math/Util.hpp"
+#include "util/StaticString.hxx"
+#include "util/Macros.hpp"
 
 using std::string_view_literals::operator""sv;
 
@@ -68,8 +71,32 @@ ParsePAAVS(NMEAInputLine &line, NMEAInfo &info)
     unsigned volume;
     if (line.ReadChecked(volume))
       info.settings.ProvideVolume(volume, info.clock);
+  } else if (type == "XPDR"sv) {
+    /*
+    $PAAVS,XPDR,<SQUAWK>,<ACTIVE>,<ALTINH>
+    <SQUAWK> Squawk code value;
+             Octal unsigned integer value between 0000 and 7777 (digits 0–7).
+    <ACTIVE> Active flag;
+             0: standby (transponder is switched off / "SBY" mode)
+             1: active (transponder is switched on / "ALT" or "ON" mode
+                dependent of ALTINH)
+    <ALTINH> Altitude inhibit flag;
+             0: transmit altitude ("ALT" mode if active)
+             1: do not transmit altitude ("ON" mode if active)
+     */
+    unsigned code_value;
+    if (line.ReadChecked(code_value)) {
+      StaticString<16> buffer;
+      buffer.Format(_T("%04u"), code_value);
+      TransponderCode parsed_code = TransponderCode::Parse(buffer);
+
+      if (!parsed_code.IsDefined())
+        return false;
+
+      info.settings.transponder_code = parsed_code;
+      info.settings.has_transponder_code.Update(info.clock);
+    }
   } else {
-    // ignore responses from XPDR
     return false;
   }
 
@@ -90,6 +117,7 @@ public:
   bool PutStandbyFrequency(RadioFrequency frequency,
                            const TCHAR *name,
                            OperationEnvironment &env) override;
+  bool PutTransponderCode(TransponderCode code, OperationEnvironment &env) override;
 };
 
 bool
@@ -119,6 +147,15 @@ ACDDevice::PutStandbyFrequency(RadioFrequency frequency,
   char buffer[100];
   unsigned freq = frequency.GetKiloHertz();
   sprintf(buffer, "PAAVC,S,COM,CHN2,%u", freq);
+  PortWriteNMEA(port, buffer, env);
+  return true;
+}
+
+bool
+ACDDevice::PutTransponderCode(TransponderCode code, OperationEnvironment &env)
+{
+  char buffer[100];
+  sprintf(buffer, "PAAVC,S,XPDR,SQUAWK,%04o", code.GetCode());
   PortWriteNMEA(port, buffer, env);
   return true;
 }
--- a/src/Dialogs/Device/DeviceListDialog.cpp
+++ b/src/Dialogs/Device/DeviceListDialog.cpp
@@ -61,7 +61,7 @@ class DeviceListWidget final
     bool alive:1, location:1, gps:1, baro:1, pitot:1, airspeed:1, vario:1, traffic:1;
     bool temperature:1;
     bool humidity:1;
-    bool radio:1;
+    bool radio:1, transponder:1;
     bool debug:1;
 
     int8_t battery_percent;
@@ -105,6 +105,7 @@ class DeviceListWidget final
       debug = device.IsDumpEnabled();
       radio = basic.settings.has_active_frequency || 
         basic.settings.has_standby_frequency;
+      transponder = basic.settings.has_transponder_code;
       battery_percent = basic.battery_level_available
         ? (int)basic.battery_level
         : -1;
@@ -404,6 +405,10 @@ DeviceListWidget::OnPaintItem(Canvas &canvas, const PixelRect rc,
     if (flags.radio) {
       buffer.append(_T("; "));
       buffer.append(_T("Radio"));
+    }
+
+    if (flags.transponder) {
+      buffer.append(_T("; XPDR"));
     }
 
     if (flags.debug) {
--- a/src/InfoBoxes/Content/Radio.hpp
+++ b/src/InfoBoxes/Content/Radio.hpp
@@ -18,3 +18,10 @@ public:
   const InfoBoxPanel *GetDialogContent() noexcept override;
   void Update(InfoBoxData &data) noexcept override;
 };
+
+class InfoBoxContentTransponderCode : public InfoBoxContent
+{
+public:
+  const InfoBoxPanel *GetDialogContent() noexcept override;
+  void Update(InfoBoxData &data) noexcept override;
+};
--- a/src/InfoBoxes/Content/Type.hpp
+++ b/src/InfoBoxes/Content/Type.hpp
@@ -152,6 +152,8 @@ namespace InfoBoxFactory
 
     e_HeartRate,
 
+    e_TransponderCode, /* Transponder code */
+
     e_NUM_TYPES /* Last item */
   };
 
--- a/test/src/TestDriver.cpp
+++ b/test/src/TestDriver.cpp
@@ -1494,8 +1494,17 @@ TestACD()
   nmea_info.Reset();
   nmea_info.clock = TimeStamp{FloatDuration{1}};
 
-  /* $PAAVS responses from XPDR must be ignored */
-  ok1(!device->ParseNMEA("$PAAVS,XPDR,7000,1,0,1697,0,0*68",nmea_info));
+  /* test XPDR response */
+  ok1(device->ParseNMEA("$PAAVS,XPDR,7000,1,0,1697,0,0*68",nmea_info));
+  ok1(nmea_info.settings.has_transponder_code);
+  ok1(equals(nmea_info.settings.transponder_code.GetCode(), 
+             TransponderCode{07000}.GetCode()));
+
+  nmea_info.Reset();
+  nmea_info.clock = TimeStamp{FloatDuration{1}};
+
+  ok1(!(device->ParseNMEA("$PAAVS,XPDR,9999,1,0,1697,0,0*6F",nmea_info)));
+  ok1(!(nmea_info.settings.transponder_code.IsDefined()));
 
   nmea_info.Reset();
   nmea_info.clock = TimeStamp{FloatDuration{1}};
@@ -1612,7 +1621,7 @@ TestFlightList(const struct DeviceRegister &driver)
 
 int main()
 {
-  plan_tests(839);
+  plan_tests(843);
 
   TestGeneric();
   TestTasman();
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -6,6 +6,7 @@ Version 7.29 - not yet released
   - in OpenAir format use AC as Class and AY as Type (#1118)
 * devices
   - ATR833: read-out active and standby frequencies
+  - AirControlDisplay: Set and readout of transponder code
 * Android
   - get rid of not usable USB interfaces in the 'Device -> Port' list
 * Kobo
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -32,6 +32,7 @@ Version 7.29 - not yet released
   - increased precision in polar edit dialogue
   - extend QuickMenu to maximum of 64 entries
   - add PEV,FlarmTraffic,FileManager to joystick QuickMenu
+  - add transponder code infobox
 
 Version 7.28 - 2022/10/29
 * data files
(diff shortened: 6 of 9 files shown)


src/TransponderSquawk.hpp
```diff
@@ -0,0 +1,74 @@
...
+#pragma once
+
+#include <cassert>
+#include <cstdint>
+
+/**
+ * This class stores a transponder squawk
+ */
+class TransponderSquawk {
+  int16_t value;
```
MaxKellermann, line 13: This shouldn't be signed. You made it signed because you needed a magic value for "undefined", but you can use any illegal (positive) value for that, e.g. UINT16_MAX.
Also, for (theoretical) correctness, I'd like new code to use the portable, always-guaranteed-to-be-available types like `uint_least16_t` instead of `uint16_t`, unless an exact size is ultimately necessary (e.g. if binary layout matters, which it does not here).

src/TransponderSquawk.hpp
```diff
@@ -0,0 +1,74 @@
...
+      return false;
+    }
+
+    // Check each digit for validity
+    for (int i = 0; i < 4; ++i) {
+      if (!IsValidDigit(code % 10)) {
+        return false;
+      }
+      code /= 10;
+    }
```
MaxKellermann, line 70: This whole check looks awfully complicated and bloated, when you missed the fact that transponder codes are just octal values, and you could easily parse them as such with `strtoul(x, &endptr, 8)` and format with `%o` instead of pretending they are decimal.
With an octal representation, you can easily have:
```
static constexpr bool IsValidCode(uint_least16_t code) { return code <= 0777; }
```
Can it be any simpler? (and faster?)

src/TransponderSquawk.hpp
```diff
@@ -0,0 +1,74 @@
...
+  constexpr void Clear() noexcept {
+    value = -1;
+  }
+
+private:
+  constexpr bool IsValidDigit(int8_t digit) const noexcept {
+    return digit >= 0 && digit <= 7;
+  }
+
+  constexpr bool IsValidCode(int16_t code) const noexcept {
```
MaxKellermann, line 59: `static`?

src/TransponderSquawk.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Set this object to "undefined".
+   */
+  constexpr void Clear() noexcept {
+    value = -1;
+  }
+
+private:
+  constexpr bool IsValidDigit(int8_t digit) const noexcept {
```
MaxKellermann, line 55: `static`?

src/TransponderSquawk.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+#pragma once
+
+#include <cassert>
+#include <cstdint>
+
+/**
+ * This class stores a transponder squawk
+ */
+class TransponderSquawk {
```
MaxKellermann, line 12: I'd like to question the nomenclature. What is a "squawk"? On radio, "squawk" is a command, an abbreviation for "switch to transponder code". But the code itself is not a "squawk" but a "code". But maybe I'm wrong, I've never owned or operated a transponder.

MaxKellermann (changes requested):
I like this PR, only the data type is much more complicated than it needs to be.

src/TransponderSquawk.hpp
lyusupov: Correct. "Squawk" is just a word commonly used in aviation language _(while talking to an ATC)_ to identify an action necessary to change the transponder code.
IMHO, use of  _TransponderCode_ is better to name a class or a file rather than _TransponderSquawk_.

Scumi: Fine for me, will change.

src/TransponderSquawk.hpp
Scumi: Yes, that was pretty much the "design choice", but you are right. Just using another illegal number (will choose 9) works as well.

Will use the new type as well.

src/TransponderSquawk.hpp
Scumi: I recognized that transponder codes are octal, but I did not manage to make use of that realization. Your remarks were helpful, I hope the updated class looks better.

src/TransponderCode.hpp
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  explicit TransponderCode(uint_least16_t code) noexcept {
+    SetCode(code);
+  }
+
+  explicit TransponderCode(const char* code) noexcept : value(ParseCode(code)) {}
+
+  void SetCode(uint_least16_t code) noexcept {
+    value = ParseCode(std::to_string(code).c_str());
```
MaxKellermann, line 41: What the f... is this? This allocates heap memory to hold the decimal representation of the code as a `std::string`, then takes the C string of it and passes it to `ParseCode()` .... this looks as confusing and wrong and error prone as it could be.

src/TransponderCode.hpp
```diff
@@ -0,0 +1,79 @@
...
+   * @return octal (base 8) unsigned
+   */
+  constexpr uint_least16_t GetCode() const noexcept {
+    assert(IsDefined());
+
+    return value;
+  }
+
+  constexpr bool IsDefined() const noexcept {
+    return value != 4096;
```
MaxKellermann, line 61: What if the value is 4097? Or 65535? Why 4096?
Why not just use `code <= 07777` here?

src/TransponderCode.hpp
```diff
@@ -0,0 +1,79 @@
...
+    value = 4096;
+  }
+
+private:
+  static constexpr bool IsValidCode(uint_least16_t code) { return code <= 07777; }
+
+  static uint_least16_t ParseCode(const char* code_str) noexcept {
+    char* endptr;
+    const uint_least16_t code = strtoul(code_str, &endptr, 8);
+    return *endptr == '\0' && IsValidCode(code) ? code : 4096;
```
MaxKellermann, line 77: What's the point of the `IsValidCode()` check here? If `IsDefined()` were implemented in a sane way, then this check wouldn't be necessary.

MaxKellermann:
This is still more complicated than it should be, and complicated code is error prone.

Scumi:
> This is still more complicated than it should be, and complicated code is error prone.

Agreed. Will do another round and hopefully come up with something more simple. Thanks for the review.

src/TransponderCode.hpp
Scumi: Done.

src/TransponderCode.hpp
Scumi: I was able to remove `IsValidCode()`. Hopefully its simpler now.

src/TransponderCode.hpp
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  explicit TransponderCode(const char* code) noexcept {
+    SetCode(code);
+  }
+
+  void SetCode(uint_least16_t code) noexcept {
+    char code_str[32];
+    snprintf(code_str, sizeof(code_str), "%u", code);
+    value = ParseCode(code_str);
```
MaxKellermann, line 48: All you did here was eliminate the heap allocation and use a fixed-size stack buffer. But this is still bad. I mean, is `value=code` really too simple for your taste of code?

MaxKellermann: > No, I would like to do that, but then `TransponderCode(7000)` would not be a valid code because

Of course it's not. But `TransponderCode(07000)` is.

Scumi: Yes, indeed. But a device or airspace file will most probably omit the leading `0`. I had a look last week on how to make 7000 a 07000 but failed (I only get back to formatting with `%05u`, which is similar to what the class does now). Can you give me a hint where to look at? Happy to dig deeper.

MaxKellermann: > But a device or airspace file will most probably omit the leading `0`.

Of course the airspace file will not have a fifth digit, because that's the wrong syntax - you're mixing up syntaxes. The leading zero is C++ syntax for octal literals. To parse from a file, which is not C++ syntax, just parse the four digits as octal, like I suggested, without requiring a leading zero.

Requiring an extra zero digit would require writing extra code, but you have a history of making things more complicated than they should be.... relax your mind, do away with this complicated sh*t that occupies your mind, make it simple!

MaxKellermann:
I'll let the code do the talking. I've cherry-picked the first commit and amended it as I imagine would be the simplest and most robust way to do it. See aed9e65e3c618f19565df04f2d6358678bd0b87b

Scumi:
> I'll let the code do the talking. I've cherry-picked the first commit and amended it as I imagine would be the simplest and most robust way to do it. See [aed9e65](https://github.com/XCSoar/XCSoar/commit/aed9e65e3c618f19565df04f2d6358678bd0b87b)

Thank you. This looks really clean indeed. I had something close to this at one point, but couldn't get over a (probably minor) issue.

MaxKellermann:
Linker failure:
```
  LINK    output/UNIX/bin/RunTrace
/usr/bin/ld: output/UNIX/dbg/driver.a(AirControlDisplay.o): in function `ParsePAAVS(NMEAInputLine&, NMEAInfo&)':
src/Device/Driver/AirControlDisplay.cpp:89: undefined reference to `TransponderCode::Parse(char const*)'
make: *** [build/test.mk:1605: output/UNIX/bin/RunTrace] Error 1
```

src/ActionInterface.cpp
```diff
@@ -369,3 +369,19 @@ ActionInterface::ExchangeRadioFrequencies(bool to_devices) noexcept
     ActionInterface::SetStandbyFrequency(old_active_freq, old_active_freq_name, to_devices);
   }
 }
+
+void
+ActionInterface::SetTransponderCode(TransponderCode code, bool to_devices) noexcept
+{
+  /* update interface settings */
```
MaxKellermann, line 376: Please add `assert(code.IsDefined())` at the start of the function so bugs passing malformed values can be detected when debug builds are used.

src/Device/Descriptor.cpp
```diff
@@ -1034,6 +1034,30 @@ DeviceDescriptor::PutStandbyFrequency(RadioFrequency frequency,
   }
 }
 
+bool
+DeviceDescriptor::PutTransponderCode(TransponderCode code,
+                                     OperationEnvironment &env) noexcept
+{
```
MaxKellermann, line 1040: assert, please

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -68,8 +69,27 @@ ParsePAAVS(NMEAInputLine &line, NMEAInfo &info)
...
+    <ALTINH> Altitude inhibit flag;
+             0: transmit altitude ("ALT" mode if active)
+             1: do not transmit altitude ("ON" mode if active)
+     */
+    unsigned transponder_code;
+    if (line.ReadChecked(transponder_code)) {
+      TCHAR buffer[12];
+      snprintf(buffer, sizeof(buffer), "%04u", transponder_code);
+      info.settings.transponder_code = TransponderCode::Parse(buffer);
+      info.settings.has_transponder_code.Update(info.clock);
```
MaxKellermann, line 90: Here, you should check whether the parsed value was valid - XCSoar must never misbehave even in the presence of arbitrary malformed input.
In the past, you could easily crash XCSoar by feeding malformed NMEA with broken GPS coordinates - that's a critical bug.

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -68,8 +69,27 @@ ParsePAAVS(NMEAInputLine &line, NMEAInfo &info)
     unsigned volume;
     if (line.ReadChecked(volume))
       info.settings.ProvideVolume(volume, info.clock);
+  } else if (type == "XPDR"sv) {
```
MaxKellermann, line 72: bad indent?

test/src/TestDriver.cpp
```diff
@@ -1494,8 +1494,11 @@ TestACD()
   nmea_info.Reset();
   nmea_info.clock = TimeStamp{FloatDuration{1}};
 
-  /* $PAAVS responses from XPDR must be ignored */
-  ok1(!device->ParseNMEA("$PAAVS,XPDR,7000,1,0,1697,0,0*68",nmea_info));
+  /* test XPDR response */
+  ok1(device->ParseNMEA("$PAAVS,XPDR,7000,1,0,1697,0,0*68",nmea_info));
+  ok1(nmea_info.settings.has_transponder_code);
+  ok1(equals(nmea_info.settings.transponder_code.GetCode(), 
+             TransponderCode{07000}.GetCode()));
```
MaxKellermann, line 1502: Would be great to have tests with malformed/bad transponder values here, to ensure that error handling in the driver keeps working properly.

src/InfoBoxes/Content/Type.hpp
```diff
@@ -144,8 +144,10 @@ namespace InfoBoxFactory
     e_NbrSat, /* Number of used Sat by GPS module */
 
     e_ActiveRadio, /* Active Radio Frequency */
-
     e_StandbyRadio, /* Standby Radio Frequency */
+
+    e_TransponderCode, /* Transponder code */
```
MaxKellermann, line 149: Adding new enum choices in the middle breaks all existing configurations. Add at the end (right before `e_NUM_TYPES`).

src/Dialogs/Device/DeviceListDialog.cpp
```diff
@@ -406,6 +407,11 @@ DeviceListWidget::OnPaintItem(Canvas &canvas, const PixelRect rc,
       buffer.append(_T("Radio"));
     }
 
+    if (flags.transponder) {
+      buffer.append(_T("; "));
+      buffer.append(_T("XPDR"));
```
MaxKellermann, line 412: If you're not going to translate the string, make it one call.

MaxKellermann (changes requested):

src/InfoBoxes/Content/Type.hpp
Scumi: done

src/Device/Descriptor.cpp
Scumi: done

src/ActionInterface.cpp
MaxKellermann: Agree!

src/Device/Driver/AirControlDisplay.cpp
Scumi: done (and error caused by TCHAR usage on WIN64 fixed).

test/src/TestDriver.cpp
Scumi: added test to check against bogus value

src/Device/Driver/AirControlDisplay.cpp
Scumi: I checked and it looks good to me. Maybe because the if-sentence above is not wrapped in curly braces?

MaxKellermann:
There are still linker errors

src/InfoBoxes/Content/Type.hpp
```diff
@@ -144,14 +144,16 @@ namespace InfoBoxFactory
     e_NbrSat, /* Number of used Sat by GPS module */
 
     e_ActiveRadio, /* Active Radio Frequency */
-
     e_StandbyRadio, /* Standby Radio Frequency */
+
```
MaxKellermann, line 148: Whitespace change?

src/InfoBoxes/Content/Radio.hpp
```diff
@@ -18,3 +18,10 @@ class InfoBoxContentStandbyRadioFrequency : public InfoBoxContent
   const InfoBoxPanel *GetDialogContent() noexcept override;
   void Update(InfoBoxData &data) noexcept override;
 };
+
+class InfoBoxContentTransponderCode : public InfoBoxContent
+{
+public:
+  const InfoBoxPanel *GetDialogContent() noexcept override;
+  void Update(InfoBoxData &data) noexcept override;
+};
```
MaxKellermann, line 27: Newline missing here

Scumi:
> There are still linker errors

Ok. Need to check how to find them. I built for three targets and everything went fine.

MaxKellermann:
`make everything`

src/InfoBoxes/Content/Radio.hpp
Scumi: fixed

PR #1170, opened 2023-04-04
Various2

lordfolken:


src/Repository/FileType.hpp
```diff
@@ -14,5 +14,6 @@ enum class FileType : uint8_t {
   FLARMNET,
   IGC,
   RASP,
-  XCI
+  XCI,
+  TASK
```
MaxKellermann, line 18: Trailing comma!

lordfolken: To extend the enum I need to add that comma. I believe this to be correct.

MaxKellermann: Of course, but I mean you should add a trailing comma! The last guy (who added `XCI`) forgot it and now you had to modify an unrelated line.

lordfolken: Got it. I split the commit in 2, one to fix the missing comma, and the second one to add the task incl. trailing comma.

GeekGoneOld:
I am pretty sure Max will insist on a comma after “TASK”.  This format was introduced to C++ over a decade ago and is preferred as it makes adding to the list and swapping items much cleaner.

PR #1171, opened 2023-04-05
Horizon: Draw Horizon with an X when no AHRS data

lordfolken:
2023-04-05-220732_638x1056_scrot


MaxKellermann:
This doesn't look right. It does draw a horizon, but if there's no data, why does it draw one with fake data?

Issue #1051, opened 2022-09-16
RASP File Download save with wrong name convension


Developer66:
The RASP files are downloaded as e.g. "Germany (by Thermalmap.info).dat".
Before moving the download to the DownloadManager, it was xcsoar-rasp.dat.

If XCSoar get restarted the file can not be found again...
Maybe not bad at all, so i have always to download the newest data but quite not very well in all cases.

@lordfolken

lordfolken:
Agreed. I will have a look.

dspreitz:
Following manual work-around works:
Manually rename a downloaded RASP file (e.g. 'CzechRepublic (by Thermalmap.info).dat') to xcsoar-rasp.dat in the /Android/data/org.xcsoar/files/ folder.
After that you can go to Info --> Weather --> RASP and select a field and time.

roscosoarer:
The Canberra Gliding club in Australia routinely glides in wave to FL245. I note the RASP function but there is no RASP map for our region.  A RASP download for 200 m radius [or to the East] for 36S 148E called 'Australia-Cooma NSW'
![7168573100784187973](https://user-images.githubusercontent.com/67410740/202574559-aaf30d20-defc-4749-9d76-c367fbd97888.jpg)
 Soaring Area would be great.

lordfolken:
The rasp maps are rendered by volunteers around the world. We merely provide a link to them.
Maybe you can convince a rasp operator in Australia: http://www.ausrasp.com/
Instructions on how to generate the correct files:
http://www.drjack.info/twiki/bin/view/RASPop/XcSoar

The urls must be added here, from then on it will be available on all XCSoar versions greater than 7.27
https://github.com/XCSoar/xcsoar-data-content/tree/master/data/remote/rasp/region

si-gr:
Feel free to use my work https://github.com/XCSoar/XCSoar/pull/1180 https://github.com/XCSoar/XCSoar/pull/1175 in a new PR as I am seemingly not able to contribute to XCSoar.

PR #1175, opened 2023-04-14
Replace old RASP data file after download

si-gr:
Replaces the previously downloaded RASP data file with the just downloaded file. Fixes https://github.com/XCSoar/XCSoar/issues/1051

I am not quite sure if this is an accepted way to solve this issue but it works for me and I wanted to share it.
This solution of course also has the drawback of users no longer being able to download multiple RASP .dat files for different regions.

### DIFF
--- a/src/Dialogs/Weather/RASPDialog.cpp
+++ b/src/Dialogs/Weather/RASPDialog.cpp
@@ -14,6 +14,8 @@
 #include "UIState.hpp"
 #include "ActionInterface.hpp"
 #include "Language/Language.hpp"
+#include "system/FileUtil.hpp"
+#include "LocalPath.hpp"
 
 #include <stdio.h>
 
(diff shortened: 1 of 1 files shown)


MaxKellermann:
The PR contains text that is missing in the commit message. It would be best to never write a PR text, because that text will be gone after merging. Commit messages should be enough to understand your changes.

MaxKellermann:
Also, a code comment explaining why this rename is necessary would be fine.

MaxKellermann:
Sure it's bad that XCSoar hard-codes the RASP filename, that should be changed. But that's not what your PR is about, and that's not what my complaint is about. No matter how your code looks like, there must be a proper commit message, and maybe a code comment.

src/Dialogs/Weather/RASPDialog.cpp
```diff
@@ -115,11 +116,16 @@ RASPSettingsPanel::OnTimeModified(const DataFieldEnum &df) noexcept
...
 {
-  auto path = DownloadFilePicker(FileType::RASP);
-  if (path == nullptr)
+  auto downloaded_file_path = DownloadFilePicker(FileType::RASP);
+  if (downloaded_file_path == nullptr)
     return;
+  
+  auto rasp_file_path = AllocatedPath::Build(downloaded_file_path.GetParent(), RASP_FILENAME);
+  
+  // Replace the old RASP file with the downloaded one
```
MaxKellermann, line 125: This code comment is useless. It only states what is obvious from reading the code, but it doesn't answer the "why" question.

MaxKellermann:
CI still fails.

si-gr:
Should I close this or will this be merged?

MaxKellermann:
Close if if you're not going to address my complaint.

si-gr:
I think I addressed every complaint you had

MaxKellermann:
No. Your latest code comment reads:
```
// Update RASP data file
```
This comment is still completely useless because it still doesn't answer the "why" question.
I don't like repeating myself, this is tiring.

si-gr:
Okay, I don't understand what you want. You asked for a comment, I wrote a comment. I even changed the comment after you complained. Obviously your feedback is not helpful to me, otherwise I would have changed the comment to satisfy you. It's sad to see this PR not being merged because of a trifle but I can't be bothered anymore.

MaxKellermann:
Bye.

PR #1180, opened 2023-04-16
Dialogs/Weather/RASP: Fix now time selection

si-gr:
Selecting "Now" in the RASP weather dialog should result in an invalid BrokenTime object being generated. However, this is only the case right after startup.

This bug is noticeable when selecting a specific time for the RASP data and then switching to "now". The BrokenTime in this case contains a valid 0:0 time but should be invalid.

### DIFF
--- a/src/Dialogs/Weather/RASPDialog.cpp
+++ b/src/Dialogs/Weather/RASPDialog.cpp
@@ -107,7 +107,7 @@ inline void
 RASPSettingsPanel::OnTimeModified(const DataFieldEnum &df) noexcept
 {
   const int value = df.GetValue();
-  time = value >= 0
+  time = value > 0
     ? BrokenTime::FromMinuteOfDay(value)
     : BrokenTime::Invalid();
 }


MaxKellermann:
The PR contains text that is missing in the commit message.

[Please don't make me repeat myself](https://github.com/XCSoar/XCSoar/pull/1175#issuecomment-1508967180)

si-gr:
I obviously do not know what I need to include in the commit message and what to write in a PR. My understanding is that the PR is used to discuss the changes and why these changes are necessary, how they relate to bugs and so on. The commit message describes what the code does.
Please enlighten me about how things are done around here.

lordfolken:
Basically your initial PR text should be empty. All your justification, why and how should be in your commits. 
The review on the implementation can be in the follow up comments. But should something change you should update the text in your commits. 

The reasoning here is that the git history will be around as long as the project. All the discussions of github will be lost once we move to a new platform. XCSoar's history is over 20 years and predates both android and ios. The git messages are very helpful when seeing what decisions where made when looking at an old piece of code.

si-gr:
Thank you! I hope this new commit message is more suitable

si-gr:
Should I include this commit https://github.com/si-gr/XCSoar/commit/e5d207ea10f93496a6b3d002e6ea0984397e1303 aswell in this PR? Without it the "now" RASP overlay still does not work. Alternatively, I could create a new PR.

PR #1181, opened 2023-04-16
Pancrosshairkeyb

lordfolken:


lordfolken:
@linuxianer99 heads up to OV Team.
With this pan mode is finaly useable with a joystick only. 

I use the firmware for the remote stick that has long and short presses.

lordfolken:
@xcnav this may also be interesting for your stick remote.

XCNav:
> @XCNav this may also be interesting for your stick remote.

Indeed,  thank you for your great work.

PR #1149, opened 2023-03-09
Device/Driver/ATR833: use data delivered by ATR833

Scumi:
Brief summary of the changes
----------------------------

Extended driver to read-out and use data delivered by ATR833.

Related issues and discussions
------------------------------

prerequisite to #594

PR #1184, opened 2023-04-18
Various code improvements

Scumi:
Various code improvements based on lessons learnt from #1167 and #1149.


MaxKellermann:
Good stuff!

PR #1185, opened 2023-04-19
sync ACD with an other device that could send an external QNH to XCSoar

Blaubart:
if, in addition to the ACD, other devices are connected to XCSoar that can send a QNH to XCSoar, then the ACD synchronizes with this device using OnCalculatedUpdate. This ensures that the ACD does not use a different value for the QNH.

### DIFF
--- a/src/Device/Driver/AirControlDisplay.cpp
+++ b/src/Device/Driver/AirControlDisplay.cpp
@@ -43,7 +43,7 @@ ParsePAAVS(NMEAInputLine &line, NMEAInfo &info)
 
     if (line.ReadChecked(value)) {
       auto qnh = AtmosphericPressure::Pascal(value);
-      info.settings.ProvideQNH(qnh, info.clock);
+      info.settings.ProvideQNH(qnh,info.clock);
     }
   } else if (type == "COM"sv) {
     /*
@@ -124,6 +124,7 @@ public:
                            OperationEnvironment &env) override;
   bool PutTransponderCode(TransponderCode code, OperationEnvironment &env) override;
   void OnSensorUpdate(const MoreData &basic) override;
+  void OnCalculatedUpdate(const MoreData &basic, const DerivedInfo &calculated) override;
 };
 
 bool
@@ -199,6 +200,19 @@ ACDDevice::OnSensorUpdate(const MoreData &basic)
     FormatGPGGA(buffer, sizeof(buffer), basic);
     PortWriteNMEA(port, buffer, env);
   }
+}
+
+void
+ACDDevice::OnCalculatedUpdate(const MoreData &basic,[[maybe_unused]] const DerivedInfo &calculated)
+{
+   NullOperationEnvironment env;
+
+   if (basic.settings.qnh_available.IsValid()){
+	 char buffer[100];
+     unsigned qnh = basic.settings.qnh.GetPascal();
+     sprintf(buffer,"PAAVC,S,ALT,QNH,%u",qnh);
+     PortWriteNMEA(port, buffer, env);
+   }
 }
 
 static Device *
--- a/src/Device/Driver/AirControlDisplay.cpp
+++ b/src/Device/Driver/AirControlDisplay.cpp
@@ -43,7 +43,7 @@ ParsePAAVS(NMEAInputLine &line, NMEAInfo &info)
 
     if (line.ReadChecked(value)) {
       auto qnh = AtmosphericPressure::Pascal(value);
-      info.settings.ProvideQNH(qnh,info.clock);
+      info.settings.ProvideQNH(qnh, info.clock);
     }
   } else if (type == "COM"sv) {
     /*
@@ -208,7 +208,7 @@ ACDDevice::OnCalculatedUpdate(const MoreData &basic,[[maybe_unused]] const Deriv
    NullOperationEnvironment env;
 
    if (basic.settings.qnh_available.IsValid()){
-	 char buffer[100];
+     char buffer[100];
      unsigned qnh = basic.settings.qnh.GetPascal();
      sprintf(buffer,"PAAVC,S,ALT,QNH,%u",qnh);
      PortWriteNMEA(port, buffer, env);


MaxKellermann:
I won't review this because this PR contains a known-bad commit and a fixup commit.

MaxKellermann:
Don't move and modify code around in one single commit. For example, it hides the fact that you added a new field that is never used.
And the PR still contains a fixup commit.

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -39,7 +43,11 @@ ParsePAAVS(NMEAInputLine &line, NMEAInfo &info)
 
     if (line.ReadChecked(value)) {
       auto qnh = AtmosphericPressure::Pascal(value);
-      info.settings.ProvideQNH(qnh, info.clock);
+      info.settings.ProvideQNH(qnh,info.clock);
+      return true;
+    }
+    else {
+      return true;
```
lordfolken, line 50: So very true, and not needed.

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -185,4 +230,4 @@ const struct DeviceRegister acd_driver = {
   _T("Air Control Display"),
   DeviceRegister::RECEIVE_SETTINGS | DeviceRegister::SEND_SETTINGS,
   AirControlDisplayCreateOnPort,
-};
+};
```
lordfolken, line 233: please fix your editor to append proper line breaks.

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -174,6 +185,40 @@ ACDDevice::ParseNMEA(const char *_line, NMEAInfo &info)
...
+  }
+}
+
+void
+ACDDevice::OnCalculatedUpdate(const MoreData &basic,const DerivedInfo &calculated)
+{
+   NullOperationEnvironment env;
+
+    if (basic.settings.qnh_available.IsValid()){
+	  char buffer[100];
```
lordfolken, line 215: indent

Blaubart:
Sorry!! I will fix this and test the driver again. If everything works I'll reopen the PR

lordfolken:
You can also concert to draft. :)

Blaubart:
I don't understand why checks were not successful. Yesterday I published almost the same code and checks were successful

lordfolken:
Scroll a bit up in the logs:
src/Device/Driver/AirControlDisplay.cpp:210:72: error: unused parameter ‘calculated’ [-Werror=unused-parameter]

Blaubart:
@lordfolken Thanks for your friendly support!!
@MaxKellermann I hat compiled before!! I build the OV image it works.

MaxKellermann:
> @MaxKellermann I hat compiled before!! I build the OV image it works.

But you ignored the compiler warnings.

The OV build is not a good test at all. OV builds are with debugging disabled, and thus with `WERROR=n`. Compiler warnings are displayed, but if you don't pay attention, you will miss them, because they are non-fatal.

Lots of mistakes can slip through if you only do an OV build. Don't.

Blaubart:
OK, thanks for this Information!! I will compile on a different way next time!!

MaxKellermann:
Just a note: I'll unsubscribe here and I won't see any further updates. If you are confident that your work is complete and your code is good enough for merging, you need to notify me explicitly.

Blaubart:
OK

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -201,6 +202,19 @@ ACDDevice::OnSensorUpdate(const MoreData &basic)
...
+void
+ACDDevice::OnCalculatedUpdate(const MoreData &basic,[[maybe_unused]] const DerivedInfo &calculated)
+{
+   NullOperationEnvironment env;
+
+   if (basic.settings.qnh_available.IsValid()){
+	 char buffer[100];
+     unsigned qnh = basic.settings.qnh.GetPascal();
+     sprintf(buffer,"PAAVC,S,ALT,QNH,%u",qnh);
+     PortWriteNMEA(port, buffer, env);
```
MaxKellermann, line 214: This function can block. It must not be called from within `OnCalculatedUpdate()`, because the latter must never block.

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -201,6 +202,19 @@ ACDDevice::OnSensorUpdate(const MoreData &basic)
   }
 }
 
+void
+ACDDevice::OnCalculatedUpdate(const MoreData &basic,[[maybe_unused]] const DerivedInfo &calculated)
+{
+   NullOperationEnvironment env;
+
+   if (basic.settings.qnh_available.IsValid()){
+	 char buffer[100];
```
MaxKellermann, line 211: bad indent

src/Device/Driver/AirControlDisplay.cpp
```diff
@@ -43,7 +43,7 @@ ParsePAAVS(NMEAInputLine &line, NMEAInfo &info)
 
     if (line.ReadChecked(value)) {
       auto qnh = AtmosphericPressure::Pascal(value);
-      info.settings.ProvideQNH(qnh, info.clock);
+      info.settings.ProvideQNH(qnh,info.clock);
```
MaxKellermann, line 46: What's the point of this change?

src/Device/Driver/AirControlDisplay.cpp
MaxKellermann: It seems @lordfolken already complained about this one, but you ignored his complaint.

src/Device/Driver/AirControlDisplay.cpp
Blaubart: reversed

src/Device/Driver/AirControlDisplay.cpp
Blaubart: Fixed

src/Device/Driver/AirControlDisplay.cpp
Blaubart: you mean this function: sprintf?

MaxKellermann: No, 
1. this isn't the line I commented on
2. how is sprintf supposed to block?

PortWriteNMEA() can block, and you circumvented the API protection by using a NullOperationEnvironment, just to make it compile, but without knowing the implications. Don't do random stuff just to make it compile! If you don't understand your own code, it's bad code.

MaxKellermann:
Closing due to lack of bugfix.

Issue #1178, opened 2023-04-16
Support for Bluetooth LE engine sensor box


fastdown101:
XCSoar version 7.9
--------------

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------
I would like to have support for a engine sensor box in the Android implementation. I'm the developer of the sensor box.

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------
The sensor box is aimed to be used by powered paragliding pilots. The sensor box gathers engine related sensor values as well as atmospheric pressure, which should be shown in xcs. Relying on the engine while flying powered paragliders, I find it crucial to have engine "health" measurements like Exhaust Gas Temperature (EGT) readings. With xcs, the sensor box would be a full blown flight computer.

What have I done so far?
-------------------------------------------------------------------
I branched locally from master. Added my "driver" implementation down to "void DeviceDescriptor::OnEngineSensboxSensors(...)" in AndroidSensors.cpp. I guess this is where the the parsed sensor values reach the native c++ code? How to proceed from there?

Where do I need help?
-------------------------------------------------------------------
How can I add support to get the sensor values displayed? Some further calculations of the sensor values are needed too, as well as calculations depending on some user configuration. How can I upstream my branch? Any help will be greatly appreciated!

Some details on the sensor values
-------------------------------------------------------------------
The sensor box does not do any calculations. For instance, the engine Revolutions Per Minute (RPM) value has to be calculated by xcs from revolutions per second. Depending on the used engine there might be 2 ignitions per revolution. So the user has to configure xcs correctly depending on the engine used, to get correct readings for the RPM value.
It would be great to have the following values displayed:
engine Revolutions Per Minute (RPM), two high temperature sensors up to 1024 Degree Celsius e.g. Cylinder Head Temperature (CHT)  and Exhaust Gas Temperature (EGT), QNH based on the barometric pressure, ambient temperature and optional sensor health flags

MaxKellermann:
How is this data going to be transmitted? As BLE characteristics?

First you need to add missing methods for transmitting those values to Java `class SensorListener`, and implementations to forward them into native C++ code to `class NativeSensorListener`, and the C++ side needs to have the same methods in `class SensorListener`.

[...]

PR #1187, opened 2023-04-20
Add engine sensor box

fastdown101:
Brief summary of the changes
----------------------------

Add support for a Bluetooth LE engine sensor box. 

### DIFF
--- a/src/NMEA/EngineState.hpp
+++ b/src/NMEA/EngineState.hpp
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The XCSoar Project
+
+#pragma once
+
+#include "Validity.hpp"
+#include "time/Stamp.hpp"
+#include "Atmosphere/Temperature.hpp"
+
+/** Information about engine sensors.
+*/
+struct EngineState
+{
+  /* The engine box measured revolutions per second on the camshaft. */
+  Validity revs_per_sec_available;
+  uint16_t revs_per_sec;
+
+  /* The engine Cylinder Head Temperature (CHT) */
+  Validity cht_temperature_available;
+  Temperature cht_temperature;
+
+  /* The engine Exhaust Gas Temperature (EGT) */
+  Validity egt_temperature_available;
+  Temperature egt_temperature;
+
+  void Clear() noexcept{
+    revs_per_sec_available.Clear();
+    cht_temperature_available.Clear();
+    egt_temperature_available.Clear();
+  }
+
+  void Reset() noexcept{
+    Clear();
+  }
+
+  void Expire(TimeStamp clock) noexcept {
+    revs_per_sec_available.Expire(clock, std::chrono::seconds(3));
+    cht_temperature_available.Expire(clock, std::chrono::seconds(3));
+    egt_temperature_available.Expire(clock, std::chrono::seconds(3));
+  }
+
+  void Complement(const EngineState &add) noexcept {
+    if (revs_per_sec_available.Complement(add.revs_per_sec_available)){
+      revs_per_sec = add.revs_per_sec;
+    }
+    if (cht_temperature_available.Complement(add.cht_temperature_available)){
+      cht_temperature = add.cht_temperature;
+    }
+    if (egt_temperature_available.Complement(add.egt_temperature_available)){
+      egt_temperature = add.egt_temperature;
+    }
+  }
+
+};
--- a/src/NMEA/Info.hpp
+++ b/src/NMEA/Info.hpp
@@ -15,6 +15,7 @@
 #include "Atmosphere/Pressure.hpp"
 #include "Atmosphere/Temperature.hpp"
 #include "DeviceInfo.hpp"
+#include "EngineState.hpp"
 #include "FLARM/Data.hpp"
 #include "Geo/SpeedVector.hpp"
 
@@ -300,6 +301,8 @@ struct NMEAInfo {
    * @see HumidityAvailable
    */
   double humidity;
+
+  EngineState engine_state;
 
   //###########
   //   Other
--- a/src/NMEA/Info.cpp
+++ b/src/NMEA/Info.cpp
@@ -155,6 +155,7 @@ NMEAInfo::Reset()
   flarm.Clear();
 
 #ifdef ANDROID
+  engine_state.Reset();
   glink_data.Clear();
 #endif
 }
@@ -222,6 +223,7 @@ NMEAInfo::Expire()
   battery_level_available.Expire(clock, std::chrono::minutes(5));
   flarm.Expire(clock);
 #ifdef ANDROID
+  engine_state.Expire(clock);
   glink_data.Expire(clock);
 #endif
   attitude.Expire(clock);
@@ -341,6 +343,7 @@ NMEAInfo::Complement(const NMEAInfo &add)
   flarm.Complement(add.flarm);
 
 #ifdef ANDROID
+  engine_state.Complement(add.engine_state);
   glink_data.Complement(add.glink_data);
 #endif
 }


MaxKellermann:
Your way of splitting commits doesn't make sense and makes code review very hard. For example, those "Refactor ..." commits refactor code that was added in previous commits in this PR. Therefore, reviewing those older commits is now a waste of time because they're outdated, but the burden of finding out what is old and outdated should not be put on reviewers.

fastdown101:
> Your way of splitting commits doesn't make sense and makes code review very hard. For example, those "Refactor ..." commits refactor code that was added in previous commits in this PR. Therefore, reviewing those older commits is now a waste of time because they're outdated, but the burden of finding out what is old and outdated should not be put on reviewers.

Okay, how can I fix it? New branch where I split the commits differently?
There is still quite some work to do. Documentation and the RPM functionality is still not finished. 
However, I thought it might be a good idea to get early reviews and comment on the code.

MaxKellermann:
> Okay, how can I fix it? New branch where I split the commits differently?

No new branch, no new PR. This only causes more confusion but will not solve the problem.

I recommend using stgit to manage patches that are being submitted. stgit allows you to create a stack of patches which you can update, and when you push that stack to GitHub, each patch is a commit. So get stgit, convert the existing commits to stgit patches (`stg uncommit -n14`), then fold those "refactor"/"minor fix" patches into the patches that are being fixed.

Put the commit that introduces the new fields to NMEAInfo at the very top (please in a separate `struct EngineState` to avoid cluttering NMEAInfo further), then the commit that implements the InfoBox content. Once you have that, I can already merge those commits, and you can continue working on the rest. A patch that lets users configure the scaling factor, a patch that adds the new method to `SensorListener` (this may include the Java bindings, or maybe you decide to make that a separate commit).

Once all of this infrastructure is in place (and maybe already merged), you can polish your new BLE driver, which should be a separate patch that contains just that, the BLE driver, and nothing else. Because the rest is already there.

fastdown101:
> please in a separate struct EngineState to avoid cluttering NMEAInfo further

Done that.

But I don't get stg... I can do the uncommit, but how to add commits to a patch? Is there a graphical tool like git-cola which shows the stg patch and diffs and stuff?

Was able to do some test flights with the engine sensbox support in xcsoar. It is great and I think it can replace my current setup.

fastdown101:
screenshot xcsoar

Absolutely loving it :)

src/NMEA/Info.hpp
```diff
@@ -301,6 +302,13 @@ struct NMEAInfo {
    */
   double humidity;
 
+  //###########
+  //   Engine sensor box
+  //###########
+
+  bool engine_state_available;
```
MaxKellermann, line 309: What's the point of this field? It's never initialised, never used, and `EngineState` has its own availability management.

fastdown101: Good point, how to proceed? Delete it with a refactor commit?

MaxKellermann: No, amend that commit! (`stg refresh`)

fastdown101: Do all commits end up in the master branch? How is the process? Don't they get squashed into one feature commit?

MaxKellermann: No, I don't squash. Having separate commits for different aspects of a PR is very valuable, and that value doesn't go away by merging. Managing the commits and their separation is your task.

MaxKellermann:
You can do it with plain git if you like, `git commit --amend`, stgit is just a tool on top of that, one that I recommend because I can't imagine working without it.

src/NMEA/EngineState.hpp
```diff
@@ -0,0 +1,62 @@
...
+    }
+    if (temp1_available.Complement(add.temp1_available)){
+      temp1 = add.temp1;
+    }
+    if (temp2_available.Complement(add.temp2_available)){
+      temp2 = add.temp1;
+    }
+  }
+
+};
```
MaxKellermann, line 62: This line does not end with a newline, thus the text file is malformed

src/NMEA/EngineState.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include "time/Stamp.hpp"
+#include "Atmosphere/Temperature.hpp"
+
+/** Information about the engine sensor box values,
+ *  temperatures and revolutions per second
+*/
+struct EngineState
+{
+  Validity sensbox_flags_available;
+  uint8_t sensbox_flags;
```
MaxKellermann, line 16: What is this? This sounds like it is very specific to your product, but it is undocumented. There should be documentation explaining that this is a proprietary attribute of your product, and there should be complete documentation on the meaning of the value.

src/NMEA/EngineState.hpp
```diff
@@ -0,0 +1,62 @@
...
+  uint16_t revs_per_sec;
+
+  Validity temp1_available;
+  Temperature temp1;
+
+  Validity temp2_available;
+  Temperature temp2;
+
+  void Clear() noexcept{
+    sensbox_flags = 0;
```
MaxKellermann, line 28: What's the point of initializing this field to this arbitrary value?

src/NMEA/EngineState.hpp
```diff
@@ -0,0 +1,62 @@
...
+    sensbox_flags_available.Clear();
+    revs_per_sec = 0;
```
MaxKellermann, line 30: Same here - why initialize?

src/NMEA/EngineState.hpp
```diff
@@ -0,0 +1,62 @@
...
+    revs_per_sec_available.Clear();
+    temp1_available.Clear();
+    temp2_available.Clear();
+  }
+
+  void Reset() noexcept{
+    Clear();
+  }
+
+  void Expire(TimeStamp clock, std::chrono::steady_clock::duration max_age) noexcept {
```
MaxKellermann, line 40: Why is max_age a parameter? In all other structs, this is hard-coded. And there is just one caller, which hard-codes the value.

MaxKellermann: btw. do you know what 3 seconds mean? Is this value your decision, or did you just copy it from somewhere else?

src/NMEA/EngineState.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include "Atmosphere/Temperature.hpp"
+
+/** Information about the engine sensor box values,
+ *  temperatures and revolutions per second.
+ *  Update rate of the engine sensor box is 1 Hz.
+*/
+struct EngineState
+{
+  /** Just one of the 8 bit is used as Flag.
+   *  Lowest bit 0b00000001 = 0x1 set to 1, indicates a valid RPM value.
```
MaxKellermann, line 17: This is redundant - we already have `revs_per_sec_available`. I think we don't need this field at all.

src/NMEA/EngineState.hpp
```diff
@@ -0,0 +1,74 @@
...
+   *  All other bits are reseved for further use.
+  */
+  Validity sensbox_flags_available;
+  uint8_t sensbox_flags;
+
+  // The engine box measured revolutions per second.
```
MaxKellermann, line 23: API documentation should be doxygen-style. And the meaning should be clarified - what rotation is being measured? The engine's rotation? Before or after the gearbox/strap?

src/NMEA/EngineState.hpp
```diff
@@ -0,0 +1,74 @@
...
+// Copyright The XCSoar Project
+
+#pragma once
+
+#include "Validity.hpp"
+#include "time/Stamp.hpp"
+#include "Atmosphere/Temperature.hpp"
+
+/** Information about the engine sensor box values,
+ *  temperatures and revolutions per second.
```
MaxKellermann, line 11: Most of this comment is just a summary of the fields below. What's the point of that? I think just "Data from sensors in the engine" is enough.

src/NMEA/EngineState.hpp
fastdown101: What will happen when there is no sensor box update for more than 3 seconds? Does the RPM reading goes into invalid without the mechanism of the Validity class?

src/NMEA/EngineState.hpp
```diff
@@ -0,0 +1,74 @@
...
+ *  Update rate of the engine sensor box is 1 Hz.
```
MaxKellermann, line 12: It would be useful to document the update rate of your product somewhere, but this is the wrong place. This should be a generic struct holding information about the engine. It should be possible to use this struct for other engine sensors. Having this information here is too specific for no reason, because what if another guy makes a sensor box with 10 Hz update rate - will they be able to use this struct, or will they have to invent a new one?

src/NMEA/EngineState.hpp
fastdown101: For future use, the sensor box will signal a emergency engine shutdown by setting bit 0x1 to 0.

src/NMEA/EngineState.hpp
```diff
@@ -0,0 +1,74 @@
...
+  Validity sensbox_flags_available;
+  uint8_t sensbox_flags;
+
+  // The engine box measured revolutions per second.
+  Validity revs_per_sec_available;
+  uint16_t revs_per_sec;
+
+  /** temp1 is a generic temperature sensor. The InfoBox showing
+   *  this value has a user configurable title. The user can choose
+   *  3 different titles: {CHT, EGT, WTR}
```
MaxKellermann, line 29: What the .... is CHT EGT WTR? (WTF)
Why does the user have to choose that? Why two temperature sensors? 
What if some piece of XCSoar code needs to have the "EGT" temperature - how will it obtain that?
Your design is bad because it requires complexity everywhere the value is used. A better design could be:
```
Temperatur cht_temperature, egt_temperature, wtr_temperature;
```
Then the driver has to figure out how to fill these fields. Not the UI code.
The UI code doesn't know about the device configuration, and should not care. All it has to do is well-defined read values from NMEAInfo.

src/NMEA/EngineState.hpp
MaxKellermann: Since you chose a 3s timeout, the value will be considered invalid 3 seconds after the most recent value was received from the sensor, and the InfoBox should show "---" instead of that value.

MaxKellermann: > For future use, the sensor box will signal a emergency engine shutdown by setting bit 0x1 to 0.

I don't understand the exact semantics from your sentence, but this still doesn't justify having this field here - because it's a one-way street. It will never be propagated back to the device. NMEAInfo is just stuff from devices/sensors into XCSoar, but never back.

src/NMEA/EngineState.hpp
fastdown101: Look, this design lets the user of xcs choose for different meanings of temp1 and temp2, depending on the wiring and use of the temperature sensors on the engine. In my mind the user should have the flexibility to enter arbitrary text for the InfoBox titles for temp1 and temp2, but I think you would not agree, right? But let the user choose would be most generic.  It could then be for instance temp1 = BAT = Battery 1 of a electric engine, or temp1 = WTR = Water Temperature of a water cooled combustion engine and so forth...

So my suggestion would be: Let it be temp1 & temp2 in the code. Let the user of xcs choose title text for it.

MaxKellermann: No, this is a bad design, and I have explained why. You did not address my criticism, and you didn't answer my question.

fastdown101: Okay, so I create variables for cht_temperature (1 cylinder machine) , egt_temperature, wtr_temperature, bat_temperature, egt2_temperature (second cylinder egt), cht2_temperature (second cylinder), ee_temperature (electrical engine temperature), esc_temperature (electric speed controller) and so on?

fastdown101: It will be obtained by a cold junction compensated MAX6675 type-K thermoelement. Two of those will measure temp1 & temp2

MaxKellermann: Your documentation explicitly said there are only 3 choices, but now you present 8 choices This is a new story. What's up?

MaxKellermann: > It will be obtained by a cold junction compensated MAX6675 type-K thermoelement. Two of those will measure temp1 & temp2

Is that supposed to be an answer to one of my questions? I don't understand.

fastdown101: Okay, I surrender. No user option. Just cht_temperature and egt_temperature. I don't understand either...

src/NMEA/EngineState.hpp
fastdown101: > It will never be propagated back to the device.

Does not need to. RPM will go "---" all other sensor box values will be != "---"  => user knows something is not in order with the engine e.g. emergency shut down of engine.

src/NMEA/EngineState.hpp
Scumi: I think it is not really clear what your box exactly does. Is it a more generic box with two temperature readings that "could" potentially anywhere? Or is your box more tied to a specific engine model (with fixed EGT and CHT readings) and other boxes "might" use WTR, BAT, etc... later? Based on that you have different design choices for the software.

fastdown101: I tried to explain... Now it is just for RPM, CHT and EGT. Specific engine.  No choices for the software

fastdown101:
Sorry, I was not aware. Switched machines and thought I configured both the same...Kind of an exhausting workflow

MaxKellermann:
Your workflow seems strange. Amending patches or creating a new patch isn't something that I can configure; it's something that I do.

src/NMEA/EngineState.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+/** Information about engine sensors.
+*/
+struct EngineState
+{
+  /* The engine box measured revolutions per second on the camshaft. */
+  Validity revs_per_sec_available;
+  uint16_t revs_per_sec;
+
+  /* The engine Cylinder Head Temperature (CHT) */
```
MaxKellermann, line 18: There's more trailing space here

fastdown101:
> Your workflow seems strange. Amending patches or creating a new patch isn't something that I can configure; it's something that I do.

Disagree, when you switch developer machine stg has to be configured the same on both machines, because stg has only local meta data, right? At least that's how I understood stgit

src/NMEA/EngineState.hpp
fastdown101: damit

MaxKellermann:
> Disagree, when you switch developer machine stg has to be configured the same on both machines, because stg has only local meta data, right? At least that's how I understood stgit

True, but your stgit configuration will not lead to a fixup commit by accident. Creating a fixup commit is an explicit action by you, not a configuration. You cannot be "not aware" of creating a new commit.

fastdown101:
> You cannot be "not aware" of creating a new commit.

True, I missed that a `stg new xyz` on a different machine creates a commit for the "new" patch...obviously I named it like the patch I am working on. There is no miracles in technical stuff and physics beats everything :)

MaxKellermann:
> True, I missed that a `stg new xyz` on a different machine creates a commit for the "new" patch

It does so an all machines, no matter if it's the one you used previously or the different one you're using now.

PR #1192, opened 2023-04-25
Add InfoBoxes for EngineState

fastdown101:
Brief summary of the changes
----------------------------
Engine status values should be shown in InfoBoxes

https://github.com/XCSoar/XCSoar/commit/3b4bbc009adf288fce5484c763ce591bb4cfe9c1


src/InfoBoxes/Content/Type.hpp
```diff
@@ -146,13 +146,20 @@ namespace InfoBoxFactory
     e_ActiveRadio, /* Active Radio Frequency */
 
     e_StandbyRadio, /* Standby Radio Frequency */
+
+    e_TransponderCode, /* Transponder code */
```
MaxKellermann, line 150: .... what?

src/InfoBoxes/Content/Engine.cpp
```diff
@@ -0,0 +1,49 @@
...
+void
+InfoBoxContentRPM::Update(InfoBoxData &data) noexcept
+{
+  const NMEAInfo &basic = CommonInterface::Basic();
+  if (!basic.engine_state.revs_per_sec_available.IsValid()) {
+    data.SetInvalid();
+    return;
+  }
+  // display as revolutions per minute / 1000, so that for instance 8400 RPM are shown as the more compact 8.40
+  data.SetValue(_T("%1.2f"), basic.engine_state.revs_per_sec * (60.0/1000.0));
```
MaxKellermann, line 48: Unit is missing, and dividing by 1000 seems rather unintuitive

src/InfoBoxes/Content/Type.hpp
fastdown101: Yea, how did that happen?!? I will fix it.

src/InfoBoxes/Content/Engine.cpp
fastdown101: Unit would be 1/(60 s) and with common scaling to 1000th of RPMs, the look of the unit gets worse. You insist on a unit for RPMs?

Scumi: I have not much say in this and will probably not use it anytime soon, but I would also prefer "RPM". Are there digital displays that show RPM as a fraction? I know analogue ones do for obvious reasons, but that does not apply here.
Anyway, we show larger values (for example RadioFrequency) in the infoboxes, so 4-5 digits for engine RPM does not sound excessive to me at least.

Scumi: #1193

MaxKellermann: > So "RPM" as unit in the InfoBox? Do I have to extend `class Units`?

In the long run, we should do that, but for now, I'd agree to merge something that sets "rpm" as "Comment". Not quite consistent, but barely good enough to get it merged.

PR #1194, opened 2023-04-25
InfoBoxes: Introduce a comment value font

### LINKED ISSUE #1193: Long InfoBox values displayed too small
Issue starts at 7bf77a3b380689f64f0ddfb4ea2dbace072f3115

Expected behavior
-----------------
Long infobox values taking all available space and are therefor shown with a normal size.

InfoBox_normal_value_size

Actual behavior
---------------
Infobox values have a margin around them, which reduces the size of long values.

InfoBox_small_value_size

Use infoboxes that can contain longer values (i.e. frequency infoboxes).

7bf77a3b380689f64f0ddfb4ea2dbace072f3115

Revert commit (quick fix) or improve logic.

lordfolken:
2023-04-25-215135_480x640_scrot

### DIFF
--- a/src/InfoBoxes/InfoBoxWindow.cpp
+++ b/src/InfoBoxes/InfoBoxWindow.cpp
@@ -140,7 +140,7 @@ InfoBoxWindow::PaintComment(Canvas &canvas)
 
   canvas.SetTextColor(look.GetCommentColor(data.comment_color));
 
-  const Font &font = look.title_font;
+  const Font &font = look.comment_font;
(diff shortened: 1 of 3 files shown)


MaxKellermann:
Now nobody uses small_value_font, so this effectively just reverts 7bf77a3b380689f64f0ddfb4ea2dbace072f3115 - just a complicated revert.

MaxKellermann:
I don't understand your response - your previous change has lost all its effect with this PR.

PR #1196, opened 2023-04-26
Add device and java bindings for EngineState

fastdown101:
Brief summary of the changes
----------------------------
Add device implementation and java bindings for EngineState

https://github.com/XCSoar/XCSoar/commit/52d32fb6913c9e0300a583168383af0d6ca135f9
https://github.com/XCSoar/XCSoar/commit/3b4bbc009adf288fce5484c763ce591bb4cfe9c1

### DIFF
--- a/android/src/NativeSensorListener.java
+++ b/android/src/NativeSensorListener.java
@@ -34,6 +34,14 @@ final class NativeSensorListener implements SensorListener {
   public native void onAccelerationSensor1(double acceleration);
 
   @Override
+  public native void onEngineSensors(boolean has_cht_temp,
+                                     int cht_temp,
+                                     boolean has_egt_temp,
+                                     int egt_temp,
+                                     boolean has_revs_per_sec,
+                                     int revs_per_sec);
+
+  @Override
   public native void onAccelerationSensor(float ddx, float ddy, float ddz);
 
   @Override
--- a/android/src/SensorListener.java
+++ b/android/src/SensorListener.java
@@ -30,6 +30,20 @@ public interface SensorListener {
   void onI2CbaroSensor(int index, int sensorType, int pressure);
   void onVarioSensor(float vario);
   void onHeartRateSensor(int bpm);
+    /**
+   * @param[in] has_cht_temp Is the Engine Cylinder Head Temperature sensor present?
+   * @param[in] cht_temp Engine Cylinder Head Temperature.
+   * @param[in] has_egt_temp Is the Engine Exhaust Gas Temperature sensor present?
+   * @param[in] egt_temp Engine Exhaust Gas Temperature.
+   * @param[in] has_revs_per_sec Is the Engine Revolutions Per Second sensor present?
+   * @param[in] revs_per_sec Engine Revolutions Per Second, rotations per seconds of the camshaft.
+   */
+  void onEngineSensors(boolean has_cht_temp,
+                       int cht_temp,
+                       boolean has_egt_temp,
+                       int egt_temp,
+                       boolean has_revs_per_sec,
+                       int revs_per_sec);
 
   void onVoltageValues(int temp_adc, int voltage_index, int volt_adc);
 
--- a/src/Android/NativeSensorListener.cpp
+++ b/src/Android/NativeSensorListener.cpp
@@ -181,6 +181,30 @@ Java_org_xcsoar_NativeSensorListener_onI2CbaroSensor(JNIEnv *env, jobject obj,
 
 gcc_visibility_default
 JNIEXPORT void JNICALL
+Java_org_xcsoar_NativeSensorListener_onEngineSensors(JNIEnv *env,
+                                                     jobject obj,
+                                                     jboolean has_cht_temp,
+                                                     jint cht_temp,
+                                                     jboolean has_egt_temp,
+                                                     jint egt_temp,
+                                                     jboolean has_revs_per_sec,
+                                                     jint revs_per_sec)
+{
+  jlong ptr = env->GetLongField(obj, NativeSensorListener::ptr_field);
+  if (ptr == 0)
+    return;
+
+  auto &listener = *(SensorListener *)ptr;
+  listener.OnEngineSensors(has_cht_temp,
+                           Temperature::FromKelvin(cht_temp),
+                           has_egt_temp,
+                           Temperature::FromKelvin(egt_temp),
+                           has_revs_per_sec,
+                           revs_per_sec);
+}
+
+gcc_visibility_default
+JNIEXPORT void JNICALL
 Java_org_xcsoar_NativeSensorListener_onVarioSensor(JNIEnv *env,
                                                    jobject obj,
                                                    jfloat vario)
--- a/src/Device/AndroidSensors.cpp
+++ b/src/Device/AndroidSensors.cpp
@@ -288,6 +288,39 @@ DeviceDescriptor::OnHeartRateSensor(unsigned bpm) noexcept
 }
 
 void
+DeviceDescriptor::OnEngineSensors(bool has_cht,
+                                  Temperature cht,
+                                  bool has_egt,
+                                  Temperature egt,
+                                  bool has_revs_per_sec,
+                                  uint16_t revs_per_sec) noexcept
+{
+  const auto e = BeginEdit();
+  NMEAInfo &basic = *e;
+  basic.UpdateClock();
+  basic.alive.Update(basic.clock);
+  if(has_revs_per_sec){
+    basic.engine_state.revs_per_sec = revs_per_sec;
+    basic.engine_state.revs_per_sec_available.Update(basic.clock);
+  }else{
+    basic.engine_state.revs_per_sec_available.Clear();
+  }
+  if(has_cht){
+    basic.engine_state.cht_temperature = cht;
+    basic.engine_state.cht_temperature_available.Update(basic.clock);
+  }else{
+    basic.engine_state.cht_temperature_available.Clear();
+  }
+  if(has_egt){
+    basic.engine_state.egt_temperature = egt;
+    basic.engine_state.egt_temperature_available.Update(basic.clock);
+  }else{
+    basic.engine_state.egt_temperature_available.Clear();
+  }
+  e.Commit();
+}
+
+void
 DeviceDescriptor::OnVoltageValues(int temp_adc, unsigned voltage_index,
                                   int volt_adc) noexcept
 {
--- a/src/Device/Descriptor.hpp
+++ b/src/Device/Descriptor.hpp
@@ -620,6 +620,12 @@ private:
                        AtmosphericPressure pressure) noexcept override;
   void OnVarioSensor(float vario) noexcept override;
   void OnHeartRateSensor(unsigned bpm) noexcept override;
+  void OnEngineSensors(bool has_cht,
+                       Temperature cht,
+                       bool has_egt,
+                       Temperature egt,
+                       bool has_revs_per_sec,
+                       uint16_t revs_per_sec) noexcept override;
   void OnVoltageValues(int temp_adc, unsigned voltage_index,
                        int volt_adc) noexcept override;
   void OnNunchukValues(int joy_x, int joy_y,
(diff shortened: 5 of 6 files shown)


src/Android/NativeSensorListener.cpp
```diff
@@ -179,6 +179,25 @@ Java_org_xcsoar_NativeSensorListener_onI2CbaroSensor(JNIEnv *env, jobject obj,
                            AtmosphericPressure::Pascal(pressure));
 }
 
+gcc_visibility_default
+JNIEXPORT void JNICALL
+Java_org_xcsoar_NativeSensorListener_onEngineSensboxSensors(JNIEnv *env,
```
MaxKellermann, line 184: No such Java method exists

src/Android/NativeSensorListener.cpp
```diff
@@ -179,6 +179,25 @@ Java_org_xcsoar_NativeSensorListener_onI2CbaroSensor(JNIEnv *env, jobject obj,
...
+                                                       jobject obj,
+                                                       jint status_flags,
```
MaxKellermann, line 186: What is `status_flags`??

src/Device/Descriptor.hpp
```diff
@@ -620,6 +620,13 @@ class DeviceDescriptor final
                        AtmosphericPressure pressure) noexcept override;
   void OnVarioSensor(float vario) noexcept override;
   void OnHeartRateSensor(unsigned bpm) noexcept override;
+  void OnEngineSensboxSensors(int status_flags,
```
MaxKellermann, line 623: Why the name `OnEngineSensboxSensors`? Why not just `OnEngineSensor()`?

src/Android/NativeSensorListener.cpp
MaxKellermann: What Java files?
Did you understand my comment? That your PR does not make sense that way?

src/Device/AndroidSensors.cpp
```diff
@@ -287,6 +288,49 @@ DeviceDescriptor::OnHeartRateSensor(unsigned bpm) noexcept
   e.Commit();
 }
 
+void
+DeviceDescriptor::OnEngineSensors(uint16_t cht_temp,
+                                      uint16_t egt_temp,
+                                      uint16_t oat_temp,
+                                      uint32_t atmospheric_pressure,
+                                      uint16_t revs_per_sec) noexcept
```
MaxKellermann, line 296: Indentation is still broken and you're still using the wrong types

src/Device/AndroidSensors.cpp
```diff
@@ -287,6 +288,49 @@ DeviceDescriptor::OnHeartRateSensor(unsigned bpm) noexcept
...
+{
+  const auto e = BeginEdit();
+  NMEAInfo &basic = *e;
+  basic.UpdateClock();
+  basic.alive.Update(basic.clock);
+  basic.engine_state.revs_per_sec = revs_per_sec;
+  basic.engine_state.revs_per_sec_available.Update(basic.clock);
+  // 0 pascal is a invalid value.
```
MaxKellermann, line 304: The documentation for the API contract does not belong into an arbitrary implementation of the interface, but in the interface API documentation.

src/Device/Descriptor.hpp
```diff
@@ -620,6 +620,12 @@ class DeviceDescriptor final
                        AtmosphericPressure pressure) noexcept override;
   void OnVarioSensor(float vario) noexcept override;
   void OnHeartRateSensor(unsigned bpm) noexcept override;
+  void OnEngineSensors(uint16_t cht_temp,
+                              uint16_t egt_temp,
+                              uint16_t oat_temp,
+                              uint32_t atmospheric_pressure,
+                              uint16_t revs_per_sec) noexcept override;
```
MaxKellermann, line 627: Bad indent

src/Device/SensorListener.hpp
```diff
@@ -44,6 +44,11 @@ class SensorListener {
                                AtmosphericPressure pressure) noexcept = 0;
   virtual void OnVarioSensor(float vario) noexcept = 0;
   virtual void OnHeartRateSensor(unsigned bpm) noexcept = 0;
+  virtual void OnEngineSensors(uint16_t cht_temp,  /* Engine Cylinder Head Temperature (CHT)*/
+                                uint16_t egt_temp,  /* Engine Exhaust Gas Temperature (EGT)*/
+                                uint16_t oat_temp,  /* Outside Air Temperature (OAT)*/
+                                uint32_t atmospheric_pressure,  /* The Static Atmosperic Pressure */
+                                uint16_t revs_per_sec) noexcept = 0;  /* Engine Revolutions Per Seconds */
```
MaxKellermann, line 51: Bad indent, and please use doxygen-style API documentation

src/Android/NativeSensorListener.cpp
fastdown101: > Did you understand my comment?

Obviously I did not. I get no error when compiling so the compiler seems not to care about the Java method. Can you give me some more context?

MaxKellermann: This piece of C++ code implements the native part of a Java method. But the Java method you're implementing doesn't exist, therefore you cannot implement it.
The C++ compiler doesn't know about Java methods, therefore it has no reason to complain. But I complain. Not because your code is syntactically wrong, but because it's logically wrong, it makes no sense.

fastdown101: So to make sens, I should add not the Java files, I should add a Java file...
Thanks

src/Device/AndroidSensors.cpp
fastdown101: What's wrong with the types? Should be valid from c++11. How would you like the types to be?

MaxKellermann: Temperature, AtmosphericPressure - your choice of plain integers is error prone because it's not type-safe. What's the unit for the temperature? What's the unit for pressure? Using the correct types makes the code more robust while not adding any bit of overhead.

android/src/NativeSensorListener.java
```diff
@@ -33,6 +33,13 @@ public native void onLocationSensor(long time, int n_satellites,
   @Override
   public native void onAccelerationSensor1(double acceleration);
 
+  @Override
+  public native void onEngineSensors(int cht_temp,
+                                       int egt_temp,
+                                       int oat_temp,
+                                       int atmospheric_pressure,
+                                       int revs_per_sec);
+
```
MaxKellermann, line 42: Bad indent

android/src/SensorListener.java
```diff
@@ -30,6 +30,11 @@ void onLocationSensor(long time, int n_satellites,
   void onI2CbaroSensor(int index, int sensorType, int pressure);
   void onVarioSensor(float vario);
   void onHeartRateSensor(int bpm);
+  void onEngineSensors(int cht_temp,
+                       int egt_temp,
+                       int oat_temp,
+                       int atmospheric_pressure,
+                       int revs_per_sec);
```
MaxKellermann, line 37: Parameters are obscure, please document the API

src/Android/NativeSensorListener.cpp
```diff
@@ -179,6 +179,28 @@ Java_org_xcsoar_NativeSensorListener_onI2CbaroSensor(JNIEnv *env, jobject obj,
...
 
+gcc_visibility_default
+JNIEXPORT void JNICALL
+Java_org_xcsoar_NativeSensorListener_onEngineSensors(JNIEnv *env,
+                                                       jobject obj,
+                                                       jint cht_temp,
+                                                       jint egt_temp,
+                                                       jint oat_temp,
+                                                       jint atmospheric_pressure,
+                                                       jint revs_per_sec)
```
MaxKellermann, line 190: Bad indent!!

src/Device/AndroidSensors.cpp
```diff
@@ -4,6 +4,7 @@
 #include "Descriptor.hpp"
 #include "DataEditor.hpp"
 #include "NMEA/Info.hpp"
+#include "NMEA/EngineState.hpp"
```
MaxKellermann, line 7: You don't need that include

src/Device/SensorListener.hpp
```diff
@@ -44,6 +44,18 @@ class SensorListener {
                                AtmosphericPressure pressure) noexcept = 0;
   virtual void OnVarioSensor(float vario) noexcept = 0;
   virtual void OnHeartRateSensor(unsigned bpm) noexcept = 0;
+  /**
+   * @param[in] cht Engine Cylinder Head Temperature (CHT).
+   * @param[in] egt Engine Exhaust Gas Temperature (EGT).
+   * @param[in] oat Outside Air Temperature.
```
MaxKellermann, line 50: Are all of these parameters mandatory?

src/Device/SensorListener.hpp
```diff
@@ -44,6 +44,18 @@ class SensorListener {
...
+   * @param[in] pressure Pressure, in Pascal, valid if pressure > 100 && pressure < 1200
+   * @param[in] revs_per_sec Engine Revolutions Per Second, rotations per seconds of the camshaft.
```
MaxKellermann, line 52: Is this parameter mandatory?

src/Device/AndroidSensors.cpp
```diff
@@ -287,6 +288,36 @@ DeviceDescriptor::OnHeartRateSensor(unsigned bpm) noexcept
...
+  }else{
+    basic.static_pressure_available.Clear();
+  }
+
+  basic.engine_state.cht_temperature = cht;
+  basic.engine_state.cht_temperature_available.Update(basic.clock);
+  basic.engine_state.egt_temperature = egt;
+  basic.engine_state.egt_temperature_available.Update(basic.clock);
+  basic.temperature = oat;
+  basic.temperature_available = true;
```
MaxKellermann, line 317: So `oat` has really nothing to do with the engine - there's already `SensorListener::OnTemperature()`, and your product should use that instead of adding a redundant parameter to `OnEngineSensors()`.

src/Device/AndroidSensors.cpp
```diff
@@ -287,6 +288,36 @@ DeviceDescriptor::OnHeartRateSensor(unsigned bpm) noexcept
...
+  basic.alive.Update(basic.clock);
+  basic.engine_state.revs_per_sec = revs_per_sec;
+  basic.engine_state.revs_per_sec_available.Update(basic.clock);
+
+  if(pressure.IsPlausible()){
+    basic.static_pressure = pressure;
+    basic.static_pressure_available.Update(basic.clock);
+  }else{
+    basic.static_pressure_available.Clear();
+  }
```
MaxKellermann, line 310: The same as for `oat` applies here - there's already `OnBarometricPressureSensor()`, making this parameter redundant.

And now you see one specific advantage of not duplicating code: the existing `OnBarometricPressureSensor()` implementation derives the vario value from pressure changes using a Kalman filter - a feature that is missing in your PR. You duplicated code, but your duplication is incomplete.

MaxKellermann:
When you transfer values from Java to C++, it must be well-defined which parameters have a value. Existing SensorListener methods have a `boolean` parameter for that, e.g.:
https://github.com/XCSoar/XCSoar/blob/7ea9fd839f7cdc5b8e6f894007735e6f83d6d941/android/src/SensorListener.java#L16-L22
Don't be tempted to use "magic" values and pass invalid values to AtmosphericPressure, trusting that its IsPlausible() method will catch them - which values are considered plausible/valid is an implementation detail that should not be exploited by Java code. Just add that boolean parameter if some values are optional. That's more robust.

android/src/NativeSensorListener.java
fastdown101: I'm just guessing what might be a good indent and doing very baldly...
It should be a multiple of 2 spaces right?

src/Android/NativeSensorListener.cpp
fastdown101: It is a multiple of 2. It is the same as for `Java_org_xcsoar_NativeSensorListener_onHeartRateSensor`. What is wrong with it?

android/src/NativeSensorListener.java
MaxKellermann: No, in this case, it must be aligned with the first parameter.

src/Android/NativeSensorListener.cpp
MaxKellermann: Aligned with the first parameter.

PR #1197, opened 2023-04-27
Change display of RPM in InfoBox

fastdown101:
Brief summary of the changes
----------------------------
Change the representation of the EngineState value RPM in it's InfoBox

https://github.com/XCSoar/XCSoar/issues/1193

### DIFF
--- a/src/InfoBoxes/Content/Engine.cpp
+++ b/src/InfoBoxes/Content/Engine.cpp
@@ -45,6 +45,6 @@ UpdateInfoBoxContentRPM(InfoBoxData &data) noexcept
     data.SetInvalid();
     return;
   }  
-  data.SetValue(_T("%d"), basic.engine_state.revs_per_sec * 60);
+  data.SetValue(_T("%.0f"), basic.engine_state.revs_per_sec * 60);
   data.SetComment(_("rpm"));
 }


fastdown101:
small-var4
small-var3
small-var2
small-var1

The one with `6.00` and `1000 x rpm` is my favorite.

MaxKellermann:
I don't think we should change the current hard-coded unit to another hard-coded unit. The unit we have now (rpm) is a "good enough" default which everybody understands. You prefer kilo-rpm, that's fine, that's your taste - but instead of hard-coding that, we should make this configurable, i.e. integrate in XCSoar's "unit" subsystem, providing the user with a set of choices (Hz, rpm, krpm, whatever).

Scumi:
A fix for the small values is being worked on (#1194).
General question: Why are the full RPM shown as float with fractions of RPM? I Don't think anyone looks at .34 RPM. Just change this and the value should look much better even before the fix.

fastdown101:
small-var5

This is how it looks with this change...
I kind of missed that there is no format `_T("%d")` , so it shows 0 with the code on `master`

MaxKellermann:
> I kind of missed that there is no format `_T("%d")` , so it shows 0 with the code on `master`

... what? There is no format what?

fastdown101:
With `data.SetValue(_T("%d"), basic.engine_state.revs_per_sec * 60);` the RPM reading is 0
With `data.SetValue(_T("%0.f"), basic.engine_state.revs_per_sec * 60);` the RPM reading is 6000, as expected.

fastdown101:
Yes, okay can I do something to get it fixed? With `%d` there is no RPM reading at all...

MaxKellermann:
Right now I'm migrating the whole InfoBoxData stuff to libfmt, and as a side effect, this fixes the bug.
If we were to stay with printf, you should have chosen FormatValue() instead, and then you could pass arbitrary types to printf,without the implicit floating point cast.

fastdown101:
New InfoBox formatting fixes small representation.

PR #1198, opened 2023-04-27
Add BLE driver for EngineState

fastdown101:
The sensor board has additional sensors for atmospheric pressure and outside air temperature. These values are passed to already implemented specific listeners.

Brief summary of the changes
----------------------------
Add BLE "driver" for EngineState values with additional atmospheric pressure and outside air temperature.

### DIFF
--- a/android/src/BluetoothSensor.java
+++ b/android/src/BluetoothSensor.java
@@ -117,6 +117,36 @@ public final class BluetoothSensor
     }
   }
 
+  /**
+   * Data in the characteristic has little endian byteorder.
+   * Lowest bit of flags indicates valid revs_per_sec reading.
+   * 0 Kelvin indicates invalid temperatures e.g 
+   * no temperature sensor present.
+  */
+  private void engineSensorDataToListeners(BluetoothGattCharacteristic c) {
+    final int flags = c.getIntValue(c.FORMAT_UINT8, 0);
+    final int cht_temp = c.getIntValue(c.FORMAT_UINT16, 1);
+    final int egt_temp = c.getIntValue(c.FORMAT_UINT16, 3);
+    final int outside_air_temperature = c.getIntValue(c.FORMAT_UINT16, 5);
+
+    if(outside_air_temperature != 0)
+      listener.onTemperature(outside_air_temperature);
+
+    final int pressure = c.getIntValue(c.FORMAT_UINT32, 7);
+
+    // Just guessing the sensor_noise_variance.
+    if(pressure != 0)
+      listener.onBarometricPressureSensor(c.getIntValue(c.FORMAT_UINT32, 7), 0.01f);
+
+    final int revs_per_sec = c.getIntValue(c.FORMAT_UINT16, 11);
+    listener.onEngineSensors(cht_temp != 0 ? true : false,
+                             cht_temp,
+                             egt_temp != 0 ? true : false,
+                             egt_temp,
+                             (flags&0x01) == 0x01 ? true : false,
+                             revs_per_sec);
+  }
+
   private void readHeartRateMeasurement(BluetoothGattCharacteristic c) {
     int offset = 0;
     final int flags = c.getIntValue(c.FORMAT_UINT8, offset);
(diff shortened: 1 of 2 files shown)


android/src/BluetoothSensor.java
```diff
@@ -156,6 +182,11 @@ public synchronized void onCharacteristicChanged(BluetoothGatt gatt,
         }
       }
 
+      if (BluetoothUuids.ENGINE_SENSORS_SERVICE.equals(c.getService().getUuid()) &&
+              BluetoothUuids.ENGINE_SENSORS_CHARACTERISTIC.equals(c.getUuid())) {
```
MaxKellermann, line 186: Bad indent

android/src/BluetoothSensor.java
```diff
@@ -117,6 +117,32 @@ private void enableNotification(BluetoothGattCharacteristic c) {
...
+   * Lowest bit of flags indicates valid revs_per_sec reading.
+   * 0 Kelvin indicates invalid temperatures e.g 
+   * no temperature sensor present.
+  */
+  private void engineSensorDataToListeners(BluetoothGattCharacteristic c) {
+    final int flags = c.getIntValue(c.FORMAT_UINT8, 0);
+    final int cht_temp = c.getIntValue(c.FORMAT_UINT16, 1);
+    final int egt_temp = c.getIntValue(c.FORMAT_UINT16, 3);
+    final int outside_air_temperature = c.getIntValue(c.FORMAT_UINT16, 5);
+    if(outside_air_temperature != 0)
```
MaxKellermann, line 131: ... and when you're fixing coding style: my eyes hurt when I see `if(` without space ... `if` is not a function.

android/src/BluetoothUuids.java
```diff
@@ -24,6 +24,18 @@ public interface BluetoothUuids {
...
 
+  /**
+   * Engine sensors service
+   */
+  UUID ENGINE_SENSORS_SERVICE =
+    UUID.fromString("D2865ECA-2C07-4610-BF03-8AEEBEF047FB");
+
+  /**
+   * Engine sensors characteristic
+   */
```
MaxKellermann, line 35: This needs a more extensive code comment. This is not a generic "Engine sensor", but your proprietary product. There should be a description of your product or a link to your website which describes the product.

android/src/BluetoothSensor.java
```diff
@@ -117,6 +117,32 @@ private void enableNotification(BluetoothGattCharacteristic c) {
...
+      listener.onTemperature(outside_air_temperature);
+    final int pressure = c.getIntValue(c.FORMAT_UINT32, 7);
+    // Just guessing the sensor_noise_variance.
+    if(pressure != 0)
+      listener.onBarometricPressureSensor(c.getIntValue(c.FORMAT_UINT32, 7), 0.01f);
+    final int revs_per_sec = c.getIntValue(c.FORMAT_UINT16, 11);
```
MaxKellermann, line 137: In your protocol design, if the flag isn't set, is the field present in the BLE packet? With a bogus value? Other BLE protocols I know just omit the field in that case, and then you shouldn't attempt to read it. The protocol is really our decision, I just want to point out that this needs to be an explicit decision, not an accident.

fastdown101: I'm just starting the project. So the developed sensor box is not jet fully functional complete. Specifically when it comes to one safety feature. There are several severe injuries and fatalities each year, due to people starting their paramotors on the ground without having them secured properly. Last year someone got half his head chopped of, while his wife was watching. The throttle cable was stuck and he did not recognize. Engine started and went crazy on full throttle.
As a future feature I developed a security circuit, monitoring the engine for the first 20 seconds and do an immediate emergency shutdown of the engine if it revs up, by pulling the primary ignition coil to GND. This circuit board is close to ready for production.  The sensor box will signal a emergency shutdown by setting the flags bit to 0. After a little while it should go back to 1, so the engine can be restarted again. That's the plan with the flag :)

PR #1201, opened 2023-04-29
Add RPM scaling for different types of engine

fastdown101:
Support for:
2-stroke, one ignition per revolution
2-stroke, two iginitions per revolution
4-stroke, one ignition per two revolutions

Brief summary of the changes
----------------------------
Added scaling factors, based on user choice to support different engine types.
Assuming, that the revolutions of any engine type are derived from ignition pulses.

### DIFF
--- a/android/src/BluetoothSensor.java
+++ b/android/src/BluetoothSensor.java
@@ -119,7 +119,7 @@ public final class BluetoothSensor
 
   /**
    * Data in the characteristic has little endian byteorder.
-   * Lowest bit of flags indicates valid revs_per_sec reading.
+   * Lowest bit of flags indicates valid ignitions_per_sec reading.
    * 0 Kelvin indicates invalid temperatures e.g 
    * no temperature sensor present.
   */
@@ -138,13 +138,13 @@ public final class BluetoothSensor
     if(pressure != 0)
       listener.onBarometricPressureSensor(pressure / 100.0f, 0.01f);
 
-    final int revs_per_sec = c.getIntValue(c.FORMAT_UINT16, 11);
+    final int ignitions_per_second = c.getIntValue(c.FORMAT_UINT16, 11);
     listener.onEngineSensors(cht_temp != 0 ? true : false,
                              cht_temp,
                              egt_temp != 0 ? true : false,
                              egt_temp,
                              (flags&0x01) == 0x01 ? true : false,
-                             revs_per_sec);
+                             ignitions_per_second);
   }
 
   private void readHeartRateMeasurement(BluetoothGattCharacteristic c) {
--- a/src/Device/AndroidSensors.cpp
+++ b/src/Device/AndroidSensors.cpp
@@ -292,18 +292,19 @@ DeviceDescriptor::OnEngineSensors(bool has_cht,
                                   Temperature cht,
                                   bool has_egt,
                                   Temperature egt,
-                                  bool has_revs_per_sec,
-                                  float revs_per_sec) noexcept
+                                  bool has_ignitions_per_second,
+                                  float ignitions_per_second) noexcept
 {
   const auto e = BeginEdit();
   NMEAInfo &basic = *e;
   basic.UpdateClock();
   basic.alive.Update(basic.clock);
-  if(has_revs_per_sec){
-    basic.engine_state.revs_per_sec = revs_per_sec;
-    basic.engine_state.revs_per_sec_available.Update(basic.clock);
+  if(has_ignitions_per_second && config.engine_type != DeviceConfig::EngineType::NONE){
+    basic.engine_state.revolutions_per_second = ignitions_per_second *
+                                                config.ignitions_to_revolutions_factors[static_cast<unsigned>(config.engine_type)];
+    basic.engine_state.revolutions_per_second_available.Update(basic.clock);
   }else{
-    basic.engine_state.revs_per_sec_available.Clear();
+    basic.engine_state.revolutions_per_second_available.Clear();
   }
   if(has_cht){
     basic.engine_state.cht_temperature = cht;
--- a/src/Device/Config.hpp
+++ b/src/Device/Config.hpp
@@ -170,6 +170,25 @@ struct DeviceConfig {
   double sensor_factor;
 
   /**
+   * User choices of engine types supported.
+   * Depending on the engine used (2-stroke, 4-stroke etc.),
+   * ignitions per second have to be scaled to revolutions per second.
+   */
+  enum class EngineType : uint_least8_t{
+    NONE = 0,
+    TWO_STROKE_1_IGN,
+    TWO_STROKE_2_IGN,
+    FOUR_STROKE_1_IGN,
+  } engine_type;
+
+  /**
+   *  Based on user choice of engine type, the measured ignitions of the
+   *  engine used, get scaled to revolutions per second. engine_type[0]
+   *  maps to ignitions_to_revolutions_factors[0] etc.
+   */
+  static constexpr float ignitions_to_revolutions_factors[4] = { 0xAFFE, 1.0f, 0.5f, 2.0f };
+
+  /**
    * Name of the driver.
    */
   StaticString<32> driver_name;
--- a/src/Dialogs/Device/DeviceEditWidget.cpp
+++ b/src/Dialogs/Device/DeviceEditWidget.cpp
@@ -16,7 +16,7 @@
 #include "Interface.hpp"
 
 enum ControlIndex {
-  Port, BaudRate, BulkBaudRate,
+  Port, EngineTypes, BaudRate, BulkBaudRate,
   IP_ADDRESS,
   TCPPort,
   I2CBus, I2CAddr, PressureUsage, Driver, UseSecondDriver, SecondDriver,
@@ -78,6 +78,22 @@ FillPress(DataFieldEnum &dfe) noexcept
   dfe.addEnumText(_T("Pitot (airspeed)"), (unsigned)DeviceConfig::PressureUse::PITOT);
 }
 
+/**
+ * The user can choose from the following engine types:
+ * None.
+ * 2S1I, 2-stroke one ignition per revolution.
+ * 2S2I, 2-stroke two ignitions per revolution.
+ * 4S1I, 4-stroke one ignition per revolution.
+*/
+static void
+FillEngineType(DataFieldEnum &dfe) noexcept
+{
+  dfe.addEnumText(_T("None"), static_cast<unsigned>(DeviceConfig::EngineType::NONE));
+  dfe.addEnumText(_T("2S1I"), static_cast<unsigned>(DeviceConfig::EngineType::TWO_STROKE_1_IGN));
+  dfe.addEnumText(_T("2S2I"), static_cast<unsigned>(DeviceConfig::EngineType::TWO_STROKE_2_IGN));
+  dfe.addEnumText(_T("4S1I"), static_cast<unsigned>(DeviceConfig::EngineType::FOUR_STROKE_1_IGN));
+}
+
 static bool
 EditPortCallback(const TCHAR *caption, DataField &df,
                  [[maybe_unused]] const TCHAR *help_text) noexcept
@@ -115,6 +131,7 @@ DeviceEditWidget::SetConfig(const DeviceConfig &_config) noexcept
   LoadValue(SyncFromDevice, config.sync_from_device);
   LoadValue(SyncToDevice, config.sync_to_device);
   LoadValue(K6Bt, config.k6bt);
+  LoadValueEnum(EngineTypes, config.engine_type);
 
   UpdateVisibilities();
 }
@@ -188,6 +205,7 @@ DeviceEditWidget::UpdateVisibilities() noexcept
     DeviceConfig::MaybeBluetooth(type, port_df.GetAsString());
   const bool k6bt = maybe_bluetooth && GetValueBoolean(K6Bt);
   const bool uses_speed = DeviceConfig::UsesSpeed(type) || k6bt;
+  const bool engine_sensor = (type == DeviceConfig::PortType::BLE_SENSOR) ? true : false;
 
   SetRowAvailable(BaudRate, uses_speed);
   SetRowAvailable(BulkBaudRate, uses_speed &&
@@ -214,6 +232,7 @@ DeviceEditWidget::UpdateVisibilities() noexcept
   SetRowVisible(SyncToDevice, DeviceConfig::UsesDriver(type) &&
                 CanSendSettings(GetDataField(Driver)));
   SetRowAvailable(K6Bt, maybe_bluetooth);
+  SetRowAvailable(EngineTypes, engine_sensor);
 }
 
 void
@@ -226,6 +245,11 @@ DeviceEditWidget::Prepare(ContainerWindow &parent,
   FillPorts(*port_df, config);
   auto *port_control = Add(_("Port"), nullptr, port_df);
   port_control->SetEditCallback(EditPortCallback);
+
+  DataFieldEnum *engine_type_df = new DataFieldEnum(this);
+  FillEngineType(*engine_type_df);
+  engine_type_df->SetValue(config.engine_type);
+  Add(_("Engine Type"), nullptr, engine_type_df);
 
   DataFieldEnum *baud_rate_df = new DataFieldEnum(this);
   FillBaudRates(*baud_rate_df);
@@ -395,6 +419,8 @@ DeviceEditWidget::Save(bool &_changed) noexcept
   bool changed = false;
 
   changed |= FinishPortField(config, (const DataFieldEnum &)GetDataField(Port));
+
+  changed |= SaveValueEnum(EngineTypes, config.engine_type);
 
   if (config.MaybeBluetooth())
     changed |= SaveValue(K6Bt, config.k6bt);
--- a/src/InfoBoxes/Content/Engine.cpp
+++ b/src/InfoBoxes/Content/Engine.cpp
@@ -29,10 +29,12 @@ void
 UpdateInfoBoxContentEGT(InfoBoxData &data) noexcept
 {
   const NMEAInfo &basic = CommonInterface::Basic();
+
   if (!basic.engine_state.egt_temperature_available.IsValid()) {
     data.SetInvalid();
     return;
   }
+
   data.FmtValue(_T("{:3.0f}"), basic.engine_state.egt_temperature.ToUser());
   data.SetValueUnit(Units::current.temperature_unit);
 }
@@ -41,11 +43,12 @@ void
 UpdateInfoBoxContentRPM(InfoBoxData &data) noexcept
 {
   const NMEAInfo &basic = CommonInterface::Basic();
-  if (!basic.engine_state.revs_per_sec_available.IsValid()) {
+
+  if (!basic.engine_state.revolutions_per_second_available.IsValid()) {
     data.SetInvalid();
     return;
   }  
 
-  data.FmtValue(_T("{}"), Units::ToUserRotation(basic.engine_state.revs_per_sec));
+  data.FmtValue(_T("{}"), Units::ToUserRotation(basic.engine_state.revolutions_per_second));
   data.SetValueUnit(Units::current.rotation_unit);
 }
--- a/src/NMEA/EngineState.hpp
+++ b/src/NMEA/EngineState.hpp
@@ -11,9 +11,16 @@
 */
 struct EngineState
 {
-  /* The engine box measured revolutions per second on the camshaft. */
-  Validity revs_per_sec_available;
-  float revs_per_sec;
+  /** The engine box measures ignitions per second of the sparkplug.
+  *   Per device setting, the user can change a scaling value to
+  *   scale, based on the engine used, to revolutions per second.
+  */
+  Validity revolutions_per_second_available;
+  float revolutions_per_second;
+
+  /* For future use. */
+  Validity ignitions_per_second_available;
+  float ignitions_per_second;
 
   /* The engine Cylinder Head Temperature (CHT) */
   Validity cht_temperature_available;
@@ -24,7 +31,7 @@ struct EngineState
   Temperature egt_temperature;
 
   void Clear() noexcept{
-    revs_per_sec_available.Clear();
+    revolutions_per_second_available.Clear();
     cht_temperature_available.Clear();
     egt_temperature_available.Clear();
   }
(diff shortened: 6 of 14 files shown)


fastdown101:
Commercial aftermarket RPM gauges usually derive RPMs non invasive to the ignition circuit.
In practice this is done by looping a wire around the high voltage wire, which goes to the spark plug. With some circuit, shaping the signal which was picked up from the looped wire, feeding to a micro controller.
Example for 2-stroke paramotor aftermarket gauge: [Polini digital racing system](https://polini-shop.de/media/pdf/PI463-Digital-Racing-System-Polini-THOR-928_830_002.pdf)
So deriving revolutions from ignitions is based on the engine type used, e.g. 2-stroke, 4-stroke, etc..

I am not sure if I did the code change properly.

MaxKellermann:
No .... that is weird!
So there's a new unit "rpm1" which is not "rounds per minute" but "rounds per 30 seconds".
And the unit "rpm2" has just as well nothing to do with "rounds per minute" but is "rounds per two minutes".

The unit system is there to choose different units, not to apply scaling factors to values. The base value must be based on the base unit, and for rotation, that's Hz, by XCSoar's definition. But you want to abuse XCSoar's unit system to assume that the base unit is not Hz, and apply weird scaling factors. That's as wrong as it can be.

fastdown101:
Okay, but how to apply the persistent scaling. Unfortunately it is needed for different engines. 
Actually I don't see this as a change of the unit, it is a adaption to the measurement. Probably the revs_per_sec should be ignitions_per_sec, because that is what gets measured...

MaxKellermann:
> Okay, but how to apply the persistent scaling. Unfortunately it is needed for different engines.

https://github.com/XCSoar/XCSoar/issues/1178#issuecomment-1516106627

> Actually I don't see this as a change of the unit, it is a adaption to the measurement.

Then... why do you implement this in the unit subsystem?

> Probably the revs_per_sec should be ignitions_per_sec, because that is what gets measured...

You could do that - the lowest level of your driver sets ignitions_per_second, and then another layer calculates revs_per_second from that, using the per-device config value.

fastdown101:
I renamed the revs_per_sec to the more precise ignitions_per_sec.
Now I'm stuck with the `DeviceEditWidget`.
I don't understand, where this widget is shown to the user?
How to show a drop down list for the user to chose from on this widget?
How to apply the chosen value to the value shown in the InfoBox for RPM?

fastdown101:
What does the `enum ControlIndex` do?

MaxKellermann:
No, don't remove revs_per_second - you're, again, trying to move complexity to the InfoBox. There should be both fields, because both may be useful, but the InfoBox should not do calculations, it should just pick the right field from the NMEAInfo structure.

src/InfoBoxes/Content/Engine.cpp
```diff
@@ -43,12 +43,14 @@ void
 UpdateInfoBoxContentRPM(InfoBoxData &data) noexcept
 {
   const NMEAInfo &basic = CommonInterface::Basic();
+  const DeviceConfig &device_settings = CommonInterface::GetSystemSettings().devices[0];
```
MaxKellermann, line 46: This is completely wrong. You can easily see why by answering the question "Why 0?"

fastdown101: > the InfoBox should not do calculations

Okay, so where to put the code which does the multiplication?

fastdown101:
Just to get this straight, there should be 2 Fields in NMEAInfo, one with revolutions_per_sec and one with ignitions_per_sec? But revolutions_per_sec is empty when data goes from java part to c++? This kind of makes no sens to me. Should the BLE driver calculate revs vom ignitions?

fastdown101:
Because of the RPM unit scaling with 60, I need scaling for the engine type of 0.5, 1.0 and 2.0. How to implement this with a DataFieldEnum?

fastdown101:
This is not a beauty
`data.FmtValue(_T("{}"), Units::ToUserRotation(basic.engine_state.ignitions_per_sec) * (device_settings.scale_to_rps / 10.0));`
, but at least it works with the `DeviceEditWidget` and the `DeviceConfig`

MaxKellermann:
> Just to get this straight, there should be 2 Fields in NMEAInfo, one with revolutions_per_sec and one with ignitions_per_sec?

Yes. In the future, both values may be useful. Right now, the UI only needs rev_per_sec (but has no use for ign_).

> But revolutions_per_sec is empty when data goes from java part to c++?

Since your sensor doesn't measure revs_, it should not pass that value, so the Java API (and the C++ class SensorListener) only transmits igns_. If somebody ever products something that measures revs_, we can add a new SensorListener method.

> This kind of makes no sens to me. Should the BLE driver calculate revs vom ignitions?

Not possible, the BLE driver doesn't have access to DeviceConfig.

The InfoBox does have access to DeviceConfig, but doesn't know from which device this was measured, so it doesn't know which DeviceConfig to use. You "solved" this by using the first DeviceConfig, but that's a mistake.

This should really be calculated by `DeviceDescriptor::OnEngineSensors()`, `DeviceDescriptor` has a `DeviceConfig` field, so it can easily obtain the scaling factor.

I designed the data flow in XCSoar in a robust way, so mistakes are difficult to make. In this case, if you remove all the things that are impossible (in BLE/Java code!) and the things that are obviously bad (using an arbitrary device index!), what remains is the proper way to do it. And it is also the simplest way to do it.

MaxKellermann:
> Because of the RPM unit scaling with 60, I need scaling for the engine type of 0.5, 1.0 and 2.0. How to implement this with a DataFieldEnum?

Not possible. There is currently no proper way to implement it, other than using symbolic names with DataFieldEnum. But maybe that's enough for now - define an enum with codes for two-stroke, four-stroke, etc. and the translation code then applies a hard-coded scaling factor for that. Not exactly elegant, but good enough for now, and easy to migrate to something better, once we need it.

src/Dialogs/Device/DeviceEditWidget.cpp
```diff
@@ -78,6 +78,14 @@ FillPress(DataFieldEnum &dfe) noexcept
   dfe.addEnumText(_T("Pitot (airspeed)"), (unsigned)DeviceConfig::PressureUse::PITOT);
 }
 
+static void
+FillRPMScaling(DataFieldEnum &dfe) noexcept
+{
+  dfe.addEnumText(_T("2S1I"), 10);
+  dfe.addEnumText(_T("2S2I"), 5);
+  dfe.addEnumText(_T("4S1I"), 20);
```
MaxKellermann, line 86: A 4-stroke engine has 20 revs per ignition?

src/Dialogs/Device/DeviceEditWidget.cpp
```diff
@@ -214,6 +224,8 @@ DeviceEditWidget::UpdateVisibilities() noexcept
   SetRowVisible(SyncToDevice, DeviceConfig::UsesDriver(type) &&
                 CanSendSettings(GetDataField(Driver)));
   SetRowAvailable(K6Bt, maybe_bluetooth);
+  SetRowVisible(EngineSensors, engine_sensor);
+  SetRowAvailable(EngineSensors, engine_sensor);
```
MaxKellermann, line 228: Don't set both. Decide which one is the best choice here - I suggest using only SetRowAvailable

src/Dialogs/Device/DeviceEditWidget.cpp
fastdown101: No, because there is no float, I use those values divided by 10.0 to get scaling from 0.5 to 2.0.
I know it is a workaround, but i could not come up with something better...sorry

src/Device/AndroidSensors.cpp
```diff
@@ -292,18 +292,21 @@ DeviceDescriptor::OnEngineSensors(bool has_cht,
...
   NMEAInfo &basic = *e;
   basic.UpdateClock();
   basic.alive.Update(basic.clock);
-  if(has_ignitions_per_sec){
-    basic.engine_state.ignitions_per_sec = ignitions_per_sec;
-    basic.engine_state.ignitions_per_sec_available.Update(basic.clock);
+  if(has_ignitions_per_second){
+    /** Workaround, because we use index of DataFieldEnum in DeviceEditWidget for scaling,
+     *  which is int, but we need scaling of 0.5, 1.0 and 2.0.
+    */
```
MaxKellermann, line 305: The API documentation of scale_to_rps doesn't explain this, but it should.
My idea, however, was to have "symbolic" values, e.g.:
```
enum class EngineType {
  TWO_STROKE_1_IGN,
  TWO_STROKE_2_IGN,
  FOUR_STROKE_1_IGN,
};
```
... but that's truly ugly. Reading your latest code again, maybe we should have two fields:
- engine type (2 = 2-stroke, 4 = 4-stroke)
- ignitions per revolution (1 or 2)

Does this concept scale well with all kinds of engines that should ever be supported by XCSoar? (I have really no idea of engines. I just want the software representation to be as simple and scalable as possible. Using weird fixed-point scaling that is not properly documented scares me a bit.)

fastdown101: Yea, I don't know either what types of engines will get supported, but it will get worse with multi cylinder. 2 Cylinder V configuration is different to inline configuration, for both 2 and 4 stroke.
I will properly document the fixed-point stuff, to me this seams the easiest approach. The engine related parts should also be documented in the manual, right? How to choose the right scaling for the used engine...

fastdown101: > maybe we should have two fields

Seems to much to me, but at the end, you decide :)

MaxKellermann: > it will get worse with multi cylinder

Then... let's make it as simple as possible for users, and go with the `enum class EngineType` with a set of hard-coded choices. The first enum choice (the one that will get assigned the `0` value by the compiler) should be the one that applies no scaling at all (factor = `1`).

This solution is ugly, but on the other hand is extensible easily (if new engine sensor products get added), is robust, and easy to use for users. Robustness & ease of use are really important, and I think that outweighs the ugliness in this particular case.

Your suggestion to use fixed-point integers can be fragile and not so easily extensible. The numeric representation has a meaning, unlike my `enum EngineType` suggestion, but that's not a real advantage here, it will never materialize.

fastdown101: So I add a 

```
enum class EngineType : unsigned{
  TWO_STROKE_1_IGN = 10,
  TWO_STROKE_2_IGN = 5,
  FOUR_STROKE_1_IGN = 20,
};
```

use it like

```
static void
FillRPMScaling(DataFieldEnum &dfe) noexcept
{
  dfe.addEnumText(_T("2S1I"), TWO_STROKE_1_IGN);
  dfe.addEnumText(_T("2S2I"), TWO_STROKE_2_IGN);
  dfe.addEnumText(_T("4S1I"), FOUR_STROKE_1_IGN);
}
```

change `unsigned scale_to_rps` in `Config.hpp` to `EngineType engine_type` and finally have a method in `AndroidSensors.cpp` `float EngineTypeToScalingFactor(EngineType engine_type)` which does the scaling?

MaxKellermann: Omit those integers (=10, =5, =20); don't pretend they have a meaning. They are just arbitrary identifiers.
If you use automatically ascending integers, then the scaling factors can be stored in a simple array (right below the `enum` definition?):
```
constexpr float ignitions_to_revs_factors[] = { 1, 0.5, 2 };
constexpr float
IgnitionsToRevs(float ignitions, EngineType type) noexcept
{
  return ignitions * ignitions_to_revs_factors[static_cast<std::size_t>(type)];
}
```

MaxKellermann: `enum class EngineType : unsigned{`
Better: `: uint_least8_t` - in the foreseeable future, we won't have more than 256 engine types ;-)

fastdown101: Okay sounds good. No `if` involved, I like :)

`std::map<EngineType, float>` would be to big of a gun for the problem?

fastdown101: I thought it of mapping a enum to a float... but little did I know about the obvious bad implementation of map...

MaxKellermann: `std::map` is an excellent container type with a good implementation, and XCSoar uses it a lot. But as with every data type, it is not universal; it has characteristics that are good for one use case, but bad for others. For your code, it would be an extremely bad choice.
"If all you have is a hammer, everything looks like a nail" - when you write high-level code, don't expect everything to be abstracted away from you. There is no universal data type for everything.

fastdown101: Oh, I see. Just ignore type safety like `enum class PressureUse` does. 
I guess at some point you have to compromise...

MaxKellermann: `DataFieldEnum` is not a template. It could be, but then every use would generate a new implementation. To avoid this, we `DataFieldEnum` does not actually use C++ enums, but plain integers. This needs to be casted, where you leave type-safe terrain. That's okay, and still better than being type-unsafe everywhere.

src/Dialogs/Device/DeviceEditWidget.cpp
```diff
@@ -188,6 +203,7 @@ DeviceEditWidget::UpdateVisibilities() noexcept
     DeviceConfig::MaybeBluetooth(type, port_df.GetAsString());
   const bool k6bt = maybe_bluetooth && GetValueBoolean(K6Bt);
   const bool uses_speed = DeviceConfig::UsesSpeed(type) || k6bt;
+  const bool engine_sensor = (type == DeviceConfig::PortType::BLE_SENSOR) ? true : false;
```
MaxKellermann, line 206: A rather complicated way of converting a `bool` to the same `bool`!

src/Dialogs/Device/DeviceEditWidget.cpp
```diff
@@ -396,8 +418,11 @@ DeviceEditWidget::Save(bool &_changed) noexcept
 
   changed |= FinishPortField(config, (const DataFieldEnum &)GetDataField(Port));
 
-  if (config.MaybeBluetooth())
+  changed |= SaveValueEnum(EngineTypes, config.engine_type);
```
MaxKellermann, line 421: Did you mean to move this inside the `MaybeBluetooth` block?
(Also, please add a space between `){`)

src/Dialogs/Device/DeviceEditWidget.cpp
```diff
@@ -443,7 +468,7 @@ void
 DeviceEditWidget::OnModified(DataField &df) noexcept
 {
   if (IsDataField(Port, df) || IsDataField(Driver, df) ||
-      IsDataField(UseSecondDriver, df) || IsDataField(K6Bt, df))
+      IsDataField(UseSecondDriver, df) || IsDataField(K6Bt, df) || IsDataField(EngineTypes, df))
```
MaxKellermann, line 471: What visibilities change when the engine type is modified?

fastdown101: I don't get the question. I thought this will update the drop down list, it does not?

src/Dialogs/Device/DeviceEditWidget.cpp
fastdown101: Actually I would like to use `DeviceConfig::BluetoothNameStartsWith("PPGM-V2")`, because my sensor advertises it's name as `PPGM-V2` but I was not able to get it to work. If I change it to `MaybeBluetooth`  but then the k6tb option is also shown for my sensor...

src/Dialogs/Device/DeviceEditWidget.cpp
MaxKellermann: No, it does not. But even if it would - why do you want to update the drop down list? The drop down list of Engine Type is static, it never needs to be updated.
Look at the code, this calls `UpdateVisibilities()`; it makes certain rows visible or invisible, depending on certain selected features. If you enable K6Bt mode, the "baud rate" row must be shown; if you disable K6Bt, it might be hidden.

src/Dialogs/Device/DeviceEditWidget.cpp
MaxKellermann: Using the name is difficult, because the name may or may not be available at the time; it's transmitted asynchronously at some random point in time *after* the device was scanned by XCSoar.

> If I change it to MaybeBluetooth but then the k6tb option is also shown for my sensor...

That description is wrong, because it sounds like the K6Bt option will *not* be shown for your sensor otherwise, but in fact it is always shown for all Bluetooth devices, no matter the condition for EngineType.

fastdown101: Screenshot_2023-05-02-12-44-39-043_org xcsoar

That's how it looks with the current implementation... No K6Bt option. Up to now there was not a single time where the BT name was not shown. Does the config not get updated on selection of BLE sensor? The BT name is stored in the config, right?

MaxKellermann: We are talking about code in `Save()`, but that method was never called when XCSoar decided whether to show K6Bt. Therefore, it cannot possibly have an influence on that decision.
If K6Bt is invisible, them something else is wrong with your code.

> The BT name is stored in the config, right?

Is it? Tell me which DeviceConfig field (or profile key), in your opinion, stores the name of the Bluetooth device?

fastdown101: This is about code in `UpdateVisibilities`. Does every BT sensor have to have the option of K6Bt?

Later on I would like to add the engine hours on this widget. If you have two or more paramotors it would be great to have engine hours of both.

MaxKellermann: > This is about code in `UpdateVisibilities`. Does every BT sensor have to have the option of K6Bt?

Not quite - the K6Bt is a Bluetooth (classic, not BLE) device, but there's no proper way to verify (from within this widget) whether the peer is really a K6Bt, that's why XCSoar asks the user to specify that.

src/Dialogs/Device/DeviceEditWidget.cpp
fastdown101: Okay something is wrong. If I move ` changed |= SaveValueEnum(EngineTypes, config.engine_type);` into the `if (config.MaybeBluetooth())` block, nothing happens. There is no change of engine_type.

android/src/BluetoothSensor.java
```diff
@@ -138,13 +138,13 @@ private void engineSensorDataToListeners(BluetoothGattCharacteristic c) {
     if(pressure != 0)
       listener.onBarometricPressureSensor(pressure / 100.0f, 0.01f);
 
-    final int ignitions_per_sec = c.getIntValue(c.FORMAT_UINT16, 11);
+    final int ignitions_per_second = c.getIntValue(c.FORMAT_UINT16, 11);
```
MaxKellermann: These changes confuse me **every time** I re-read an amended version of your PR. Doesn't belong here.

src/NMEA/EngineState.hpp
```diff
@@ -11,9 +11,12 @@
 */
 struct EngineState
 {
-  /* The engine box measured ignitions per second of the sparkplug. */
-  Validity ignitions_per_sec_available;
-  float ignitions_per_sec;
```
MaxKellermann: As I wrote earlier, I'd prefer to have both fields (ignitions and revs) in the struct. Maybe one day the raw "ignitions" value is useful for something. Maybe we have some "engine on" InfoBox which works even if the user didn't configure an engine type. Whatever. But we can really spare those few bytes of RAM.

android/src/BluetoothSensor.java
fastdown101: If I put this in an own commit, can you merge it? Its just renaming of a variable...

src/NMEA/EngineState.hpp
fastdown101: This does not make any sens to me, because now I have to add dead code which is never used. 
You told me to calculate in the Device part the rpm from ign, this is where rpm data is put into NMEAInfo. Now I should add ignitions which is not used in the code?!?

Actually I thing it would be much more easy if you could implement some of this stuff by your own. Very time consuming this way. Implementation details I have to follow are only in your head...

src/Device/Config.hpp
```diff
@@ -169,6 +169,25 @@ struct DeviceConfig {
...
+    TWO_STROKE_2_IGN,
+    FOUR_STROKE_1_IGN,
+  } engine_type;
+
+  /**
+   *  Based on user choice of engine type, the measured ignitions of the
+   *  engine used, get scaled to revolutions per second. engine_type[0]
+   *  maps to ignitions_to_revolutions_factors[0] etc.
+   */
+  static constexpr float ignitions_to_revolutions_factors[4] = { 1.0f, 1.0f, 0.5f, 2.0f };
```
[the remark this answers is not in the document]
fastdown101: Changed "None" to "---"

src/NMEA/EngineState.hpp
MaxKellermann: Please re-read https://github.com/XCSoar/XCSoar/pull/1201#issuecomment-1529423498 - yesterday, in that post, I told you to keep both. You don't have to "add dead code, you decided to remove that code, which already existed. The problem is not that implementation details are "in my head" - if I already told you, it's not just in my head, but also in yours, but you decided to ignore it.

fastdown101: Yes, because it does not make any sens to me. RPMs are in the NMEAInfo struct. Now I have to add ignitions... your comment is on "Outdated" code?

fastdown101: But no problemo, I add ignitions into NMEAInfo for future use...

MaxKellermann: Yesterday, when I asked you to keep both, you didn't disagree. Sure, you didn't explicitly agree, but by silently doing something else, you make it harder than necessary to work with you. Because that's "only in your head"...

src/Device/Config.hpp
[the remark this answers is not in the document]
MaxKellermann: No, you changed the DeviceConfigWidget text to "---", which is wrong - that's not the InfoBox. The InfoBox still shows bogus values if the user didn't select an engine type.

src/NMEA/EngineState.hpp
fastdown101: Yea sorry, I did not see the point of keeping ignitions_per_second...but anyway, they are back

src/Device/Config.hpp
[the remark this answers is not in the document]
fastdown101: Okay sorry, that InfoBox thing I missed

MaxKellermann:
I fixed it for you: a2593e705c329dbc5a1dbdb4ccebd65a7278aedd (and added a commit to fix a bug in your initial PR).
And here's why ignitions_per_second is useful: 6ddb5eaf08f866cbfa0af81d731b578d216b7627 - the list shows "Engine" even if the user did not select an engine type.

fastdown101:
Thanks Max, very much appreciate your help :)

PR #1203, opened 2023-05-02
New type of start at PEV inside cylinder proposed to IGC and used during PL Nationals

marek1313:
I have created an implementation proposition of new type of start. Functionality described in README at https://github.com/marek1313/XCSoar-fork-for-PEV-start. 
The fork contains only changes related to this functionality and is up to date (was at 21:00 today).
Additionally I have created copy of 7.28 version with this modification available here: https://gihub.com/marek1313/XCSoarPEV

To implement this type of start I had to modify a little logic of how PEV is handled. Logic of start window and wait time was moved to the task inside SetPEV method from InputEventActions.
Current identification of start is generally based on 2 situations: transition enter or transition exit. I decided not to add new type of situation, use transition exit and change its logic for ordered start point: being inside the start zone and receiving valid PEV causes transition exit. Task restarts do not happen afterward despite being inside the zone, since a new valid PEV would be needed for that. 
Methods to provide location of the start for solvers where updated to return position of start instead of minimum zone distance in case of this kind of start.
Logic to calculate finish alt is moved to new method in finish point - as there are two places where the same code would repeat in OrderedTask otherwise: during task start and during rules update propagation after rules update from dialog. 

The officially presented rules to IGC are as below:
The Start with use of Departure Zone
1. The organizer may implement the Start with use of Departure Zone. It must be stated in
Local Rules.
2. Start Point is the center of Start Ring.
3. The only start option is a circle.
4. Minimum circle radius is 10 km.
5. A Start is valid if the flight log shows a valid fix inside Start Ring, after the opening of the
Start or if the flight log shows a straight line between following fixes that crosses the Start,
after the opening of the Start.
6. To make a Start competitor has to mark a fix via activation of EM function. Any activation of
EM makes previous EM invalid. Start Time is the time of EM activation (Start Moment). The
fix at Start Moment should be inside Start Zone.
7. A distance of the first leg is a distance between the fix at Start Moment and first Waypoint
or first Zone.
8. If the last activation of EM is before start opening, then Start Moment will be counted at
start opening (if exactly at start opening there is no recorded fix or fix is not inside Start
Zone, Start Moment is the first fix recorded after start opening and inside Start Zone).
9. If the last activation of EM has happened before the glider has entered Start Zone, the Start
Moment is the first fix inside Start Zone.
10. If Start Moment has happened when the glider has left Start Zone, the first leg distance is
calculated form Start Moment, and Start Time is the time of last fix inside the Start Zone.
This is valid only if Start Moment is not later than 1 minute after leaving Start Zone.
11. Minimum time interval between following starts is 10 min.
12. If the time interval between last activation of EM and previous activation is less than interval
(each Start Moment is taken into account), the valid Start is the last Start but with penalty –
competitor task time is increased with the result of following subtraction:
13. (10 minutes) – (difference between last Start Moment and previous Start Moment).
14. The above subtraction should be not less than zero. The rule is to protect competitor for
inaccurate counting of interval and decries workload.
15. If the time interval between last activation of EM and previous activation is less than 2
minutes, that second start is valid. Only one start repeating within 2 minutes is allowed.
This is to allow immediate repeating of start if competitor is unsure about previous one
validity.
16. If no Start has been declared via EM or Start Moment was later than 1 minute after Start
Zone leaving, than valid Start is the last leaving of Start Zone (crossing the Start Zone
boundary), but the penalty is:
o First offence – task time + 5 minutes,
o Any next – task time + 10 minutes.
17. If a competitor has declared more than 2 GNSS recorders (in competition or any task), only
declared as a 1st (primary) and 2nd (secondary) will be taken for EM scoring.
18. The Competitor has to deliver flight log from 1st recorder. If, from any circumstances, there
is a need to check 2nd recorder, the last EM activation from both recorders will be used for
scoring.

19. The Organizer may specify in the Task Sheet the Maximum Loss of Height (MLH) between
the Start Moment and the Finish Ring. In this case, a ground speed limit at the Start
Moment should be specified (penalty is 1 point for every 1 km/h, but not more than 30
points).
20. If a competitor crosses the Finish Ring with a loss of height greater than the MLH, penalty
points are applied. The sum of the penalty points so applied may not be greater than the
number of speed points achieved by the competitor in the given task.
21. Penalty - 1 point for every 3 m above the MLH, but not more than 15% of points achieved
by the winner of the day.
22. If, after crossing the Finish Ring, the competitor climbs above the MLH, then the time of
reaching the MLH could be considered as the finish time.
23. It is recommended to adjust MLH to the cloud base/thermal height during task start.

### DIFF
--- a/build/ov.mk
+++ b/build/ov.mk
@@ -2,6 +2,7 @@ OV_MENU_SOURCES = \
 	$(SRC)/Version.cpp \
 	$(SRC)/Asset.cpp \
 	$(SRC)/LocalPath.cpp \
+	$(SRC)/FlightInfo.cpp \
 	$(SRC)/Formatter/HexColor.cpp \
 	$(SRC)/Formatter/TimeFormatter.cpp \
 	$(SRC)/Hardware/CPU.cpp \
@@ -14,6 +15,8 @@ OV_MENU_SOURCES = \
 	$(SRC)/Look/ButtonLook.cpp \
 	$(SRC)/Look/CheckBoxLook.cpp \
 	$(SRC)/Renderer/TwoTextRowsRenderer.cpp \
+	$(SRC)/Renderer/FlightListRenderer.cpp \
+	$(SRC)/Logger/FlightParser.cpp \
 	$(SRC)/Gauge/LogoView.cpp \
 	$(SRC)/Dialogs/DialogSettings.cpp \
 	$(SRC)/Dialogs/WidgetDialog.cpp \
@@ -24,6 +27,7 @@ OV_MENU_SOURCES = \
 	$(SRC)/Dialogs/KnobTextEntry.cpp \
 	$(SRC)/Dialogs/TouchTextEntry.cpp \
 	$(SRC)/Dialogs/ProcessDialog.cpp \
+	$(SRC)/Dialogs/Error.cpp \
 	$(TEST_SRC_DIR)/Fonts.cpp \
 	$(TEST_SRC_DIR)/FakeLanguage.cpp \
 	$(TEST_SRC_DIR)/FakeLogFile.cpp \
--- a/src/OV/OpenVarioMenu.cpp
+++ b/src/OV/OpenVarioMenu.cpp
@@ -5,17 +5,23 @@
 #include "Dialogs/Message.hpp"
 #include "Dialogs/WidgetDialog.hpp"
 #include "Dialogs/ProcessDialog.hpp"
+#include "Dialogs/Error.hpp"
+#include "Widget/DrawWidget.hpp"
 #include "Widget/RowFormWidget.hpp"
+#include "Renderer/FlightListRenderer.hpp"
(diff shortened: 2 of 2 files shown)


MaxKellermann:
Please clean up the PR branch. This is a big mess. A PR must not contain merge commits, and commit messages like "Update FILENAME" are not acceptable.

marek1313:
Sure, I will do cleanup. All files required by modification are only is src/. Rest of modifications were only to stop automatic workflows.

marek1313:
Fork cleaned up. Commit contain only modifications required by new start method with to additional to be considered:
 - start window time and wait time removed from serializer/deserializer
 - angle::asdelta - small calculation instead of assert

lordfolken:
This is too much for one commit. 
Please split things into smaller commits and add meaningful commit messages of what you are trying to implement and how.
Also stuff that touches central functions such as calulation of the angle, should be individual commits.

PR #1204, opened 2023-05-03
Add engine hours for engine maintenance

fastdown101:
Count seconds while the engine has rpm/ignitions. Displayed in format hhh:mm.

Brief summary of the changes
----------------------------
Add a counter for engine hours, to track maintenance intervals. After engine rebuild, engine hours should be resettable by user of xcs.

### DIFF
--- a/src/io/EngineLogAccess.cpp
+++ b/src/io/EngineLogAccess.cpp
@@ -1,0 +1,78 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The XCSoar Project
+
+#include "EngineLogAccess.hpp"
+#include "io/FileLineReader.hpp"
+#include "io/KeyValueFileReader.hpp"
+#include "io/KeyValueFileWriter.hpp"
+#include "io/FileOutputStream.hxx"
+#include "io/BufferedOutputStream.hxx"
+#include "lib/fmt/ToBuffer.hxx"
+#include "util/PrintException.hxx"
+#include "LocalPath.hpp"
+
+bool
+EngineLogAccess::ReadLogFile()
+{
+  try{
+    FileLineReaderA file(LocalPath(log_file_name));
+    KeyValueFileReader kvreader(file);
+    KeyValuePair pair;
+
+    while (kvreader.Read(pair)) {
+      SensorToLog(std::string(pair.key), std::atof(pair.value));
+    }
+  } catch (...){
+    PrintException(std::current_exception());
+    return false;
+  }
+  return true;
+}
+
+bool
+EngineLogAccess::WriteLogFile()
+{
+  try{
+    FileOutputStream file(LocalPath(log_file_name), FileOutputStream::Mode::CREATE);
+    BufferedOutputStream buffered(file);
+    KeyValueFileWriter kvwriter(buffered);
+
+    for (auto const& entry : engine_log) {
+      kvwriter.Write(entry.first.c_str(), std::to_string(entry.second).c_str());
+    }
+
+    buffered.Flush();
+    file.Commit();
+  } catch (...){
+    PrintException(std::current_exception());
+    return false;
+  }
+  return true;
+}
+
+bool
+EngineLogAccess::WriteSingleUpdateToLogFile(std::string bt_mac, double engine_runtime)
+{
+  SensorToLog(bt_mac, engine_runtime);
+
+  try{
+    FileOutputStream file(LocalPath(log_file_name), FileOutputStream::Mode::APPEND_OR_CREATE);
+    const auto line = FmtBuffer<64>("{}={}\n", bt_mac, engine_runtime);
+    file.Write(line, strlen(line)); // @brief is 0 terminated by FmtBuffer implementation.
+    file.Commit();
+  } catch (...){
+    PrintException(std::current_exception());
+    return false;
+  }
+  return true;
+}
+
+void EngineLogAccess::SensorToLog(std::string bt_mac, double engine_runtime)
+{
+  auto result_pair = engine_log.insert({bt_mac, engine_runtime});
+
+  // does entry for bt_mac exist, update entry?
+  if (!result_pair.second) {
+    result_pair.first->second = engine_runtime;
+  }
+}
--- a/src/io/EngineLogAccess.hpp
+++ b/src/io/EngineLogAccess.hpp
@@ -1,0 +1,66 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The XCSoar Project
+
+#pragma once
+
+#include "system/Path.hpp"
+
+#include <tchar.h>
+#include <map>
+#include <string>
+
+/**
+ * @brief Runtime Log Access. For reading and writing engine runtime to a file.
+ * The log file might have duplicated entries for the keys,
+ * with different values. Key is the BT MAC of a engine sensor,
+ * with engine runtime in seconds, as value.
+ */
+class EngineLogAccess{
+
+/**
+ * @brief Filename of the LocalPath file.
+ */
+const TCHAR *log_file_name = _T("engine_runtime.log");
+
+/**
+ * @brief The logfile data. BT Mac is key, engine runtime is value.
+ */
+std::map<std::string, double, std::less<>> engine_log {};
+
+public:
+  explicit EngineLogAccess() noexcept = default;
+
+public:
+  /**
+   * @brief Read the whole engine_runtime.log into the engine_log map.
+   * If there are dublicate entries in the log file, the highest
+   * engine runtime value, for the particular bt mac, gets stored in the map.
+   */
+  bool ReadLogFile();
+
+  /**
+   * @brief Write the whole engine log map into the engine_runtime.log.
+   * Overwrites the existing log.
+   */
+  bool WriteLogFile();
+
+  /**
+   * @brief Write one key value pair to the engine_runtime.log.
+   * Might be a duplicate. Reading the log at app start takes care of
+   * duplicates.
+   * @param[in] bt_mac Bluetooth MAC of the engine sensor.
+   * @param[in] engine_runtime Total runtime of the engine in seconds.
+   */
+  bool WriteSingleUpdateToLogFile(std::string bt_mac, double engine_runtime);
+
+  /**
+   * @brief Add or update engine sensor to the log map, identified by it's BT MAC.
+   * If the sensor is already in the log map, the engine_runtime gets updated.
+   * It does not get written to the engine_runtime.log file. Use
+   * WriteSingleUpdateToLogFile, to get it added and written imediately.
+   * @param[in] bt_mac Bluetooth MAC of the engine sensor.
+   * @param[in] engine_runtime Total runtime of the engine in seconds.
+
+   */
+  void SensorToLog(std::string bt_mac, double engine_runtime);
+};
--- a/build/main.mk
+++ b/build/main.mk
@@ -167,6 +167,7 @@ XCSOAR_SOURCES := \
 	$(IO_SRC_DIR)/MapFile.cpp \
 	$(IO_SRC_DIR)/ConfiguredFile.cpp \
 	$(IO_SRC_DIR)/DataFile.cpp \
+	$(IO_SRC_DIR)/EngineLogAccess.cpp \
 	$(SRC)/Airspace/ProtectedAirspaceWarningManager.cpp \
 	$(SRC)/Airspace/ActivePredicate.cpp \
 	$(SRC)/Task/Serialiser.cpp \


fastdown101:
There is the `struct DeviceConfig`, and support for 6 devices, right? So there are 6 `DeviceConfig` stored persistently in the profile?
That would be great, because then we can have 6 different engine monitoring devices, which all could have their own engine hours counter.
Here is some questions I have regarding implementing this.
I guess counting hours would happen in the Device part, per device. But how to know which hour counter to display in a InfoBox? Where to implement the reset of the hours counter?

MaxKellermann:
DeviceConfig is for user configuration, but how is this new field supposed to be used? The user configures the engine runtime? What's the point of that?
The data type is wrong - it's not type-safe. Don't use raw integers to store durations - there's `std::chrono::duration` for that. Use the specialization with the appropriate precision.
The name is also bad. I mean, what should `engine_seconds` mean? Okay, you didn't use a proper data type, so the name contains the unit of the quantity - which is not necessary with proper data types - but how can "engine" be measured in "seconds"? Maybe call it "engine_runtime" or something. "Engine hours" is just as wrong - can you measure engines in hours? I know people are used to saying "engine hours" but they really mean "engine runtime". Let's try to make the names precise to avoid ambiguities and confusion.

fastdown101:
> I know people are used to saying "engine hours" but they really mean "engine runtime"

Why invent a new name, when I google "engine runtime" vs. "engine hours", hours clearly win. Anyway, when writing code I'm not a dogmatic person. So "engine runtime" it is. 

> how is this new field supposed to be used?

Use case: I want to monitor my engine runtime. If I switch engine or do a overhaul, I would like to be able to reset the runtime counter. Lets keep it simple and support just one engine for now, not multiple devices.

Implementation: 
- Have some entity count the seconds the engine is running (has revs or ign) and store it persistently (on app closure?)
- Display the time the engine is running in total
- Give the xcs user the ability to reset the counter

Where to implement the above mentioned aspects in xcs?

MaxKellermann:
DeviceConfig or the profile are certainly the wrong tool to manage variable data. The profile is for configuration settings set by the user; not only that: switching to a different profile will give you a new engine runtime counter, and moving the device to a different position in the device list also resets the counter.

I think there is currently no XCSoar feature that automatically manages file contents (without user interaction) - except for xcsoar.log, maybe, but that's a different animal. This needs a new concept, and it should be robust; switching profiles should not reset the counters, and multiple different sensors should not be mixed up. Maybe have some database with per-engine counters? The primary key is the Bluetooth MAC? That "database" could be a plain text file (with lines "mac=runtime"), or maybe a sqlite database, or ... what do you think?

By the way - this is where the raw "ignitions" value will be useful - with this feature I demanded from you, it is possible to count engine hours even if the user didn't configure an engine type. For you, it made no sense...

fastdown101:
> switching profiles should not reset the counters, and multiple different sensors should not be mixed up

Totally agree.

> primary key is the Bluetooth MAC? That "database" could be a plain text file

Sounds good. How about a json file. Has xcs a json parser, if not, should it have one? With json it would be a little bit more structured than plane file and less complex than sqlite. I guess there is already sqlite api, if so and no json, then maybe go for sqlite.

> By the way - this is where the raw "ignitions" value will be useful - with this feature I demanded from you

Kind of agree. If you do not configure a engine type you do not have rpm readings. Why be interested in engine runtime then? 

How to begin with coding? Maybe with the part which handles the bt-mac:runtime (key:value) pairs?

MaxKellermann:
> Sounds good. How about a json file. Has xcs a json parser, if not, should it have one? With json it would be a little bit more structured than plane file and less complex than sqlite. I guess there is already sqlite api, if so and no json, then maybe go for sqlite.

Yes, we have boost::json.

The disadvantage of JSON is that for each modification, you have to read the whole file, parse it, modify the JSON tree, serialize and write the whole file again. Atomically, or else data is lost if something bad happens in between. That much unnecessary overhead always worries me - though it's probably negligible for this feature. sqlite is more efficient, and it's battle-proven, virtually everything on Android is stored in sqlite databases - but XCSoar doesn't have any sqlite bindings yet (they need to work on all platforms, not just Android).

For your feature, I wonder if the simplest approach would be: have a text file, let's call it `engine_runtime.log`, which looks like:
```
00:11:22:33:44:55=3600
```
That engine sensor has already logged 1 hour. Every minute or so, XCSoar appends a new line (with `O_APPEND`), after that the file looks like:
```
00:11:22:33:44:55=3600
00:11:22:33:44:55=3660
```
That line is a dupe now, but only the last one shall be used, all others with the same MAC are obsolete.

Every now and then (maybe every time XCSoar reads the file), it gets "compressed", i.e. rewritten but obsolete lines omitted.

That allows XCSoar to update the file frequently, so barely anything gets lost even if XCSoar crashes or the phone battery dies, but this frequent I/O has almost no overhead, it's a simple `write()` to an `O_APPEND` file.

JSON can be fine for some things, but I feel it's really overkill for your feature - it does too much and puts unnecessary constraints on the implementation, making optimizations like this one impossible.

> Kind of agree. If you do not configure a engine type you do not have rpm readings. Why be interested in engine runtime then?

You connect the sensor and go ahead flying, not minding about the RPM InfoBox, but you can be sure XCSoar already logs engine runtime; eventually, you're interested in setting it up properly, but no runtime is lost.

There's just no reason to not count runtime even if the user hasn't set up XCSoar properly. Not setting up software properly is a very common thing. Let's design our software that it does its best to compensate, and let's not throw away information for no reason that may be helpful for that.

> How to begin with coding? Maybe with the part which handles the bt-mac:runtime (key:value) pairs?

Your feature consists of several pieces:
- some code in the Computer classes that counts engine runtime since XCSoar started and puts it, together with the engine's MAC address, somewhere in DerivedInfo (todo: check if the Computer code runs even if no GPS is available)
- a library for managing that text log file; it can read the text file into a `std::map<std::string,double,std::less<>>`; it can write that map into the text file (replacing the current version atomically); it can append to an already-opened appending file descriptor
- read the file during startup into a global variable
- a timer in the UI thread that appends to the file

(I'll be offline until next week, but I guess you have enough to do for now?)

fastdown101:
> I guess you have enough to do for now?

Yes I do have :)
I like the idea with the file.

fastdown101:
My build environment is some how broken?!?
I usually build with `docker run --mount type=bind,source="$(pwd)",target=/opt/xcsoar -it ghcr.io/xcsoar/xcsoar/xcsoar-build:latest xcsoar-compile ANDROID`

But now I get the error message `make: *** No rule to make target 'android/ioio/IOIOLibCore/src/main/java/ioio/lib/api/AnalogInput.java', needed by 'output/ANDROID/android/classes.zip'.  Stop.`

I already did a `make clean` and reloaded the docker image... did not help. Any sugetions?

MaxKellermann:
My psychic powers tell me that you forgot `git submodule update`.

MaxKellermann:
btw. the last time I did `make clean` is 14 years ago.

src/io/EngineLogAccess.cpp
```diff
@@ -0,0 +1,76 @@
...
+EngineLogAccess::WriteSingleUpdateToLog(std::string bt_mac, double engine_runtime)
+{
+  SensorToLog(bt_mac, engine_runtime);
+
+  try{
+    FileOutputStream file(LocalPath(log_file_name), Mode::APPEND_OR_CREATE);
+    BufferedOutputStream buffered(file);
+    KeyValueFileWriter kvwriter(buffered);
+    kvwriter.Write(bt_mac.c_str(), std::to_string(engine_runtime).c_str());    
+    buffered.Flush();
```
MaxKellermann, line 56: This just creates big amounts of overhead for no reason. The buffer is not needed because there will be only one write, and the KeyValueFileWriter is useless here, and `std::to_string` has more useless overhead for allocating another buffer on the heap.
Why not just
```
const auto line = FmtBuffer<64>("{}={}\n", bt_mac, engine_runtime);
file.Write(line, strlen(line));
```
A very small fixed-size buffer allocation on the stack where libfmt produces the string, and a single `write()` system call from that buffer.

src/io/EngineLogAccess.cpp
```diff
@@ -0,0 +1,76 @@
...
+    file.Commit();
+  } catch (...){
+    return false;
```
MaxKellermann, line 59: I don't like suppressing exceptions. On I/O error, this silently loses data, but data loss must never be silent. If you write a library, just let the caller deal with exceptions.

src/io/EngineLogAccess.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+unsigned
+EngineLogAccess::ReadLog()
+{
+  try{
+    FileLineReaderA file(LocalPath(log_file_name));
+    KeyValueFileReader kvreader(file);
+    KeyValuePair pair;
+    while (kvreader.Read(pair)) {
+      std::string bt_mac(pair.key);
```
MaxKellermann, line 18: This allocates an unnecessary `std::string` on the heap!

src/io/EngineLogAccess.cpp
```diff
@@ -0,0 +1,76 @@
...
+}
+
+void EngineLogAccess::SensorToLog(std::string bt_mac, double engine_runtime)
+{
+  // is entry not yet in map log?
+  if (engine_log.find(bt_mac) != engine_log.end()) {
+    engine_log.insert({bt_mac, engine_runtime});
+  } else {    
+    // is the entry larger than the one in the log map?
+    double er = engine_log[bt_mac];
```
MaxKellermann, line 71: You already did this lookup - what's the point of doing it again?

src/io/EngineLogAccess.cpp
```diff
@@ -0,0 +1,76 @@
...
+void EngineLogAccess::SensorToLog(std::string bt_mac, double engine_runtime)
+{
+  // is entry not yet in map log?
+  if (engine_log.find(bt_mac) != engine_log.end()) {
+    engine_log.insert({bt_mac, engine_runtime});
+  } else {    
+    // is the entry larger than the one in the log map?
+    double er = engine_log[bt_mac];    
+    if (er < engine_runtime) {
+      engine_log[bt_mac] = engine_runtime;
```
MaxKellermann, line 73: And here's the third lookup!

src/io/EngineLogAccess.cpp
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+}
```
MaxKellermann, line 76: Newline missing

src/io/EngineLogAccess.hpp
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The XCSoar Project
+
+#pragma once
+
+#include "io/FileOutputStream.hxx"
```
MaxKellermann, line 6: Why include this here?

src/io/EngineLogAccess.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include <string_view>
+
+/**
+ * Engine Runtime Log access. For reading and writing engine runtime to a file.
+ * The log file might have duplicated entries for the key,
+ * with different values. Key is the BT MAC of a engine sensor,
+ * with engine runtime in seconds, as value.
+ */
+class EngineLogAccess{
+private:
```
MaxKellermann, line 19: Redundant `private`

src/io/EngineLogAccess.hpp
```diff
@@ -0,0 +1,62 @@
...
+/**
```
MaxKellermann, line 20: Bad indent

src/io/EngineLogAccess.hpp
```diff
@@ -0,0 +1,62 @@
...
+ * Filename of the LocalPath file.
+ */
+constexpr std::string_view log_file_name{"engine_runtime.log"};
```
MaxKellermann, line 23: This will fail to compile on Windows. Also this allocates 16 bytes in each instance for no reason.

src/io/EngineLogAccess.hpp
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * The logfile data. BT Mac is key, engine runtime is value.
+ */
+std::map<std::string, double, std::less<>> engine_log {}; 
+
+public:
+  explicit EngineLogAccess() noexcept {}
```
MaxKellermann, line 30: What's the point of writing this constructor?

src/io/EngineLogAccess.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+public:  
+  /**
+   * Read the whole engine_runtime.log into the engine_log map.
+   * If there are dublicate entries in the log file, the hightest 
+   * engine runtime value, for the particular bt mac, gets stored in the map.
+   */
+  unsigned ReadLog();
```
MaxKellermann, line 38: What does the return value mean and how is it useful?

MaxKellermann (changes requested):

src/io/EngineLogAccess.cpp
fastdown101: Okay, you mean throw an I/O exception?

src/io/EngineLogAccess.hpp
fastdown101: Should there be more than one instance of this?
What would be a better type?

src/io/EngineLogAccess.cpp
MaxKellermann: No. You don't need to throw anything. Exceptions are already being thrown, but your code catches and discards them.

src/io/EngineLogAccess.hpp
MaxKellermann: I don't understand the question - it was your choice to make this a per-instance variable, so it's you who should answer that question.
The canonical type for file name literals is `Path`.

fastdown101:
How to fix the codacy CWE-126 strlen error?

MaxKellermann:
> How to fix the codacy CWE-126 strlen error?

Not. It's bullshit.

MaxKellermann:
Is anything happening here?

MaxKellermann:
No reply

PR #1107, opened 2023-02-08
implement submenu "settings" to the XCSoar-based menu

Blaubart:
I have added to the XCSoar-based menu the submenu for settings. Now the rotation and the brightness of the display, the language, the timeout for the autostart, as well as the activation or deactivation of SSH, variod and sensord can be made.

PR #1214, opened 2023-05-09
Improvements to OpenVarioMenu

Scumi:
Brief summary of the changes
----------------------------
Work-in-progress implementation update for the OpenVarioMenu.

#1107 
#1207
Openvario/meta-openvario#199

### DIFF
--- a/src/OV/OpenVarioMenu.cpp
+++ b/src/OV/OpenVarioMenu.cpp
@@ -29,6 +29,7 @@
 
 #include <cassert>
 #include <cstdio>
+#include <thread>
 
 enum Buttons {
   LAUNCH_SHELL = 100,
@@ -467,6 +468,9 @@ Main()
 
 int main()
 {
+  /*the x-menu is waiting a second to solve timing problem with display rotation */
+  std::this_thread::sleep_for(std::chrono::seconds(1));
+
   try {
     InitialiseDataPath();
     have_data_path = true;
--- a/src/OV/OpenVarioMenu.cpp
+++ b/src/OV/OpenVarioMenu.cpp
@@ -25,6 +25,7 @@
 #include "util/PrintException.hxx"
 #include "util/ScopeExit.hxx"
 #include "LocalPath.hpp"
+#include "System.hpp"
 
 #include <cassert>
 #include <cstdio>
@@ -441,6 +442,12 @@ Main()
 
   UI::TopWindowStyle main_style;
   main_style.Resizable();
+
+  DisplayOrientation orientation = DisplayOrientation::DEFAULT;
+  try {
+    orientation = OpenvarioGetRotation();
+  } catch (...) {}
+  main_style.InitialOrientation(orientation);
 
   UI::SingleWindow main_window{screen_init.GetDisplay()};
   main_window.Create(_T("XCSoar/OpenVarioMenu"), {600, 800}, main_style);
--- a/src/OV/System.cpp
+++ b/src/OV/System.cpp
@@ -75,9 +75,9 @@ OpenvarioGetRotation()
 
   switch (result) {
   case 0: return DisplayOrientation::LANDSCAPE;
-  case 1: return DisplayOrientation::PORTRAIT;
+  case 1: return DisplayOrientation::REVERSE_PORTRAIT;
   case 2: return DisplayOrientation::REVERSE_LANDSCAPE;
-  case 3: return DisplayOrientation::REVERSE_PORTRAIT;
+  case 3: return DisplayOrientation::PORTRAIT;
   default: return DisplayOrientation::DEFAULT;
   }
 }
@@ -95,13 +95,13 @@ OpenvarioSetRotation(DisplayOrientation orientation)
   case DisplayOrientation::LANDSCAPE:
     rotation = 0;
     break;
-  case DisplayOrientation::PORTRAIT:
+  case DisplayOrientation::REVERSE_PORTRAIT:
     rotation = 1;
     break;
   case DisplayOrientation::REVERSE_LANDSCAPE:
     rotation = 2;
     break;
-  case DisplayOrientation::REVERSE_PORTRAIT:
+  case DisplayOrientation::PORTRAIT:
     rotation = 3;
     break;
   };


src/OV/System.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+#pragma once
+
+enum class SSHStatus {
+  ENABLED,
+  DISABLED,
+  TEMPORARY,
+};
+
+[[gnu::const]]
```
MaxKellermann, line 35: this attribute is wrong. Do you know the difference between "const" and "pure"?

src/OV/System.cpp
```diff
@@ -3,10 +3,33 @@
...
+  return result;
+}
+
+void
+OpenvarioSetBrightness(uint_least8_t value)
+{
+  if (value < 1) { value = 1; }
+  if (value > 10) { value = 10; }
+
+  File::WriteExisting(Path("/sys/class/backlight/lcd/brightness"), std::to_string(value).c_str());
```
MaxKellermann, line 67: No `std::to_string()` please - `fmt::format_int` is much faster and does not need to allocate anything.

src/OV/System.hpp
```diff
@@ -9,6 +9,12 @@ enum class SSHStatus {
   TEMPORARY,
 };
 
+uint_least8_t
+OpenvarioGetBrightness();
```
MaxKellermann, line 24: noexcept?

src/OV/System.cpp
```diff
@@ -4,10 +4,45 @@
...
+}
+
+void
+WriteConfigFile(ProfileMap &map, Path path)
+{
+  FileOutputStream file(path);
+  BufferedOutputStream buffered(file);
+
+  for (const auto &i : map)
+    buffered.Fmt("{}={}\n", i.first.c_str(), i.second.c_str());
```
MaxKellermann, line 42: Why `c_str()`?

src/OV/System.hpp
```diff
@@ -3,12 +3,21 @@
...
+class Path;
+
 enum class SSHStatus {
   ENABLED,
   DISABLED,
   TEMPORARY,
 };
 
+void
+LoadConfigFile(ProfileMap &map, Path path);
```
MaxKellermann, line 18: Why use `ProfileMap` here? Is this a XCSoar profile? (There is no documentation explaining what this library is about)

src/OV/System.hpp
Scumi: added

src/OV/System.cpp
Scumi: Agreed, forgot about our new library.

src/OV/System.cpp
```diff
@@ -4,10 +4,45 @@
...
+LoadConfigFile(ProfileMap &map, Path path)
+{
+  try {
+    FileLineReaderA reader(path);
+    KeyValueFileReader kvreader(reader);
+    KeyValuePair pair;
+    while (kvreader.Read(pair))
+      map.Set(pair.key, pair.value);
+  } catch (...) {
+    ShowError(std::current_exception(), "Config File");
```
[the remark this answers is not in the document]
Scumi: Agreed, done.

src/OV/System.hpp
Scumi: `Profile/Map.hpp` is a nice library that I wanted to use to read and write `config.uEnv` - mainly for updating the `rotation` variable. However, Its probably too much for this rather small task and I can just use `std::map` and a few of its member functions for the intended job.

Documentation will be added.

src/OV/System.cpp
Scumi: Not needed, removed.

src/OV/System.hpp
Scumi: Not sufficiently apparently. After educating myself a bit, I'm not sure if either fits here. I'd remove that one entirely.

MaxKellermann: "pure" would be correct here - the return value can change if global state changes, but the return value will stay the same if you call it twice right way. "const" ignores global state.

Scumi:
> ```
> src/OV/SystemDialog.cpp:111:28: error: lambda capture 'this' is not used [-Werror,-Wunused-lambda-capture]
> ```

Did not observe this during make build with gcc, but could re-produce with clang. Fixed.

MaxKellermann:
More compiler complaints:
```
In file included from ./src/io/BufferedOutputStream.hxx:11,
                 from src/OV/System.cpp:10:
In constructor ‘fmt::v9::format_int::format_int(int)’,
    inlined from ‘void OpenvarioSetRotation(DisplayOrientation)’ at src/OV/System.cpp:112:41:
/usr/include/fmt/format.h:3674:54: error: ‘rotation’ may be used uninitialized [-Werror=maybe-uninitialized]
 3674 |   explicit format_int(int value) : str_(format_signed(value)) {}
      |                                         ~~~~~~~~~~~~~^~~~~~~
src/OV/System.cpp: In function ‘void OpenvarioSetRotation(DisplayOrientation)’:
src/OV/System.cpp:92:7: note: ‘rotation’ was declared here
   92 |   int rotation;
      |       ^~~~~~~~
```

src/OV/System.cpp
```diff
@@ -113,6 +113,13 @@ OpenvarioSetRotation(DisplayOrientation orientation)
   WriteConfigFile(map, Path("/boot/config.uEnv"));
 }
 
+void
+OpenvarioSetLanguage(const char* lang)
+{
+  Run("localectl", "set-locale", lang);
```
MaxKellermann, line 119: Did you test this? Does this work?

src/OV/System.cpp
```diff
@@ -113,6 +113,13 @@ OpenvarioSetRotation(DisplayOrientation orientation)
...
+  Run("export", fmt::format("LANG={}", lang).c_str());
```
MaxKellermann, line 120: This has zero effect, even if this function call were able to invoke a shell built-in (which it does not). So you spawn a shell, use `export` to set an environment variable in this process, and then the process exits. The environment change is gone.

src/OV/System.cpp
Scumi: No, not tested yet. I should've marked that commit as WIP as well. Was just an idea as a first shot without further looking into it. I can keep such things out of this Draft PR if it wastes your time looking at it.

Scumi:
> More compiler complaints:
> 
> [...]

May I ask what your compile command is?
I'm wondering why I do not see those compiler complaints.

Also you merged the rotation commit as is, right? I need to issue a fixup commit to have this variable initialized properly.

src/OV/OpenVarioMenu.cpp
```diff
@@ -467,6 +468,9 @@ Main()
 
 int main()
 {
+  /*the x-menu is waiting a second to solve timing problem with display rotation */
+  std::this_thread::sleep_for(std::chrono::seconds(1));
```
MaxKellermann, line 472: This sounds really horrible, and the justification for such a horrible kludge must be documented really well. Just saying "timing problem" isn't enough.

lordfolken:
Cherry-picked the portrait rotation mapping fix onto a fresh branch (OV/System: align OpenvarioGetRotation/SetRotation with fbcon/DetectInitialOrientation). Left out the 1-second startup sleep and OpenVarioMenu InitialOrientation wiring — those can be a follow-up if needed. The rotation API was dead code on master; this makes it correct for future use.

Scumi:
The OpenVarioMenu will not be needed anymore once everything is integrated into XCSoar anyway.

PR #1215, opened 2023-05-10
Added volume up and volume down keys definitions to input events.

egzumer:
Brief summary of the changes
----------------------------

On Android, when external keyboard is detected the zoom control from volume buttons is disabled. This PR adds volume input definitions, to enable user to configure volume buttons events.

I tried to use Bluetooth remote media controller (with volume up, volume down, previous, next, play/pause buttons), as an external controller for XCSoar. I could configure all buttons except volume buttons. This fixes this issue.

### DIFF
--- a/doc/input_events.rst
+++ b/doc/input_events.rst
@@ -258,9 +258,21 @@ The key type can have the following possible values:
 
 - ``APP1-APP6``: Hardware key on pocket pc
 - ``F1-F12``: Standard function keys
-- ``LEFT, RIGHT, UP, DOWN, RETURN``: Mapped to arrow keys - joystick
-  on organisers
+- ``LEFT, RIGHT, UP, DOWN, RETURN, ESCAPE, MENU, TAB``: Mapped to arrow 
+  keys - joystick on organisers
 - ``A-Z, 0-9``: and other possible keyboard buttons (case is ignored)
+
+Android only:
+
+- ``BUTTON_R1, BUTTON_R2, BUTTON_L1, BUTTON_L2, BUTTON_A, BUTTON_B, 
+  BUTTON_C, BUTTON_X, BUTTON_Y, BUTTON_Z, MEDIA_NEXT, MEDIA_PREVIOUS, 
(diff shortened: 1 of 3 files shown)


MaxKellermann:
Bad commit message and the second commit doesn't appear to make sense.

egzumer:
I tested on android, it compiled and worked. The test have failed because there is no key code definition for other systems, the second commit moves text definition to apply only to android. Both together should work ok. I'm leaving now will check later. If you want I can try later to change this so there are definitions for all OSs.

MaxKellermann:
There should not be a second commit in this case - if the first one is bad, amend it instead of adding fixup commits. Known-bad commits will not be merged.

egzumer:
I think I fixed all issues with this PR.

MaxKellermann:
> Bad commit message

egzumer:
I already changed it, if it is still bad than I don't know what to change it to, sorry, did my best :)

MaxKellermann:
You changed the PR text, but I mean the commit message. Rule of thumb: if the PR contains text, it means that the commit message is bad. Any text present in the PR must be the same in the commit message or else why would you write it in the first place?

egzumer:
I pushed the commit message before I changed the PR title. It's the same now.

MaxKellermann:
I thought this would be obvious enough after I explained the PR text thing, but let's compare. This is the full commit message:

>  Added volume up and volume down keys definitions to input events.

And this is the PR text:

> Added volume up and volume down keys definitions to input events.
> On Android, when external keyboard is detected the zoom control from volume buttons is disabled. This PR adds volume input definitions, to enable user to configure volume buttons events.
> I tried to use Bluetooth remote media controller (with volume up, volume down, previous, next, play/pause buttons), as an external controller for XCSoar. I could configure all buttons except volume buttons. This fixes this issue.

Do we agree that there's a difference?

(It was you who decided that this text is relevant for describing your work, but this text is missing in the commit message.)

MaxKellermann:
Being overly descriptive is good. Though your text doesn't mention the limited scope of your change (only applies to Android).

If you're not used to writing proper commit messages, then please do if you submit code to XCSoar. Being able to comprehend past changes is important.

PR #1216, opened 2023-05-10
Show "No Data" text on the horizon page when required data is missing

r0w4n:
Brief summary of the changes
----------------------------

- Added "No Data" text when there is no artificial horizon data available on the horizon screen
- Text adherers to the colour invert function 

Screenshot 2023-05-10 at 20 13 09

### DIFF
--- a/src/HorizonWidget.cpp
+++ b/src/HorizonWidget.cpp
@@ -8,6 +8,8 @@
 #include "ui/window/AntiFlickerWindow.hpp"
 #include "ui/canvas/Canvas.hpp"
 #include "Renderer/HorizonRenderer.hpp"
+#include "Language/Language.hpp"
+#include "Screen/Layout.hpp"
 
 /**
  * A Window which renders a terrain and airspace cross-section
@@ -42,10 +44,27 @@ protected:
       canvas.ClearWhite();
 
     if (!attitude.bank_angle_available && !attitude.pitch_angle_available)
-      // TODO: paint "no data" hint
-      return;
+      PaintNoData(canvas, canvas.GetRect());
+    else
+      HorizonRenderer::Draw(canvas, canvas.GetRect(), look, attitude);
+  }
 
-    HorizonRenderer::Draw(canvas, canvas.GetRect(), look, attitude);
+  /**
+   * @todo languages
+  */
+  void PaintNoData(Canvas &canvas, PixelRect rc) const noexcept {
+    const TCHAR *str = _("No Data");
+    canvas.Select(look.no_data_font);
+    PixelSize textSize = canvas.CalcTextSize(str);
+    canvas.SetTextColor(inverse ? COLOR_WHITE : COLOR_BLACK);
+    
+    const PixelPoint center{
+      int(rc.GetWidth()) / 2 - int(textSize.width) / 2,
+      int(rc.GetHeight()) / 2 - int(textSize.height) / 2
+    };
+
+    canvas.SetBackgroundTransparent();
+    canvas.DrawText(center, str);
   }
 };
 
--- a/src/HorizonWidget.cpp
+++ b/src/HorizonWidget.cpp
@@ -49,22 +49,13 @@ protected:
       HorizonRenderer::Draw(canvas, canvas.GetRect(), look, attitude);
   }
 
-  /**
-   * @todo languages
-  */
   void PaintNoData(Canvas &canvas, PixelRect rc) const noexcept {
     const TCHAR *str = _("No Data");
     canvas.Select(look.no_data_font);
-    PixelSize textSize = canvas.CalcTextSize(str);
+    PixelSize text_size = canvas.CalcTextSize(str);
     canvas.SetTextColor(inverse ? COLOR_WHITE : COLOR_BLACK);
-    
-    const PixelPoint center{
-      int(rc.GetWidth()) / 2 - int(textSize.width) / 2,
-      int(rc.GetHeight()) / 2 - int(textSize.height) / 2
-    };
-
     canvas.SetBackgroundTransparent();
-    canvas.DrawText(center, str);
+    canvas.DrawText(rc.CenteredTopLeft(text_size), str);
   }
 };
 
--- a/src/Look/HorizonLook.cpp
+++ b/src/Look/HorizonLook.cpp
@@ -3,6 +3,7 @@
 
 #include "HorizonLook.hpp"
 #include "Screen/Layout.hpp"
+#include "FontDescription.hpp"
 
 void
 HorizonLook::Initialise()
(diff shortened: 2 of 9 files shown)


MaxKellermann:
The first commit is broken. Known-bad commits will not be merged - amend broken commits instead of adding fixup commits.

MaxKellermann:
About "added data reference in language files" ... are you aware that the PO/POT files are generated, and that you shouldn't edit those manually?

r0w4n:
> The first commit is broken. Known-bad commits will not be merged - amend broken commits instead of adding fixup commits.

I didn't, that would have saved me some time.

src/Look/HorizonLook.cpp
```diff
@@ -14,4 +14,7 @@ HorizonLook::Initialise()
 
   terrain_brush.Create(terrain_color);
   terrain_pen.Create(Layout::Scale(1), DarkColor(terrain_color));
+
+  no_data_font.Load(FontDescription(Layout::FontScale(22)));
```
MaxKellermann, line 19: You copied that from the FlarmTrafficLook. That's bad. Don't submit duplicate code! This is not only duplicate code, but also adds resource usage because the font is loaded twice. Find a better solution to avoid code/data duplication.

src/Look/HorizonLook.hpp
```diff
@@ -5,10 +5,12 @@
 
 #include "ui/canvas/Pen.hpp"
 #include "ui/canvas/Brush.hpp"
+#include "ui/canvas/Font.hpp"
 
 struct HorizonLook {
   Pen aircraft_pen;
 
+
```
MaxKellermann, line 12: What's the point of adding this empty line?

src/HorizonWidget.cpp
```diff
@@ -42,10 +44,27 @@ class HorizonWindow : public AntiFlickerWindow {
...
+  /**
+   * @todo languages
+  */
+  void PaintNoData(Canvas &canvas, PixelRect rc) const noexcept {
+    const TCHAR *str = _("No Data");
+    canvas.Select(look.no_data_font);
+    PixelSize textSize = canvas.CalcTextSize(str);
+    canvas.SetTextColor(inverse ? COLOR_WHITE : COLOR_BLACK);
+    
+    const PixelPoint center{
```
MaxKellermann, line 61:
```
const auto center = rc.CenteredTopLeft(textSize);
```
Why so complicated?

src/HorizonWidget.cpp
```diff
@@ -42,10 +44,27 @@ class HorizonWindow : public AntiFlickerWindow {
...
+  }
 
-    HorizonRenderer::Draw(canvas, canvas.GetRect(), look, attitude);
+  /**
+   * @todo languages
+  */
+  void PaintNoData(Canvas &canvas, PixelRect rc) const noexcept {
+    const TCHAR *str = _("No Data");
+    canvas.Select(look.no_data_font);
+    PixelSize textSize = canvas.CalcTextSize(str);
```
MaxKellermann, line 58: Bad variable name - coding style violation

src/HorizonWidget.cpp
```diff
@@ -42,10 +44,27 @@ class HorizonWindow : public AntiFlickerWindow {
...
-      // TODO: paint "no data" hint
-      return;
+      PaintNoData(canvas, canvas.GetRect());
+    else
+      HorizonRenderer::Draw(canvas, canvas.GetRect(), look, attitude);
+  }
 
-    HorizonRenderer::Draw(canvas, canvas.GetRect(), look, attitude);
+  /**
+   * @todo languages
```
MaxKellermann, line 53: What??

src/HorizonWidget.cpp
```diff
@@ -92,4 +111,4 @@ void
 HorizonWidget::OnGPSUpdate(const MoreData &basic) noexcept
 {
   Update(basic);
-}
+}
```
MaxKellermann, line 114: You broke this text file by removing the trailing newline from the last line

MaxKellermann:
Now your PR branch contains known-bad commits and revert commits.

MaxKellermann:
> I didn't, that would have saved me some time.

You didn't what? What would have saved you some time? I don't get it.

MaxKellermann (changes requested):

src/Look/HorizonLook.cpp
r0w4n: I'm not quite sure what to do here. There are two options that come to mind:
1. go for inline in the required function e.g.,:
`const unsigned no_data_font_size = Layout::FontScale(12);`
2. or somehow create a shared statement between horizon and traffic

I don't like two and I'm not sure if 1 is better than the current approach.

Are there other options, just a link to an example that I can see or is 1 preferable?

r0w4n:
> > I didn't, that would have saved me some time.
> 
> You didn't what? What would have saved you some time? I don't get it.

I meant to reply to the langue files not needing to be manually edited...

r0w4n:
> Now your PR branch contains known-bad commits and revert commits.

Hopefully resolved?

MaxKellermann:
I don't like "update news" commits. A commit that does a user-visible change should come with the NEWS line. That way, I can `git blame NEWS` and see all commits which have user-visible effects.

src/Look/HorizonLook.cpp
r0w4n: @MaxKellermann just checking around your git commit suggestion - are you suggesting to do this in my current branch but just keep them as separate commits or something else, like insert this commit earlier in the sequence of commits?

MaxKellermann: Yes, insert new commit in this branch. (If you don't know stgit - get it, it helps a lot with managing proper PR branches.)

MaxKellermann:
Never do merges in a PR branch. Always rebase, never merge.

src/Look/HorizonLook.cpp
r0w4n: Hi @MaxKellermann 

I've struggled with implementing your suggestion. I just wanted to check my understanding of your suggestion is this:
 1. Create a new struct called GaugeLook
 2. Move the "no data" or "no traffice" font creation into GaugeLook e.g.,
 `no_data_font.Load(FontDescription(Layout::FontScale(22)));`
 3. In the FlarmTrafficLook.cpp create a variable called no_traffic_font that points to the no_data_font in GaugeLook.
 4. Then do this for HorizonLook

 Did I follow this correctly?

r0w4n: I've committed my WIP change to so you can see my attempt. This obviously doesn't work and is broken but I thought it might be helpful for the conversation.

MaxKellermann:
Converted to draft because it contains a "WIP" commit. That's obviously not ready for review or merging.

MaxKellermann:
No progress in 4 months.

PR #1219, opened 2023-05-15
Fixed horizon issues for roll clipping and showing only ground when pitched up at 50 Degrees

r0w4n:
- Fixed roll clipping which meant that banks of > 90 degrees were not shown
- Fixed pitch visual issues which resulted in ground being show when pitched up greater than 50 degrees

Before:

https://github.com/XCSoar/XCSoar/assets/6553343/d8b0188f-1a4e-4899-ac22-6733523f54ec

After:

https://github.com/XCSoar/XCSoar/assets/6553343/cdefbf86-3cd4-4e2c-9298-2392f741dea1

Brief summary of the changes
----------------------------

- Removed bank angle clamp as this was restricting the visualisation of rolls greater than 90 degrees
- Changed to conditionally drawing ground and sky circle segments.

The reason to remove both the sky and the ground for their respective conditions is that the same issue is present for the sky when pitched down by > 50 degrees - this issue is masked because the ground layer is above the sky layer. 

You could argue that this is a bug with the DrawSegment method. But the reason I chose to fix it here were:

1. I felt it was better not to be attempting to render something that was not visible therefore less wasteful on resources
2. The DrawSegment method is used by many other modules and altering might have inadvertent consequences 

https://github.com/XCSoar/XCSoar/discussions/1212

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -1,4 +1,6 @@
 Version 7.32 - not yet released
+* user interface
+  - fix horizon roll, show bank angles greater than 90°
 
 Version 7.31 - 2023/05/12
 * user interface
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -1,6 +1,7 @@
 Version 7.32 - not yet released
 * user interface
   - fix horizon roll, show bank angles greater than 90°
+  - fix horizon pitch, show sky instead of ground when pitch is greater than 50°
 
 Version 7.31 - 2023/05/12
 * user interface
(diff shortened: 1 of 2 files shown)


lordfolken:
The explanation you put here in the GitHub pr text should be in your commit message. All the PR text is lost on github. Commit messages stay forever in git. 

Words like "updated" and "Fixed" are not descriptive and do not explain what you have done.

That said: Thank your for your work, in correcting the math. :-)

r0w4n:
> The explanation you put here in the GitHub pr text should be in your commit message. All the PR text is lost on github. Commit messages stay forever in git.
> 
> Words like "updated" and "Fixed" are not descriptive and do not explain what you have done.
> 
> That said: Thank your for your work, in correcting the math. :-)

Amended commit title and description in response to your comment

r0w4n:
I did think about separating out the two changes.

I'll do that.

lordfolken:
Nice!

MaxKellermann:
Hey, now the two commits looks very simple and obvious - that's how commit diffs should look like!

PR #1228, opened 2023-05-23
FLARM: Parse AircraftType field as hex in PLFAA

lordfolken:
Beware: I wasn't able to test this with a real flarm yet.


MaxKellermann:
This deserves a better commit message which points to the FLARM protocol documentation where one can verify that the change is correct.

lordfolken:
I've updated the commit message. I hope its OK to include the test in the commit.

Issue #910, opened 2022-04-23
Infobox background color


gldrpilot:
XCSoar version
--------------
7.xx

7.19

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------

Offer a gray background for infoboxes (with text remaining the colors they are now) in addition to the current white or inverted (black with white text).

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------

I have a Samsung phone I've been using a long time with XCSoar and have found screen burn-in (this due to white background with maximum display brightness necessary for outdoor use). I used the inverted option for a while, however I find the Flarm display is not easy to read.
I think having a gray ( which is of neutral brightness) background will improve screen life.

lordfolken:
so we have a GlassRenderer.cpp in src/Renderer/|GlassRenderer.cpp that supports a custom color as a background. But currenlty that color isn't specifyable.

PR #1233, opened 2023-05-26
UI color params

MarkusBagge:
Brief summary of the changes
----------------------------
Added a contrast option for InfoBoxes. 6 levels 0-5 can be selected. 5 is default and highest contrast. At 5, foreground is black (0x00) and background white (0xFF). At 0, foreground is dark gray (0x64) and background light gray (0x9B).

Proposal for issue #910.

### DIFF
--- a/src/Look/InfoBoxLook.cpp
+++ b/src/Look/InfoBoxLook.cpp
@@ -21,15 +21,15 @@
 #define COLOR_INVERSE_GREEN COLOR_GREEN
 #define COLOR_INVERSE_MAGENTA COLOR_MAGENTA
 
+
 void
-InfoBoxLook::Initialise(bool _inverse, bool use_colors,
+InfoBoxLook::Initialise(const UISettings &settings,
                         unsigned width)
 {
-  inverse = _inverse;
 
-  value.fg_color = title.fg_color = comment.fg_color =
-    inverse ? COLOR_WHITE : COLOR_BLACK;
-  background_color = inverse ? COLOR_BLACK : COLOR_WHITE;
+  bool use_colors = settings.info_boxes.use_colors;
+  inverse = UISettings::GetDarkMode(settings);
+
   caption_background_color = inverse
     ? Color(0x40, 0x40, 0x40)
     : Color(0xe0, 0xe0, 0xe0);
@@ -44,9 +44,7 @@ InfoBoxLook::Initialise(bool _inverse, bool use_colors,
   Color border_color = Color(128, 128, 128);
   border_pen.Create(BORDER_WIDTH, border_color);
 
-  ReinitialiseLayout(width);
-
-  unit_fraction_pen.Create(1, value.fg_color);
+  ReinitialiseLayout(settings, width);
 
   colors[0] = border_color;
   if (HasColors() && use_colors) {
@@ -60,8 +58,17 @@ InfoBoxLook::Initialise(bool _inverse, bool use_colors,
 }
 
 void
-InfoBoxLook::ReinitialiseLayout(unsigned width)
+InfoBoxLook::ReinitialiseLayout(const UISettings &settings,
+      unsigned width)
 {
+
+  uint8_t bg = UISettings::GetBgColor(settings);
+  background_color = Color(bg,bg,bg);
+
+  uint8_t fg = inverse ? 255 : 0;
+  value.fg_color = title.fg_color = comment.fg_color = Color(fg,fg,fg);
+  unit_fraction_pen.Create(1, value.fg_color);
+
   const unsigned max_font_height = Layout::FontScale(12);
 
   FontDescription title_font_d(8);
--- a/src/Look/InfoBoxLook.cpp
+++ b/src/Look/InfoBoxLook.cpp
@@ -28,7 +28,7 @@ InfoBoxLook::Initialise(const UISettings &settings,
 {
 
   bool use_colors = settings.info_boxes.use_colors;
-  inverse = UISettings::GetDarkMode(settings);
+  inverse = settings.GetDarkMode();
 
   caption_background_color = inverse
     ? Color(0x40, 0x40, 0x40)
@@ -62,7 +62,7 @@ InfoBoxLook::ReinitialiseLayout(const UISettings &settings,
       unsigned width)
 {
 
-  uint8_t bg = UISettings::GetBgColor(settings);
+  uint8_t bg = settings.GetBgColorVal();
   background_color = Color(bg,bg,bg);
 
   uint8_t fg = inverse ? 255 : 0;
--- a/src/Look/InfoBoxLook.hpp
+++ b/src/Look/InfoBoxLook.hpp
@@ -18,8 +18,7 @@ struct InfoBoxLook {
   bool inverse;
 
   Pen border_pen;
-  Color foreground_color, background_color,
-        focused_background_color, pressed_background_color;
+  Color background_color, focused_background_color, pressed_background_color;
 
   /**
    * Used only by #InfoBoxSettings::BorderStyle::SHADED.
--- a/src/Look/InfoBoxLook.hpp
+++ b/src/Look/InfoBoxLook.hpp
@@ -7,6 +7,8 @@
 #include "ui/canvas/Brush.hpp"
 #include "ui/canvas/Font.hpp"
 #include "util/Macros.hpp"
+#include "InfoBoxes/InfoBoxSettings.hpp"
+#include "UISettings.hpp"
 
 class Font;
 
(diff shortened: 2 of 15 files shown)


MarkusBagge:
For some reason this builds on all platforms except Windows. I have no idea why.
This is the code that fails. Maybe someone can advise on how it should have been done? 
The format strings "%d" where deliberately not translatable _("%d") because there should be no need.

`
src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp: In member function ‘virtual void LayoutConfigPanel::Prepare(ContainerWindow&, const PixelRect&)’:
src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp:210:14: error: cannot convert ‘const char [3]’ to ‘const TCHAR*’ {aka ‘const wchar_t*’}
  210 |              "%d", "%d", 0, 5, 1, ui_settings.info_boxes.contrast);
      |              ^~~~
      |              |
      |              const char [3]
In file included from src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp:14:
./src/Widget/RowFormWidget.hpp:356:40: note:   initializing argument 3 of ‘WndProperty* RowFormWidget::AddInteger(const TCHAR*, const TCHAR*, const TCHAR*, const TCHAR*, int, int, int, int, DataFieldListener*)’
  356 |                           const TCHAR *display_format,
      |                           ~~~~~~~~~~~~~^~~~~~~~~~~~~~
`

MarkusBagge:
Hi

> * making this specific to InfoBoxes; this should affect all UI components which use full-white, like the vario gauge

That's a good point. I will see if I can add that.

> * I think black should remain black and not gray; you don't explain why you want to spend more battery on lighting up those pixels

Not sure what you mean. Black would be Color(0,0,0) and white Color(255,255,255), right? The Background is calculated by 
`bg = inverse ? 20*(5-contrast) : 255 - 20*(5-contrast);`
So contrast=5 yield a color value of 255 (non-inverted) or 0 (inverted). 
Or do you mean contrast should only affect background but not foreground (which could be full white or black)?

> Maybe we should never use full white? Would using 80% white solve the problem for everybody, and the rest can use the global brightness knob?
Full white is definitely good for Kobo's. Should be kept. Global brightness knob? What's that?

MarkusBagge:
Rebased my PR. Here are my comments.

- Contrast setting affects only background brightness rather than "white" values. So might still need some rework. _However_, I think it does work quite good. So try it out before proceeding?
- Let UISettings.hpp declare GetDrakMode() (moved it from Look.hpp) and new GetBgColor(). The latter returns the color value based on dark mode and contrast.
- When (Re)Initializing VarioLook and InfoBoxLook provide the whole UISettings thing rather that each individual setting for each widget. These now calls GetDrakMode() and GetBgColor().
- No need to restart XCS when changing contrast.
- I do not use any Android system setting for contrast yet. Not sure how to do that (but also not sure if that's what is really  wanted).

Even though dithering is used, it still works ok on Kobo. Let it be a personal preference. Note that I also made each contrast step 0-5 to change color by 16. That way, I was hoping to make the best of the 16 shades of gray available on most (?) eInk screens.

src/UISettings.hpp
```diff
@@ -78,6 +78,12 @@ struct UISettings {
   constexpr unsigned GetPercentScale() const noexcept {
     return scale;
   }
+
+  [[gnu::pure]]
+  static bool GetDarkMode(const UISettings &settings) noexcept;
+
+  static uint8_t GetBgColor(const UISettings &settings) noexcept;
```
MaxKellermann, line 85: What's the point of declaring these `static` and explicitly passing a `this` pointer under a different name?

MaxKellermann: `GetBgColor()` sound like it returns a `Color` but it returns an integer. This is confusing.

src/Look/InfoBoxLook.cpp
```diff
@@ -60,8 +58,17 @@ InfoBoxLook::Initialise(bool _inverse, bool use_colors,
...
 void
-InfoBoxLook::ReinitialiseLayout(unsigned width)
+InfoBoxLook::ReinitialiseLayout(const UISettings &settings,
+      unsigned width)
 {
+
+  uint8_t bg = UISettings::GetBgColor(settings);
+  background_color = Color(bg,bg,bg);
+
+  uint8_t fg = inverse ? 255 : 0;
```
MaxKellermann, line 68: This whole section of code added by you looks confusing - but what confuses me most is that even in low-contrast mode, it's 100% white.

src/Look/InfoBoxLook.hpp
```diff
@@ -16,7 +18,8 @@ struct InfoBoxLook {
   bool inverse;
 
   Pen border_pen;
-  Color background_color, focused_background_color, pressed_background_color;
+  Color foreground_color, background_color,
```
MaxKellermann, line 21: This new field is never used, is it?

src/UISettings.cpp
```diff
@@ -35,3 +36,31 @@ UISettings::SetDefaults() noexcept
...
+{
+  switch (settings.dark_mode) {
+  case UISettings::DarkMode::OFF:
+    break;
+
+  case UISettings::DarkMode::ON:
+    return true;
+
+  case UISettings::DarkMode::AUTO:
+    return GlobalSettings::dark_mode;
```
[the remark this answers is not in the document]
MarkusBagge: That's GetDarkMode()-code was copy/paste from Look.cpp. Ok there but not in UISettings.cpp then?

src/Look/InfoBoxLook.cpp
MarkusBagge: That is because, right now, contrast only changes the background. So foreground is just black or white.

src/Look/InfoBoxLook.hpp
MarkusBagge: True, foreground_color is not used! Will fix that.

src/UISettings.hpp
MarkusBagge: So I change declarations to below. Looks right?

`bool GetDarkMode() const noexcept;`

`uint8_t GetBgColorVal() const noexcept;`

When I update my PR, should I always do "git push --force" or just "git push" on my side? Sorry, I am not professional in neither git nor software so let me know when I am out sailing :-).

MarkusBagge:
Ok, pushed a separate commit with updates. GetDarkMode() and GetBgColorVal() is not static anymore and removed unused variable foreground_color. (I can do a merge and git push --force later if wanted).
Don't know what I can do about "Layering violation" or what it really means.
Also, I don't get why there is Windows build error. All other OS seems to compile fine.

MaxKellermann:
I've been wary of merging this because

> If we have a setting, maybe it should be "Full contrast" (disabled by default); enabled means use 255,255,255 (for those who like to be blinded) and 80% or so if disabled. An on/off knob is easier to implement, easier to understand and comes without the complicated formula your code contains.

It is unfortunate that your PR is more complicated than it needs to be and should be. Your formula means if we ever decide to go back to something simpler, we need to migrate all existing profiles, and keep that migration code forever. If we do something simple first, we can always make it more complicated if we feel we really need more contrast levels.

That kind of over-engineering is what complicates the work of long-term maintainers, and I don't get why people always try to over-engineer from the start.

MarkusBagge:
Hi,
It was only with good intention I tried to improve XCS since other people - as I understood - also asked for it. I think that the concept of contrast level selection would be fairly well understood by end users. Using an integer value would allow users to select a level suitable for any screen that XCS might be displayed on (which we cannot know). I didn't anticipate the formula being too hard to comprehend considering the level of programming that must have been required to create XCS prior to this.

I suppose my implementation for contrast levels might have been done differently, this was just my best effort. Fully understand you have no obligations to approve any pull requests!  I conclude this won't make it's way into XCS so will close this PR now.

DanD222:
@MarkusBagge Why didn’t you ask for a hint on how to do a simpler solution?

MarkusBagge:
Hi Dan and Max,
Programming is not my profession but, believe it or not, I have spent time on this too. My understanding was that the proposal I had was not satisfying. In a hindsight, it might have been better to discuss the solution on the forum more to see if it was of any interest before making the PR. (Although I got the feeling that might have taken Max time anyhow). I did not ignore your suggestions but did not agree and motivated why. Possibly I misunderstood your suggestions. The PR-code is still available and free for anyone to pickup. If so, hopefully, the PR was a help on the way and not for nothing then.

And well... If I am "yet another guy ...  caused damage ... wasting time", perhaps the XCS team should reflect a bit about attracting future contributors. Relating to damaging the project here.

Sincerely
Markus

DanD222:
Well, I see a missed opportunity to:
- move the issue forward
- potentially learn something from someone who's been doing this for a while

which is a shame, as, like you say, you've obviously spent time on this too.

I don't think the forum will help you with support, but doing a PR because you'd like to see a feature included is perfectly fine from my point of view (and probably the best/only way to get a feature implemented). 

So ultimately you're on your own here.. well, with Max, who will more likely than not challenge your code. It's primarily then that how that is dealt with that matters.

PR #1262, opened 2023-07-06
Provisioning script - guard to prevent accidentally running with sudo

andrestotorica:
Brief summary of the changes
----------------------------

As title says, prevent manually running the script incorrectly


andrestotorica:
However this change breaks the Docker image build, as it only has a root user and everything is installed there.
I don't know how to nicely make it work on both cases, so best leave the guard out.

PR #1263, opened 2023-07-09
PolarStore: Switch v_no (max speed for normal operation) from m/s to km/h ..

### LINKED ISSUE #1134: max. cruise speed of polar
XCSoar version
--------------
head/all so far

PoalrStore has a m/s representation of the max cruise speed, whereas other horizontal speeds are stored as km/h. Units are arbitrary, but the mix here is a bit awkward.
The polare store entry should use km/h also for the max cruise speed.

Get a better readability for the polar store data.

Any objection changing this?

hjr:
… in the build-in list.

Not touching the XCS extension to the plr file format.

Brief summary of the changes
----------------------------
Internal change of unit not touching any UI or interface.

Closes #1134


MaxKellermann:
The code should never have (arbitrary) user units but only SI units. Your commit message does not contain an explanation for why this change is necessary.

PR #1283, opened 2023-08-14
hide mouse cursor after 10 seconds of inactivity

remcoderksen:
Brief summary of the changes
----------------------------
On builds like OpenVario based on Cubieboard the rendering of a mouse cursor is enabled. But even when no mouse is attached the mouse cursor is rendered on screen. With this contribution the mouse cursor will be hidden when there is no mouse activity for 10 seconds. On the event of new mouse activity the mouse will be displayed again for 10 seconds.

For now this is hardcoded for 10 seconds. It is open for debate if this is acceptable or a higher/lower value should be used. 
And if an extra settings should be introduced to let te user decide to use this feature or not.

### DIFF
--- a/src/ui/window/TopWindow.hpp
+++ b/src/ui/window/TopWindow.hpp
@@ -139,8 +139,8 @@ class TopWindow : public ContainerWindow {
 #ifdef DRAW_MOUSE_CURSOR
   uint8_t cursor_size = 1;
   bool invert_cursor_colors = false;
+  std::chrono::steady_clock::time_point cursor_visible_until = std::chrono::steady_clock::now();
 #endif
-
 #ifndef USE_WINUSER
   TopCanvas *screen = nullptr;
 
--- a/AUTHORS
+++ b/AUTHORS
@@ -64,6 +64,7 @@ The following people and organizations have contributed code to XCSoar:
  Keith Laidlaw <laidlaw1546@gmail.com>
  Tim Stuhler <thermalmap@stuhler.de>
  Dave Spragg <dave.spragg@gmail.com>
+ Remco Derksen <remco@viperlook.com>
 
 Documentation:
 
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -1,5 +1,7 @@
 Version 7.40 - not yet released
-
+* user interface
+  - hide mouse cursor after 10 seconds of inactivity
+  
 Version 7.39 - 2023/07/28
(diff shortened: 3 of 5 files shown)


src/ui/window/TopWindow.hpp
```diff
@@ -139,8 +140,8 @@ class TopWindow : public ContainerWindow {
 #ifdef DRAW_MOUSE_CURSOR
   uint8_t cursor_size = 1;
   bool invert_cursor_colors = false;
+  BrokenDateTime cursor_visible_until = BrokenDateTime::NowUTC();
```
MaxKellermann, line 143: This should be `std::chrono::steady_clock::time_point` and not `BrokenDateTime`.
- `BrokenDateTime` is a very complicated structure that should only be used when you really need it broken down - the nice C++ class with its overloaded operators hides all the complexity for you, but it's there.
- `BrokenDateTime` is a system clock, not a monotonic clock. System clocks should only be used when there's a reason to; a monotonic ("steady") clock is safe against time warps and should always be preferred.

MaxKellermann (changes requested):

src/ui/window/TopWindow.hpp
remcoderksen: I simply searched and reused the code was already there on the code. First thing I found was BrokenDateTime so I never knew about the other time related code you mention. But if it is better then I will look into it and will adjust it to that.

remcoderksen: Done. 
I changed the usage of `BrokenDateTime` to `std::chrono::steady_clock::time_point` and removed the previously added include for `BrokenDateTime`.

MaxKellermann:
There are now 4 commits with the same commit message. I won't review that.

remcoderksen:
I have rewritten all four commit messages and supplied them with detailed information on the changes that were made.

MaxKellermann:
If you need to fix something in an unmerged commit, amend the commit instead of adding a fixup commit.
Look at this PR: you currently suggest merging 4 commits, the first of which you already know is bad, but if you know it is bad, why do you still suggest merging it? Somebody who reviews your work will see the bad commit, will rightfully complain about it, and only after that will see the fixups, when it's too late already. That is just unnecessary churn.

remcoderksen:
Done... Now there is one commit. Hopefully this suits your needs.

Issue #969, opened 2022-07-14
Wrong altitude from Lx Eos variometer


jirimaier:
XCSoar versions having and not having the problem
this problem was likely always present

Expected behavior:
Lx Eos (used with LXNAV driver) is sending altitude in the NMEA sentence, this altitude is already recalculated for actual QNH pressure, xcsoar should use it "as is".

Actual behavior
XCSoar is treating the value as standard-pressure-altitude (relative to 1013 hPa), if correct QNH is set in XCSoar, the altitude shown will be wrong, because it is compensated twice (in both Lx Eos as well as XCSoar)

Steps to reproduce the behavior
(When sea level pressure is different from 1013) connect Lx Eos vario to XCSoar computer. Set correct altitude and QNH on Eos. Then set correct  QNH in XCSoar. Altitude shown in XCSoar will be wrong.

Do you have any idea what may have caused this?
XCSoar expects standard pressure altitude, while Lx Eos is providing QNH altitude.

Do you have an idea how to solve the issue?
Create a modified driver that would handle the altitude value correctly.

A decent workaround is to set XCSoar QNH pressure to 1013 regardless of real value, this results in correct altitude. It will however cause wrong values in vertical distance to airspace which is defined in flight levels (independent of QNH pressure).

Scumi:
Duplicate of #415 ?

lordfolken:
LXNavigation products are not supported at this time. 

And until lxnvaigation fix their device's firmware it just won't be.

(current firmware, now outputs the QNH offset in feet, but only until you takeoff. If you change the qnh thereafter the value won't be corrected and not show you the correct altitude.)

jirimaier:
It would be nice to at least have an option to "lock" the QNH at 1013 for the workaround to be more reliable. Because the QNH seems to be synced with the vario so when the correct one is set on vario, it will change the value in XCSoar and altitude goes wrong. After initial setup on ground, you will notice that AGL is way off, but if this happens in flight, you are unlikely to notice.

jirimaier:
I noticed that it works well with the Condor Simulator driver, which seems to be able to receive true altitude and then calculate the standard altitude given QNH set in XCSoar. As such a method already exists, it shouldn't be hard to duplicate the LXNAV  driver and make this small change.

lordfolken:
Sure then make the change, and try it.

lordfolken:
Closing this as we currently don't have support for LX EOS. That is being tracked in the issue #415

Issue #415, opened 2020-05-22
ERA80-Xcsoar


vargazsolt20e:
Hello,
I fly with an Era80 with connection to xcsoar and i have the following problem;
The Era sends the barometric altitude with QNH correction, but the Xcsoar needs now the standard batometric altitude. The xcsoar thinks that it is the standard altitud and corrects this for qnh. So it shows a completely wrong height because it is corrected 2x to qnh.
it would be really great a driver for era, because now the era-xc combination is useless.. Thank you!
Regards
Zsolt

rawtaz:
Related: #389

lordfolken:
So both companies (lxnav and lxnavigation) spec the LXWP0 sentence:
Lxnav barometric altitude:
https://github.com/XCSoar/XCSoar/blob/48b435a81c2f25297169367f2b65c0b32c8e1c10/src/Device/Driver/LX/Parser.cpp#L50

Lxnavigation in their doc in #389  say:
True altitude in meters

While we could work around it according to the device initialization, such kind of code is a kludge to maintain. Lets talk to lxnavigation and ask for a firmware update.

vargazsolt20e:
Hello, thank you for your message. I contacted lx navigation about the problem and they answered the following:

"we have created a Dataport document for our devices and sent it to the development teams of XC Soar, LK8000, Naviter, etc.
It is now on them to correct the calculations they perform and how they receive this info.

​We're still waiting on the development team from XC soar to give us a reply, as the LK8000 and Naviter teams are already working on it."

lordfolken:
All the basics (data exchange, bugs, mc) should work out of the box with the LXNAV driver. 
Its just that ERA device is not outputting what is defined in their own specification.

The LXWP0 sentence definition goes back all the way to filser... so changing it for that one device isn't correct. Btw we had the same issue with LXNAV s7 and they also updated the firmware to correct the output.

vargazsolt20e:
I understand. It's a really pitty because I think the Era is a great device. I also paid a lot of money for it, but until the problem is not solved, an Era-Xc combination makes no sense at all. It's desperate ... what a pitty. I hope for a happy ending

lordfolken:
I'm talking to them via Whatsapp. Lets see if what they think.

vargazsolt20e:
Thank you very much!!

MaxKellermann:
Wait, what.
So the LXWP0 sentencen was redefined with a different meaning after a firmware update?

[...]

Issue #389, opened 2020-05-07
New generation LX Navigation Device


K-Factor-1961:
I upgraded the new version on Data port document
[LX navigation communication protocols LX_CP_R1.pdf](https://github.com/XCSoar/XCSoar/files/4624893/LX.navigation.communication.protocols.LX_CP_R1.pdf)
I propose to develope a specific driver to communicate with all the new generation LX Navigation Devices.
In attach you can find the data port protocol for alla the new LX Navigation device as:
EOS57
EOS80
ERA57
ERA80
LX10K
COLIBRI X

I attach there also a draft of ERA manual to understand how it works.
The LX device has:
1-user port rs232 nmea standard
2-Bluetooth port that can work as server or as client
3-WiFi port that can work as server or as client
4-Integrated radio bridge tha can comunicate by rs232 port to all the most popular radio
5-integrated inertial platform
6-AHRS option
7-Voice module
8-G-meter

For example:
If Xcsoar send a GOTO to the LX Device, this last sent the eventually frequency to the radio and the voice indicates frequency, type and runway direction.
You can also to provide a special and friendly radio interface on LK8000 and mange direcly the radio by the LX Device. (Xcsoar must be connected to the LX device by bluetooth because rs232 is connected to the radio.

These devices in this moment exchange the protocol data to bluethooth port and to user port (NMEA RS232 port). (May be to the WiFi port in the future)

I'm also involved in feature developing for these devices.

[Era manual_rev_2.pdf](https://github.com/XCSoar/XCSoar/files/4594878/Era.manual_rev_2.pdf)

These devices in this moment exchange the protocol data to bluethooth port and to user port (NMEA RS232 port).
May be implemented bidirectional data interchange from device to external device connected with Xcsoar running.
[LX navigation communication protocols.pdf](https://github.com/XCSoar/XCSoar/files/4594852/LX.navigation.communication.protocols.pdf)
I can provide a EOS80 demo device to make tests
In the future may be possible also to use the LX Navigation Joy remote control mount on the glider stick to manage xcsoar app. 
In fact the LX Navigation Joy remote control can work in three different states. One of the states manages the Vario device, a second state could be used to manage the xcsoar app on the external device if you think usefull.

With  this driver Xcsoar will became integrated with LX Device, add to this device the AAT management (these LX device has no AAT manager). A very powerful glider navigation system wit a very cheap price!!

rawtaz:
[...]

PR #1284, opened 2023-08-16
Driver for LX Navigation Eos57

jirimaier:
Brief summary of the changes
----------------------------
Driver for LX Navigation Eos57 variometer.

Features
----------------------------
- Receiving NMEA data, correctly interpreting altitude
- Task declaration, including observation zones shape
- Flight log download
- MacCready and bugs and ballast synchronization

Compatibility
----------------------------
- Tested on Eos57
- Might work well with Era and Eos80 too. That is not tested, and it is possible that some features will not work properly. Anyway, it will most likely work better than the LXNAV driver. 

Known issues
----------------------------
- Declaration does not set the Landing entry in the task settings. This behavior is the same as with the LXNAV driver and is probably caused by a bug in Eos's firmware. It does not pose any problem for usage, as the Landing point doesn't have any practical usage.
- Ballast synchronization may not be precise. The value provided by vario is relative to polar reference mass. The reference mass of vario's polar (can be different from XCSoar's polar) cannot be directly obtained and is estimated from polar coefficients. 

New generation LX Navigation Device #389
Wrong altitude from Lx Eos variometer #969
ERA80-Xcsoar #415


MaxKellermann:
This PRs copies lots of existing code. That is not acceptable and will not get merged.

jirimaier:
Thank you for the feedback.

I modified the code to call the LXNAV driver's methods for parsing LXWP1 and LXWP3 sentences instead of copying the code (I think this was the main issue with copying existing code).

I also modified the tests to use an LXWP1 sentence which is given as an example in the LX Navigation Communication Protocol documentation (instead of copying the one from the LXNAV driver test).

I think that now there is no copied code. Some parts may be similar because the protocol is similar, but not exactly the same. I don't think that I could reuse other code from the LXNAV driver without making modifications to that driver. 

The Codacy Analysis shows issues regarding unused variables in src/Device/RecordedFlight.hpp, I believe these are false positives. These variables are used in DownloadFlight and GetFlightInfo methods of LXEosDevice.

lordfolken:
This ignores the vario's qnh setting. 
We can hope that somehow xcsoars autoqnh and the varios undocumented autoqnh somehow arrive at the same conclusion.

lordfolken:
Also please test the driver post launch detection. I.e. with pressure on the tas. The vario behaves different post launch.

jirimaier:
> This ignores the vario's qnh setting. We can hope that somehow xcsoars autoqnh and the varios undocumented autoqnh somehow arrive at the same conclusion.

The QNH of XCSoar is set based on the "alt_offset" from LXWP3 sentence.

I am not sure what you mean by "varios autoqnh". When the vario is turned on, the user sets the current altitude (airport elevation), and the QNH of the vario is set accordingly. The altitude offset in LXWP3 can then be converted to the QNH setting so that the QNH in XCSoar gets set to the same value as on the vario.

lordfolken:
> > This ignores the vario's qnh setting. We can hope that somehow xcsoars autoqnh and the varios undocumented autoqnh somehow arrive at the same conclusion.
> 
> The QNH of XCSoar is set based on the "alt_offset" from LXWP3 sentence.
> 
> I am not sure what you mean by "varios autoqnh". When the vario is turned on, the user sets the current altitude (airport elevation), and the QNH of the vario is set accordingly. The altitude offset in LXWP3 can then be converted to the QNH setting so that the QNH in XCSoar gets set to the same value as on the vario.

The alt_offset will remain the same value, after launch detection on the nmea side, no matter what you set the qnh on the varios ui to.

jirimaier:
I will test the in-flight behavior soon (probably tomorrow).

jirimaier:
I never changed the QNH on the vario during flight.
If I understand the behavior of the vario correctly:
1) After turning on the vario, pilot sets the elevation.
2) Vario calculates QNH based on the elevation
3) Pilot can change the QNH value, but that doesn't change the altitude (which is intentional, because you just want to tweak the QNH while keeping the elevation that you set)
4) After confirming this settings, the elevation cannot be changed, but it is possible to change the QNH in settings.
5) Changing QNH before takeoff has no effect on altitude (same as in initial setup)
6) Changing QNH in-fligh will change the true altitude 

And you say that the alt_offset doesn't change when the QNH is changed in flight (This is a bug in the vario, right?).

So this means that if the pilot changes QNH on the vario during flight, they also have to change QNH in the XCSoar. This is not so big problem I guess. I just need to make sure that the new (manually set) QNH in XCSoar doesn't get overwritten to the old value by the next LXWP3 sentence (that has outdated alt_offset).

lordfolken:
That summarizes it pretty well. 
Yes, its a bug at least in the era i have.
Basically stop accepting the alt_offset post takeoff.

jirimaier:
I tested it in flight today. 

When changing QNH on the vario, the QNH in XCSoar also got changed. 

So it seems that the bug is not present in Eos. 

The baro altitude and FL shown by XCSoar always matched the values shown by vario, so the testing is successful. 

To make it compatible with the bug in Era, I can modify the LXWP3 parsing so that the QNH will be set only if the alt_offset is different from the one received in previous LXWP3 (or if it is the first sentence since connecting). 
That way, change in alt_offset will cause change of QNH, but constant value will be ignored. It wouldn't affect the behavior in Eos, and also would fix the problem with overwriting new QNH se by user in case of Era. 

Anyway, I noticed that if I set the QNH in XCSoar manually, all changes of QNH in the vario are ignored.
Is this intended behavior of XCSoar (that it ignores attempts to change QNH from the driver once the QNH was changed by user)?

jirimaier:
It is a bit strange that in the initial setup of the vario, the elevation and QNH can be set independently. In XCSoar, the QNH and the difference between true and standard altitude are totaly dependent on each other. 

The alt_offset in LXWP3 only depends on the initial elevation and on in-flight changes of QNH, it doesn't depend on the initial QNH.

If the initial QNH on vario is changed to different value than the one calculated from elevation (for whatever reason), the QNH in XCSoar (based on alt_offset) will be different from the QNH in vario. But this is fine, because the QNH in XCSoar will lead to correct conversion of true altitude to standard altitude,as it is based on the alt_offset.

jirimaier:
I noticed one more issue:

If the Eos is connected together with another device (Flarm, for example), the altitude provided by the other dece is prioritized regardless of the order in the device list.

In other words: the info.ProvideBaroAltitudeTrue(altitude_true) is "weaker" than info.ProvidePressureAltitude(value).

Can this be considered a bug in XCSoar, or is it a feature?

src/Device/Driver/LX_EOS/LXEosDevice.hpp
```diff
@@ -16,6 +16,85 @@
...
+  const int16_t taskid = 0; // Not used
+
+  // Number of TP without Takeoff, Start, Finish and Landing.
+  char num_of_tp;
+
+  // 1=Turnpoint (also Start and Finish), 2=Landing, 3=Takeoff
+  uint8_t prg[12];
+
+  // TP Longitude in degrees multiplied by 60000.0f
+  int32_t lon[12];
```
MaxKellermann, line 59: What's the byte order?

src/Device/Driver/LX_EOS/LXEosDevice.hpp
```diff
@@ -16,6 +16,85 @@
 
 using std::string_view_literals::operator""sv;
 
+struct EosDeclarationStruct
```
MaxKellermann, line 21: This should have a `static_assert` checking the size of the struct at compile time to verify its correctness

src/Device/Driver/LX_EOS/LXEosDevice.hpp
```diff
@@ -16,6 +16,85 @@
...
+
+  // TP number 1=Start, 2 = TP1, 3=TP2, 4=Finish
+  uint8_t uiTpNr;
+
+  // direction 0= Symmetric, 1=Fixed, 2=Next, 3=Previous
+  uint8_t uiDirection;
+
+  uint8_t bAutoNext = 1; // Is this auto next TP or AAT TP
+  uint8_t bIsLine = 0;   // Is this line flag
+  float fA1;             // Angle A1 in radians
```
MaxKellermann, line 82: Byte order? What bit format is being used? Is this bit format guaranteed on all CPU architectures?

src/Device/Driver/LX_EOS/LXEosDevice.cpp
```diff
@@ -137,6 +137,68 @@ LXEosDevice::SendNewSettings(OperationEnvironment& env)
...
+    for (int count = 8; --count >= 0; d <<= 1) {
+      uint8_t tmp = result ^ d;
+      result <<= 1;
+      if ((tmp & 0x80) != 0)
+        result ^= 0x69;
+    }
+  }
+
+  return result;
+}
```
MaxKellermann, line 200: Duplicate code, already exists in the LXNAV driver

src/Device/Driver/LX_EOS/LXEosDevice.hpp
```diff
@@ -95,6 +97,34 @@ struct EosClassStruct
...
+  uint8_t crc;       // CRC
+} gcc_packed;
+
+struct EosRequestFlightBlock
+{
+  const uint8_t syn = 0x02;
+  const uint8_t cmd = 0xF1;
+
+  // Flight ID is the index (1 = newest), not the ID from FlightInfo
+  uint16_t flight_id;
```
MaxKellermann, line 122: Byte order?

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,270 @@
...
+  std::memcpy(&file_size, &response[89], 4);
+  flight.date = julianToDate(julianDate);
+  flight.start_time = BrokenTime::FromSecondOfDay(time_takeoff);
+  flight.end_time = BrokenTime::FromSecondOfDay(time_landing);
+  flight.internal.lx_eos.file_size = file_size;
+  return true;
+}
+
+bool
+LXEosDevice::GetFlightLogBlock(AllocatedArray<std::byte>& block,
```
MaxKellermann, line 192: This function should return the `AllocatedArray` instead of passing it by reference; throw an exception on error.

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+  /* Flight ID for downloading is the index (1 = newest),
+  not the ID from this message */
+  flight.internal.lx_eos.flight_id = index;
+
+  uint32_t julianDate, time_takeoff, time_landing, file_size;
+  std::memcpy(&julianDate, &response[13], 4);
+  std::memcpy(&time_takeoff, &response[17], 4);
+  std::memcpy(&time_landing, &response[21], 4);
+  std::memcpy(&file_size, &response[89], 4);
```
MaxKellermann, line 183: Byte order? Why are you using a byte array with hard-coded offsets here, instead of declaring a (documented) struct, as you did in other parts of this PR?

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+  if (!WriteAndWaitForACK(message, env))
+    return false;
+
+  try {
+    port.FullRead(std::span{ responseHeader + 1, sizeof(responseHeader) - 1 },
+                  env,
+                  std::chrono::milliseconds(5000));
+  } catch (...) {
+    return false;
```
MaxKellermann, line 215: What's the point of catching this exception? You discard all information about the failure by converting it to`false` here.

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  // Message has 3 parts: header, block, and CRC byte.
+  // Calculate CRC from all 3 parts, result should be 0x00
+  uint8_t crc = CalculateCRC(responseHeader, sizeof(responseHeader), 0xFF);
+  crc = CalculateCRC(block.data(), block.size(), crc);
+  if (CalculateCRC(crc_byte, 1, crc) != 0)
+    return false;
```
MaxKellermann, line 247: Throw exception instead of returning false

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+  return true;
+}
+
+BrokenDate
+LXEosDevice::julianToDate(uint32_t julian_date)
```
MaxKellermann, line 253: I think it could be useful to move this to class BrokenDate, e.g. `BrokenDate::FromJulianDay(unsigned)`.

src/Device/RecordedFlight.hpp
```diff
@@ -49,6 +49,14 @@ struct RecordedFlightInfo : FlightInfo {
      */
     unsigned flytec;
 
+    struct {
+      /**
+       * Flight id and igc file size, used by the LX Eos driver.
+       */
```
MaxKellermann, line 55: A doxygen comment documents only the following field, not both. Split this comment.

src/Device/Driver/LX_EOS/LXEosDeclare.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+  data.bIsLine = (turnpoint.shape == Declaration::TurnPoint::LINE) ? 1 : 0;
+
+  /* We cannot be sure about the format of float on the platform, hardcoded
+   * bytes are used for the angles */
+
+  // Bytes of float values for 45, 90 and 180 degree angles in radians
+  uint8_t angle180[] = { 0xdb, 0x0f, 0x49, 0x40 };
+  uint8_t angle90[] = { 0xdb, 0x0f, 0xc9, 0x3f };
+  uint8_t angle45[] = { 0xdb, 0x0f, 0x49, 0x3f };
```
MaxKellermann, line 139: This is extremely obscure and for no good reason, it allocates 12 bytes on the stack at runtime on every call and initializes them. If you really need constants like that, make them `static constexpr`, but I doubt this is the right way to do it.

src/Device/Driver/LX_EOS/LXEosDeclare.cpp
```diff
@@ -0,0 +1,218 @@
...
+      float elev = declaration.GetWaypoint(tp_nr - 1).GetElevationOrZero();
+      uint32_t elev_bytes = 0;
+      memcpy(&elev_bytes, &elev, 4);
+      elev_bytes = ByteSwap32(elev_bytes);
+      memcpy(&data.fElevation, &elev_bytes, 4);
+    } else {
+      // Set zero if the format is something exotic
+      memset(&data.fElevation, 0, 4);
+    }
+  }
```
MaxKellermann, line 188: All of this is ... awful. This adds runtime overhead by forcing the compiler to generate memcmp() calls, even on CPUs where nothing needs to be changed. All of these checks should be at compile-time and generate only runtime code if runtime code is really necessary. And it should be encapsulated in a generic class that can be reused everywhere. Solve this once, properly, and then reuse this class everywhere you need a float in a specific bit format with a specific byte order.

src/Device/Driver/LX_EOS/LXEosDevice.cpp
```diff
@@ -133,6 +133,50 @@ LXEosDevice::SendNewSettings(OperationEnvironment& env)
...
+  } catch (...) {
+    return false;
+  }
+  return true;
+}
+
+void
+LXEosDevice::CopyStringSpacePadded(char dest[],
+                                   const TCHAR src[],
+                                   const uint8_t len)
```
MaxKellermann, line 159: A buffer length declared as `uint8_t`? Huh? `std::size_t`!

src/Device/Driver/LX_EOS/LXEosDevice.hpp
```diff
@@ -173,14 +250,14 @@ class LXEosDevice : public AbstractDevice
...
- * @return false: coefficients are different or at least one polar is invalid
- */
+  /**
+   * @brief Compare coefficients of polars
+   *
+   * @param polar1
+   * @param polar2
+   * @return true: coefficients are the same
+   * @return false: coefficients are different or at least one polar is invalid
+   */
```
MaxKellermann: Why change indentation here?

src/Device/Driver/LX_EOS/LXEosDevice.hpp
```diff
@@ -247,8 +324,7 @@ class LXEosDevice : public AbstractDevice
    *
    * @param declaration
    * @param env
-   * @return true
-   * @return false
+   * @return true if successful
```
MaxKellermann: ?

src/Device/Driver/LX_EOS/LXEosDevice.hpp
```diff
@@ -259,11 +335,47 @@ class LXEosDevice : public AbstractDevice
    * @param env OperationEnvironment
    * @return true if ACK was received,
    * @return false if time out or other exception occurred
-   * @note timeout is 1 second
+   * @note timeout is 3 seconds
```
MaxKellermann: ?

src/Device/Driver/LX_EOS/LXEosDevice.hpp
```diff
@@ -259,11 +335,47 @@ class LXEosDevice : public AbstractDevice
...
+  /**
+   * @brief Get the Number Of Flights in logger. Maximum number that can be
+   * read is 0xFF
+   *
+   * @param flight_count output number of flights (max 0xFF)
+   * @param env
+   * @return true if successful
+   * @note(flight_count not changed on unsuccess)
+   */
+  bool GetNumberOfFlights(uint8_t& flight_count, OperationEnvironment& env);
```
MaxKellermann, line 352: Return the number of flights instead of passing a reference variable for the result.

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,245 @@
+#include "Device/Error.hpp"
```
MaxKellermann, line 1: License header is missing

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  env.SetProgressRange(flight_count + 1);
+  env.SetProgressPosition(1);
+
+  if (success) {
+    for (uint8_t i = 0; i < flight_count; i++) {
+      RecordedFlightInfo flight;
+
+      // Try up to 5 times before giving up
+      for (uint8_t attempt = 0; attempt < 5; attempt++) {
```
MaxKellermann, line 39: Are you aware that declaring this `uint8_t` is not faster, but actually adds overhead? It's a fallacy.

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,245 @@
...
+  // Download blocks until remaining bytes reach zero
+  for (uint16_t block_id = 0; bytes_remaining > 0; block_id++) {
+    AllocatedArray<std::byte> block;
+
+    // Try up to 5 times before giving up
+    for (uint8_t attempt = 0; attempt < 5; attempt++) {
+      try {
+        block = GetFlightLogBlock(flight_id, block_id, env);
+      } catch (...) {
+        success = false;
```
MaxKellermann, line 86: This discards details about the error condition. Reporting error details to the user (or the log file) is very useful, don't throw that away!

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,245 @@
...
+}
+
+bool
+LXEosDevice::GetNumberOfFlights(uint8_t& flight_count,
+                                OperationEnvironment& env)
+{
+  // Request
+  EosGetNumOfFlights data;
+  auto data_array = reinterpret_cast<std::byte*>(&data);
+  std::span<std::byte> message(data_array, sizeof(data));
```
MaxKellermann, line 131: Complicated and fragile way of casting to a span.

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  // Read response (ACK byte already read)
+  EosNumOfFlightsResponse response;
+  try {
+    port.FullRead(
+      { reinterpret_cast<std::byte*>(&response) + 1, sizeof(response) - 1 },
+      env,
+      std::chrono::milliseconds(5000));
+  } catch (...) {
+    return false;
```
MaxKellermann, line 143: Again: what's the point of this exception handler All it does is discard details about the error condition, but why?

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  // Read response (ACK byte already read)
+  EosFlightInfoResponse info;
+  try {
+    port.FullRead(
+      { reinterpret_cast<std::byte*>(&info) + 1, sizeof(info) - 1 },
+      env,
+      std::chrono::milliseconds(5000));
+  } catch (...) {
+    return false;
```
MaxKellermann, line 182: Again: what's the point of this exception handler All it does is discard details about the error condition, but why?

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  if (Calculate8bitCRC(
+        reinterpret_cast<uint8_t*>(&info), sizeof(info), 0xFF, 0x69) !=
+      0x00) // Check CRC
+    return false;
```
MaxKellermann, line 188: Throw exception instead of returning false with no details.

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,245 @@
...
+  flight.internal.lx_eos.file_size = FromLE32(info.size);
+  return true;
+}
+
+AllocatedArray<std::byte>
+LXEosDevice::GetFlightLogBlock(uint16_t flight_id,
+                               uint16_t block_id,
+                               OperationEnvironment& env)
+{
+  AllocatedArray<std::byte> block;
```
MaxKellermann, line 205: Declare this variable where it is first used; declaring it here bloats the code and adds runtime overhead!

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  // Request
+  EosRequestFlightBlock data;
+  data.flight_id = ToLE16(flight_id);
+  data.block_id = ToLE16(block_id);
+  data.crc = Calculate8bitCRC(
+    reinterpret_cast<uint8_t*>(&data), sizeof(data) - 1, 0xFF, 0x69);
+  std::span<std::byte> message(reinterpret_cast<std::byte*>(&data),
+                               sizeof(data));
```
MaxKellermann, line 214: Clumsy span cast

src/Device/Driver/LX_EOS/LXEosDownload.cpp
```diff
@@ -0,0 +1,245 @@
...
+  std::byte crc_byte[1];
+  port.WaitAndRead(crc_byte, env, std::chrono::milliseconds(5000));
+
+  // Message has 3 parts: header, block, and CRC byte.
+  // Calculate CRC from all 3 parts, result should be 0x00
+  uint8_t crc = Calculate8bitCRC(
+    reinterpret_cast<uint8_t*>(&header), sizeof(header), 0xFF, 0x69);
+  crc = Calculate8bitCRC(block.data(), block.size(), crc, 0x69);
+  if (Calculate8bitCRC(crc_byte, 1, crc, 0x69) != 0)
+    throw("Bad CRC");
```
MaxKellermann, line 242: Don't throw a C string; also the parentheses are unnecessary. `throw` is not a function.

src/Device/RecordedFlight.hpp
```diff
@@ -49,8 +49,16 @@ struct RecordedFlightInfo : FlightInfo {
...
+    struct {
+      uint16_t flight_id;
+      uint32_t file_size;
+    } lx_eos;
+
   } internal;
 };
 
-class RecordedFlightList : public StaticArray<RecordedFlightInfo, 128u> {
+class RecordedFlightList : public StaticArray<RecordedFlightInfo, 128u>{
```
MaxKellermann, line 63: ??

MaxKellermann (changes requested):

src/Device/Driver/LX_EOS/LXEosDeclare.cpp
```diff
@@ -0,0 +1,218 @@
...
+   * bytes are used for the angles */
+
+  // Bytes of float values for 45, 90 and 180 degree angles in radians
+  uint8_t angle180[] = { 0xdb, 0x0f, 0x49, 0x40 };
+  uint8_t angle90[] = { 0xdb, 0x0f, 0xc9, 0x3f };
+  uint8_t angle45[] = { 0xdb, 0x0f, 0x49, 0x3f };
+
+  // Init A21 and A2 to zero
+  memset(&data.fA21, 0, 4);
+  memset(&data.fA2, 0, 4);
```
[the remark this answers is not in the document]
jirimaier: 4 is the size of float in the structure that is transmitted to the vario (if it wasn't, the static_assert of structure size would fail). 
In the previous review, you pointed out that the number format may not be guaranteed on all architectures, so I wanted to explicitly set all bytes to zero rather than expecting that float zero is represented like that on any platform.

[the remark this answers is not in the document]
MaxKellermann: This is all true, but it has nothing to do with my complaint. In this line of code, you don't see what `fA21` is, and you don't know how big it is. This obscures the code. You could have used `sizeof(data.fa21)` or just do `data.fA21 = {};`, but you didn't - instead you hard-coded `4` which makes your code rather obscure.

src/Device/Driver/LX_EOS/LXEosDeclare.cpp
jirimaier: Can I assume that all platforms use IEEE 754 float? How is float affected by big/little endian?

src/Device/Driver/LX_EOS/LXEosDevice.cpp
```diff
@@ -133,6 +133,50 @@ LXEosDevice::SendNewSettings(OperationEnvironment& env)
...
+  port.FullWrite(message, env, std::chrono::milliseconds(1000));
+
+  try {
+    port.WaitForByte(
+      // ACK is 0x06 (NACK is 0x15)
+      std::byte{ 0x06 },
+      env,
+      std::chrono::milliseconds(3000));
+  } catch (...) {
+    return false;
```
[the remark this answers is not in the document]
jirimaier: When calling the method, I am interested in whether the transmission was successful (ACK was received afterward) or not. Why should I keep the details?

src/Device/Driver/LX_EOS/LXEosDeclare.cpp
MaxKellermann: > Can I assume that all platforms use IEEE 754 float?

You need to check https://en.cppreference.com/w/cpp/types/numeric_limits/is_iec559 - I agree to make this a `static_assert`, until we see that a currently supported target doesn't support it.

> How is float affected by big/little endian?

I don't know, I never cared. But if you write code that transmits floats in a binary stream, it's up to you to figure out what byte order (and bit format) is expected/emitted by the device you're talking to, and you must figure out the byte order (and bit format) of all CPUs supported by XCSoar.

As I wrote in another comment, this shouldn't be sprinkled everywhere, but implemented in one central type, like I did with classes like `PackedLE32`.

src/Device/Driver/LX_EOS/LXEosDevice.hpp
jirimaier: big-endian, I added that to the comment

src/Device/Driver/LX_EOS/LXEosDevice.cpp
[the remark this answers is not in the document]
MaxKellermann: To differentiate between "there was no reply at all" or "there was a malformed reply, what the hell is this device saying?" or "there was a nack".

src/Device/Driver/LX_EOS/LXEosDevice.hpp
jirimaier: It returns false if ACK is not received within 3 seconds

src/Device/Driver/LX_EOS/LXEosDevice.cpp
```diff
@@ -133,6 +133,50 @@ LXEosDevice::SendNewSettings(OperationEnvironment& env)
...
+                                OperationEnvironment& env)
+{
+  port.FullFlush(
+    env, std::chrono::milliseconds(50), std::chrono::milliseconds(200));
+  port.FullWrite(message, env, std::chrono::milliseconds(1000));
+
+  try {
+    port.WaitForByte(
+      // ACK is 0x06 (NACK is 0x15)
+      std::byte{ 0x06 },
```
MaxKellermann, line 147: There should be a central protocol definitions giving these magic values a good name. That way, you can avoid adding those comments everywhere. Look at  the LXNAV driver, it already has such a definition! And the Volkslogger driver has another one. Bad duplicate code, of course, but still better than your code.

src/Device/Driver/LX_EOS/LXEosDevice.hpp
MaxKellermann: Then why not use a safe class like PackedBE32 here? That way, it's impossible to misuse it, and you don't even need a comment.

src/Device/Driver/LX_EOS/LXEosDevice.hpp
MaxKellermann: But why is the documentation changed in this commit? Did you change the timeout? Why?

src/util/ByteOrder.hxx
```diff
@@ -263,7 +264,7 @@ public:
 			x = ByteSwap16(x);
 		return x;
 	}
-};
+} __attribute__((packed));
```
MaxKellermann, line 267: This is a proprietary compiler extension that should not be used unconditionally, but why use it here??

MaxKellermann: Also, adding it to this existing structure is unrelated to this commit and is not even explained in the commit message.
Please don't hide unrelated changes. I'll find crap like that!

src/util/ByteOrder.hxx
```diff
@@ -516,7 +517,55 @@ public:
...
+    uint32_t unsigned_int = 0;
+    float floating_point;
+  } __attribute__((packed)) value;
+
+public:
+  PackedFloatLE() = default;
+
+  PackedFloatLE(float f)
+  {
+    value.floating_point = f;
```
MaxKellermann, line 538: This will crash the process with SIGBUS on certain CPUs.

src/util/ByteOrder.hxx
```diff
@@ -516,7 +517,55 @@ public:
...
+    if (IsBigEndian()) {
+      value.unsigned_int = ByteSwap32(value.unsigned_int);
```
MaxKellermann, line 540: This will also crash.

src/util/ByteOrder.hxx
```diff
@@ -516,7 +517,55 @@ public:
...
+} __attribute__((packed));
 
 static_assert(sizeof(PackedLE64) == sizeof(uint64_t), "Wrong size");
 static_assert(alignof(PackedLE64) == 1, "Wrong alignment");
+
+class PackedFloatLE
+{
+  union Value
+  {
+    uint32_t unsigned_int = 0;
```
MaxKellermann, line 529: What's the point of this initializer?

src/util/ByteOrder.hxx
```diff
@@ -516,7 +517,55 @@ public:
...
+    float floating_point;
+  } __attribute__((packed)) value;
+
+public:
+  PackedFloatLE() = default;
+
+  PackedFloatLE(float f)
```
MaxKellermann, line 536: Your implementation is not `constexpr`, that's bad.

src/util/ByteOrder.hxx
```diff
@@ -516,7 +517,55 @@ public:
...
+  {
+    value.floating_point = f;
+    if (IsBigEndian()) {
+      value.unsigned_int = ByteSwap32(value.unsigned_int);
+    }
+  }
+} __attribute__((packed));
+
+static_assert(std::numeric_limits<float>::is_iec559);
```
MaxKellermann, line 545: Why this? I know I suggested doing that, but the next guy won't understand it. And you didn't even document this requirement in the class's documentation. Oh, there's no API documentation at all!

src/util/ByteOrder.hxx
jirimaier: When I used the PackedLE32 (or others) in my structures (which are using this attribute to ensure no padding), the compiler gave error that the attribute (of the struct) is ignored because it contains "unpacked non-POD". Adding the attribute to the classes fixed that.

src/util/ByteOrder.hxx
```diff
@@ -516,7 +517,55 @@ public:
...
+static_assert(sizeof(PackedFloatLE) == sizeof(uint32_t), "Wrong size");
```
MaxKellermann, line 546: This is an obscure requirement. Why should the size of this class be the same as `uint32_t`? It's not a 32-bit integer. It's a 32-bit float. Sure, in the end, both are 4 bytes, but you're not testing for it to be 32 bit or 4 bytes, you're testing this to be uint32_t, but why this choice? This is confusing.

src/util/ByteOrder.hxx
MaxKellermann: So you have no idea what you're doing, and you did some random thing and this made the compiler shut up, so you figured this must be the solution to your problem?

src/util/ByteOrder.hxx
jirimaier: I know for sure that the result (to be sent to the vario) should be 4 bytes. What if the float is not 4 bytes on some platform? Maybe it can't be, I don't know. But it was the reason why I did it like this.

src/util/ByteOrder.hxx
jirimaier: The __attribute__((packed)) is used in an example of data structure in the protocol documentation from LX Navigation. So it is not entirely random. But you are right that I have no deep understanding of what this does, just that it somehow ensures that there is no padding between the variables.

jirimaier: I previously used this for my structures in the LXEosDevic, which was already reviewed by you I think.

jirimaier: Should I rather use gcc_packed (as in the LXNAV driver)? What is the difference?

jirimaier: The LXNAV drive is using the unit and int types in the structures, not the PackedBE32 and others. I think they are actually used nowhere in the code.

src/util/ByteOrder.hxx
jirimaier: However, when the std::numeric_limits<float>::is_iec559 is true, it probably guarantees that sizeof(float)==4

src/util/ByteOrder.hxx
MaxKellermann: > The __attribute__((packed)) is used in an example of data structure in the protocol documentation from LX Navigation.

And LX Navigation's documentation follows which standard?

> So it is not entirely random.

You're missing the point, and it sounds like you don't know what I'm talking about. You havn't even tried to answer my question. I asked why you use it. You have not explained.

Saying it's in the LX Navigation documentation in example code doesn't answer the question, unless you want to tell me that you just copy somebody else's code without understanding it. Code that nobody can explain will never be merged into XCSoar.

> Should I rather use gcc_packed (as in the LXNAV driver)?

Why would you want to do that? This simple question, you're still sidestepping it wherever you can, instead of answering it.

src/util/ByteOrder.hxx
MaxKellermann: Sigh. I already said that you're "**not testing for it to be 32 bit or 4 bytes**". Now you say you do it because you want to test whether it's 4 bytes. This sounds like you didn't read my comment.

MaxKellermann: > However, when the std::numeric_limits::is_iec559 is true, it probably guarantees that sizeof(float)==4

"Probably" isn't good enough, even less so if you're wrong.

src/util/ByteOrder.hxx
```diff
@@ -516,7 +517,61 @@ public:
...
+{
+  uint8_t a, b, c, d;
+
+public:
+  PackedFloatLE() = default;
+
+  constexpr PackedFloatLE(float f)
+  {
+    uint32_t value_as_uint;
+    value_as_uint = *reinterpret_cast<uint32_t*>(&f);
```
MaxKellermann, line 544: This is the worst possible way to express this because the compiler is forced to allocate stack space for the parameter (cannot be an a register) and it violates C++'s aliasing rules, so it's UB.

src/util/ByteOrder.hxx
jirimaier: My understanding is: When I have a structure that contains variables with sizes different than 4 bytes (on 32-bit platform), the compiler may decide to pad the less-than-4-byte variable with empty bytes so that the next variable begins at an address that is multiple of 4, because the CPU can work with that efficiently. This is unwanted for my data structures, because it must exactly match the communication protocol bytes. I already wrote 

> it somehow ensures that there is no padding between the variables

I know about two ways to achieve this: gcc_packed and the attribute((packed)) 
I don't know which of them is preferred. 

And it seems that the attribute cannot work on a struct that contains "unpacked non-POD", so I added the attribute to the PackedBE32 and others too. It makes no practical difference for them (if they had any padding, the static_assert of their size would fail), but it 

> made the compiler shut up

jirimaier: BTW the reason I chose atribute((packed)) rather than gcc_packed is that when using gcc_packed, my IDE was complaining about the asserts checking the size of the structs. Of course, this is just some wrong setting of the IDE, which is not a good reason for deciding which to use. But I wasn't aware of any reason to prefer one over the other anyway.

MaxKellermann: Your understanding is not correct.

jirimaier: You can help me better understand it by telling me if the following claims are true or not:

The compiler can place unused bytes between variables to ensure that variables are properly aligned in the memory.

Aligned means that they can be accessed efficiently. Misaligned causes lower performance or even exceptions.

On a 32-bit platform, properly aligned variables will start at a memory address that is multiple of 4

The data structure used for the communication protocol must not contain any padding.

This can be achieved by telling the compiler not to do any alignment in the structure.

Which is the purpose of the "packed" attribute.

Structure with this attribute cannot have "unpacked non-POD" members.

These are classes or structures that do not have the packed attribute.

So I need to add the attribute to the Packed32BE (and other types I use)

MaxKellermann: No, you can't just put the burden on me and throw this wall of statements at me, making it my responsibility to verify them. I don't have time for that.
And no again, PackedBE32 doesn't need the "padded" attribute.
I'm tired of this PR, I'm going to unsubscribe. When you're completely done, have fixed all the problems and it's no longer in "Draft" status, you may ask for another review, but until then, I decide my time is too precious for you to waste it.

jirimaier: Honestly, I was not expecting you to review it while it is still in "draft" status...

jirimaier: I still don't understand what is going on

The Packed32BE indeed is 4 byte even if the "packed" attribute is not used. 

However, when I try to use the Packed32BE in the struct of data that are transmitted to vario (this struct obviously must be packed), the compiler will say that the packed attribute is ignored because the structure contains "unpacked non-POD". 

Adding the packed attribute to Packed32BE solves that, but you say that it is not correct (or at least not needed) to do so. So what would be a good solution? The Packed32BE is not used anywhere else in the code; I thought this was the intended use-case of Packed32BE, was it? Was it ever used (and working)?

jirimaier: I still think that the attribute needs to be there (to be able to use the Packed32BE inside another struct that is packed)

jirimaier: So if I try to answer my previous questions:

> The compiler can place unused bytes between variables to ensure that variables are properly aligned in the memory.

True (or at least close-enough to true)

> Aligned means that they can be accessed efficiently. Misaligned causes lower performance or even exceptions.

True (or at least close-enough to true)

> On a 32-bit platform, properly aligned variables will start at a memory address that is multiple of 4

False, it is a multiple of the size of the data type, so for example `uint8_t a,b,c,d;` will have no padding.

> The data structure used for the communication protocol must not contain any padding.

True

> This can be achieved by telling the compiler not to do any alignment in the structure.
> Which is the purpose of the "packed" attribute.

True (or at least close-enough to true)

> Structure with this attribute cannot have "unpacked non-POD" members.

That's what the compiler is complaining about

> These are classes or structures that do not have the packed attribute.

Maybe false, but seems like the compiler will be satisfied only if I add the attribute

> So I need to add the attribute to the Packed32BE (and other types I use)

That's what I think. If it is not true, I would be glad for any suggestions on how to solve it in other way.

src/Device/Driver/LX_EOS/LXEosDevice.cpp
```diff
@@ -0,0 +1,199 @@
...
+    return true;
+  } else
+    return false;
+}
+
+bool
+LXEosDevice::HasReliableAltOffset(DeviceInfo device)
+{
+  if (device.product == "LX Eos"sv) {
+    if (device.software_version == "1.7"sv)
```
lordfolken, line 178: Only Version 1.7 or equal or bigger then?

src/Device/Driver/LX_EOS/LXEosDevice.cpp
```diff
@@ -0,0 +1,199 @@
...
+bool
+LXEosDevice::PutMacCready(double mc, OperationEnvironment& env)
+{
+  const std::lock_guard lock{ settings_mutex };
+  vario_settings.mc = mc;
+  return SendNewSettings(env);
+}
+
+bool
+LXEosDevice::PutBallast([[maybe_unused]] double fraction,
```
lordfolken, line 64: Beware, I think xcsoar is currently sending the wrong ratio to the external devices. Its falsley using dry mass instead of polar reference mass while talking to external devices. See my overload2 PR.

lordfolken: Internally its using polar reference mass correctly.

src/Device/Driver/LX_EOS/LXEosDevice.cpp
lordfolken: I would even consider refusing service on a non supported firmware, instead of working around bugs on the device.

src/Device/Driver/LX_EOS/LXEosRegister.cpp
```diff
@@ -13,6 +13,7 @@ LXEosCreateOnPort([[maybe_unused]] const DeviceConfig& config,
 const struct DeviceRegister lx_eos_driver = {
   _T("EOS"),
   _T("Lx Eos 57"),
-  DeviceRegister::RECEIVE_SETTINGS | DeviceRegister::SEND_SETTINGS,
-  LXEosCreateOnPort,
+  DeviceRegister::DECLARE | DeviceRegister::RECEIVE_SETTINGS |
+    DeviceRegister::SEND_SETTINGS,
+  LXEosCreateOnPort
 };
```
lordfolken, line 19: Missing new line

src/Device/Driver/LX_EOS/LXEosDeclare.cpp
```diff
@@ -0,0 +1,174 @@
...
+  [[maybe_unused]] const Declaration& declaration,
+  OperationEnvironment& env)
+{
+  EosClassStruct data;
+
+  // Declaration does not provide class information, set empty
+  CopyStringSpacePadded(data.name, _T(""), sizeof(data.name));
+
+  // CRC is calculated only from data bytes (excluding the SYN and CMD)
+  // Using poly 0x69 and initial 0xFF
```
lordfolken, line 167: Multilin comment should be witg /*

src/Device/Driver/LX_EOS/LXEosParser.cpp
```diff
@@ -0,0 +1,186 @@
...
+    vario_settings.device_polar = polar;
+    CalculateDryMassToReferenceMassRatio(vario_settings);
+  }
+
+  if (vario_settings.dry_mass_to_ref_mass_uptodate) {
+    float ballast_overload = bal / vario_settings.dry_mass_to_ref_mass;
+
+    // Negative result can happen when the empty weight and/or crew weight
+    // setting in XCSoar is higher than in the vario
+    if (ballast_overload < 1)
```
lordfolken, line 138: This can be smaller than 1. Lxnav limits it to .8

NEWS.txt
```diff
@@ -20,6 +20,7 @@ Version 7.42 - 2024/03/01
   - ignore unit for radio frequency in airspace files
 * devices
   - validate wind direction
+  - added driver for LX Navigation Eos57 variometer
```
lordfolken, line 23: The news entry should be on the unrelased version 7.43 not 7.42

lordfolken (dismissed):
I would be willing to merge this, as long as you support this driver. 
Would you be ready for merge?

lordfolken:
> I noticed one more issue:
> 
> If the Eos is connected together with another device (Flarm, for example), the altitude provided by the other dece is prioritized regardless of the order in the device list.
> 
> In other words: the info.ProvideBaroAltitudeTrue(altitude_true) is "weaker" than info.ProvidePressureAltitude(value).
> 
> Can this be considered a bug in XCSoar, or is it a feature?

A feature that doesn't consider your current usecase:
https://github.com/XCSoar/XCSoar/blob/57ab4a6934ebeeecb66740736a104c6fa115cb71/src/Computer/BasicComputer.cpp#L67-L85

jirimaier:
> > I noticed one more issue:
> > If the Eos is connected together with another device (Flarm, for example), the altitude provided by the other dece is prioritized regardless of the order in the device list.
> > In other words: the info.ProvideBaroAltitudeTrue(altitude_true) is "weaker" than info.ProvidePressureAltitude(value).
> > Can this be considered a bug in XCSoar, or is it a feature?
> 
> A feature that doesn't consider your current usecase:
> 
> https://github.com/XCSoar/XCSoar/blob/57ab4a6934ebeeecb66740736a104c6fa115cb71/src/Computer/BasicComputer.cpp#L67-L85

This shouldn't matter in the current implementation, because now it uses the QNH offset from LXWP3 to calculate standard pressure altitude and provides it using ProvidePressureAltitude. ProvideBaroAltitudeTrue is only used in the fallback case used for ERA and older EOS versions (where we don't know if the bug is present or not).

jirimaier:
> I would be willing to merge this, as long as you support this driver. Would you be ready for merge?

Yes, I am ready.

lordfolken:
* I renamed the driver to LxEos simply because it supports the 57 and 80 mm versions of the device?
* The Era I tested as well as i could... 
* In device Register.cpp you have to add the driver to the END of the enum, otherwise you mess up people's configurations
* rebased the code on current master for merge. 

@jirimaier  Do you agree?

lordfolken:
@jirimaier could you have a look at the past two comments?

lordfolken:
Woohoo! Way to go! :-) Thanks for sticking with it.

Issue #970, opened 2022-07-14
Automatically replace AAT delta time infobox by something else for racing task


jirimaier:
AAT delta time is a "must have" infobox for AAT task, but totaly useless for racing task. 

It would be nice to have infobox that would show the delta time for AAT task, but automatically switch to something different for racing task.

I think that a good candidate for that "something different" would be the "task progress" circle, which is more relevant for racing task than AAT.

It should be possible to either:

Create infobox that would be combined for AAT dT and task progress (automatically change depending on task type)

or have independent infobox set(s) for AAT, so that user can define a different layout for different task types 

Note: "racing" type should also apply to the "FAI badges/diplomas" type

PR #1285, opened 2023-08-21
Infobox that combines ETA with AAT dT

jirimaier:
Brief summary of the changes
----------------------------
Added an infobox that shows ETA and Delta time during AAT task and only ETA in case of a racing task.

Automatically replace AAT delta time infobox by something else for racing task #970


jirimaier:
Looks like the checks failed because of problem with downloading third party libs, not because of my code. Can you re-run it?

PR #1286, opened 2023-08-23
Changing order of waypoints being read

### LINKED ISSUE #1122: Changing number of markers changes Home field

kobedegeest:
Markers will now be added after other waypoints instead of before. This will keep the id of home waypoint field and fix #1122

Brief summary of the changes
----------------------------
On start up read user.cup as last waypoint file instead of first file. This will add the markers at the end of the waypoint list instead of at the begining which will keep the id of the home waypoint consistent.

Closes #1122

### DIFF
--- a/src/Waypoint/WaypointGlue.cpp
+++ b/src/Waypoint/WaypointGlue.cpp
@@ -76,6 +76,10 @@ LoadWaypoints(Waypoints &way_points, const RasterTerrain *terrain,
   // Delete old waypoints
   way_points.Clear();
 
+  LoadWaypointFile(way_points, LocalPath(_T("user.cup")),
+                   WaypointFileType::SEEYOU,
+                   WaypointOrigin::USER, terrain, progress);
+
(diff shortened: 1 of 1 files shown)


kobedegeest:
made a mistake and then all commits where messed up after trying to fix it so just made a new pr.

PR #1296, opened 2023-09-03
Lxnav list fix

### LINKED ISSUE #1237: Impossible to download newest flight from LXNAV S8/S80/S10/S100 when more than 128 flights in logger

All versions, probably

Expected behavior
-----------------

When trying to download flight from the logger,  the list should show all flights in the logger, or at least the most recent ones

Actual behavior
---------------

The list shows the 128 *oldest* flights, newer flights can not be downloaded.

XCSoar puts an arbitrary limit for the download list at 128 entries

Easiest: We just increase the number of entries to a value that is higher than the max number of stored flights (which is unfortunately unknown). I tested 512, and so far it works, but in a few years we might reach 512 and it might fail again. I did not find a way to artificially generate flights to test the limit without doing actual flights.

(See pull request for this quick solution)

Otherwise: Use a list without fixed length.

Better solution: Only download the *newest* N flights rather than the oldest. Not sure if the protocol allows that easily, I was not able to understand it in reasonably short time by reading the quite complicated download code.

groundhogxc:
Brief summary of the changes
----------------------------

- Display the latest 128 flights in the flight download dialog for LXNav devices, instead of the 512 oldest ones
- Display the list with newest additions first so you do not have to scroll down to the latest flight

Closes #1237 
(this time for good, I hope)

I made the functional test (with an actual lx device) based on version 7.36, and have since rebased to 7.39 for the PR (still compiles and runs, but have not had access to logger since). Hope that is OK,

List reversal happens in driver rather than in UI as different devices/manufacturers might deliver flight lists in different orders.

### DIFF
--- a/src/Device/Driver/LX/NanoLogger.cpp
+++ b/src/Device/Driver/LX/NanoLogger.cpp
@@ -205,7 +205,12 @@ Nano::ReadFlightList(Port &port, RecordedFlightList &flight_list,
 
   env.SetProgressRange(nflights);
 
-  unsigned requested_tail = 1;
+  /* Start download at first flight in logger if capacity of flight_list is
+     enough for all flights in logger. Otherwise, calculate the starting
+     point to fill flight_list to capacity with only the latest flights. */
+  unsigned requested_tail = (unsigned) std::max(1,
+                     (signed) nflights - (signed) flight_list.max_size() + 1);
+
   while (true) {
     const unsigned room = flight_list.max_size() - flight_list.size();
     const unsigned remaining = nflights - requested_tail + 1;
--- a/src/Device/RecordedFlight.hpp
+++ b/src/Device/RecordedFlight.hpp
@@ -51,5 +51,5 @@ struct RecordedFlightInfo : FlightInfo {
   } internal;
 };
 
-class RecordedFlightList : public StaticArray<RecordedFlightInfo, 512u> {
+class RecordedFlightList : public StaticArray<RecordedFlightInfo, 128u> {
 };
--- a/src/Logger/ExternalLogger.cpp
+++ b/src/Logger/ExternalLogger.cpp
@@ -274,6 +274,9 @@ ExternalLogger::DownloadFlightFrom(DeviceDescriptor &device)
     return;
   }
 
+  /* Reverse flight list so that most recent flights are shown first */
+  std::reverse(flight_list.begin(),flight_list.end());
+
   const auto logs_path = MakeLocalPath(_T("logs"));
 
   while (true) {


groundhogxc:
Forced pushed a new version, just to fix indentation (spaces -> tabs)

MaxKellermann:
The TrivialArray::reverse commit is pointless.

groundhogxc:
Awww, so sad, it was such a nice method. :)
Ok I removed it. It did seem in line with other methods in TrvialArray, to the untrained eye...

MaxKellermann:
It's not always clear which operations should be a freestanding function and which ones should be a method, but in this case, no C++ standard container has it, and the C++ standard says it's a freestanding function. So adding a wrapper to this function as a method doesn't follow any existing example.

src/Device/Driver/LX/NanoLogger.cpp
```diff
@@ -205,7 +205,7 @@ Nano::ReadFlightList(Port &port, RecordedFlightList &flight_list,
 
   env.SetProgressRange(nflights);
 
-  unsigned requested_tail = 1;
+  unsigned requested_tail = (unsigned) std::max(1, (signed) nflights - (signed) flight_list.max_size() + 1);
```
MaxKellermann, line 208: This looks awfully complicated and obscure. Can you rephrase this in a way that's easier to understand, or at least add an explanatory code comment?

src/Device/RecordedFlight.hpp
```diff
@@ -51,5 +51,5 @@ struct RecordedFlightInfo : FlightInfo {
   } internal;
 };
 
-class RecordedFlightList : public StaticArray<RecordedFlightInfo, 512u> {
+class RecordedFlightList : public StaticArray<RecordedFlightInfo, 128u> {
```
MaxKellermann, line 54: This is an unrelated change which I don't understand, and your commit message doesn't mention it.

src/Device/Driver/LX/NanoLogger.cpp
groundhogxc: Well, this is what happens with different backgrounds. For me this code is very easy to read, but the serious C++ things very hard :-).
But I will happily amend - would you like me to write a longer code that is easier to read instead, or just to add explanatory comments? I will add more documentation in both cases though.

src/Device/RecordedFlight.hpp
groundhogxc: This is in response to your concern that downloading a list of 512 flights might take very long, in the discussion on the issue #1237.
It is mentioned in the commit message that the max list length is reduced from 512 to 128.

src/Device/Driver/LX/NanoLogger.cpp
```diff
@@ -225,6 +225,8 @@ Nano::ReadFlightList(Port &port, RecordedFlightList &flight_list,
     env.SetProgressPosition(requested_tail - 1);
   }
 
+  std::reverse(flight_list.begin(),flight_list.end());
```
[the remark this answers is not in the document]
groundhogxc: No - there is no concern (yet) about what one specific user prefers in this code, just like it was before. It is just assumed that the user would prefer to see the most recent flight first, without having to scroll.

The stated reason is that other types of devices such as a FLARM or Butterfly or some future devices might already deliver the flights in reverse order in which case the reversal is not required, so instead of putting it into the UI affecting all drivers at the same time it is in the driver to act on LXnav devices specifically. If you prefer another solution then that is also fine of course, but this is how I am able to implement it.

src/Device/Driver/LX/NanoLogger.cpp
MaxKellermann: If you think the code is good, add a comment explaining what this formula does (and why it's there).

src/Device/RecordedFlight.hpp
MaxKellermann: True, you can read the commit message that way (I didn't), but what it does is reduce the number of flights for ALL drivers, which is out of this commit's scope. I think that reducing the number of flights again is a good idea now that you've found a new approach, but this global change deserves a separate commit, instead of being mixed in this one.

src/Device/Driver/LX/NanoLogger.cpp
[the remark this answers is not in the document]
MaxKellermann: It is indeed a good idea to make it most easy for the user to select the most recent flight, but that is purely a UI thing and the driver should not make any assumptions.
The ordering of flights is an implementation detail of the logger firmware, and I don't know if any vendor has documented a guaranteed ordering of flights. This means it can change with any minor firmware update, breaking your code.
I think it really belongs in the UI and not in the driver. I'd agree to merge a PR that adds sorting to the UI, but let's keep UI things out of the driver code.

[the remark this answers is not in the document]
groundhogxc: I only know the LXNav devices, I never got FLARM to work with my hardware, and never used other types of loggers. Which order do other drivers provide? Is it true that all of them deliver oldest flight first? That is the relevant question.  (And in that case UI is the better place)

(At any rate I will not be able to do the UI thing for the foreseeable future, I will keep my implementation for my own flights)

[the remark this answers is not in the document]
groundhogxc: > (At any rate I will not be able to do the UI thing for the foreseeable future, I will keep my implementation for my own flights)

Actually, it is not so bad. The same code can be applied on the UI level in the dialog code as it uses still the same data type. I have that commit, it compiles fine. It is the same line of code in a different file. So, now it is implemented in the UI, I just hope that users of other logger types now do not complain about suddenly having oldest flight first.

So that is fine, however I am not able to test this version with an actual device. The same code in the driver was tested with an actual device. Of course I have tested it even less with other types of devices, which are now all affected. If you consider my previous test good enough, then use the commit. Otherwise I remove it from the PR until I have tested my own XC version, which includes it.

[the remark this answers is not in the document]
MaxKellermann: No, that's bad. You can't just reverse the order. The order of flights from the logger (or from the driver) is undefined. If stuff works now, it is by coincidence, but you must not break that. You can replace it with a well-defined order, but not randomly with something else.
The right function you need is `std::sort()`, not `std::reverse()`.

PR #1299, opened 2023-09-13
add .vscode to .gitignore

GeorgeDowning20:
Add .vscode to .gitignore


MaxKellermann:
No. Add this to your own global gitignore. There's no point in adding each and everybody's private ignores to the project - that's a bottomless pit.

PR #1301, opened 2023-09-14
Android: update android:targetSdkVersion to 33

zinuzoid:
Google enforce android-33 now. We won't be able to upload APK version without this. https://developer.android.com/google/play/requirements/target-sdk
https://developer.android.com/about/versions/13/behavior-changes-13

Some dependency is already updated in https://github.com/XCSoar/XCSoar/commit/db34adf9bd4e85d1023098ba0592d2420cb2bafc, but not Manifest file. 

I tested in my fork and it's working fine in my Pixel 7 Pro.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -77,7 +77,7 @@
 
   </application>
 
-  <uses-sdk android:minSdkVersion="21" android:targetSdkVersion="31"/>
+  <uses-sdk android:minSdkVersion="21" android:targetSdkVersion="33"/>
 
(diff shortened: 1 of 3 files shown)


MaxKellermann:
What are the consequences for raising the targetSdkVersion? Is there a behavior change like there was with other changes of that field?

zinuzoid:
Screenshot 2023-09-18 at 01 36 26
Screenshot 2023-09-18 at 01 36 44
Screenshot 2023-09-18 at 01 36 38


Based on the documentation, these are the most related but I don't think we need any changes.

Issue #1206, opened 2023-05-03
Feature request: Display range circles on map


DanD222:
What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------
It would be nice if XCSoar could optionally display range circles on the map, centred around the glider.

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------
This would allow a rough distance estimate to points all around the aircraft.

I would imagine the map circles to be enabled via a checkbox in Settings / Map Display, and also with a QuickMenu button.

The spacing increments should be automatically switched, for example 500m, 1km, 5km, 10km, 20km and 50km, depending on the map zoom level. 

Each spacing circle could have the distance marked next to the circle, for example in the 6 o'clock position below the circle.

xtophe80:
I have a very preliminary implementation in my local tree. I haven't test flown it yet. I'll push it to my own repo later this week in case you want to use/contribute

lordfolken:
Implementation in progress: #2522

WolfgangDrescher:
Thank you for linking this issue, @lordfolken. I had missed that one already existed. Do @kobedegeest and @DanD222 have any feedback on the distance ring presets in `DistanceRingsRenderer::SelectRingSet`? I also implemented basic settings for nautical miles and statute miles.

DanD222:
Awesome, thanks! I've had a quick look, and I'm not sure about the `50000, 75000` step in `RINGS_KM_WIDE`; I'd probably prefer it if would continue in a linear fashion, i.e. 
`10000, 20000, 30000, 40000, 50000, 60000, 70000, 80000, 90000, 100000`

(also @lordfolken not sure if https://github.com/XCSoar/XCSoar/pull/2526 should be linked to this?)

PR #1303, opened 2023-09-19
thermal assistant icon position based on 30s avg

GeorgeDowning20:
Adjust the position of the glider icon in the thermal assistant according to the vario average. makes the thermal assistant more intuitive to use.

### DIFF
--- a/src/Gauge/ThermalAssistantRenderer.cpp
+++ b/src/Gauge/ThermalAssistantRenderer.cpp
@@ -88,6 +88,25 @@ ThermalAssistantRenderer::NormalizeLift(double lift, double max_lift)
   return std::clamp(lift, 0., 1.);
 }
 
+static void
+DrawCircleLabel(Canvas &canvas, PixelPoint p,
+                tstring_view text) noexcept
+{
+  const auto size = canvas.CalcTextSize(text);
+  p.x -= size.width / 2;
+  p.y -= size.height * 3 / 4;
+
+  canvas.DrawText(p, text);
+}
+
+static void
+DrawCircleLabelVSpeed(Canvas &canvas, PixelPoint p, double value) noexcept
+{
+  TCHAR buffer[10];
+  FormatUserVerticalSpeed(value, buffer, ARRAY_SIZE(buffer));
+  DrawCircleLabel(canvas, p, buffer);
+}
+
 void
 ThermalAssistantRenderer::PaintRadarPlane(Canvas &canvas) const
 {
@@ -106,25 +125,6 @@ ThermalAssistantRenderer::PaintRadarPlane(Canvas &canvas) const
                        +Layout::FastScale(small ? 2 : 4)),
                   p.At(-Layout::FastScale(small ? 2 : 4),
                        +Layout::FastScale(small ? 2 : 4)));
-}
-
-static void
-DrawCircleLabel(Canvas &canvas, PixelPoint p,
-                tstring_view text) noexcept
-{
-  const auto size = canvas.CalcTextSize(text);
-  p.x -= size.width / 2;
-  p.y -= size.height * 3 / 4;
-
-  canvas.DrawText(p, text);
-}
-
-static void
-DrawCircleLabelVSpeed(Canvas &canvas, PixelPoint p, double value) noexcept
-{
-  TCHAR buffer[10];
-  FormatUserVerticalSpeed(value, buffer, ARRAY_SIZE(buffer));
-  DrawCircleLabel(canvas, p, buffer);
 }
 
 void
--- a/src/Gauge/ThermalAssistantRenderer.cpp
+++ b/src/Gauge/ThermalAssistantRenderer.cpp
@@ -108,11 +108,13 @@ DrawCircleLabelVSpeed(Canvas &canvas, PixelPoint p, double value) noexcept
 }
 
 void
-ThermalAssistantRenderer::PaintRadarPlane(Canvas &canvas) const
+ThermalAssistantRenderer::PaintRadarPlane(Canvas &canvas, double max_lift) const
 {
+  int normalised_average = NormalizeLift(vario.average, max_lift) * radius;
+
   canvas.Select(look.plane_pen);
 
-  PixelPoint p = mid.At(circling.TurningLeft() ? (int)radius : (int)-radius,
+  PixelPoint p = mid.At(circling.TurningLeft() ? normalised_average : -normalised_average,
                         0);
 
   canvas.DrawLine(p.At(+Layout::FastScale(small ? 5 : 10),
@@ -209,5 +211,5 @@ ThermalAssistantRenderer::Paint(Canvas &canvas)
   PaintPoints(canvas, lift_points);
   PaintAdvisor(canvas, lift_points);
 
-  PaintRadarPlane(canvas);
+  PaintRadarPlane(canvas,max_lift);
 }
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -3,6 +3,7 @@ Version 7.40 - not yet released
   - Added infobox that combines ETA with AAT dT
   - FLARM: Add 100m zoom option.
   - hide mouse cursor after 10 seconds of inactivity
+  - Move glider icon on thermal assist according to 30s average.
 * Linux
   - fix the /etc/xcsoar data directory
 * Android
--- a/src/Gauge/ThermalAssistantRenderer.hpp
+++ b/src/Gauge/ThermalAssistantRenderer.hpp
@@ -80,7 +80,7 @@ protected:
 
   void CalculateLiftPoints(LiftPoints &lift_points, double max_lift) const;
   double CalculateMaxLift() const;
-  void PaintRadarPlane(Canvas &canvas) const;
+  void PaintRadarPlane(Canvas &canvas, double max_lift) const;
   void PaintRadarBackground(Canvas &canvas, double max_lift) const;
   void PaintPoints(Canvas &canvas, const LiftPoints &lift_points) const;
   void PaintAdvisor(Canvas &canvas, const LiftPoints &lift_points) const;


DanD222:
@GeorgeDowning20 I’m having a hard time imagining what the difference is, would you have a screenshot by any chance please?

GeorgeDowning20:
> @GeorgeDowning20 I’m having a hard time imagining what the difference is, would you have a screenshot by any chance please?
Screenshot from 2023-09-19 23-59-56
Screenshot from 2023-09-19 23-59-12

The position of the glider icon is the 30s average:

1) Where the thermal assistant trace is normalized to fit in the circle, this upgrade helps the user quantify the magnitude of the trace relative to their average. useful for gauging the significance of spikes in the trace. 

2) The assistant is more intuitive: 
-if the lift is stronger than your average (best bit), the glider is shown to be inside the trace, and highlighted blue.
-if the lift is less strong than your trace (worst bit), the glider is shown to be outside of the trace.

This makes it significantly clearer to identify which way you need move to get better centered.

lordfolken:
I like the change very much.
Two things
  Lets document it somewhere 
  A ring would be indeed very nice

GeorgeDowning20:
> Lets document it somewhere

as in the manual?

lordfolken:
Ideally both. And the help text in the enable disable feature.
But best would be intuitive recognition. 
Potential idea: Maybe a 30s label beside the plane icon?

GeorgeDowning20:
> plane
Screenshot from 2023-09-20 18-45-21

not sure if I like the Macready circle. What do you think? It could be useful but clutters things a bit.
I have also added a smoothing function.

lordfolken:
Instead of MC, how about you hook the circle to the plane icon and adapt its size to the 30s average?

Otherwise the circle would have to have a different color, or change the color whenever the climb value exceeds the average.

DanD222:
> 
> not sure if I like the Macready circle. What do you think? It could be useful but clutters things a bit.

I quite like the idea, I think it would look a lot better if the colour of the MC circle was different though, and in a contrasting colour to the thermal shape colour.

GeorgeDowning20:
> > not sure if I like the Macready circle. What do you think? It could be useful but clutters things a bit.
> 
> I quite like the idea, I think it would look a lot better if the colour of the MC circle was different though, and in a contrasting colour to the thermal shape colour.

i have had a play with this... also not a huge fan. can't find a colour scheme that doesn't clash. my plan is to change the color of the plane icon

MaxKellermann:
I don't get it. I have no idea what the change is and how it is supposed to improve something. I don't understand the screenshots.

And in my opinion, the thermal assistant certainly should not have this MC ring. Ideally, the MC should be the same as the average lift, and then both shapes overlap in confusing ways, adding clutter, but without any practical value.

GeorgeDowning20:
> I don't get it. I have no idea what the change is and how it is supposed to improve something. I don't understand the screenshots.

commit 1:
-Apply a smoothing filter to the thermal assistant trace. This improves the readability, where it can be difficult in weak or -gusty thermals.

Commit 2:
- The glider is positioned on the graph according to the average climb rate. This is a change from the original where the glider is positioned at the "max lift" mark, which is a useless measure. This improves the feature because the thermal assistant becomes more intuitive to use.

> And in my opinion, the thermal assistant certainly should not have this MC ring. Ideally, the MC should be the same as the average lift, and then both shapes overlap in confusing ways, adding clutter, but without any practical value.

MC ring idea dropped. in theory, this is useful for observing the best part of the lift. If your average climb rate starts to drop, you can interpolate if it will increase again in the next turn or if you just leave. However, I can't find a non-cluttered way of implementing this.

MikeBravo15:
How about to have only MC ring?

MaxKellermann:
>     * Apply a smoothing filter to the thermal assistant trace. This improves the readability, where it can be difficult in weak or gusty thermals.

Sure, smoothing makes things look nicer, but also increases latency a lot and discards a lot of detail. I don't think this is a good idea. Showing all the gory details (including all the gusts), is the whole point of the thermal assistant.

>     * The glider is positioned on the graph according to the average climb rate. This is a change from the original where the glider is positioned at the "max lift" mark, which is a useless measure. This improves the feature because the thermal assistant becomes more intuitive to use.

I have no idea what that means. How can a glider be positioned according to a climb rate?

> MC ring idea dropped. in theory, this is useful for observing the best part of the lift. If your average climb rate starts to drop, you can interpolate if it will increase again in the next turn or if you just leave. However, I can't find a non-cluttered way of implementing this.

Maybe choose the line color or fill color depending on whether that part of the line is above or below the MC (or this thermal's average). Make it more similar to the trace.

DanD222:
> How can a glider be positioned according to a climb rate?

Here's my understanding of the situation - the red glider icons are where the glider would be positioned, given the glider's 30 second average climb was 0 kts and 7.8 kts.

The green circle is my understanding of the MC circle, which I don't think is all that bad - it shows me where the thermal is stronger than MC; the question is whether the additional information of `thermal > MC` in addition to `thermal > now` is worth the slight additional degradation in clarity.

image


> Showing all the gory details (including all the gusts), is the whole point of the thermal assistant.

My first thought regarding this is whether gusts that happened in the last circle would be useful/relevant to coring the thermal - I would think that only the best _consistent_ part of the climb is ultimately of interest in centering a thermal, so in practice I foresee the need to mentally "unimagine" them to get the best results in centering.

GeorgeDowning20:
> My first thought regarding this is whether gusts that happened in the last circle would be useful/relevant to coring the thermal - I would think that only the best _consistent_ part of the climb is ultimately of interest in centering a thermal, so in practice I foresee the need to mentally "unimagine" them to get the best results in centering.

if the sampling rate was much higher, seeing gust could be really useful. You would be able to see the turbulence signature around cores. In the mean time, gust are not useful to centre cores.

GeorgeDowning20:
> Sure, smoothing makes things look nicer, but also increases latency a lot and discards a lot of detail. I don't think this is a good idea. Showing all the gory details (including all the gusts), is the whole point of the thermal assistant.

the smoothing filter has been applied, around the unit circle, Y = 0.5z + 0.25z+1 + 0.25z-1. Therefore introduces no latency, just smooths out gusts.

if you don’t like I will remove.

The trouble with the existing systems is: although the positions on the graph for gusts is technically correct, guests are represented by too much surface area. This masks where the less strong but consistent lift is.

DanD222:
> if the sampling rate was much higher

This is determined by the GPS position update rate, if I understand correctly? Probably tending to over-complicate things again, but I like the idea of the smoothing to be switchable in preferences - one could try if one prefers the thermal assistant with or without smoothing, and could make use of the gust information in case of a high update rate position source.

src/Gauge/ThermalAssistantRenderer.cpp
```diff
@@ -88,12 +88,33 @@ ThermalAssistantRenderer::NormalizeLift(double lift, double max_lift)
...
+  canvas.DrawText(p, text);
+}
+
+static void
+DrawCircleLabelVSpeed(Canvas &canvas, PixelPoint p, double value) noexcept
+{
+  TCHAR buffer[10];
+  FormatUserVerticalSpeed(value, buffer, ARRAY_SIZE(buffer));
+  DrawCircleLabel(canvas, p, buffer);
+}
```
MaxKellermann, line 108: Creating duplicate code is not acceptable. Don't do that.

MaxKellermann:
> Here's my understanding of the situation - the red glider icons are where the glider would be positioned, given the glider's 30 second average climb was 0 kts and 7.8 kts. As the glider's 30 second average changes, I would expect the glider icon to move along the x axis (please correct me if I'm wrong).

I don't expect it to do that. This feature does not make sense to me at all. I don't get why the glider should move anywhere. The only reason why it's there is to show the pilot in one quick glance whether he's on the left edge of the thermal (i.e. circling right) or on the right edge (circling left).

Moving the glider symbol around makes it harder to see that with one quick look. Not only that, this position doesn't even encode any useful information.

Also I don't get why this commit diff is so large. All it should do, according to the description, is change the X position, but there is so much more code. What does it do? Is it worth taking the time to try to read and understand it? I don't have time to waste, and I wasted so much time on reading PRs which the author would never finish, so I'm cautious where I spend the little time I have.

GeorgeDowning20:
> > Here's my understanding of the situation - the red glider icons are where the glider would be positioned, given the glider's 30 second average climb was 0 kts and 7.8 kts. As the glider's 30 second average changes, I would expect the glider icon to move along the x axis (please correct me if I'm wrong).
> 
> I don't expect it to do that. This feature does not make sense to me at all. I don't get why the glider should move anywhere. The only reason why it's there is to show the pilot in one quick glance whether he's on the left edge of the thermal (i.e. circling right) or on the right edge (circling left).
> 
> Moving the glider symbol around makes it harder to see that with one quick look. Not only that, this position doesn't even encode any useful information.
> 
> Also I don't get why this commit diff is so large. All it should do, according to the description, is change the X position, but there is so much more code. What does it do? Is it worth taking the time to try to read and understand it? I don't have time to waste, and I wasted so much time on reading PRs which the author would never finish, so I'm cautious where I spend the little time I have.

Something weird is going on with the gitt diff... i have moved a function in-order to access an undeclared function. it should look like this...

Screenshot from 2023-09-22 20-23-51

I will declare the existing undeclared functions. and move the edited function back. the code should be really simple

src/Gauge/ThermalAssistantRenderer.cpp
```diff
@@ -17,6 +17,13 @@
 
 #include <algorithm> // for std::clamp()
 
+static void
+DrawCircleLabel(Canvas &canvas, PixelPoint p,
+                tstring_view text) noexcept;
+
+static void
+DrawCircleLabelVSpeed(Canvas &canvas, PixelPoint p, double value) noexcept;
```
MaxKellermann, line 25: Undeclared? What does that mean?

GeorgeDowning20:
fixed up the git diff, so is now more presentable. Sorry about that.

> I don't expect it to do that. This feature does not make sense to me at all. I don't get why the glider should move anywhere. The only reason why it's there is to show the pilot in one quick glance whether he's on the left edge of the thermal (i.e. circling right) or on the right edge (circling left).


The reason is: that the thermal assistant is effectively a graph. The icon is placed on the graph and therefore represents a value. At present this value is the 'max lift in the last turn'. This is a useless bit of information. By encoding the 30s average as this value, produces a handful of advantages as described above.

src/Gauge/ThermalAssistantRenderer.cpp
```diff
@@ -113,6 +115,14 @@ ThermalAssistantRenderer::PaintRadarPlane(Canvas &canvas) const
...
                   p.At(-Layout::FastScale(small ? 2 : 4),
                        +Layout::FastScale(small ? 2 : 4)));
+
+  if (small)
+    return;
+
+  if (!circling.TurningLeft()) 
+    DrawCircleLabelVSpeed(canvas, p + PixelSize{-45, 0}, vario.average);
+  else 
+    DrawCircleLabelVSpeed(canvas, p + PixelSize{+45, 0}, vario.average);
```
MaxKellermann, line 125: Why is this here? This is not described anywhere, your commit message doesn't mention it.

src/Gauge/ThermalAssistantRenderer.cpp
```diff
@@ -96,11 +96,13 @@ ThermalAssistantRenderer::NormalizeLift(double lift, double max_lift)
...
 void
-ThermalAssistantRenderer::PaintRadarPlane(Canvas &canvas) const
+ThermalAssistantRenderer::PaintRadarPlane(Canvas &canvas, double max_lift) const
 {
+  int normalised_average = NormalizeLift(vario.average, max_lift) * radius;
+
   canvas.Select(look.plane_pen);
 
-  PixelPoint p = mid.At(circling.TurningLeft() ? (int)radius : (int)-radius,
+  PixelPoint p = mid.At(circling.TurningLeft() ? (int)normalised_average : (int)-normalised_average,
```
MaxKellermann, line 105: What's the point of casting an `int` to an `int`?

src/Gauge/ThermalAssistantRenderer.cpp
GeorgeDowning20: lordfolken suggested it? i will refactor PR to include

DanD222:
> Not only that, this position doesn't even encode any useful information.

Here's my interpretation of the additional information:
image

src/Gauge/ThermalAssistantRenderer.cpp
GeorgeDowning20: perhaps my terminology is wrong. Initially the 'PaintRadarPlane' radar function did not have access to these functions. My initial approach was to move the function but this caused diff problems. Therefore I have declared the function at the top. please can you suggest a better method?

GeorgeDowning20:
> > Not only that, this position doesn't even encode any useful information.
> 
> Here's my interpretation of the additional information: image

I have removed the MC ring as Max strongly objects "And in my opinion, the thermal assistant certainly should not have this MC ring."

src/Gauge/ThermalAssistantRenderer.cpp
MaxKellermann: You can move functions if they need to be moved. Just make it a separate commit (and let the commit message explain why this move is needed), or else you confuse reviewers.
Adding unnecessary forward declarations just clutters the code.

GeorgeDowning20: roger. thanks for your help. sorted

src/Gauge/ThermalAssistantRenderer.cpp
```diff
@@ -127,6 +127,14 @@ ThermalAssistantRenderer::PaintRadarPlane(Canvas &canvas, double max_lift) const
...
                   p.At(-Layout::FastScale(small ? 2 : 4),
                        +Layout::FastScale(small ? 2 : 4)));
+
+  if (small)
+    return;
+
+  if (!circling.TurningLeft()) 
+    DrawCircleLabelVSpeed(canvas, p + PixelSize{-45, 0}, vario.average);
+  else 
+    DrawCircleLabelVSpeed(canvas, p + PixelSize{+45, 0}, vario.average);
```
MaxKellermann, line 137: These two lines are duplicate, the only difference is the sign. This should be parametrized. Also where does this `45` come from? Don't hard-code pixel sizes!

MaxKellermann:
I've cherry-picked the first two commits, even though I don't like the feature. The third commit needs to be fixed, but I still don't quite get the point, and I don't like it either; it just adds more graphical clutter for something that is supposed to be very simple.

GeorgeDowning20:
> I've cherry-picked the first two commits, even though I don't like the feature. The third commit needs to be fixed, but I still don't quite get the point, and I don't like it either; it just adds more graphical clutter for something that is supposed to be very simple.

Thank you for your help getting this up to scratch. The first two commits represent my original idea.

The label and MC ring, are additional stuff that has come out of this thread. I will continue working on them if wanted else i will focus my attention elsewhere.

DanD222:
> I have removed the MC ring as Max strongly objects "And in my opinion, the thermal assistant certainly should not have this MC ring." but thank you, this shows the additional information well.
> 
> The label and MC ring, are additional stuff that has come out of this thread. I will continue working on them if wanted else i will focus my attention elsewhere.

Fair enough, let it serve "for illustration purposes only" then. I still see some value in the additional information gained, and I like the idea of possible better coring with the smoothed thermal graph on low-update screens, but Better flarm colours are significantly more important than either of those.

There's also potentionally low-hanging fruit here? https://github.com/XCSoar/XCSoar/issues/1206

PR #1305, opened 2023-09-30
Csv newlines

twhitehead:
Technically the CSV standard lets you have newlines inside quoted items.

Adding support for this makes some minimal layout formatting of desc and user fields possible, which make them a lot easier to digest at a glance.

Here is an example of using the decs field for multi-line data

image

### DIFF
--- a/build/libio.mk
+++ b/build/libio.mk
@@ -11,6 +11,7 @@ IO_SOURCES = \
 	$(IO_SRC_DIR)/MemoryReader.cxx \
 	$(IO_SRC_DIR)/BufferedReader.cxx \
 	$(IO_SRC_DIR)/BufferedLineReader.cpp \
+	$(IO_SRC_DIR)/BufferedCsvReader.cpp \
 	$(IO_SRC_DIR)/FileDescriptor.cxx \
 	$(IO_SRC_DIR)/FileMapping.cpp \
 	$(IO_SRC_DIR)/FileReader.cxx \
--- a/build/libwaypointfile.mk
+++ b/build/libwaypointfile.mk
@@ -9,6 +9,6 @@ WAYPOINTFILE_SOURCES = \
 	$(SRC)/Waypoint/WaypointFileType.cpp \
 	$(SRC)/Waypoint/WaypointReader.cpp
 
-WAYPOINTFILE_DEPENDS = WAYPOINT CUPFILE UNITS
+WAYPOINTFILE_DEPENDS = WAYPOINT CUPFILE UNITS IO
 
 $(eval $(call link-library,libwaypointfile,WAYPOINTFILE))
--- a/src/Task/TaskFileSeeYou.cpp
+++ b/src/Task/TaskFileSeeYou.cpp
@@ -3,7 +3,9 @@
 
 #include "Task/TaskFileSeeYou.hpp"
 #include "io/BufferedReader.hxx"
+#include "io/BufferedCsvReader.hpp"
 #include "io/FileReader.hxx"
+#include "io/StringConverter.hpp"
 #include "Engine/Waypoint/Waypoints.hpp"
 #include "Waypoint/CupParser.hpp"
 #include "Waypoint/WaypointReaderSeeYou.hpp"
@@ -425,18 +427,8 @@ static bool
 ParseSeeYouWaypoints(BufferedReader &reader, Waypoints &way_points)
 {
   const WaypointFactory factory(WaypointOrigin::NONE);
-  WaypointReaderSeeYou waypoint_file(factory);
 
-  while (true) {
-    char *line = reader.ReadLine();
-    if (line == nullptr)
-      return false;
-
-    if (StringIsEqualIgnoreCase(line, "-----Related Tasks-----"))
-      return true;
-
-    waypoint_file.ParseLine(line, way_points);
-  }
+  return ParseSeeYou(factory, way_points, reader);
 }
 
 static char *
--- a/src/Waypoint/WaypointReader.cpp
+++ b/src/Waypoint/WaypointReader.cpp
@@ -28,7 +28,7 @@ CreateWaypointReader(WaypointFileType type, WaypointFactory factory)
     return new WaypointReaderWinPilot(factory);
 
   case WaypointFileType::SEEYOU:
-    return new WaypointReaderSeeYou(factory);
+    break;
 
   case WaypointFileType::ZANDER:
     return new WaypointReaderZander(factory);
@@ -56,6 +56,9 @@ ReadWaypointFile(Reader &file_reader, WaypointFileType file_type,
   BufferedReader buffered_reader{progress_reader};
 
   switch (file_type) {
+  case WaypointFileType::SEEYOU:
+    ParseSeeYou(factory, way_points, buffered_reader);
+    break;
   default:
     std::unique_ptr<WaypointReaderBase> reader { CreateWaypointReader(file_type,
                                                                       factory) };
--- a/src/Waypoint/WaypointReader.cpp
+++ b/src/Waypoint/WaypointReader.cpp
@@ -47,15 +47,24 @@ CreateWaypointReader(WaypointFileType type, WaypointFactory factory)
 }
 
 static void
-ReadWaypointFile(Reader &file_reader, uint_least64_t total_size,
-                 WaypointReaderBase &waypoint_reader,
-                 Waypoints &way_points,
+ReadWaypointFile(Reader &file_reader, WaypointFileType file_type,
+                 uint_least64_t total_size,
+                 Waypoints &way_points, WaypointFactory factory,
                  ProgressListener &progress)
 {
   ProgressReader progress_reader{file_reader, total_size, progress};
   BufferedReader buffered_reader{progress_reader};
 
-  waypoint_reader.Parse(way_points, buffered_reader);
+  switch (file_type) {
+  default:
+    std::unique_ptr<WaypointReaderBase> reader { CreateWaypointReader(file_type,
+                                                                      factory) };
+    if (!reader)
+      throw std::runtime_error{"Unrecognised waypoint file"};
+
+    reader->Parse(way_points, buffered_reader);
+    break;
+  }
 }
 
 void
@@ -63,14 +72,9 @@ ReadWaypointFile(Path path, WaypointFileType file_type,
                  Waypoints &way_points,
                  WaypointFactory factory, ProgressListener &progress)
 {
-  std::unique_ptr<WaypointReaderBase> reader(CreateWaypointReader(file_type,
-                                                                  factory));
-  if (!reader)
-    throw std::runtime_error{"Unrecognised waypoint file"};
-
   FileReader file_reader{path};
-  ReadWaypointFile(file_reader, file_reader.GetSize(),
-                   *reader, way_points, progress);
+  ReadWaypointFile(file_reader, file_type, file_reader.GetSize(),
+                   way_points, factory, progress);
 }
 
 void
@@ -86,12 +90,7 @@ ReadWaypointFile(struct zzip_dir *dir, const char *path,
                  WaypointFileType file_type, Waypoints &way_points,
                  WaypointFactory factory, ProgressListener &progress)
 {
-  std::unique_ptr<WaypointReaderBase> reader(CreateWaypointReader(file_type,
-                                                                  factory));
-  if (!reader)
-    throw std::runtime_error{"Unrecognised waypoint file"};
-
   ZipReader file_reader{dir, path};
-  ReadWaypointFile(file_reader, file_reader.GetSize(),
-                   *reader, way_points, progress);
+  ReadWaypointFile(file_reader, file_type, file_reader.GetSize(),
+                   way_points, factory, progress);
 }
--- a/src/Waypoint/WaypointReaderSeeYou.cpp
+++ b/src/Waypoint/WaypointReaderSeeYou.cpp
@@ -2,12 +2,13 @@
 // Copyright The XCSoar Project
 
 #include "WaypointReaderSeeYou.hpp"
-#include "CupParser.hpp"
 #include "Units/System.hpp"
 #include "Waypoint/Waypoints.hpp"
 #include "util/DecimalParser.hxx"
 #include "util/IterableSplitString.hxx"
 #include "util/NumberParser.hxx"
+#include "io/StringConverter.hpp"
+#include "io/BufferedCsvReader.hpp"
 
 #include <stdlib.h>
 
@@ -176,9 +177,11 @@ ParseStyle(std::string_view src, Waypoint::Type &type)
   return true;
 }
 
-bool
-WaypointReaderSeeYou::ParseLine(const char *line, Waypoints &waypoints)
-{
+bool ParseSeeYou(WaypointFactory factory, Waypoints &waypoints, BufferedReader &reader) {
+  StringConverter string_converter;
+
+  // 2018: name, code, country, lat, lon, elev, style, rwydir, rwylen, freq, desc
+  // 2022: name, code, country, lat, lon, elev, style, rwdir, rwlen, rwwidth, freq, desc, userdata, pics
   enum {
     iName = 0,
     iShortname = 1,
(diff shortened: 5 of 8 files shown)


MaxKellermann:
Your code is very clumsy, but before we go into details: where can I obtain this "CSV standard" and where can I see in the specification of each file format you changed that this file format is supposed to support this "CSV standard? Where can I find code that writes those files according to these standards? Where can I find example files that are approved by the standardization authors that XCSoar should be able to parse?

twhitehead:
The CSV standard is [RFC 4180](https://www.ietf.org/rfc/rfc4180.txt). Specifically Section 2, _Definition of the CSV Format_. Quoting double quotes, commas, and newlines are item 6

* comma: `"hi, there"`
* double quote: `"hi ""there"""`
* newline: `"hi\nthere"`

All of these are currently implemented in [_ExtractParameters.cpp_](../blob/master/src/util/ExtractParameters.cpp#L45). The only reason newlines don't currently work is we only feed the extractor a single line.

There is nothing in the [CUP format](https://downloads.naviter.com/docs/SeeYou_CUP_file_format.pdf) that specifically says that it is actually as CSV file. It is pretty clear that it is a CSV file though, including using double quotes to quote strings with commas.

What happens if you have a double quote or a newline though? The CUP format doesn't say. Both XCSoar and the LXxxxx software currently follow RFC 4180 (as above). This makes sense as this is the output a spreadsheet gives you when you edit your file in it and insert a double quote.

I would suggest following RFC 4180 for newlines also makes sense too. This completes compatibility with spreadsheets, and it does no harm, as a strict reading of the CUP format would mean files with quoted newlines (or quoted quotes for that matter) are simply invalid.

twhitehead:
With regard to input files, I have written [code](https://github.com/twhitehead/soaring-cfs) to automatically a CUP file for all the Canadian Aerodromes from the Canadian Flight Supplement (CFS). Here is an example CUP file from this process with just the entry for Arthur ([York Soaring](https://www.yorksoaring.com/)
```
name,code,country,lat,lon,elev,style,rwdir,rwlen,rwwidth,freq,desc,userdata,pics
Arthur (Walter’s Field),CPC3,CA,4351.000N,08027.000W,1550ft,4,140,2800ft,300ft,123.400,"Rwy 14/32 2800x300 turf
Rwy 12/30 2650x150 turf
Rwy 17/35 2550x170 turf
Rwy 05/23 1900x330 turf
Rwy 03/21 1750x250 turf",,CPC3.jpg
```
And here is the associated _CPC3.jpg_ image (right click and save it)
image

src/io/CSVLineReader.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include <vector>
+#include <memory>
+
+/**
+ * Adapter which handles CSV escaped newlines
+ */
+
+class CSVLineReader : public TLineReader {
+  std::unique_ptr<TLineReader> source;
+  std::vector<TCHAR> line;
```
twhitehead, line 18: `CSVLineReader::line` is most certainly used. See `CSVLineReader::ReadLine()` in _CSVLineReader.cpp_.

src/io/CSVLineReader.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+public:
+  CSVLineReader(std::unique_ptr<TLineReader> &&_source);
```
twhitehead, line 21: This follows the `ConverterLineReader` constructor.

lordfolken:
> With regard to input files, I have written [code](https://github.com/twhitehead/soaring-cfs) to automatically generate up-to-date CUP files for all the Canadian Aerodromes from the [Canadian Flight Supplement](https://www.navcanada.ca/en/ecfs_04_en.pdf) (CFS). Here is an example CUP file from this process with just the entry for Arthur ([York Soaring](https://www.yorksoaring.com/) home field)
> 
> ```
> name,code,country,lat,lon,elev,style,rwdir,rwlen,rwwidth,freq,desc,userdata,pics
> Arthur (Walter’s Field),CPC3,CA,4351.000N,08027.000W,1550ft,4,140,2800ft,300ft,123.400,"Rwy 14/32 2800x300 turf
> Rwy 12/30 2650x150 turf
> Rwy 17/35 2550x170 turf
> Rwy 05/23 1900x330 turf
> Rwy 03/21 1750x250 turf",CAUTION: extv glider activity. Simultaneous rgt & left hand circuits in use by gliders and towplanes Apr-Dec (CAR 602.96).,CPC3.jpg
> ```
> 
> using tho `RWY DATA` (runway data) section for `desc` and the `PRO` (procedure) section for the `useredata`.
> 
> And here is the associated _CPC3.jpg_ image (right click and save it) image

Make me a PR on xcsoar-data-content then we can include it in the xcsoar file manager.

twhitehead:
Had a look at the various waypoint reader test failures and the issue is this sort of entry
```
"Bergneustadt","",,5103.117N  ,00742.367E,  488.0m,  5  ,040,590m,"123.650" , "Rabbit holes, 20" ditch south end of rwy"
```
The issue being that the [_ExtractParamaters.cpp_](../blob/master/src/util/ExtractParameters.cpp) is actually more tolerant than [RFC 4180](https://www.ietf.org/rfc/rfc4180.txt). Specifically, it also allows a single quote not followed by a optional space and then a comma to pass through (e.g., `... holes, 20" ditch ...`) .

twhitehead:
As a small aside, one item I noticed while looking at this routine is that the code under the comment _Quote inside quoted string (but not end), copy to dst_ is actually _Quote inside ~quoted~ string (but not end), copy to dst_.

This mean that spreadsheet (RFC 4180) `""` → to `"` quote escaping for quoted strings is also being applied to unquoted strings. From the comments I don't think this was intended. Changing
```
        // Quote inside quoted string (but not end), copy to dst
        last_non_discardable_char = d;
        *d++ = *s;
        if (*(s+1) == quote_char) {
          // Spreadsheets use "" while exporting a single " contained in
          // a textfield to a CSV file - skip second "
          s++;
        }
```
to
```
        // Quote inside string (but not end), copy to dst
        last_non_discardable_char = d;
        *d++ = *s;
        if (in_quote && *(s+1) == quote_char) {
          // Spreadsheets use "" while exporting a single " contained in
          // a textfield to a CSV file - skip second "
          s++;
        }
```
would switch to only enabling quote escaping inside of quotes, which I expect was the intention.

twhitehead:
I believe the code should now handle quotes in a manner wholly compatible with the prior. That is, the only difference should now be that newlines inside of quotes do not end fields. In particular, it now accepts single quotes occurring inside of quotes like `"Rabbit holes, 20" ditch south end of rwy"` and so passes the _WaypointReader_ checks.

With regard to the code design, I am not an expert on the XCSoar code, but I did study it for a bit before writing what I did. My goals were (1) to not introduce any memory bugs, (2) to be minimal, readable, and not duplicated in many places, and (3) to not slow the common case. My decisions were as follows

* I made it an io LineReader converter in the style of `ConvertLineReader` to
  - allows the same code to be enabled on any of the waypoint reader types (currently only CUP, but can be done on any others by changing one line now), and
  - not reimplement a bunch of underlying readers, but rather work whether it is a `ZipLineReader` (`CUP` file in a `XCM` zip map file or, maybe someday, a `CUPX` file), `LineReader` (standard `CUP` file), or anything else.
* I observed that their is increasing use of modern C++ constructs and library usage throughout the XCSoar code, so I used `std::vector` instead of `DynamicFifoBuffer` to implement the buffering. This was easier and less code (at least for me) and seemed more inline with the transition to more std library instead of custom code.
* I made the `CSVLineReader` code a bit more duplicate and verbose so that the standard case of no newline occurring inside a quoted section is just a pass through without any additional allocating and copying of the input.

As I said, I am not an expert of the XCSoar code base, and I really don't care how this is done. I would just like it done so I can get way more detailed and up-to-date landables information into my glide computer. :smile:

MaxKellermann:
> There is nothing in the [CUP format](https://downloads.naviter.com/docs/SeeYou_CUP_file_format.pdf) that specifically says that it is actually as CSV file. It is pretty clear that it is a CSV file though, including using double quotes to quote strings with commas.

The CUP specification says: "Each line represents one waypoint".
That is very specific: CUP files **cannot** contain multi-line values, which clearly invalidates your assumption that CUP files are RFC-compliant CSV files.

> What happens if you have a double quote or a newline though?

Impossible with the CUP format. The spec says: "Fields containing a comma must be embraced in double quotes (")."
It mentions no under circumstances under which quotes should be used and *allows* quotes for other reasons only in "frequency". What XCSoar currently does (allow escaping via backslash) thus violates the CUP format specification.

About your actual code: I think your implementation is bad because adds unnecessary overhead, contains an unbounded memory allocation based on input files, is overly complicated (the whole `transition` table, a 4x4 matrix for something as simple as CSV parsing, wtf?) and CSVLineReader duplicates code that already lives somewhere else, which is a fragile thing to do, because both copies need to be exactly in sync, but who's going to take care of that?
Even if I was convinced that the CUP format is capable of having multi-line records (it clearly isn't), I wouldn't merge this until all those warts are eliminated from your code.

twhitehead:
> The CUP specification says: "Each line represents one waypoint". That is very specific: CUP files **cannot** contain multi-line values, which clearly invalidates your assumption that CUP files are RFC-compliant CSV files.
> ...
> Impossible with the CUP format. The spec says: "Fields containing a comma must be embraced in double quotes (")." It mentions no other circumstances under which quotes should be used and _allows_ quotes for other reasons only in "frequency". What XCSoar currently does (allow escaping via backslash) thus violates the CUP format specification.

I believe, from reading the [field reader logic](../blob/master/src/util/ExtractParameters.cpp#L44) and testing, your current extension for a quote inside of quotes is not a backslash but rather either a single quote not followed by optional spaces and a comma (i.e., accepted so long as it is obviously not an end-of-field) or an RFC compliant double quote. 

With regard to the spec saying that each line represents one waypoint, I had missed that. As you said, a quoted multi-line entry is then obviously not a compliant CUP file. Although, as you also point out, neither is the current escaping of quotes inside of quotes. I guess then, the question is what forms of non-compliant CUP files do you want XCSoar to be able to load (given, of course, that support for non-compliant files should never break compliant ones).

> About your actual code: I think your implementation is bad because adds unnecessary overhead, contains an unbounded memory allocation based on input files, is overly complicated (the whole `transition` table, a 4x4 matrix for something as simple as CSV parsing, wtf?) and CSVLineReader duplicates code that already lives somewhere else, which is a fragile thing to do, because both copies need to be exactly in sync, but who's going to take care of that? Even if I was convinced that the CUP format is capable of having multi-line records (it clearly isn't), I wouldn't merge this until all those warts are eliminated from your code.

Thank you for your feedback. I had written some more explanations as to why I used a state transition table and such, but, upon reflection, I think the better thing is to ask

1. given XCSoar already support non-standard CUP extensions, can you be convinced that also supporting multi-line entities would be a good thing, and
2. if you can be convinced, can you lay out for me, as someone who isn't an XCSoar contributor, how I should implement this so it would be acceptable to you (e.g., would it be acceptable if I did it by making `ReadLine` in `BufferedLineReader` an abstract function and then implement a new derived class to support CSV multiline reading?).

With regard to the first of these, my reason for wanting newline support is so I can have the following sort of level of detail available in XCSoar for airfieds. I want to have the contact, the full communication details, the complete set of runways and their surface information, the procedure, and the caution information all at my finger tips. I hope this strikes you as valuable too.

image

kobedegeest:
Just fyi, if you just want info of airfields maybe the waypoint-details file is the way to go with that you can link other files like pdf's or any other file your device can open, to your waypoints. Sure it does not show in the app itself but since xcsoar again supports acces to gps data when running in background i see no reason why this would be a problem. 

Or addapt that code so csv/txt files linked would open in the app just like figures do now.

twhitehead:
Thanks. I believe I understand how you would like this done, and it seems very reasonable. :+1: 

If I run into further questions, I will ask them here.

twhitehead:
> Just fyi, if you just want info of airfields maybe the waypoint-details file is the way to go with that you can link other files like pdf's or any other file your device can open, to your waypoints. Sure it does not show in the app itself but since xcsoar again supports acces to gps data when running in background i see no reason why this would be a problem.

I didn't know or had forgotten about the waypoint-details file. It does seem like it would give a similar result. I am still going to implement the multi-line support though as it is nice to distribute this in a single file. Maybe even with pictures if we can do [CUPX](https://downloads.naviter.com/docs/SeeYou_CUPX_file_format.pdf) at some point (whoever said concatenating two zip files instead of just putting everything in one was crazy, I totally agree!).

MaxKellermann:
> The `BufferedReader` however reads just `char` lines, not `TCHAR` lines; but as long as numbers are being parsed, we can just keep it `char` and only convert human-readable strings (like waypoint names) to `TCHAR` strings. This will avoid a good amount of charset conversion overhead.

Migration to `char` finished: 41eb5393b1273f2b9776bdd7d8c7a132b43a660e

twhitehead:
Okay. I believe I have coded up something more like what you are looking for. One item left to resolve is the loading of the SeeYou task files. It can not use the new buffered CSV reader as is because it distinguishes the start of a task definition by there being a quoted task name field, and the new buffered CSV reader does expose whether a field it extracted was quoted or not.

Technically the [CUP standard](https://downloads.naviter.com/docs/SeeYou_CUP_file_format.pdf) does not say anything about the task name having to be quoted, although both examples it gives are. So, either I can

* extend the new buffered CSV reader to also tell whether the fields extracted were quoted or not (which is pretty easy), or
* come up with some new logic to separate the start of a task from subsequent options, observation zones, and alternative starts lines (each of these start with a fixed string, so this is also not difficult).

MaxKellermann:
Your API is too complicated. Your CSV parser has a dependency on `BufferedReader`, and that is not necessary. Just pass a `std::string_view` to it that shall be parsed, and the function need not care where this string comes from. Once you have that, the API not only becomes simpler, but you can also inspect the string and check whether there are quotes.

MaxKellermann:
> Your API is too complicated

Ah, you know what's funny? It's my API. I suggested it. But it wasn't good. Sorry for blaming you for my own design ;-)

twhitehead:
Thanks Max. One of the thing about the `BufferedReader` is that I could do a call out if I needed more data and at the end I could advance it over the data that I used. Was wondering how to handle these items with a `std::span<char>`?

For example, I could return an optional<std::span> that indicates (1) that their was sufficient data (i.e., if it returns empty, then you need to pass in more data), and (2) the portion of the input that wasn't consumed?

MaxKellermann:
If you need more data, call `Fill(true)` and then the next `Read()` call will give you a larger span. Or `Fill(true)` throws an exception if no more data is available in the underlying stream.

twhitehead:
Thanks for getting back to me on this. I am currently [calling `Fill(true)` followed by another `Read()`](https://github.com/twhitehead/XCSoar/blob/7788718/src/io/BufferedCsvReader.cpp#L31) if I need more data
```
size_t
ReadCsvRecord(BufferedReader &reader, std::span<std::string_view> views) {
  ...
  do {
    input = reader.Read();
    ... incremental parse that aborts loop upon completion ...
  } while ( reader.Fill(true) );
  ...
}
```
To do this though, I need the `BufferedReader`. If I switch to a `std::string_view`/`std::span<char>`, as you requested, I won't be able to call these functions.

In this case I will need to move the above outer loop to the call sites. The inner incremental parser would then need to tell the calling site with its return type that it has completed, and here are the results, or that it could use more data. If it says it needed more data, then the call sites would then need to do the `Fill(true)` followed by `Read()` and re-invoke the inner parser to give it more data or tell it there is no more data. The inner parser would have to restart the parse from scratch with the original unmodified input or retain its state and input modifications so it could resume where it left off.

I can do this if you want, but I wanting to clarify with you first that this is what you really want. It is additional complexity and it duplicates the feeder code at all the call sites. So maybe not worth it unless you want to use it on thing other than `BufferedReader`? I also wanted to know your thoughts on what the interface between the split off outer loop and the inner parser should look like, and what I should do with respect to resuming given that we are modifying the input for in-place efficiency.

Without direction, I would likely
* return `-1` to ask for more input
* call again with exact same input to indicate no more input
* make the parser resuming by changing it to a lambda with captured state (get the compiler to make it a class with state variables and an `operator()(...)` to invoke/re-invoke it)
  ```
  auto CreateCsvParser() {
    ... initialized state ...
    return [](std::span<char> &input_buffer,
              bool input_final,
              std::span<std::string_view> views) -> size_t {
      ... parser code (implicitly captures state) ...
    };
  }
  ```

MaxKellermann:
(btw. Fill() does not throw if there is no more data; ReadFull() does but that's a different method.)

> So maybe not worth it unless you want to use it on thing other than `BufferedReader`?

Maybe the current API which does the refill loop inside the CSV library is currently the easiest way forward. We can later move the loop out if we feel we need to make it more generic.

One thing that is unacceptable currently is the use of `goto`. Can we please not revert to such a thing?

If you remove it, I'll have a closer look again and see what remains before we can merge this. The use of BufferedReader inside the parser can be merged.

twhitehead:
The `goto` is now gone. :+1:

twhitehead:
Noticed I had an unused variable I had forgot to remove. That's gone now too.

MaxKellermann:
```
src/io/BufferedCsvReader.cpp:72:79: error: arithmetic on a null pointer treated as a cast from integer to pointer is a GNU extension [-Werror,-Wgnu-null-pointer-arithmetic]
   72 |           views[fields_parsed] = std::string_view(static_cast<char*>(nullptr) + field_start.value(),
      |                                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
```

MaxKellermann:
Now the linker fails:
```
  LINK    output/UNIX/bin/RunMapWindow
/usr/bin/ld: output/UNIX/dbg/libwaypointfile.a(WaypointReaderSeeYou.o): in function `ParseSeeYou(WaypointFactory, Waypoints&, BufferedReader&)':
src/Waypoint/WaypointReaderSeeYou.cpp:208: undefined reference to `ReadCsvRecord(BufferedReader&, std::span<std::basic_string_view<char, std::char_traits<char> >, 18446744073709551615ul>)'
/usr/bin/ld: src/Waypoint/WaypointReaderSeeYou.cpp:224: undefined reference to `ReadCsvRecord(BufferedReader&, std::span<std::basic_string_view<char, std::char_traits<char> >, 18446744073709551615ul>)'
make: *** [build/test.mk:1743: output/UNIX/bin/RunMapWindow] Error 1
```

twhitehead:
Thanks for the heads up. :+1: I wasn't building the `everything` target on my machine, so I had missed that linking issue. Hadn't been aware you shouldn't technically use `nullptr` as a base pointer for arithmetic either to store offsets.

PR #1306, opened 2023-10-10
External logger flight download: Reverse flight list

groundhogxc:
This PR reverses the list of flights available in an external logger such as Flarm, Nano, or LX Nav and LX Navigation devices. These typically show up with the oldest flights first on the list, as opposed to the typical way they are displayed, e.g. in the device screen.
By reversing the list scrolling and searching for the latest flight is usually not necessary any more.

Further discussion in #1296 - In that older PR the current implementation (in the UI) was present but not tested on an actual device in a glider. In this version, the commit has been fully tested on the current version of the Master branch.

### DIFF
--- a/src/Logger/ExternalLogger.cpp
+++ b/src/Logger/ExternalLogger.cpp
@@ -274,6 +274,9 @@ ExternalLogger::DownloadFlightFrom(DeviceDescriptor &device)
     return;
   }
 
+  /* Reverse flight list so that most recent flights are shown first */
+  std::reverse(flight_list.begin(),flight_list.end());
+
   const auto logs_path = MakeLocalPath(_T("logs"));
 
   while (true) {


DanD222:
Does this just reverse, without checking the fike creation date? I’m not sure if the Flarm flights weren’t ordered correctly as opposed to LXNav, so wondering if a blanket reversal wouldn’t cause issues elsewhere.

MaxKellermann:
Reversing isn't good. But I'm repeating myself. Read https://github.com/XCSoar/XCSoar/pull/1296#discussion_r1318559570

groundhogxc:
> Does this just reverse, without checking the fike creation date? I’m not sure if the Flarm flights weren’t ordered correctly as opposed to LXNav, so wondering if a blanket reversal wouldn’t cause issues elsewhere.

Yes, it is doing that, and one reason is precisely the issue with the fake creation date, which I have seen myself. In the case of a fake creation date this means that your most recent flight (the one you just landed from) will typically in fact be displayed in the top *even if* the creation date is wrong. Otherwise you would not be able to figure out which flight is the one you just did between the different flights with 00-00-0000 date. (Which is not impossible considering that e.g. the LXNav devices can save apparently a high number of flights in the hundred(s).).

DanD222:
Did you manage to check this with both Flarm and LXNav devices? Because in case one is ordered ‘latest first’ and the other is ordered ‘earliest first’ then this just moves the problem from one device to the other.

MaxKellermann:
What I won't do is argue about loggers that lie about the date and send fake data to XCSoar. Shit in, shit out.
Also don't bullshit me with stuff like "you gave no logical reason for this claim". This is so much bullshit, I don't know where to start. I could post a rebuttal and prove my point, but that would just waste my time, and would make your post a DoS attack on my free time. Don't play chess with a pigeon, you know.
I'm annoyed enough by your attitude that I'm closing this PR now. You may disagree, but then you can fork XCSoar or find a new maintainer for XCSoar.

groundhogxc:
> Also don't bullshit me with stuff like "you gave no logical reason for this claim". This is so much bullshit, I don't know where to start.

Well, I can not change that you closed the PR, but just for completeness - I am referring to this (in #1296 and that was referencing #1237):
> The order of flights from the logger (or from the driver) is undefined. If stuff works now, it is by coincidence, but you must not break that. You can replace it with a well-defined order, but not randomly with something else.

The reversal operation is mathematically speaking a reversible (undo-able) operation, and as an operation it is well-defined, without any additional parameters. Therefore, under normal mathematical proceedings, the order following the reversal is *just as well-defined* as it was before the reversal.
Hence, mathematically speaking, your claim that it is not a "well-defined order" or even "random" is simply mathematically incorrect, hence not logical. I cannot change that either. 

(But the main criterion should anyway not be maths, but whether a change improves the program for the user or not - this is what in my opinion the PR does.)

MaxKellermann:
> The reversal operation is mathematically speaking a reversible (undo-able) operation, and as an operation it is well-defined, without any additional parameters. Therefore, under normal mathematical proceedings, the order following the reversal is _just as well-defined_ as it was before the reversal. Hence, mathematically speaking, your claim that it is not a "well-defined order" or even "random" is simply mathematically incorrect, hence not logical. I cannot change that either.

That's is not math, it's just bullshit. I don't want to waste any more time with you, so I'm blocking you now. Bye.

PR #1310, opened 2023-11-04
doc / build.rst - add cmake and ninja to macOS Homebrew packages

DanD222:
From https://forum.xcsoar.org/viewtopic.php?f=7&t=4213

### DIFF
--- a/doc/build.rst
+++ b/doc/build.rst
@@ -289,7 +289,9 @@ Install the required Homebrew packages::
     libtiff libgeotiff proj \
     c-ares \
     curl \
-    lua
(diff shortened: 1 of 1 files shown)


MaxKellermann:
Your commit message doesn't explain why cmake is needed, it only contains a link to a lengthy forum discussion, but I don't have the time to read it all. Even if there is an explanation why cmake and ninja are needed, that explanation should be in the commit message.
I doubt cmake is really necessary, therefore I assume this PR is a mistake and I close it, until there is a proper explanation proving me wrong.

PR #1311, opened 2023-11-07
Topography/shapelib/mapserver-config: functions strlcpy and strlcat  have been added to glibc version 2.38

ubx:
The functions strlcpy and strlcat have been added to glibc 2.38. This pull request reflects these changes. 
XCSoar can now be built on Linux box with gcc 13.2.0


MaxKellermann:
The new `#if` expression is obviously wrong and makes no sense.

ubx:
The sense of the `#if` changes is, as stated in the commit message, to define HAVE_STRLCAT and HAVE_STRLCPY for all platforms, except for glibc versions earlier than 2.38 and Win32.

PR #1313, opened 2023-11-14
Kobo: fix Wifi connection status not visible

### LINKED ISSUE #1309: WIFI connection status no more visible on KOBO version 7.40

OK 7.38
NOK 7.40

System information
------------------

KOBO GLO

see the following 2 screenshots from 7.38 and 7.40

- Connect KOBO to local network
- Go to "Wifi" with the list of network devices

Expected behavior
-----------------

Connected network should be shown with status "connected" and network address

Actual behavior
---------------

Connection is established with "Wifi on" but the status is only "Saved and visible"

This is confusing compared to the previous version.

Connection is also shown with WPA 213 (7.40) and WPA 210 (7.38)

Bug or feature?

no

no

ubx:
This PR fixes the issue where the WiFi connection status is not visible on KOBO.

Closes #1309

### DIFF
--- a/src/Kobo/WPASupplicant.cpp
+++ b/src/Kobo/WPASupplicant.cpp
@@ -73,9 +73,9 @@ ParseStatusLine(WifiStatus &status, std::string_view src) noexcept
   if (value.data() == nullptr)
     return false;
 
-  if (src == "bssid"sv)
+  if (name == "bssid"sv)
     status.bssid = value;
-  else if (src == "ssid"sv)
+  else if (name == "ssid"sv)
     status.ssid = value;
   return true;
 }
--- a/src/Kobo/WPASupplicant.cpp
+++ b/src/Kobo/WPASupplicant.cpp
@@ -136,8 +136,9 @@ ParseScanResultsLine(WifiVisibleNetwork &dest, std::string_view line) noexcept
 
   dest.bssid = bssid;
 
-  if (const auto value = ParseInteger<unsigned>(signal_level))
-    dest.signal_level = *value;
+  if (const auto value = ParseInteger<signed>(signal_level))
+    // TODO: The signal level is in decibels (dBm), we should treat it that way
+    dest.signal_level = std::abs(*value);
   else
     return false;
 


MaxKellermann:
This PR mixes two unrelated changes in one commit, and there is no explanation why those two changes are necessary and how they fix the bug.

ubx:
Split PR into two commits and explain the changes.

MaxKellermann:
I still don't get this:
> WPASupplicant: Fix the bug introduced in 442517f where the signed signal_level was parsed as an unsigned integer, which was never successful

Old code:
```
dest.signal_level = ParseUnsigned(src, &endptr);
```
This was never signed!

MaxKellermann:
I cherry-picked the first commit. I believe the second commit is wrong. If you disagree and believe the second commit is necessary, please explain. Until that happens, this PR is closed (and Wifi scanning does work indeed).

ubx:
The second commit is also necessary (Wifi scanning doesn't work on my Kobo Libra 2). The commit message was partly incorrect, and I have made the changes.

The issue was that before commit 442517f1c2f4737bb26da4703ae5df14f9d250d0, the sign in the signal_level string was skipped: `++src` and parsed as an unsigned.

`git diff 4e8f23bd 442517f1`

```
diff --git a/src/Kobo/WPASupplicant.cpp b/src/Kobo/WPASupplicant.cpp
index 50d3657921..1aae3d3bc5 100644
--- a/src/Kobo/WPASupplicant.cpp
+++ b/src/Kobo/WPASupplicant.cpp
@@ -7,6 +7,7 @@
 #include "net/AllocatedSocketAddress.hxx"
 #include "util/IterableSplitString.hxx"
 #include "util/NumberParser.hpp"
+#include "util/NumberParser.hxx"
 #include "util/StringCompare.hxx"
 #include "util/StringSplit.hxx"
 
@@ -126,90 +127,61 @@ WPASupplicant::Status(WifiStatus &status)
  * - ssid ascii ssid. ssid could be empty if ssid broadcast is disabled.
  */
 static bool
-ParseScanResultsLine(WifiVisibleNetwork &dest, char *src)
+ParseScanResultsLine(WifiVisibleNetwork &dest, std::string_view line) noexcept
 {
-  char *tab = strchr(src, '\t'); // seek "frequency"
-  if (tab == nullptr)
-    return false;
-
-  *tab = 0;
-  dest.bssid = src;
-
-  src = tab + 1;
+  const auto [bssid, rest1] = Split(line, '\t');
+  const auto [frequency, rest2] = Split(rest1, '\t');
+  const auto [signal_level, rest3] = Split(rest2, '\t');
+  const auto [flags, rest4] = Split(rest3, '\t');
+  const auto [ssid, _] = Split(rest4, '\t');
 
-  src = strchr(src + 1, '\t'); // seek "signal level"
-  if (src == nullptr)
-    return false;
-
-  ++src;
-
-  char *endptr;
-  dest.signal_level = ParseUnsigned(src, &endptr);
-  if (endptr == src || *endptr != '\t')
+  if (bssid.empty() || frequency.empty() || signal_level.empty() || ssid.empty())
     return false;
 
-  src = endptr + 1;
+  dest.bssid = bssid;
 
-  tab = strchr(src, '\t'); // seek "ssid"
-  if (tab == nullptr)
+  if (const auto value = ParseInteger<unsigned>(signal_level))
+    dest.signal_level = *value;
+  else
     return false;
 
-  *tab = 0;
-
-  // src points to the flags.
-  if (strstr(src, "WPA") != NULL)
+  if (flags.find("WPA"sv) != flags.npos)
     dest.security = WPA_SECURITY;
-  else if (strstr(src, "WEP") != NULL)
+  else if (flags.find("WEP"sv) != flags.npos)
     dest.security = WEP_SECURITY;
   else
     dest.security = OPEN_SECURITY;
 
-  src = tab + 1;
-
-  tab = strchr(src, '\t');
-  if (tab != nullptr)
-    *tab = 0;
-
-  // src points to ssid or if empty we assume a hidden ssid.
-  if (StringIsEmpty(src)) {
-    dest.ssid.clear();
-    return true;
-  }
-
-  dest.ssid = src;
+  dest.ssid = ssid;
   return true;
 }
 
 static std::size_t
-ParseScanResults(WifiVisibleNetwork *dest, std::size_t max, char *src)
+ParseScanResults(WifiVisibleNetwork *dest, std::size_t max, std::string_view src)
 {
-  if (memcmp(src, "bssid", 5) != 0)
+  if (!src.starts_with("bssid"sv))
     throw std::runtime_error{"Malformed wpa_supplicant response"};
 
-  src = strchr(src, '\n');
-  if (src == nullptr)
+  src = Split(src, '\n').second;
+  if (src.data() == nullptr)
     throw std::runtime_error{"Malformed wpa_supplicant response"};
 
-  ++src;
-
   std::size_t n = 0;
-  do {
-    char *eol = strchr(src, '\n');
-    if (eol != nullptr)
-      *eol = 0;
+  for (const auto line : IterableSplitString(src, '\n')) {
+    if (line.empty())
+      break;
 
-    if (!ParseScanResultsLine(dest[n], src))
+    if (!ParseScanResultsLine(dest[n], line))
       break;
 
     // skip hidden ssid
-    if (!dest[n].ssid.empty())
+    if (!dest[n].ssid.empty()) {
       ++n;
 
-    if (eol == nullptr)
-      break;
-
-    src = eol + 1;
-  } while (n < max);
+      if (n >= max)
+        break;
+    }
+  }
 
   return n;
 }
@@ -223,13 +195,11 @@ WPASupplicant::ScanResults(WifiVisibleNetwork *dest, unsigned max)
   SendCommand("SCAN_RESULTS");
 
   char buffer[4096];
-  ssize_t nbytes = ReadTimeout(buffer, sizeof(buffer) - 1);
-  if (nbytes <= 5)
-    throw std::runtime_error{"Unexpected wpa_supplicant response"};
-
-  buffer[nbytes] = 0;
+  const std::size_t nbytes = ReadTimeout(buffer, sizeof(buffer));
+  if (nbytes == 0)
+    throw std::runtime_error{"wpa_supplicant closed the socket"};
 
-  return ParseScanResults(dest, max, buffer);
+  return ParseScanResults(dest, max, {buffer, nbytes});
 }
 
 unsigned
```

ubx:
aa

MaxKellermann:
> the sign in the signal_level string was skipped: `++src` and parsed as an unsigned

That's wrong. The `++src` skips the tab, not the sign.

MaxKellermann:
> Can't reopen, why ?

Me neither, probably because you changed the PR branch. But do you really want to reopen? So far, you have not brought up anything that could demonstrate the correctness of the remaining commit.

ubx:
> That's wrong. The ++src skips the tab, not the sign.

I didn't catch.

The only finding I have after remote debugging on a Kobo Libra 2 for a `signal_level` of '-39' is:

`If (const auto value = ParseInteger<unsigned>(signal_level))`, it gives `false`.
`If (const auto value = ParseInteger<signed>(signal_level))`, it gives `true`.

ulipo:
I compiled v7.41 and flashed on my Kobo GloHD and the wifi status is now working as expected, so probably the cherry-picked commit is enough to fix the issue

ubx:
> I compiled v7.41 and flashed on my Kobo GloHD and the wifi status is now working as expected, so probably the cherry-picked commit is enough to fix the issue

But for my Kobo Libra 2 it does not.

MaxKellermann:
> The only finding I have after remote debugging on a Kobo Libra 2 for a `signal_level` of '-39' is:

Now that's the first time you post actual data. So negative signal levels are possible? I've never seen that, that's why I used ParseUnsigned() to implement this. So far, you have not claimed that it is possible.

That means the code was buggy from the beginning. And even the variable declaration is wrong:
https://github.com/XCSoar/XCSoar/blob/f0437208763b629de30d459f8f9ff0fb584477b5/src/Kobo/WPASupplicant.hpp#L32
... and your second commit is still wrong, because it keeps the `unsigned`. It would be a bug to assign a negative value to an `unsigned` variable.

In any case, it was the right decision not to merge your commit. It may make one user-visible symptom disappear, but it's still buggy. Let's fix all the bugs instead!

ubx:
> In any case, it was the right decision not to merge your commit. 

I disagree. The commit belongs to the PR because it fixes #1309. Without it, the status connection is not visible, at least on the Kobo Libra 2. I'm aware of the problem with negative signal levels and have even included a notice: `// TODO: The signal level is in decibels (dBm), and we should treat it that way`. Therefore, I didn't want to mix unrelated changes in that PR.

> Let's fix all the bugs instead!

Certainly, I'm willing to address the issue with negative signal levels, but let's handle it in a new separate issue.

MaxKellermann:
> Certainly, I'm willing to address the issue with negative signal levels, but let's handle it in a new separate issue.

But negative signal levels (and the lack of support for them) ARE the core issue here.

You are thinking in symptoms, I am thinking in problems. That's why we perceive the solutions and their approaches differently. But if we fix the problem, all the symptoms go away - even those that you are not paying attention to currently.

PR #1320, opened 2023-12-26
Kobo/WPASupplicant: for new models, parse and interpret the signal level as decibels (dBm).

ubx:
For the Wifi setup page, this PR implements parsing and interpreting the signal level as decibels (dBm) for new Kobo models.

Tested on a Kobo Mini and Kobo Libra 2.


ulipo:
Tested on Kobo GloHD, works as expected

Issue #1319, opened 2023-12-26
LXNav S10 incorrect ballast synchronization


tatro77:
XCSoar versions having and not having the problem
-------------------------------------------------

Current version 7.38 

System information
------------------

 Stefly Nav 7.0, Android 11.

Steps to reproduce the behavior
-------------------------------

 It is not possible to match the water ballast setting of the LXNav S10 and XCsoar. 

Expected behavior
-----------------

Ideally the settings for water ballast should match and drain at the same rate.

Actual behavior
---------------

Currently  when 100L of water ballast are  set in  XCsoar (flight setup-custom), the LXNAV S10 shows 70L

Do you have any idea what may have caused this?
-----------------------------------------------

no

Do you have an idea how to solve the issue?
-------------------------------------------

no

tatro77:
![IMG_9465](https://github.com/XCSoar/XCSoar/assets/107053423/d3041fcd-c737-4f94-9a91-92123f1f14d3)

lordfolken:
Please switch on debug on the serial port and attach the resulting log.

tatro77:
Hi. 
Here is the log. :)

tatro77:
[xcsoar.log](https://github.com/XCSoar/XCSoar/files/13778855/xcsoar.log)

lordfolken:
The ballast is communicated as an overload factor. 
$PLXV0, BAL,R/W,Overload factor*CS, where overload factor is defined as ratio between current mass and mass of glider, where polar was calculated

PLXV0,BA L,W,1.50

Please show us the configured plane profile.

tatro77:
![IMG_9496](https://github.com/XCSoar/XCSoar/assets/107053423/b069c21a-39ac-46e3-b5fa-701f9bc95672)
![IMG_9497](https://github.com/XCSoar/XCSoar/assets/107053423/79d86db6-3494-46d1-a70a-5658b892e120)
![IMG_9498](https://github.com/XCSoar/XCSoar/assets/107053423/4edce8d5-94dc-4bbb-a740-b64ba6c2daa0)

lordfolken:
So according to src/ActionInterface.ccp overload is defined as:
https://github.com/XCSoar/XCSoar/blob/f0437208763b629de30d459f8f9ff0fb584477b5/src/ActionInterface.cpp#L105-L107

However the LX Manual gives a different formula:
[...]

PR #1326, opened 2024-01-08
Overload2

lordfolken:


lordfolken:
This needs some testing... 
@iltis42 @Blaubart could you check with your devices please, wheter the ballast is still syncing correclty?   

- [ ] Vaulter
- [ ] CAI302
- [x] LXNAV S80
- [] LX1600
- [] XCVario
- [] OpenVario

Blaubart:
Yes, I am welcome to test on the OpenVario. Can you briefly describe what exactly you would like to have tested?

iltis42:
XCVario i can do in the second half of this week.

MaxKellermann:
I don't think we should change XCSoar's definition of overload factor. I rather think we should remove XCSoar's definition completely, because it's a useless metric. See https://github.com/XCSoar/XCSoar/issues/1319#issuecomment-1878567839

PR #1334, opened 2024-01-27
Icons: mountain pass with usual icon, and bridge icon looking like a bridge

mlep:
Currently, in XCsoar:
- the bridge icon is map_bridge-ini.
This icon is usually used on maps for mountain passes. For example :
    -   Pass-OpenStreetMap  (source: https://www.openstreetmap.org/#map=19/46.57269/8.41550&layers=C)
    -  Mountain_pass_12x12_ne (source: https://en.wikipedia.org/wiki/File:Mountain_pass_12x12_ne.svg)
    So this is confusing.
- the pass icon is map_pass, which, on small screens, is difficult to differentiate from the mountain top icon (map_mountain_top)

Hence, this PR proposes that:
- the new pass icon gets the shape of a pass icon (map_bridge-ini)
- a new bridge icon is used: map_bridge  (this drawing is my own creation).

PR #1335, opened 2024-01-29
[by @mlep] Icons: mountain pass with usual icon, and bridge icon looking like a bridge

llauner:
Re-creating a PR for @mlep  until he gains furhter knowledge on Git. This replaces #1334 
Currently, in XCsoar:

the bridge icon is image.
This icon is usually used on maps for mountain passes. For example :
image (source: https://www.openstreetmap.org/#map=19/46.57269/8.41550&layers=C)
Mountain_pass_12x12_ne (source: https://en.wikipedia.org/wiki/File:Mountain_pass_12x12_ne.svg)
So this is confusing.
the pass icon is image, which, on small screens, is difficult to differentiate from the mountain top icon image
Hence, this PR proposes that:

the new pass icon gets the shape of a pass icon (image)
a new bridge icon is used: image (this drawing is my own creation = @mlep ).


lordfolken (approved):

llauner:
Is there anything I need to provide / do to get this PR merged ?
Thanks for your help ...

Issue #1345, opened 2024-02-10
WIN64: TestAirspaceParser.exe fails


ubx:
XCSoar versions having and not having the problem
-------------------------------------------------
WIN64, latest from master (9149c338f94a03f5e596265759e3cf34b07e5e4b)

System information
------------------
[see here]( https://github.com/ubx/XCSoar/actions/runs/7854201754/job/21434955765)

Steps to reproduce the behavior
-------------------------------
On a Linux box: 

`make TARGET=WIN64 DEBUG=y USE_CCACHE=y V=2 everything`

`wine output/WIN64/bin/TestAirspaceParser.exe`

Expected behavior
-----------------
All tests should PASS

Actual behavior
---------------
```
not ok 37 - airspace.GetRadioFrequency() == RadioFrequency::FromMegaKiloHertz(130, 125)
#     Failed test (test/src/TestAirspaceParser.cpp:TestOpenAir() at line 116)
```
and
```
not ok 79 - airspace.GetRadioFrequency() == RadioFrequency::FromMegaKiloHertz(130, 125)
#     Failed test (test/src/TestAirspaceParser.cpp:TestTNP() at line 224)

```

[and here]( https://github.com/ubx/XCSoar/actions/runs/7854201754/job/21434955765)

Do you have any idea what may have caused this?
-----------------------------------------------

Parsing radio frequency on Windows os.

Do you have an idea how to solve the issue?
-------------------------------------------

lordfolken:
Jup, saw that when fixing the toolchain. Care to bisect?
There is also #1336 and #1332

ubx:
```
78766892c2eb3e52b07b4225b18d423949c309db is the first bad commit
commit 78766892c2eb3e52b07b4225b18d423949c309db
Author: Max Kellermann <max.kellermann@gmail.com>
Date:   Sun Oct 8 16:11:45 2023 +0200

    Airspace/Parser: migrate from TLineReader to BufferedReader

 build/test.mk                         |   8 +-
 src/Airspace/AirspaceGlue.cpp         |  18 +-
 src/Airspace/AirspaceParser.cpp       | 312 +++++++++++++++++-----------------
 src/Airspace/AirspaceParser.hpp       |   6 +-
 test/src/RunAirspaceParser.cpp        |   6 +-
 test/src/RunAirspaceWarningDialog.cpp |  12 +-
 test/src/RunMapWindow.cpp             |   9 +-
[...]

PR #1348, opened 2024-02-13
RadioFrequency: ignore unit ("MHz") in Parse(std::string_view)

ubx:
This PR fixes the reading of radio frequency with unit "MHz" in the openAir file on Windows os.

Fix for #1345

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -3,6 +3,8 @@ Version 7.42 - not yet released
   - fix bogus "Latin-1 to UTF-8 conversion failed" error
 * Kobo
   - Wifi setup, display signal level in dBm for new models
+* Windows
+  - fix reading of radio frequency with MHz in the openAir file
 
 Version 7.41 - 2023/12/21
 * data files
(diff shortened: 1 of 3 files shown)


lordfolken (commented):
I'm OK with the approach.
The thing I'd like to understand is, why the tchar version of the function was in use on *nix platforms. 
Also do we need to handle the unit variable as its basically unused.

ubx:
> Also do we need to handle the unit variable as its basically unused.

do you mean:
`  const auto [val, _] = Split(src, ' ');`
?

lordfolken:
> > Also do we need to handle the unit variable as its basically unused.
> 
> do you mean: ` const auto [val, _] = Split(src, ' ');` ?
Elegant. 

> > Also do we need to handle the unit variable as its basically unused.
> 
> do you mean: ` const auto [val, _] = Split(src, ' ');` ?
Elegant.

MaxKellermann:
Also, the commit message doesn't explain why this should be ignored. I'll revert the commit.

lordfolken:
The unit isn't specified in the standard. It breaks when there is something else on the line.

MaxKellermann:
What standard? What how is that standard relevant for XCSoar's `RadioFrequency` class? What what is "something else" and who put it there?

ubx:
Ok, should be handled in AirspaceParser.

lordfolken:
@ubx will you provide another implementation?

ubx:
Yes.

PR #1350, opened 2024-02-17
test/TestTaskSave: add a test for saving and loading a task

### LINKED ISSUE #1346: Windows: Task file produced by XCSoar contains invalid values in XML attributes

Bad: 7.41 (win64)
Bad: 7.40 (win64)
Good: 7.38(win64)

System information
------------------

 Desktop PC, Windows 11

Screenshot Fehler
Screenshot Editor
Screenshot XCSoar
Screenshot WP_Datei

 I created a task and stored it. 
      Loading was not possible. Failure message was: see screenshot

Expected behavior
-----------------

 
File should contain the name, comment and correct type as xml attributes instead of simply being replaced by "1".
The resulting tsk file should also be parseable on next start of xcsoar. 

Actual behavior
---------------

 In the file in which the task was stored, the name of WP wasn't included. The name is "1". All other data of the waypoints look good, including the index of the waypoint.
--> on the map the task is shown correctly. See screenshot
--> I tried to use all

 no idea, I found some bad datasets in the waypoint-file, but this was not the root-cause
it doesn't matter wether the name of the waypoint is set between quotion marks or not

ubx:
May help to fix  #1346

### DIFF
--- a/build/test.mk
+++ b/build/test.mk
@@ -93,7 +93,7 @@ TEST_NAMES = \
 	TestColorRamp TestGeoPoint TestDiffFilter \
 	TestFileUtil TestPolars TestCSVLine TestGlidePolar \
 	test_replay_task TestProjection TestFlatPoint TestFlatLine TestFlatGeoPoint \
-	TestMacCready TestOrderedTask TestAATPoint \
+	TestMacCready TestOrderedTask TestAATPoint TestTaskSave\
 	TestPlanes \
 	TestTaskPoint \
 	TestTaskWaypoint \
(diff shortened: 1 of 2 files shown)


lordfolken:
Thank you. I was thinking of adding that.

MaxKellermann:
This test fails randomly on GH actions, but only on Windows. Please fix that.

ubx:
> This test fails randomly on GH actions, but only on Windows. Please fix that.

Fixed in d1acbf41f202dd990f1ea2e51afc0432ac7fba31

PR #1352, opened 2024-02-19
Cppperf

lordfolken:
Findings from cppcheck

### DIFF
--- a/src/Engine/GlideSolvers/GlidePolar.hpp
+++ b/src/Engine/GlideSolvers/GlidePolar.hpp
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 // Copyright The XCSoar Project
- 
+
 #pragma once
 
 #include "PolarCoefficients.hpp"
@@ -17,17 +17,17 @@ struct SpeedVector;
 
 /**
  * Class implementing basic glide polar performance model
- * 
+ *
  * Implements aircraft-specific glide performance, including
  * bugs/ballast, MacCready setting and cruise efficiency.
  *
- * Cruise efficiency is the ratio of actual cruise speed to 
+ * Cruise efficiency is the ratio of actual cruise speed to
  * target to the classical MacCready speed.
  * Cruise efficiency is stored in this class for convenience,
  * it is used in MacCready class.
- * 
+ *
  * The MacCready class uses this GlidePolar data to calculate
- * specific GlideSolutions. 
+ * specific GlideSolutions.
  *
  * This uses a parabolic glide polar:
  * \f[ w = a.V^2+b.V+c \f]
@@ -254,7 +254,7 @@ public:
   void SetBugs(const double clean) noexcept;
 
   /**
-   * Retrieve bugs 
+   * Retrieve bugs
    * @return Cleanliness of glider (0-1]
    */
   constexpr double GetBugs() const noexcept {
@@ -275,7 +275,7 @@ public:
   void SetBallastLitres(const double litres) noexcept;
 
   /**
-   * Retrieve ballast 
+   * Retrieve ballast
    * @return Proportion of possible ballast [0-1]
    */
   constexpr double GetBallast() const noexcept {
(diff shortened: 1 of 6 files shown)


src/Math/Angle.hpp
```diff
@@ -106,7 +106,7 @@ class Angle
     }
   };
 
-  constexpr Angle(DMS dms) noexcept:Angle(Degrees(dms.ToFloat())) {}
+  constexpr Angle(const DMS &dms) noexcept : Angle(Degrees(dms.ToFloat())) {}
```
MaxKellermann, line 109: What's the point?

src/Topography/TopographyFileRenderer.cpp
```diff
@@ -159,7 +159,7 @@ TopographyFileRenderer::Paint(Canvas &canvas,
 
   UpdateArrayBuffer();
   array_buffer->Bind();
-  const ShapePoint *const buffer = nullptr;
+  const ShapePoint *const buffer = 0;
```
MaxKellermann, line 162: That's wrong

src/Topography/Convert.hpp
```diff
@@ -5,9 +5,7 @@
 
 #include "shapelib/mapprimitive.h"
 
-constexpr GeoBounds
-ImportRect(const rectObj r)
-{
+constexpr GeoBounds ImportRect(const rectObj &r) {
```
MaxKellermann, line 8: What's the point?

src/Engine/GlideSolvers/GlidePolar.hpp
```diff
@@ -552,7 +552,7 @@ class GlidePolar
   }
 
   /** Sets the ideal polar coefficients */
-  void SetCoefficients(PolarCoefficients coeff, bool update=true) noexcept {
+  void SetCoefficients(const PolarCoefficients &coeff, bool update=true) noexcept {
```
MaxKellermann: Why?

src/Topography/TopographyFileRenderer.cpp
lordfolken: What i was trying to fix here, is its first use in whats otherwise a nullptr addition:
https://github.com/XCSoar/XCSoar/blob/a7bffc83982970d2cc2d204803ec5b576ffc15b8/src/Topography/TopographyFileRenderer.cpp#L205

src/Math/Angle.hpp
lordfolken: The value that's passed, isn't modified in the function. Why would you reinstantiate it?

src/Topography/TopographyFileRenderer.cpp
MaxKellermann: 1. your change doesn't "fix" this, it's still a null pointer, just not using the `nullptr` keyword, but the `0` constant which is effectively the same here
2. not a bug; this isn't an actual pointer, but an offset into an OpenGL buffer (`GL_ARRAY_BUFFER`), see https://registry.khronos.org/OpenGL-Refpages/es3.0/html/glVertexAttribPointer.xhtml

src/Math/Angle.hpp
MaxKellermann: I asked first :-)
Clarification: I'm not asking about adding `const` - adding that would be okay (though all it would really do is add noise with no actual use, because it's trivial to see nobody modifies the value); my question was about the `&`, i.e. why it's useful to make this a reference. You proposed to make it a reference, and I ask why. You can't reply with another question, it's your proposal and it's up to you to explain it.

lordfolken: I want to pass it by reference so the value isn't copied every time, wasting memory and adding additional operations. 
The const is a safety so nothing ever in that function touches the referenced value, as there are potential other consumers for that value.

Its also marking the function clearly as not modifying of any arguments, which makes it easier to understand when reading / understanding the code. 

While one could argue that this is a small object and the indirection overhead may be a disadvantage, the compiler optimization such as parameter passing will decide on that. 

Like this its clear that we want the place in the code to be optimized.

PR #1354, opened 2024-02-20
Airspace/AirspaceParser: ignore unit for radio frequency

ubx:
This PR is a reimplementation for #1345


lordfolken (approved):
For me this + #1353, all the Tests pass on both WIN64 and UNIX.

PR #1357, opened 2024-02-28
update 'white list' of USB devices with more PIDs for SoftRF Academy

lyusupov:
Brief summary of the changes
----------------------------

add more USB VID/PID pairs for SoftRF Academy into (Android) 'white list' of USB devices


lyusupov:
Thank you, Max!

Issue #1359, opened 2024-03-02
Huge icons with v7.42


Turbo87:
XCSoar versions having and not having the problem
-------------------------------------------------

> Which XCSoar version are you using when seeing the problem?

v7.42

> Which XCSoar version is the last one you know that did not show the problem?

v7.41

System information
------------------

Android on OnePlus 8

Steps to reproduce the behavior
-------------------------------

Start the app and look at the map 😅 

Actual behavior
---------------

![image](https://github.com/XCSoar/XCSoar/assets/141300/6a82894b-c3d1-4bb2-8b9b-ead508d207c3)

Expected behavior
-----------------

Not quite as huge icons... :D

Do you have any idea what may have caused this?
-----------------------------------------------

https://github.com/XCSoar/XCSoar/commit/3b08f28a5addab8bd3655ef1a3d6e9fc636caac0 implemented high-DPI icons. I guess there might be a downscaling factor missing somewhere?

Do you have an idea how to solve the issue?
-------------------------------------------

MaxKellermann:
3b08f28a5addab8bd3655ef1a3d6e9fc636caac0 is not the cause. I bet a bisect would track it down to 0f0d6a63343c7f33e086f8086de0903a616c5ac5, which reverts a bad commit, but meanwhile another commit had modified the icon sizes to something else. For those map icons, we really need to get a proper concept how to size them. We have none, so each change randomly changes something, and nobody has any idea what is going on.

We need to define physical sizes for each icon; how big are they really supposed to be. Nobody has ever done that, and previous XCSoar versions just happened to appease most people.

lordfolken:
This has all been implemented. Just remove the zoom factor from svg convert.

MaxKellermann:
Can somebody please explain f2d0e21a95bba191f6e211456b6e6b06331622b2 to me? @Sundown3867 / @lordfolken ?

It enlarged map icons from 5x6 pixels to 22x22 pixels, for some definition of "pixel".
But what definition of "pixel"?
And why make all icons 4 times larger?

@Sundown3867 's commit message only says "Unification of icon size [...] to 22x22 for Map icons", but this isn't just a "unification", it's a 4x growth. But why?

What is the physical size of these icons? Why didn't you specify a physical size instead of a pixel size?

And, @Sundown3867, what does the rest of the commit diff do?
```
--- a/Data/icons/map_tower.svg
+++ b/Data/icons/map_tower.svg
@@ -1,84 +1,49 @@
[...]

PR #1367, opened 2024-03-05
Resizing of incorrectly sized icons by 1/3

Sundown3867:
Fixes https://github.com/XCSoar/XCSoar/issues/1359


lordfolken:
So the icons are now the defined size. Looks like the dpi value wasn't evaluated correctly in 7.41 and the larger dpi icons where used.

lordfolken:
Android

Sundown3867:
I have tested on Android device with 220dpi and Android device with 443 dpi and both are OK. As I do not have low dpi device, can anybody test it please? Test builds are here https://github.com/Sundown3867/XCSoar/actions/runs/8161686719. Thank you

lordfolken (approved):

PR #1390, opened 2024-03-31
draft: monitor and record wind readings

bomilkar:
writes the content of the Wind InfoBox to xcsoar.log
use it to compare different wind algorithms results.


bomilkar (commented):
It needs a fixed maximum size of buffer space.
This change shall be removed for the production branch.

bomilkar:
I want to take that back!
I can't figure out how to add an AddBoolean to the wind settings panel. I can't follow the convoluted code of how data flows between setting, profile, panel etc. And I don't want to spend time on something that isn't intended to make it into production anyway.
I'll take a new approach.

PR #1393, opened 2024-04-03
new infobox showing last turn average.

GeorgeDowning20:
This looks at the time to complete the last turn and then uses that time to calculate an average. This stops the aliasing that you get when in an uneven core, doing 20s turns. similar to clearnav.

thoughts on the idea?

### DIFF
--- a/src/Computer/GlideComputerAirData.cpp
+++ b/src/Computer/GlideComputerAirData.cpp
@@ -37,7 +37,9 @@ GlideComputerAirData::ResetFlight(DerivedInfo &calculated,
   average_vario.Reset();
 
   lift_database_computer.Reset(calculated.lift_database,
-                               calculated.trace_history.CirclingAverage);
+                               calculated.trace_history.CirclingAverage,
+                               calculated.trace_history.TurnAverage);
+
   calculated.trace_history.circling_available.Clear();
 
   thermallocator.Reset();
@@ -119,6 +121,12 @@ GlideComputerAirData::ProcessVertical(const MoreData &basic,
   lift_database_computer.Compute(calculated.lift_database,
                                  calculated.trace_history.CirclingAverage,
                                  basic, calculated);
+
+  lift_database_computer.Compute(calculated.trace_history.turn_time,
+                                 calculated.trace_history.TurnAverage,
+                                 basic, calculated);
+
+
   calculated.trace_history.circling_available.Update(basic.clock);
 
   circling_computer.MaxHeightGain(basic, calculated.flight, calculated);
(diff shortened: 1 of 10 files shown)


lordfolken:
I'm sorry no one replied to you. It was marked as a draft and I didn´t really see it. 
@bomilkar you are on top of this code at the moment. Do you see merit?

bomilkar:
Is there an Issue related to this PR?
I'd like to understand why 20s average is better than 30s or 10s ?
My personal preference is to compare the 30s average to the climb performance since the beginning of the thermal. That helps me to decide should I continue or should I move on. Often times I regret the last (couple of) cycles because they weren't worth the time. But all that doesn't depend on the average being calculated over a fixed time vs. one full circle (more or less).

GeorgeDowning20:
> Is there an Issue related to this PR? I'd like to understand why 20s average is better than 30s or 10s ? My personal preference is to compare the 30s average to the climb performance since the beginning of the thermal. That helps me to decide should I continue or should I move on. Often times I regret the last (couple of) cycles because they weren't worth the time. But all that doesn't depend on the average being calculated over a fixed time vs. one full circle (more or less).

20s isn’t necessarily better. The idea is the time constant shouldn’t be fixed. Instead it is constantly adjusted to the time of your last turn. 

If you are not properly centered (eg in a gaggle) you sometimes get aliasing where one side of the thermal the 30s average reads 4knts and on the other side 3knts. For example if you are in the “good bit” doing say 20s turns, the 30s average includes the good bit twice and the bad bit once, hence the averager over reads the true climb rate. Like wise in the opposite respect you might be fooled by the averages out of a perfectly good thermal.

bomilkar:
I understand your point, but I don't agree with what you are proposing. Admittedly, that's my opinion, therefor I suggest you open an Issue to discuss the arguments.

I find only very few thermals are "round enough" that your proposal would actually make a noticeable difference to my decision making. When I look at the flight trail while circling the "good" parts aren't always in the same sector of the circles. So, you can add yet another InfoBox, but is going make a difference. Not in my opinion.
Maybe one alternative I can think of: modify the "TC 30s" InfoBox such that instead of a fixed period of 30 sec make that a parameter in the Gauges->Vario panel, with the choices of 10s, 20s, 30s, 40s, and limit to one full circle while circling. Be aware: "TC 30s" also works in cruise. Please also explain the suggested behavior in corner cases, e.g. during the first circle, very wide circles, in cruise, ... And please discuss all this in an Issue before jumping to an implementation.

lordfolken:
Closing as stale draft for now. The idea (last-turn time window for a smoother average, similar to ClearNav) is still valid — if anyone wants to pick this up, feel free to reopen this PR or start a fresh branch from current master.

PR #1406, opened 2024-04-21
NMEAChecksum: initializer list is ambigous on OSx

### LINKED ISSUE #1365: IOS/MacOSX Targets fail with Checksum.hpp:17:1: note: candidate template ignored: couldn't infer template argument '_src:auto'
Current master fails to build on ios and macos targets. IOS doesn't report an error interestingly. 

```
src/Device/Util/NMEAReader.cpp:53:39: error: no matching function for call to 'NMEAChecksum'
  const uint8_t calculated_checksum = NMEAChecksum({start, asterisk});
                                      ^~~~~~~~~~~~
./src/NMEA/Checksum.hpp:43:1: note: candidate function not viable: cannot convert initializer list argument to 'std::string_view' (aka 'basic_string_view<char>')
NMEAChecksum(std::string_view src) noexcept
^
./src/NMEA/Checksum.hpp:17:1: note: candidate template ignored: couldn't infer template argument '_src:auto'
NMEAChecksum(std::convertible_to<const char *> auto &&_src) noexcept
^
1 error generated.
gmake: *** [build/compile.mk:104: output/OSX64/dbg/src/Device/Util/NMEAReader.o] Error 1
gmake: *** Waiting for unfinished jobs....
gmake: 
```

lordfolken:


MaxKellermann:
This is a terrible way to deal with this Apple toolchain bug. Be careful who to pick code from.

lordfolken:
Hey max, thanks for chipping in. :-)

I'm aware that this is a POS. If this was a primary target i wouldn't have merged this approach. The issue is also not considered solved, until a decent solution is implemented. Respectively the bug is fixed in the tool chain.
At the moment I need the CI target to function to catch perhaps other bugs, as i'd like to send out a release. (btw did you see my mail concerning the playstore release, to @blarg.de 2024-03-17?)

PR #1408, opened 2024-04-24
Make copy and past of infobox set more clear in UI

### LINKED ISSUE #214: InfoBox Sets Copy then Paste don't seem to overwrite destination infobox
When wanting to change the infoboxes positions using the InfoBox Sets Config screens the paste function does not overwrite the selected infobox. 

Test Case:
Navigate to 
* Config -> System -> Look -> InfoBox Sets -> {any infobox set}
* Select an infobox
* press Copy
* select another infobox
* press Paste
* Press Yes when asked to overwrite

The destination infobox does not change.

Tested in:
XCS7 on Android & XCS6 on Android, Mac and Linux

kobedegeest:
closes #214

### DIFF
--- a/src/Dialogs/Settings/dlgConfigInfoboxes.cpp
+++ b/src/Dialogs/Settings/dlgConfigInfoboxes.cpp
@@ -247,9 +247,9 @@ InfoBoxesConfigWidget::Prepare(ContainerWindow &parent,
   button_style.TabStop();
 
   const auto &button_look = GetLook().button;
-  copy_button.Create(parent, button_look, _("Copy"), layout.copy_button,
+  copy_button.Create(parent, button_look, _("Copy Set"), layout.copy_button,
                      button_style, [this](){ OnCopy(); });
-  paste_button.Create(parent, button_look, _("Paste"), layout.paste_button,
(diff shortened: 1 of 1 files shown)


src/Dialogs/Settings/dlgConfigInfoboxes.cpp
```diff
@@ -247,9 +247,9 @@ InfoBoxesConfigWidget::Prepare(ContainerWindow &parent,
   button_style.TabStop();
 
   const auto &button_look = GetLook().button;
-  copy_button.Create(parent, button_look, _("Copy"), layout.copy_button,
+  copy_button.Create(parent, button_look, _("Copy \n Set"), layout.copy_button,
```
lordfolken, line 250: Please don't try to format the text of the button. 
If it overflows the button's label rendering should take care. 
Its also bad for translations.

lordfolken (changes requested):
Clear and conciece commit messages are essential, especially in a code base that spawns over 20 years.

first commit should be named:

dlgConfigInfoboxes.cpp: Rename button label copy/paste to copy/paste set

second commit:

dlgConfigInfoboxes.cpp: Improve warning text when pasting infobox sets

lordfolken (approved):

PR #1414, opened 2024-04-29
Improved button layout

kobedegeest:
Many menus and dialogs did not follow the developer manual policy regarding button placement (place of close and cancel button) 
Also naming of close cancel and ok was not consistent accros all menus
Some menus hide inactive buttons instead of setting them disabled, hiding the buttons often leaves gaps between the remaining buttons.


lordfolken:
Good idea. I'll test it. 

FYI, you spelled visible and disabled wrong in your commit texts. (just if you wanted to correct that.)

lordfolken (changes requested):
So i compiled and tried it. 
For me this makes a lot of sense and clarifies the ui. 
If you could correct the spelling mistakes in the commit messages, then we can merge this.

lordfolken (approved):

PR #1417, opened 2024-04-30
Dialog for setting user-defined frequencies

jirimaier:
XCSoar can set the frequency of a radio. However, this is currently only possible for frequencies that are tied to an airfield (defined in a waypoints file). This PR proposes a dialog that shows a list of user-defined frequencies. the frequency list is editable via GUI and stored in user.freq file in XCSoar's root directory.

This will be particularly useful for frequencies like:
- glider air-to-air
- FIC
- emergency

image

image

image

image

## Summary by CodeRabbit

* **New Features**
  * Added frequency list management dialog to browse, create, edit, and manage custom radio frequencies
  * Added radio frequency entry dialog for editing individual frequency values
  * Added new "Frequencies" input event to access the frequency management interface
  * Support for saving and loading custom frequency lists to a persistent user file

### DIFF
--- a/src/Dialogs/Frequency/dlgFrequencyEdit.cpp
+++ b/src/Dialogs/Frequency/dlgFrequencyEdit.cpp
@@ -1,0 +1,42 @@
+#include "dlgFrequencyEdit.hpp"
+
+FrequencyEditResult
+dlgFrequencyEditShowModal(FrequencyListItem &freq)
+{
+    const DialogLook &look = UIGlobals::GetDialogLook();
+    TWidgetDialog<FrequencyEditWidget>
+        dialog(WidgetDialog::Auto{}, UIGlobals::GetMainWindow(),
+               look, _("Frequency Editor"));
+    dialog.AddButton(_("OK"), mrOK);
+    dialog.AddButton(_("Cancel"), mrCancel);
+    dialog.SetWidget(look, freq);
+    const int result = dialog.ShowModal();
+
+    if (result != mrOK)
+        return FrequencyEditResult::CANCEL;
+
+    if (!dialog.GetChanged())
+        return FrequencyEditResult::UNMODIFIED;
+
+    freq = dialog.GetWidget().GetValue();
+    return FrequencyEditResult::MODIFIED;
+}
+
+void FrequencyEditWidget::Prepare(ContainerWindow &, const PixelRect &) noexcept
+{
+    AddText(_("Name"), nullptr, value.name.c_str(), this);
+    AddText(_("Short Name"), nullptr, value.short_name.c_str(), this);
+    AddFrequency(_("Frequency"), nullptr, value.freq, this);
+}
+
+bool FrequencyEditWidget::Save(bool &_changed) noexcept
+{
+    value.name = GetValueString(NAME);
+    value.short_name = GetValueString(SHORTNAME);
+    value.freq = GetValueRadioFrequency(FREQ);
+    if(!value.freq.IsDefined())
+        return false;
+
+    _changed |= modified;
+    return true;
+}
--- a/src/Dialogs/Frequency/dlgUserFrequencyList.cpp
+++ b/src/Dialogs/Frequency/dlgUserFrequencyList.cpp
@@ -1,0 +1,316 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The XCSoar Project
+
+#include "dlgUserFrequencyList.hpp"
+#include "LogFile.hpp"
+#include "system/Path.hpp"
+#include "io/FileOutputStream.hxx"
+#include "io/BufferedCsvReader.hpp"
+#include "io/FileReader.hxx"
+#include "io/BufferedOutputStream.hxx"
+#include "io/BufferedReader.hxx"
+#include "LocalPath.hpp"
+#include "io/DataFile.hpp"
+#include "io/StringConverter.hpp"
+#include "ActionInterface.hpp"
+#include "UIGlobals.hpp"
+#include "dlgFrequencyEdit.hpp"
+#include "Dialogs/Message.hpp"
+#include <string_view>
+
+void UserFrequencyListWidget::ReadFile()
+{
+  std::unique_ptr<Reader> reader;
+  list.clear();
+  try
+  {
+    reader = OpenDataFile(file_name);
+  }
+  catch (...)
+  {
+    // file does not exist, list stays empty
+    ListUpdated();
+    return;
+  }
+
+  BufferedReader buffered_reader{*reader};
+  StringConverter string_converter(Charset::UTF8);
+
+  std::array<std::string_view, 4> params;
+
+  int line = 0;
+  bool has_invalids = false;
+  while (1)
+  {
+    size_t fields_parsed = ReadCsvRecord(buffered_reader, params);
+    if (fields_parsed == 0)
+      break;
+    line++;
+    try
+    {
+      if (fields_parsed != 3)
+        throw std::runtime_error("Invalid number of fields");
+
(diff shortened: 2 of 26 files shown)


lordfolken:
Hey, 
So i'm all for expanding the radio control situation, which is currently providing little advantage over entry directly on the device. 

I think a favourites list like this, that can be activated by tapping on the radio infobox, or called up via a menu option, is a good thing.

What i would like to avoid, is yet another file that has to be managed by the user,  outside of xcsoar. 

So please consider:
  * User editable from the application
  * Saving stations in the profile

DanD222:
@lordfolken Would you be OK with an iterative approach, i.e. get the functionality of reading from a file and sending the frequency to a radio  done first, and then expand the XCSoar editing features?

jirimaier:
Hello,

This is a draft that is far from finished. I mainly wanted to have such a discussion about proper implementation, so thank you for sharing your ideas.

I already have an intention to add GUI controls for editing the list. New/Edit/Delete, maybe Copy would also be useful.

I also don't like storing it in a separate file; storing it in preferences may be a good option. However, I would like to have a way to load/transfer a larger amount of frequencies quickly rather than typing them manually (for example there are going to be 8 air-to-air frequencies for cross-country gliding in Czechia this year). Maybe I can store it in preferences and add some import/export button that would load/create the frequencies file.

Or I can think of a hybrid approach, where the frequencies will be loaded from both preferences and a file. In that case, it is questionable how to order them in the list (I'd like to let the user select the order rather than sorting them by name, but when they come from multiple sources, this is not so simple), and also whether the user should be able to edit the separate file (from GUI) or only those in preferences.

lordfolken:
> @lordfolken Would you be OK with an incremental approach, i.e. get the functionality of reading from a file and sending the frequency to a radio done first, and then expand the XCSoar editing features?
> 
> Edit - incremental.. shouldn’t try to use fancy words.

The problem with that is, that you would have to code the migration path as well. As users expect "everything to just work". So i'd rather have the functionality first.

lordfolken:
> I also don't like storing it in a separate file; storing it in preferences may be a good option. However, I would like to have a way to load/transfer a larger amount of frequencies quickly rather than typing them manually (for example there are going to be 8 air-to-air frequencies for cross-country gliding in Czechia this year). Maybe I can store it in preferences and add some import/export button that would load/create the frequencies file.

Potentially you could also create an import/export text field, where a csv or multiline file could be pasted/copied. The result of which would go to the profile. 

An additional option would be to add a "to favorite" button in airspace and waypoint dialogs.

jirimaier:
What about taking the Waypoit Editor as a precedent? 
The user would create/edit the list in GUI and it will get saved in a file (let's name it user.freq).
There is no need for the user to ever touch the file, but they can still copy it to another device if they want to.

src/Dialogs/Frequency/dlgFrequencyEdit.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+    freq = dialog.GetWidget().GetValue();
+    return FrequencyEditResult::MODIFIED;
+}
+
+void FrequencyEditWidget::Prepare(ContainerWindow &, const PixelRect &) noexcept
+{
+    AddText(_("Name"), nullptr, value.name.c_str(), this);
+    AddText(_("Short Name"), nullptr, value.short_name.c_str(), this);
+    AddFloat(_("Frequency"), nullptr, _("%.3f"), _("%.3f"), 118.000, 137.000, 0.005, false, value.freq.GetKiloHertz() / 1000.0, this);
```
lordfolken, line 29: To make it fully clean we should add an AddFrequency() in EditRowFormWidget  that deals with frequencies.
Then add a check in RadioFrequency to parse the vaildify of the Frequency.

src/Dialogs/Frequency/dlgUserFrequencyList.cpp
```diff
@@ -0,0 +1,239 @@
...
+void UserFrequencyListWidget::CreateButtons(WidgetDialog &dialog)
+{
+  form = &dialog;
+  if (!edit_mode)
+  {
+    set_button = dialog.AddButton(_("Set Active"), [this]()
+                                  { SetClicked(true); });
+    set_standby_button = dialog.AddButton(_("Set Stand-by"), [this]()
+                                          { SetClicked(false); });
+    edit_mode_button = dialog.AddButton(_("Edit"), [this]()
```
lordfolken, line 105: I think edit here is imprecise as it would lead the user to believe that this edits the current frequency. Please use "Manage" for this stage, Then edit for editing the items.

lordfolken (changes requested):
When saving the frequencies on manage mode somehow the MessageBox blanks the background
image

Please make the frequencies dialog also accessible from the Active/Standby infoboxes via the touch menu.

lordfolken:
A numpad kind of keyboard entry widget for numbers/frequencies would be a welcome addtion.

src/Dialogs/Frequency/dlgUserFrequencyList.cpp
```diff
@@ -0,0 +1,239 @@
...
+                                        { dlgUserFrequencyListWidgetShowModal(true);
```
lordfolken, line 106: Whitespace

src/Dialogs/Frequency/dlgUserFrequencyList.cpp
```diff
@@ -0,0 +1,239 @@
...
+void UserFrequencyListWidget::EditClicked()
+{
+  assert(GetList().GetCursorIndex() < list.size());
+
+  FrequencyListItem item = list[GetList().GetCursorIndex()];
+
+  if (dlgFrequencyEditShowModal(item) == FrequencyEditResult::MODIFIED)
+  {
+    list[GetList().GetCursorIndex()] = item;
+    listUpdated();
```
lordfolken, line 189: Whitespace

lordfolken:
Invalid lines in user.freq get erased when saving. The question is whether this should be handled or not.

lordfolken:
Ive had several crashes when saving on android. Guess we more error handling.

jirimaier:
> When saving the frequencies on manage mode somehow the MessageBox blanks the background 

This behavior is same as in the Waypoint Editor. Perhaps it can't be avoided, because the message box is show after the dialog is closed. I may show the message box after clicking Close button and closing the dialog after Yes/no is selected. However, the behavior would then be different for Close button and closing by escape key.

jirimaier:
> Invalid lines in user.freq get erased when saving. The question is whether this should be handled or not.

I have an idea: if the file contains invalid line and user opens the edit/manage screen, show message box saying "There are invalid lines and they will be deleted if changes are saved" and immediately set the 'modified' flag to show the "save changes" prompt after closing even if no changes were made.

src/Dialogs/RadioFrequencyEntry.cpp
```diff
@@ -0,0 +1,65 @@
...
+  dialog.SetWidget(std::move(frequency_entry));
+
+  if (dialog.ShowModal() != mrOK)
+    return false;
+
+  auto &entry = ((DigitEntry &)dialog.GetWidget().GetWindow());
+  value = RadioFrequency(entry.GetRadioFrequency());
+  
+  return true;
+}
```
lordfolken, line 65: missing newline

lordfolken:
Good work on the Frequency dialog. Really cool. :-)

xtophe80:
In case it is useful, I have pushed my rebased version of #867 mentioned last week in my repo: https://github.com/xtophe80/XCSoar/tree/FrequencyCard2

lordfolken:
@xtophe80 can you colaborte with @jirimaier ?

PR #1429, opened 2024-05-16
Infoboxes/Content: add estimated task speed infobox

ericcarden:
This change adds estimated task speed (the value currently shown as "Speed estimated" on the Status-Task panel) to the list of available infoboxes. This could be a desirable infobox in a speed record attempt flight. I recently missed such a record by less than the duration of one circle, and I'd have probably left the last thermal at least one circle sooner (and gotten the record) had I been able to see this value without having to go back and forth to the Status-Task panel to see it.

This change also changes "cross country" to "cross-country" in every infobox description that contained "cross country". Previously, infobox descriptions were inconsistent regarding whether "cross country" or "cross-country" was used. It also tidies up the Type enum in Type.hpp (to make the most recent additions look like the earlier ones - just whitespace and comment changes), since a new item was being added here, anyway.

PR #1431, opened 2024-05-16
Infoboxes/Content: add estimated task speed infobox

ericcarden:
This change adds estimated task speed (the value currently shown as "Speed estimated" on the Status-Task panel) to the list of available infoboxes. This could be a desirable infobox in a speed record attempt flight. I recently missed such a record by less than the duration of one circle, and I'd have probably left the last thermal at least one circle sooner (and gotten the record) had I been able to see this value without having to go back and forth to the Status-Task panel to see it.

### DIFF
--- a/src/InfoBoxes/Content/Factory.cpp
+++ b/src/InfoBoxes/Content/Factory.cpp
@@ -1128,6 +1128,14 @@ static constexpr MetaData meta_data[] = {
     UpdateInfoTaskETAorAATdT,
   },
 
+  // e_SpeedTaskEst
+  {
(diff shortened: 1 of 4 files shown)


ericcarden:
I just added one more condition (!task_stats.total.IsAchievable()) to the new function in Task.cpp that can cause the infobox to show "---": MC>0 but too much wind and/or MC too low (lift too weak) to finish the task. This same check is used for infoboxes like "Fin ETA" and "Fin ETE".

ericcarden:
I'm closing this PR as requested by @lordfolken in PR #1429.

PR #1454, opened 2024-05-22
ui/canvas/memory/Export.cpp: fix compile error using USE_FB=y

ubx:


lordfolken:
@ubx do you think we should configure one of the targets in the CI to USE_FB=y? 
That way we  should find the compilation failures.

lordfolken (approved):

PR #1456, opened 2024-05-22
Waypoint/WaypointCommandsWidget.cpp: Dissable frequency buttons

kobedegeest:
Did all these buttons in one commit because as far as my understanding goes i had to use enum Buttons {...}
So needed fixed number of buttons i assume. 
Either way those if statements to not add the buttons seemed to be the wrong way of doing it.

### DIFF
--- a/src/Dialogs/Waypoint/WaypointCommandsWidget.cpp
+++ b/src/Dialogs/Waypoint/WaypointCommandsWidget.cpp
@@ -22,6 +22,7 @@
 #include "Operation/MessageOperationEnvironment.hpp"
 #include "Profile/Current.hpp"
 #include "ActionInterface.hpp"
+#include "Widget/RowFormWidget.hpp"
 
 static bool
 ReplaceInTask(ProtectedTaskManager &task_manager,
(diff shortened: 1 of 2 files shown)


src/Dialogs/Waypoint/WaypointCommandsWidget.cpp
```diff
@@ -202,78 +204,92 @@ ActivatePan(const Waypoint &waypoint)
   return PanTo(waypoint.location);
 }
 
+void
+WaypointCommandsWidget::UpdateButtons()
+{
+  has_freq = waypoint->radio_frequency.IsDefined();
+  SetRowEnabled(REPLACE_IN_TASK, task_manager != nullptr);
+  SetRowEnabled(INSERT_IN_TASK, task_manager != nullptr);
```
lordfolken, line 212: INSERT_IN_TASK creates a new "goto task" currently. So this shouldn't be disabled if there is no current task defined.

lordfolken: The same is true for APPEND_TO_TASK

lordfolken: Ah I see. I think this is part of Max's defensive programming. Basically the task_manager is initialized to a pointer. If this pointer should be invalid for any reason, the button would be disabled.

lordfolken (changes requested):
In general, function and implemenation wise ok. 
Spelling error in commit message. Dissable->Disable
Please separate code and formatting changes into two different commits.
The button variables should be named _button

src/Dialogs/Waypoint/WaypointCommandsWidget.cpp
```diff
@@ -22,6 +22,8 @@
 #include "Operation/MessageOperationEnvironment.hpp"
 #include "Profile/Current.hpp"
 #include "ActionInterface.hpp"
+#include "UIGlobals.hpp"
```
lordfolken, line 25: This include is unnecessary?

lordfolken:
One thing i noticed with the "remove from task" button: You cannot have less than two waypoints, and the task manager complains that it would be an invalid task. If there is only one waypoint, i would expect it to be reconverted to a goto task.

lordfolken:
> > Please separate code and formatting changes into two different commits.
> You mean the indentation changes after removing the if's? I will try to separate (and remember for the furure).

Yes please. This makes reviewing changes easier, and future bug search as well. 

> 
> > The button variables should be named _button
> 
> This just some naming convention? or does it also do something? Just to learn because first i tried button->SetEnabled(bool) but could not get that to work.

I meant the format like edit_button insert_button etc. Yes on one hand convention, on the other less naming conflicts. Clean readability; Because insert and edit are too generic as variable names. 

src/Dialogs/Device/DeviceListDialog.cpp is a good example. 

> 
> > One thing i noticed with the "remove from task" button: You cannot have less than two waypoints, and the task manager complains that it would be an invalid task. If there is only one waypoint, i would expect it to be reconverted to a goto task.


> This is a separate issue, or do you want it to be adressed in this pr? I am not sure how to do that yet. So there is this part in the code.
> 
> ```
> case MapTaskManager::INVALID:
>     ShowMessageBox(_("Task would not be valid after the change."), _("Error"),
>                 MB_OK | MB_ICONEXCLAMATION);
>     break;
> 
>   case MapTaskManager::MUTATED_FROM_GOTO:
>   case MapTaskManager::MUTATED_TO_GOTO:
> ```
> 
> But it seems thus that MUTATED_TO_GOTO is not triggerd but INVALID instead.

That's indeed a separate issue and already present in the current head. So no need to handle in this pr.

PR #1460, opened 2024-05-23
Dialogs/StatusPanels: On the Status-Task panel, make “Speed estimated” blank when “Speed remaining” is blank.

ericcarden:
This change makes “Speed estimated” on the Task-Status panel blank when “Speed remaining” on that panel is blank. The latter is blank if MC>0 but the task can’t be finished due to either too much wind or MC setting being too low given the amount of wind. Like “Speed remaining”, “Speed estimated” (the predicted start-to-finish average task speed) has no meaning if the task can’t be finished. This also makes this field more like infoboxes that show “---” in this can’t-finish-task scenario (e.g., “V Task Est” and “Fin ETA”).

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -5,6 +5,7 @@ Version 7.43 - not yet released
   - TC 30s Infobox shows now climb rate since start of thermal
   - TL Gain Infobox shows now the overall climb rate of last thermal
   - add new Infobox V Task Est (Speed task estimated)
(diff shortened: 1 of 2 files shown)


ericcarden:
By the way, the new (in this commit) logical test used to control whether “Speed estimated” is shown (task_stats.total.remaining_effective.IsDefined()) is the same as the one used to control whether “Speed remaining” is shown.

lordfolken (approved):

PR #1465, opened 2024-05-30
lua/Http.cpp: Use common curl settings

styblope:
Applies the same common Curl options that are used elsewhere in XCSoar for the Lua HTTP client. These currently include disabling host and peer SSL verification on Android to circumvent connection errors due to missing access to Android's own CA certs (#1469)

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -22,6 +22,8 @@ Version 7.43 - not yet released
 * airspace
   - add classes according the OpenAir extended format AY tag
   - coloring is done according to the AC tag, or if present and allowed, according to the AY tag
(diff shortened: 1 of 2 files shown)


lordfolken (changes requested):
What shall i say? Thanks  I hate it. ;-)

Yes this is currently needed, as we don't have the certificate checks enabled on a diverse array of platforms. 

Please:
 - Open a new issue that documents this fact that we currently do not check certificates an d where this is located in the code - prefeerably all instances
 - add a NEWS.txt entry for this change.

lordfolken (approved):

PR #1467, opened 2024-05-30
Add pan to

### LINKED ISSUE #1443: Waypoint Details Dialog: should show the PanTo Button on the first Page (When in Target mode)
image
image

kobedegeest:
Guess this is what is wanted
closes #1443 

I did notice when testing that the panto function could be improved. The infobox pannel is still open when you press panto so you end up in pan mode but with that panel widget still open. 

Or maybe this pannel should close when "details" is pressed.


lordfolken (approved):
Nice. Ty.

PR #1475, opened 2024-06-02
Dialogs/StatusPanels: consolidate duplicate IF statements

ericcarden:
This change moves some code to eliminate two repeated IF statements in the Refresh() function. It also adds a comment describing the effect of this IF statement.


ericcarden:
Never mind. I have other things in mind for this file, and things will get messy if these are grouped like this.

PR #1476, opened 2024-06-02
Dialogs/StatusPanels: clear some Task panel fields after task finish

ericcarden:
Make Status-Task panel fields "Remaining time", "Remaining distance", "Speed estimated", and "Speed remaining" blank after task finish. These fields have no meaning after the task is finished. The same could be said for "Estimated task time", by the way, but I left it visible even after task finish as the only way on this panel to see the actual task time after finishing the task.


lordfolken:
fe70abed3432ec918225d65d7a7892eab511cf3d i merged this some time ago..

PR #1478, opened 2024-06-03
Correct SGS 2-33 polar based on mph instead of knots

stewi2:
This corrects the polar for the Schweizer SGS 2-33, which apparently had originally been entered assuming the polar in the handbook was using knots instead of mph. This means all velocities were artificially high, resulting in better performance than actual.

This takes the speeds from the original polar, and simply scales them back down.

To verify, I have digitized the polar from the handbook, and plotted it along with the old, and new quadratic polars:

image

I also corrected the reference weight. The polar in the handbook is based on 1040 lbs, which equals 471.74 kg.


lordfolken:
Hi,  
Thx for having a look and taking the time to correct this polar. 

Please update the commit message to the following:

PolarStore: correct SGS-233 polar
This corrects the polar for the Schweizer SGS 2-33, which apparently had originally been entered assuming the polar in the handbook was using knots instead of mph. This means all velocities were artificially high, resulting in better performance than actual.

lordfolken (approved):

Issue #1487, opened 2024-06-07
Defaulted equality comparison may not work as expected


bomilkar:
XCSoar versions having the problem
-------------------------------------------------

7.42 in combination with compiler version g++-10 .

XCSoar versions not having the problem
-------------------------------------------------

The issues was introduced when specific C++20 language features were used.

System information
------------------

Affected are platforms with older compilers, for instance g++-10. But also platforms with g++-12 are affected.

Steps to reproduce the behavior
-------------------------------

Issue #1253 demonstrates an issue in one place. Other places might be affected, too.

Expected behavior
-----------------

Actual behavior
---------------

Do you have any idea what may have caused this?
-----------------------------------------------

Some problems have been reported that are due to shortcomings in certain compilers. For example, it turned out that Issues #1253  is compiler version specific: g++ version 10.2 does not implement the default “operator==” construct in the same way as g++ version 12. This becomes clear if you replace the default “operator==” with the explicit “operator==” declaration.

One additional issue is the declaration of the inequality operator „operator!=“ together with the declaration of the „operator==“ be it defaulted or not. The reference manual says “per the rules for operator==, this will also allow inequality testing.” In other words declaring the equality operator implicitly declares the inequality operator. The reference manual is not explicit what happens if both operators are declared.

These are the affected files:
src/Geo/Flat/FlatGeoPoint.hpp
src/Geo/GeoPoint.hpp
src/Math/Point2D.hpp
src/PageSettings.hpp

Do you have an idea how to solve the issue?
-------------------------------------------

To address this second issue I recommend to remove the declarations of the inequality operator where it comes together with the corresponding equality comparison. This will remove ambiguity.

To address the first issue I recommend to replace the defaulted equality comparison by the explicit declaration. (If the defaulted equality comparison causes bad code.) We can fix it either here in the xcsoar code or we can create a patch which should be applied on platforms which are still on g++-10. One affected platform is OpenVario. Are there any others (we should care about)?

Details to the “default comparisons” are here: https://en.cppreference.com/w/cpp/language/default_comparisons

lordfolken:
@August2111 heads up to you, as this affects ov builds.

bomilkar:
> Can we revert these commits? Documentation also mentions GCC 10.

[...]

PR #1493, opened 2024-06-09
remove redundant inequality comparison operators

bomilkar:
This PR addresses a part of Issue #1487 .

The LRM for C++20 says "per the rules for operator==, this will also allow inequality testing." In other words declaring the equality operator implicitly declares the inequality operator. The reference manual is not explicit about what happens if both operators are declared. To avoid ambiguity and inconsistent declarations of inequality operators (be it defaulted or not) I suggest to remove them.
Details to the “default comparisons” are here: https://en.cppreference.com/w/cpp/language/default_comparisons


lordfolken (changes requested):
Ok let's remove the excessive declations. 
Please squash all three commits and 
add your PR text as reasoning to the commit. Then we can merge this.

PR #1499, opened 2024-06-11
src/Waypoint: restore check for header in user.cup

ericcarden:
This change restores the check for a header (a line containing only field names) in a CUP-format waypoint file when reading the file. This fixes a problem introduced in v7.42 via PR #1305, part of which was the removal of such a check. This (probably inadvertently) made it so that XCSoar now assumes there to always be a header. The XCSoar-created user.cup file, though, historically hasn’t had a header – and currently isn’t given one when XCSoar creates the file. So the changes in PR #1305 made it so that the first record (waypoint) in user.cup is ignored. If a new waypoint is added via the Waypoint Editor’s “New” or “Import” function, the result is worse: the first waypoint in user.cup is deleted. Each time the Waypoint Editor is exited after creating a new waypoint or importing a waypoint, another existing waypoint in user.cup is deleted. Restoring the check for a header fixes these problems.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -4,6 +4,7 @@ Version 7.43 - not yet released
 * ui
   - TC 30s Infobox shows now climb rate since start of thermal
   - TL Gain Infobox shows now the overall climb rate of last thermal
+  - add new Infobox V Task Est (Speed task estimated)
   - add missing airspace to Select Airspace filter
   - NumberEntry dialog value can now be accepted by enter
   - Vario center gross label
--- a/src/InfoBoxes/Content/Factory.cpp
+++ b/src/InfoBoxes/Content/Factory.cpp
@@ -1128,6 +1128,14 @@ static constexpr MetaData meta_data[] = {
     UpdateInfoTaskETAorAATdT,
   },
 
+  // e_SpeedTaskEst
+  {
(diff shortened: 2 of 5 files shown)


ericcarden:
@lordfolken: It looks like this bug fix didn't make it into the v7.43 release. Could you tell me why not, please? Did I foul things up by (prematurely?) deleting the public branch associated with this PR? Thanks.

lordfolken:
I don't know what went wrong here... but i had to forcepush history a couple of times. Probably accidently dropped the patch. 

Could you rebase and reopen? 
https://github.com/XCSoar/XCSoar/commit/4693ee2bb1070e9f8e588e30133b414b2018cf6a is the commit.

lordfolken:
I think you need to restore your branch or create one named the same. A PR is must that pull from my repo/branch.

ericcarden:
Thanks. I just now created a branch with the same name (header_check), applied the changes to it (as I had saved in a patch), tested the changes, and pushed the new branch to my online repo. I don't know how to associate it with this PR, though, in order to reopen this PR. Can you help, please, or should I create a new PR? Thanks.

lordfolken:
Yes create a new PR.

ericcarden:
Done. It's PR #1553.

PR #1510, opened 2024-06-27
InfoBoxes/Content: add Home Altitude Difference infobox

ericcarden:
This change adds a “Home altitude difference” (arrival altitude to home waypoint) infobox. This idea came from discussion 1464 (https://github.com/XCSoar/XCSoar/discussions/1464). Top Hat has had this infobox for 11 years. I agree with the requestor (a Top Hat user wanting to switch to XCSoar) in the discussion that the most important thing a flight computer can do is inform you about your reach to a safe landing place, and having this info in an infobox makes the info easier to get. Currently, one typically needs to pan and zoom or visit the alternates list or waypoints list in order to get this info while flying a task.

### DIFF
--- a/src/Computer/Settings.cpp
+++ b/src/Computer/Settings.cpp
@@ -20,6 +20,7 @@ PlacesOfInterestSettings::ClearHome()
 {
   home_waypoint = -1;
   home_location_available = false;
+  home_elevation_available = false;
 }
 
 void
@@ -28,6 +29,11 @@ PlacesOfInterestSettings::SetHome(const Waypoint &wp)
   home_waypoint = wp.id;
   home_location = wp.location;
   home_location_available = true;
+  if (wp.has_elevation) {
+    home_elevation = wp.elevation;
+    home_elevation_available = true;
+  } else
+    home_elevation_available = false;
 }
 
 void
--- a/src/Waypoint/HomeGlue.cpp
+++ b/src/Waypoint/HomeGlue.cpp
@@ -27,6 +27,13 @@ FindHomeId(Waypoints &waypoints,
 
   settings.home_location = wp->location;
   settings.home_location_available = true;
+
+  if (wp->has_elevation) {
+    settings.home_elevation = wp->elevation;
+    settings.home_elevation_available = true;
+  } else
+    settings.home_elevation_available = false;
+
   waypoints.SetHome(wp->id);
   return wp;
 }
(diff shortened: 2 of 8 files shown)


ericcarden:
I don't understand the Codacy "issues" found. They say two variables/members are never used, but that isn't so. They're used in the new function in Places.cpp.

lordfolken:
> I don't understand the Codacy "issues" found. They say two variables/members are never used, but that isn't so. They're used in the new function in Places.cpp.

You can ignore those.

src/Computer/Settings.cpp
```diff
@@ -28,6 +29,11 @@ PlacesOfInterestSettings::SetHome(const Waypoint &wp)
   home_waypoint = wp.id;
   home_location = wp.location;
   home_location_available = true;
+  if (wp.has_elevation) {
+    home_elevation = wp.elevation;
+    home_elevation_available = true;
+  } else
+    home_elevation_available = false;
 }
```
lordfolken, line 37: This else isn't necessary.

src/Waypoint/HomeGlue.cpp
```diff
@@ -27,6 +27,13 @@ FindHomeId(Waypoints &waypoints,
 
   settings.home_location = wp->location;
   settings.home_location_available = true;
+
+  if (wp->has_elevation) {
```
lordfolken, line 31: You are tracking the state in separate variables. Please check home_elevation for value or nullptr instead.

lordfolken (changes requested):
Comment should be in commit message, describing the change and the infobox.

ericcarden:
> Comment should be in commit message, describing the change and the infobox.

Thanks for reviewing this PR. Can you clarify what you mean by this comment, please? It appears to me that this commit's message/comment already describes the change and the infobox. It reads...

"add infobox Home Altitude Difference (arrival altitude to home waypoint)"

Thanks.

src/Computer/Settings.cpp
ericcarden: Are you sure? Without this, if the user changes home waypoints using the "Set as new home" button - changing from a waypoint WITH elevation data to one WITHOUT elevation data - wouldn't home_elevation_available wrongly remain TRUE (when it should be changed to FALSE)?

src/Waypoint/HomeGlue.cpp
ericcarden: Thanks for reviewing and commenting. I copied this approach from Top Hat, and it's like how elevation is managed in the Waypoint structure (in XCSoar). This structure has "has_elevation" and "elevation" members, and the latter is to be used only if the former is true. "home_elevation" isn't a pointer, so it can never be nullptr, right? I suppose we could choose some impossible value (e.g., -999m) to which to set home_elevation to signify that it's an invalid/unavailable value (thus negating the need for the separate home_elevation_available variable), but this is inconsistent with how it's done in the Waypoint structure. Would you like me to do something like that... or leave it as is? Thanks.

nealedj:
Any chance this could be reviewed again please @lordfolken ? It's a feature I went looking for and was glad to see had been implemented. It looks like @ericcarden has responded to your requested changes.

lordfolken:
merged in 9b19fb97adf1034861ff1516bb5e71e5e2dbc5ca

ericcarden:
> merged in [9b19fb9](https://github.com/XCSoar/XCSoar/commit/9b19fb97adf1034861ff1516bb5e71e5e2dbc5ca)

Thanks. I don't see it in 9b19fb9, but I see it in 2e12c13.

PR #1513, opened 2024-07-02
Engine/Task: fix calculation of achieved task distance

ericcarden:
This change fixes the calculation of achieved task distance, making it total planned task distance minus task distance remaining. Without this fix, achieved task distance is calculated as the sum of completed legs’ distances plus the distance from the start of the current leg to the aircraft’s current location, which is only accurate if the aircraft is exactly on the course line. This calculation affects the values shown in infoboxes “Speed task achieved”, “Speed task average”, and “Speed task last hour” and in the “Speed average” and “Achieved speed” fields on the Status-Task panel. A simple way to see the problem with the current calculation is to start a task and then turn and fly perpendicular to the task line. The “Speed task average” infobox, for example, will show a speed roughly equal to ground speed even though you’re getting no closer to the active task point – and even after you’re farther from the active task point than you were when you started the task.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -15,6 +15,11 @@ Version 7.43-rc1 - not yet released
 Version 7.43-rc0 - 2024-06-02
 * map
   - reachable map labels show correct altitude with head wind #1424
+* calculations
+  - Fix calculation of achieved task distance (making it total planned task
+    distance minus task distance remaining), thus fixing the values "Speed
+    task achieved", "Speed task average", and "Speed task last hour" shown in
+    infoboxes and/or Status-Task panel fields.
 * ui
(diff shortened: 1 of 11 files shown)


ericcarden:
Please don't review this yet. I just came upon a problem I think is probably due to an error in my code, and I'm trying to find and fix it.

lordfolken:
Moving to v7.45 for more time.

lordfolken:
@ericcarden Bumping this to the **v7.46** milestone for more time to finish the fix and validation.

When you had a chance to look at the issue you spotted in your own branch, did you find the root cause? No rush on review from our side until you're happy with it — but it would be good to know whether the `planned − remaining` approach is still the direction you want, or if you ended up on a different fix.

PR #1516, opened 2024-07-10
Engine/Task/Stats: clarify comment describing time_remaining_start

ericcarden:
Clarify that time_remaining_start starts at the task start only if the task hasn't been started.

The current comment is accurate only if the task hasn't been started. If the task has been started, time_remaining_start counts from NOW, not from when the task was started. Between the variable's somewhat confusing name and this incorrect comment, I was chasing my tail while working on something that involves time_remaining_start. After studying how it's used in the code, I found the error in the comment description. Hopefully this comment correction/clarification will help future developers avoid some confusion.

I made the revised comment exactly like the comment for time_remaining_now except to change "including" to "excluding".

### DIFF
--- a/src/Engine/Task/Stats/ElementStat.hpp
+++ b/src/Engine/Task/Stats/ElementStat.hpp
@@ -52,8 +52,8 @@ struct ElementStat
   FloatDuration time_remaining_now;
 
   /**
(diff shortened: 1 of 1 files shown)


lordfolken:
merged in 351403ca0ca4a717f409e1ecea0698eed7e788b0

PR #1528, opened 2024-08-08
LocalPath: Android - Use Android/media/ directory for XCSoar's Data

### LINKED ISSUE #992: XCSoarData dir not accessible due to scoped storage policy of Android 11

### LINKED ISSUE #1524: Android 14: not able to copy airspace and waypoint files into the XCsoar folder
On Android 14 it is not possible to copy something into Android/data/org.xcsoar/files because this folder is blocked from Android

 7.42

It worked on Android 13

System information
Android 14, Xiaomi 14

Expected behavior
Would like to be able to copy WP files and Airspace

lordfolken:


lordfolken:
Merged

zinuzoid:
Do we need to provide the migration part?

Imagine the user just upgrade to new APK and suddenly all their config is disappeared?

zinuzoid:
Do we need to provide the migration path?

Imagine the user just upgrade to new APK and suddenly all their config is disappeared?

lordfolken:
This could be indeed a problem for the jet edition. 
XCSoar itself anyway has now new signature keys, so it won't upgrade. Except maybe the F-droid release.

lordfolken:
@zinuzoid i hacked something together. Can you test and put into a final form? https://github.com/lordfolken/XCSoar/commit/7e387c3d13935bfb23836ba260091b5b2e4e5be3

lordfolken:
Lets continue in #1544 @zinuzoid

PR #1529, opened 2024-08-09
Improve readability and fix airspace coloring bugs

kobedegeest:
I believe that with this all the initial bugs with the coloring of the new extended format are fixed. 
But maybe we should add some documentation somewhere about what is used type vs class and how to ignore the type.

Also i find this a logical way of dealing with it but that is just based on my use case.

So to summarize the logic:
If an AY tag is present (which is not other) and that type has its display setting to true the color settings of that airspace type will be used 

If an AY tag is present (which is not other) and that type has its display setting to false the color settings of that airspace class will be used instead allowing to basically ignore certain/all airspace types based on user preferences as the type is not always relevant TMA's for example.

If no AY tag is present or the AY tag is 'other' the airspace class will be used for the colors of the airspace.


lordfolken:
Thx

PR #1536, opened 2024-08-11
Fixed compatibility with ImageMagick v7

yorickreum:
ImageMagick v7 was throwing `WARNING: The convert command is deprecated in IMv7, use "magick"` during compilation, updated the `build/imagemagick.mk` file to obey the new syntax.


yorickreum:
Sadly, seems to cause trouble on Linux since ImageMagick v7 can not be installed via apt (yet).
On macOS with homebrew, this is default by now.

lordfolken:
You could variabelize it and set it depending on target.

yorickreum:
> You could variabelize it and set it depending on target.

Okay.

yorickreum:
I could not reopen this PR for some reason, hence I created https://github.com/XCSoar/XCSoar/pull/1747 instead.

Issue #1461, opened 2024-05-23
No module named 'build.download'


waarmond:
XCSoar versions having the problem
-------------------------------------------------

git master (> 7.42)

System information
------------------

ArchLinux
(but) Python 3.10

Steps to reproduce the behavior
-------------------------------

```
	git submodule update --init --recursive
	make DEBUG=n
```

Expected behavior
-----------------

```
  GET     output/download/boost_1_85_0.tar.bz2
Traceback (most recent call last):
  File "./XCSoar/./build/download.py", line 9, in <module>
    from build.download import download_and_verify
```
vendored boost 1.85 would be downloaded

Actual behavior
---------------

`ModuleNotFoundError: No module named 'build.download'`
no download

Do you have any idea what may have caused this?
-----------------------------------------------

Do you have an idea how to solve the issue?
-------------------------------------------

I've just installed boost 1.85 globally on my system. Would it be possible to add an override switch to https://github.com/XCSoar/XCSoar/blob/master/build/libboost.mk for users which already have the right version at `/usr/include/boost` ?

lordfolken:
Note this isn't just for boost, but also other libraries. Also other things will fail if this doesn´t  work. 

The question is why this is failing on your system:

build/download.py:
https://github.com/XCSoar/XCSoar/blob/94f330bf08eabee603d4caba0e3e793e0bb10341/build/download.py#L6-L7

do you have the file in build/python/download.py ?

acollign:
I faced the same today and found that adding the __init__.py module file solves the issue. See PR#1537.

lordfolken:
fixed in https://github.com/XCSoar/XCSoar/pull/1537/commits/4fc839e9b46c377bf9899f8feea11c1fee4c698d

PR #1537, opened 2024-08-15
fix: python build module is missing __init__.py file

acollign:


acollign:
This fixes the issue reported under #1461. Hope it helps.

PR #1544, opened 2024-08-22
LocalPath: move files from application storage to media storage

lordfolken:
Your PR should:
  * no fixup commits

### DIFF
--- a/src/LocalPath.hpp
+++ b/src/LocalPath.hpp
@@ -5,6 +5,10 @@
 
 #include <tchar.h>
 
+#ifdef ANDROID
+#include <jni.h>  // Required for JNIEnv definition on Android
+#endif
+
 class Path;
 class AllocatedPath;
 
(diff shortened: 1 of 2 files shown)


MaxKellermann:
Not using `std::filesystem` was an explicit decision I made for XCSoar. Like iostreams, it's a badly designed API and the implementation is bloated and slow. Look at how much your executable size grows with this piece of code. Better stay away from `std::filesystem`.
If you want a small and fast implementation for copying a directory recursively, look at https://github.com/CM4all/libcommon/blob/master/src/io/RecursiveCopy.hxx (Linux-only though, but that's all you need here anyway)

src/LocalPath.hpp
```diff
@@ -17,40 +21,34 @@ namespace File {
  *
  * Throws on error.
  */
-void
-InitialiseDataPath();
+void InitialiseDataPath();
```
CazYokoyama, line 24: Why you modify? There is no effect, just cosmetic.

lordfolken: And its wrong code style in top of it.

PR #1545, opened 2024-08-23
More extended airspace fixes

kobedegeest:
some minor changes. In the airspace list have the filter option be listed as Unknown since it is Unknown anywhere else in the app (it is other in the code) 
and have AY Unclassified change to unknown as Unclassified is used for AC field makes it better to set the correct colors and filters for the map display this way.


lordfolken (changes requested):
Please fix the indenting.

lordfolken:
57ab4a6934ebeeecb66740736a104c6fa115cb71

lordfolken:
broken in master too. lets see if 
https://github.com/XCSoar/XCSoar/pull/1563 
passes. then we can merge yours on top.

PR #1549, opened 2024-08-25
Quickmenutouch

lordfolken:


lordfolken:
https://forum.xcsoar.org/viewtopic.php?p=25036#p25036 was mainly the motivation.

Issue #1526, opened 2024-08-07
Plane icon color in thermal assistant


kobedegeest:
XCSoar versions having the problem
-------------------------------------------------
7.43-rc0

XCSoar versions not having the problem
-------------------------------------------------

System information
------------------
Android

Steps to reproduce the behavior
-------------------------------
In settings: Look> screen layout: set dark mode to on
Gauges > Flarm, Other : set thermal assistant to anything but off

Replay a flight (or fly) with a thermal 

Expected behavior
-----------------
Plane icon in the thermal assistant to be easy visible. Since the theraml assistant background is white expect the plane to be black

Actual behavior
---------------
With dark mode on the plane is collord white on a white background and can not be easily spotted during flight.

Do you have any idea what may have caused this?
-----------------------------------------------
In src\Look\ThermalAssistantLook.cpp 
ThermalAssistantLook::Initialise(bool small, bool inverse)
plane_pen.Create(width, inverse ? COLOR_WHITE : COLOR_BLACK);

Probably the dark mode sets this inverse to true causing the plane to be colored white

Do you have an idea how to solve the issue?
-------------------------------------------
Remove the inverse check and always have the color black.

lordfolken:
Well the gauge should be dark in dark mode. But there doesn´t  seem to be any config as of yet.

lordfolken:
Lets do it this way: lets remove the inverse check for now as you suggested.

Then we make a new ticket for gauges in general.
* There should only one gauge implementation serving both purposes.
* The placement of the gauges should be able to avoid infoboxes
* flarm gauge config should be on the gauges setup page
*

PR #1551, opened 2024-08-27
Better thermal and flarm gauge

kobedegeest:
Based on comment in #1526


lordfolken:
Cool Thx. I will review it asap.

lordfolken:
image
I like it.

lordfolken (changes requested):
So its very cleanly done. All good. 
3 Points we should look at:
  - I think we should change current default overlayin the infoboxen. 90% don't change the defaults, we shoudl set it to avoid the infoboxen for example bottom left ib for thermal assistant and bottom right for the traffic gauge. 
  - In the current configuration dialogs the flarm configuration is under Look -> ScreenLayout, we should move  it to the Gauges->Flarm, other page.
  - Please add a NEWS entry so we can merge this.

src/Dialogs/Settings/Panels/GaugesConfigPanel.cpp
```diff
@@ -31,6 +32,36 @@ static constexpr StaticEnumChoice final_glide_bar_display_mode_list[] = {
...
+  { TrafficSettings::GaugeLocation::TOP_RIGHT_AVOID_IB,
+    N_("Top Right (Avoid Infoboxes)") },
+  { TrafficSettings::GaugeLocation::BOTTOM_LEFT_AVOID_IB,
+    N_("Bottom Left (Avoid Infoboxes)") },
+  { TrafficSettings::GaugeLocation::BOTTOM_RIGHT_AVOID_IB,
+    N_("Bottom Right (Avoid Infoboxes)") },
+  { TrafficSettings::GaugeLocation::CENTER_TOP_AVOID_IB,
+    N_("Center Top (Avoid Infoboxes)") },
+  { TrafficSettings::GaugeLocation::CENTER_BOTTOM_AVOID_IB,
+    N_("Center Bottom (Avoid Infoboxes)") },
```
lordfolken, line 61: Can you align the Text strings with the one for the Flarm gauge (case etc.) This would mean less translation strings.

lordfolken: "Top left". Pretty much all the other strings are written in that fashion.

lordfolken (approved):

PR #1552, opened 2024-08-27
Analysis improvment to charts

kobedegeest:


lordfolken (changes requested):
Analysis: Climb graph:
I get why you want to do this, and i see the point. Just the 45 minutes seem a bit arbituary. Many people look at the graphs post flight, so maybe the old data is still relevant? Would there be a possiblity to show full flight / vs last 45 minutes?

Scumi:
I'd like to speak out against cutting the amount of visible thermals from the climb graph. I personally use the full graph during the flight to review weather developments. Looking at the data post-flight is also done by a lot of my club members, as @lordfolken pointed out.
Without knowing how big of a userbase is using this feature to its full extend, I would advise against changing the logic. Make it an option in the (already vast and complex) settings, then it would be fine I guess.

Scumi:
@kobedegeest I agree on the improvement idea in general. If "zooming", x-axis scrolling or zoom toggle would be possible, I would be happy with it. I would not make an additional page as there are quite a few stat pages already. Also for "zooming" or "scrolling": How would that be done? With a stick remote, up/down and right/left is already used. A toggle might be the easiest option.

lordfolken:
> @lordfolken did you test the post flight situation? Based on reading the code i would expect it to work I mean to show the whole flight after landing. As there already was a check for is flying.

I saw it. I'm partially happy with it. 
While you cover the use cases, its another one of those features in XCSoar that you basically only know its there, because you read the source code. 

The question is how do we communicate these 2 different states to the user?

lordfolken:
Thinking about it some more, we have plenty of analysis pages that behave different during / post flight. 
Add a text on the page that explains the behaviour and i think we can take it.

Scumi:
Currently, three pages are display a flight-time based graph (barogram, thermal history, task speed). Introducing a different time frame for just one of these graphs will lead to confusion for users. Also I don't think the approach should be "because some of those behave differently, all of them could". 
Furthermore, the choice of 45 minutes seems arbitrary? Why not 60 (like last hour speed), 90 (cuz I like this number), 120 (reflecting league speed time)? Its in the end a magic number that might work for one person but doesn't for the next. 
While I agree on the basic idea, I disagree making this non-configurable and therefor force it on everyone.

lordfolken:
I'll cherry pick the last commit.

lordfolken:
Closing as stale draft for now.

The **TaskLegRenderer** clipping fix is already on master (08ad2e9c7a) — thanks for that.

The **climb chart sliding window** (last 45 minutes while flying) was discussed but not merged; the in-flight vs post-flight split is a reasonable direction if someone wants to pick it up again. A fresh branch from current `master` with named constants, NEWS entry, and rebased history would be the easiest path — feel free to reopen this PR or start a new one anytime.

PR #1566, opened 2024-09-16
Max. selectable value for crew weight increased to 500.

bennsch:
Max. selectable value for crew weight increased to 500 (both in "Flight Setup" and "Logger Setup").

Reason:
The drop-down selection for "Crew Weight" was limited to a value of 300, regardless of unit for mass. 
If the user has selected "lb" as unit, the max. possible crew weight would be 300 lb (=136 kg), which is easily exceeded if both seats are occupied. 
Most gliders are limited to a max. seat weight of 242 lb, so using 500 as max. value should
cover all use cases (2 * 242 lb = 484 lb). I am not aware of a glider with a useful load of more than 500 lbs anyway.

### DIFF
--- a/src/Dialogs/Settings/Panels/LoggerConfigPanel.cpp
+++ b/src/Dialogs/Settings/Panels/LoggerConfigPanel.cpp
@@ -13,6 +13,7 @@
 #include "Components.hpp"
 #include "BackendComponents.hpp"
 #include "Units/Group.hpp"
+#include "Units/Units.hpp"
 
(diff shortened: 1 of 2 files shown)


bennsch:
Thanks for your suggestion! That's a much better solution.

lordfolken (approved):

lordfolken:
Can you rebase, and drop the first commit please?

bennsch:
Done

lordfolken (approved):

Issue #1575, opened 2024-09-24
Flarm radar hide infoboxes 


Joerg-CH:
XCSoar versions having the problem
-------------------------------------------------
XCSoar v7.44

XCSoar versions not having the problem
-------------------------------------------------
XCSoar v7.42

System information
------------------
XCTouchNav 7", android 11, landscape layout with 10 infoboxes.

Steps to reproduce the behavior
-------------------------------
1. Use your device in landscape layout
2. Define the position of flarm radar in 'Config-System-Gauges-FLARM, Other' as "Bottom right (avoid infoboxes)"
3. Have some FLARM traffic (fly or use it on ground with other gliders nearby) to show the flarm radar.
4. See, that the flarm radar covers the lower right infoboxes.
5. Try different positions of flarm radar. See, that both "left" and "right" have no difference if set to "avoid infoboxes" or not. Only "Center ..." keeps the flarm radar out of the infoboxes, but this is shifted left of center.
6. The (similar feature) thermal assistant can be positioned correctly as expected.

Expected behavior
-----------------
If set to "avoid infoboxes", the flarm radar should not cover the infoboxes both in screenlayout landscape and portrait. It works with device in screenlayout portrait.

Actual behavior
---------------
In screenlayout landscape, the position of flarm radar does not avoid the infoboxes, independant of the setting.

Do you have any idea what may have caused this?
-----------------------------------------------
The calculation of the positions of the flarm radar is done differently to the thermal assistant. There, the calculation use "rc.left = GetMainRect().left" instead of "rc.left" only as in flarm radar. This is needed for landscape layout.
In portrait layout, the important formula is "rc.bottom = GetMainRect().bottom", and this is used both in flarm radar and in thermal assistant.

Do you have an idea how to solve the issue?
-------------------------------------------
Update MainWindow.cpp in:
case TrafficSettings::GaugeLocation::TOP_LEFT_AVOID_IB: add
    rc.left = GetMainRect().left 
for TOP_RIGHT_AVOID_IB add
    rc.right = GetMainRect().right 
for BOTTOM_LEFT_AVOID_IB add
    rc.left = GetMainRect().left
for BOTTOM_RIGHT_AVOID_IB add
   rc.right = GetMainRect().right

Modify calculation of center positions in 
TrafficSettings::GaugeLocation::CENTER_TOP_AVOID_IB: and CENTER_BOTTOM_AVOID_IB:
[...]

PR #1576, opened 2024-09-24
MainWindow.cpp: Improved code for flarm radar positions

Joerg-CH:
Your PR should:
  * no fixup commits

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -417,51 +417,55 @@ MainWindow::ReinitialiseLayout_flarm(PixelRect rc,
     break;
 
   case TrafficSettings::GaugeLocation::CENTER_TOP:
-    rc.left = (rc.left + rc.right) / 2 - width - 1;
+    rc.left = (rc.left + rc.right - width) / 2 - 1;
     rc.right = rc.left + width;
     rc.bottom = rc.top + height;
     break;
 
   case TrafficSettings::GaugeLocation::CENTER_BOTTOM:
-    rc.left = (rc.left + rc.right) / 2 - width - 1;
+    rc.left = (rc.left + rc.right - width) / 2 - 1;
     rc.right = rc.left + width;
     rc.top = rc.bottom - height;
     break;
 
   case TrafficSettings::GaugeLocation::TOP_LEFT_AVOID_IB:
     rc.top = GetMainRect().top;
+    rc.left = GetMainRect().left; 
     rc.right = rc.left + width;
     rc.bottom = rc.top + height;
     break;
 
   case TrafficSettings::GaugeLocation::TOP_RIGHT_AVOID_IB:
     rc.top = GetMainRect().top;
+    rc.right = GetMainRect().right;
     rc.left = rc.right - width;
     rc.bottom = rc.top + height;
     break;
 
   case TrafficSettings::GaugeLocation::BOTTOM_LEFT_AVOID_IB:
     rc.bottom = GetMainRect().bottom;
+    rc.left = GetMainRect().left;
     rc.right = rc.left + width;
     rc.top = rc.bottom - height;
     break;
 
   case TrafficSettings::GaugeLocation::CENTER_TOP_AVOID_IB:
     rc.top = GetMainRect().top;
-    rc.left = (rc.left + rc.right) / 2 - width - 1;
+    rc.left = (GetMainRect().left + GetMainRect().right - width) / 2 - 1;
     rc.right = rc.left + width;
     rc.bottom = rc.top + height;
     break;
 
   case TrafficSettings::GaugeLocation::CENTER_BOTTOM_AVOID_IB:
     rc.bottom = GetMainRect().bottom;
-    rc.left = (rc.left + rc.right) / 2 - width - 1;
+    rc.left = (GetMainRect().left + GetMainRect().right - width) / 2 - 1;
     rc.right = rc.left + width;
     rc.top = rc.bottom - height;
     break;
 
   case TrafficSettings::GaugeLocation::BOTTOM_RIGHT_AVOID_IB:
     rc.bottom = GetMainRect().bottom;
+    rc.right = GetMainRect().right;
     rc.left = rc.right - width;
     rc.top = rc.bottom - height;
     break;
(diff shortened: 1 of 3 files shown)


Joerg-CH:
This will solve issue  #1575

src/MainWindow.cpp
```diff
@@ -430,38 +430,42 @@ MainWindow::ReinitialiseLayout_flarm(PixelRect rc,
...
   case TrafficSettings::GaugeLocation::TOP_RIGHT_AVOID_IB:
     rc.top = GetMainRect().top;
+	rc.right = GetMainRect().right;
     rc.left = rc.right - width;
     rc.bottom = rc.top + height;
     break;
 
   case TrafficSettings::GaugeLocation::BOTTOM_LEFT_AVOID_IB:
     rc.bottom = GetMainRect().bottom;
+	rc.left = GetMainRect().left;
```
lordfolken, line 447: Please remove indent.

src/MainWindow.cpp
```diff
@@ -430,38 +430,42 @@ MainWindow::ReinitialiseLayout_flarm(PixelRect rc,
...
   case TrafficSettings::GaugeLocation::TOP_LEFT_AVOID_IB:
     rc.top = GetMainRect().top;
+	rc.left = GetMainRect().left;
     rc.right = rc.left + width;
     rc.bottom = rc.top + height;
     break;
 
   case TrafficSettings::GaugeLocation::TOP_RIGHT_AVOID_IB:
     rc.top = GetMainRect().top;
+	rc.right = GetMainRect().right;
```
lordfolken, line 440: Please remove indent.

src/MainWindow.cpp
```diff
@@ -430,38 +430,42 @@ MainWindow::ReinitialiseLayout_flarm(PixelRect rc,
 
   case TrafficSettings::GaugeLocation::TOP_LEFT_AVOID_IB:
     rc.top = GetMainRect().top;
+	rc.left = GetMainRect().left;
```
lordfolken, line 433: Please remove indent.

src/MainWindow.cpp
```diff
@@ -430,38 +430,42 @@ MainWindow::ReinitialiseLayout_flarm(PixelRect rc,
...
     rc.bottom = GetMainRect().bottom;
-    rc.left = (rc.left + rc.right) / 2 - width - 1;
+    rc.left = (rc.left + rc.right - width) / 2 - 1;
     rc.right = rc.left + width;
     rc.top = rc.bottom - height;
     break;
 
   case TrafficSettings::GaugeLocation::BOTTOM_RIGHT_AVOID_IB:
     rc.bottom = GetMainRect().bottom;
+	rc.right = GetMainRect().right;
```
lordfolken, line 468: Please fix indent

lordfolken (changes requested):
So just style fixes. 

Please do a git pull --rebase to rebase on current branch so we can avoid the merges.

src/MainWindow.cpp
Joerg-CH: Hope I succeeded now. With "edit in place", every try to indent gave a wrong indentation. With GitHub Desktop, the file edited locally in notepad++ looked fine. Now with I try it with GitHub Dev and the web editor.

src/MainWindow.cpp
Joerg-CH: done.

src/MainWindow.cpp
Joerg-CH: done.

src/MainWindow.cpp
Joerg-CH: done.

Joerg-CH:
Quite difficult to achieve a clean layout, when the editor does not show differences between tab and space. Is there a possibility to show unvisible symbols?

Jörg
schuongf@hotmail.com

lordfolken:
Hmm I tested it, but i don't think it does what i would expect the setting to do:

This is center top (avoid ib)

image

I would expect it to be in the center of the map. 

Testing can be done like this btw:

```
make ./output/UNIX/bin/FeedNMEA
cat test/data/driver/FLARM/pflaf01.nmea | ./output/UNIX/bin/FeedNMEA tcp 4353 
```

Joerg-CH:
You are right, it should be centered in the map window. I did not test unsymmetric IB-layouts, and was happy to find a solution for the (symmetric) landscape layouts. I wonder how to determine the position of the infoboxes, to incorporate this in the calculation of the map window center.

Joerg-CH:
I think I found a solution now. I will update the file MainWindow.cpp tomorrow and commit it to this PR.
image

Joerg-CH:
Latest commit 25e0939 tested with different infobox layouts. Both with "avoid IB" and without. Hope this is OK now.
Test different IB-layouts

lordfolken:
So its good now from a postition point of view. However the size calculation seems to still take the dimentions of GetClientRect() instead of the map, resulting in overlap with the thermal assistant (if configured left/right avoid ib).

lordfolken:
Yes I think thisvis how we should proceed.

Joerg-CH:
I did a mistake: Used the original MainWindow.cpp as basis for the last commit, without the previous changes

lordfolken (approved):

lordfolken:
image
 Was able to test in in VR. Dark mode doesn't really work in Windows it seems.

lordfolken:
@Joerg-CH i cleaned up the git history, added news and added you to authors. 
Does this look ok for you? Then we could merge it.

Joerg-CH:
OK for me. Thanks for cleaning the history!

Joerg-CH:
@lordfolken 
I am a bit lost, because I do not find anymore the resize of the gauges in the code to avoid overlapping the thermal assistant with the flarm radar. I commited these changes after you answered "Yes I think thisvis how we should proceed."
But strange, I do not find the changes also in my branch 'position-of-flarm-radar'.

lordfolken:
They are all in one commit. I added them as fixup to have one package.

lordfolken:
Wait you may be right. Ill check my got.

Joerg-CH:
I have a local copy of the MainWindow.cpp with ALL my modifications integrated, so it would be not much work to add another commit.

lordfolken:
Please do

Issue #1577, opened 2024-09-27
scale InfoBox title and comment font size


bomilkar:
XCSoar version
--------------

7.44

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------

The font size of the title line and the comment line should be adjustable.

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------

Many pilots find it hard to read the text of the title and the comment line of the InfoBoxes.
I'm proposing to scale the font for these relative to the default font size.
I'm NOT proposing to change the size of the title or comment box. Therefore a bigger scale factor might lead to clipped text. So the pilot has to tradeoff font size vs. the amount of clipping.
There is Look->Language_Input->Text_size. But that doesn't change the title or comment text. Even if it did, what suits for the Map area isn't necessarily the right choice for the title or comment text.

PR #1578, opened 2024-09-27
scale InfoBox title and comment font

bomilkar:
This PR implements a solution to Issue #1577

### DIFF
--- a/src/Look/InfoBoxLook.cpp
+++ b/src/Look/InfoBoxLook.cpp
@@ -23,7 +23,7 @@
 
 void
 InfoBoxLook::Initialise(bool _inverse, bool use_colors,
-                        unsigned width)
+                        unsigned width, unsigned scale_title_font)
 {
   inverse = _inverse;
 
@@ -44,7 +44,7 @@ InfoBoxLook::Initialise(bool _inverse, bool use_colors,
   Color border_color = Color(128, 128, 128);
   border_pen.Create(BORDER_WIDTH, border_color);
 
-  ReinitialiseLayout(width);
+  ReinitialiseLayout(width, scale_title_font);
 
   unit_fraction_pen.Create(1, value.fg_color);
 
@@ -60,14 +60,11 @@ InfoBoxLook::Initialise(bool _inverse, bool use_colors,
 }
 
 void
-InfoBoxLook::ReinitialiseLayout(unsigned width)
+InfoBoxLook::ReinitialiseLayout(unsigned width, unsigned scale_title_font)
 {
-  const unsigned max_font_height = Layout::FontScale(12);
-
   FontDescription title_font_d(8);
-  AutoSizeFont(title_font_d, width, _T("0123456789"));
-  if (title_font_d.GetHeight() > max_font_height)
-    title_font_d.SetHeight(max_font_height);
+  AutoSizeFont(title_font_d, (width * scale_title_font) / 100U,
+               _T("12345678901234"));
 
   title_font.Load(title_font_d);
 
--- a/src/Look/Look.cpp
+++ b/src/Look/Look.cpp
@@ -46,9 +46,8 @@ Look::InitialiseConfigured(const UISettings &settings,
   thermal_band.Initialise(dark_mode,
                           cross_section.sky_color);
   trace_history.Initialise(dark_mode);
-  info_box.Initialise(dark_mode,
-                      settings.info_boxes.use_colors,
-                      infobox_width);
+  info_box.Initialise(dark_mode, settings.info_boxes.use_colors, infobox_width,
+                      settings.info_boxes.scale_title_font);
   vario.Initialise(dark_mode,
                    settings.info_boxes.use_colors,
                    infobox_width,
@@ -66,12 +65,12 @@ Look::InitialiseConfigured(const UISettings &settings,
 }
 
 void
-Look::ReinitialiseLayout(unsigned infobox_width)
+Look::ReinitialiseLayout(unsigned infobox_width, unsigned scale_title_font)
 {
   /* dialog fonts have an upper bound depending on the window size,
      and thus they might need to be reloaded */
   dialog.LoadFonts();
 
-  info_box.ReinitialiseLayout(infobox_width);
+  info_box.ReinitialiseLayout(infobox_width, scale_title_font);
   vario.ReinitialiseLayout(infobox_width);
 }
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -15,6 +15,7 @@ Version 7.44 - not yet released
   - add new gesture ULDR to access quick menu
   - add quick menu page buttons in case of touch screen
   - add quit button to quick menu
+  - infoboxen: new setting, allow scaling of title/comment font
 * Android
   - fix crash on startup when loading icons on ldpi screens
 
--- a/src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp
+++ b/src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp
@@ -35,6 +35,7 @@ enum ControlIndex {
   MapOrientation,
   DarkMode,
   AppInfoBoxGeom,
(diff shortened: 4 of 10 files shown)


bomilkar:
Can someone help me, please:
Why is Codacy choking on this:
 src/InfoBoxes/InfoBoxSettings.hpp#L116
struct member 'InfoBoxSettings::scale_title_font' is never used.

The struct member is being used in other files. How can I help Codacy to see this?

lordfolken:
Ignore it. The cpplinter isnt good enough to recognize this.

bomilkar:
> Ignore it. The cpplinter isnt good enough to recognize this.

Thanks! I assume the "macos-12, OSX64" issue is a false negative, too.

But I'd like to understand: The struct has ~100 elements and it ONLY complains on the one element which I added.

src/Look/InfoBoxLook.cpp
```diff
@@ -60,14 +60,19 @@ InfoBoxLook::Initialise(bool _inverse, bool use_colors,
 }
 
 void
-InfoBoxLook::ReinitialiseLayout(unsigned width)
+InfoBoxLook::ReinitialiseLayout(unsigned width, unsigned scale_title_font)
 {
-  const unsigned max_font_height = Layout::FontScale(12);
+  // const unsigned max_font_height = Layout::FontScale(12);
```
lordfolken, line 65: Please remove this line, as its no longer needed.

src/Look/InfoBoxLook.cpp
```diff
@@ -60,14 +60,19 @@ InfoBoxLook::Initialise(bool _inverse, bool use_colors,
...
 
   FontDescription title_font_d(8);
-  AutoSizeFont(title_font_d, width, _T("0123456789"));
+  AutoSizeFont(title_font_d,
+    (width * scale_title_font) / 100U,
```
lordfolken, line 69: Where do you get this 100U from?

src/Look/InfoBoxLook.cpp
```diff
@@ -60,14 +60,19 @@ InfoBoxLook::Initialise(bool _inverse, bool use_colors,
...
+    _T("123456789012345"));
+
+  /*****
   if (title_font_d.GetHeight() > max_font_height)
     title_font_d.SetHeight(max_font_height);
+  *****/
```
lordfolken, line 75: Please remove.

src/Look/InfoBoxLook.cpp
```diff
@@ -60,14 +60,19 @@ InfoBoxLook::Initialise(bool _inverse, bool use_colors,
...
+InfoBoxLook::ReinitialiseLayout(unsigned width, unsigned scale_title_font)
 {
-  const unsigned max_font_height = Layout::FontScale(12);
+  // const unsigned max_font_height = Layout::FontScale(12);
 
   FontDescription title_font_d(8);
-  AutoSizeFont(title_font_d, width, _T("0123456789"));
+  AutoSizeFont(title_font_d,
+    (width * scale_title_font) / 100U,
+    _T("123456789012345"));
```
lordfolken, line 70: I think you can limit this to 14Chars. This will slightly increase the title size, and it even in current default you wouldn't see more than 14 Chars.

src/Look/Look.cpp
```diff
@@ -66,12 +66,12 @@ Look::InitialiseConfigured(const UISettings &settings,
...
 void
-Look::ReinitialiseLayout(unsigned infobox_width)
+Look::ReinitialiseLayout(unsigned infobox_width, unsigned scale_title_font)
 {
   /* dialog fonts have an upper bound depending on the window size,
      and thus they might need to be reloaded */
   dialog.LoadFonts();
 
-  info_box.ReinitialiseLayout(infobox_width);
+  info_box.ReinitialiseLayout(infobox_width,scale_title_font);
```
lordfolken, line 75: space between the arguments please

lordfolken (changes requested):

src/Look/InfoBoxLook.cpp
lordfolken: Thx for clarifying. Multiline comment should be with /* */

lordfolken:
Just some style corrections, squashing the commits, adding news, and rewording commit.

lordfolken:
This seems to be true on android. Before it was slightly larger on the Unix target.

bomilkar:
The scaling happens in src/Look/InfoBoxLook.cpp line 69. How can this simple calculation be target-dependant?
I made a new build right after the merge and it worked as expected on ANDROID, UNIX and WIN64. 100% is the default and results in same size as before,

lordfolken:
New

Screenshot_20241007-184946

Previous
Screenshot_20241007-184930

bomilkar:
There was an additional change: from 15 chars to 14 chars.
And the whole InfoBox seem smaller.

bomilkar:
new
old
This is on WIN64: no difference. UNIX is the same, too. ???

bomilkar:
> I just updated my build and noticed this change. To me it feels like the 150% setting is the size i had before. So not sure if this achieves your original intend, as 150% is the max setting. Also i feel like 100% should match the previous behavior as to not confuse users when updating (or make 150 default). The new default was noticeably smaller. (15 infoboxes 3 rows, portrait mode is what i use).

I can't reproduce the problems you report. Several of my club mates can't either. So it's either your phone or your device.
If you would like to try my build:
https://xcs.fliegerclub-moosburg.de/Apps/XCSoar-debug_7.44.2024-10-07.apk
It was cloned on October 7.

lordfolken:
Its a DPI issue. (Also apperant in the linked apk)
Phones have 450 dpi.

This divison doesent scale linearly with the dpi depth. 
https://github.com/bomilkar/XCSoar/blob/fee17f00ca632125059a2c8cbbea7c604812b8af/src/Look/InfoBoxLook.cpp#L66

One option would be to use FastScale(). But i noticed thst thid function seems to only be calculated once at startup.

bomilkar:
There are 2 distinct issues:
1. what [kobedegeest](https://github.com/kobedegeest) reported: "To me it feels like the 150% setting is the size i had before."
2. the scaling factor on ANDROID

I don't think anybody else can reproduce the first issue.

Regarding issue 2: Why would there be a different result than the previous, non-scaled version if `scale_title_font / 100 == 1`  ? Isn't this what we see?
The scaling may not be linear. I don't see this as a problem as long as it's monoton and increasing with input values. Maybe the range for `scale_title_font `should be a bit wider (for ANDROID) to acheive the goal: readability .

lordfolken:
So i tested a  debug build outputing all the values: 

10-09 21:41:42.127  7402  7420 I XCSoar  : Width: 270
10-09 21:41:42.128  7402  7420 I XCSoar  : scale_title_font: 100
10-09 21:41:42.129  7402  7420 I XCSoar  : width_passed 270

10-09 21:47:11.820  7588  7607 I XCSoar  : Width: 270
10-09 21:47:11.825  7588  7607 I XCSoar  : scale_title_font: 145
10-09 21:47:11.829  7588  7607 I XCSoar  : width_passed 391

The code looks like it works as implemented.

When I try builds before the commit, there are only 11 characters shown in the title on android. Meaning that the initial value is already larger than the 100%. 

-  const unsigned max_font_height = Layout::FontScale(12);
-
   FontDescription title_font_d(8);
-  AutoSizeFont(title_font_d, width, _T("0123456789"));
-  if (title_font_d.GetHeight() > max_font_height)
-    title_font_d.SetHeight(max_font_height);

So previously the font size calculation was calculated to 10 chars only, and we only saw more chars, because of the max_font_height clamping.

lordfolken:
i made a pr after trying different things. reducing the calculation to 11 chars actually pretty much restores  the behaviour as it was. 
#1588

bomilkar:
Then we should change the range for the scale factor: 50 ... 150 instead of 80 ... 150
Line 188 src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp

Issue #1084, opened 2022-12-08
Add support for Argentinean 95% distance rule


andrestotorica:
XCSoar version
--------------
7.28

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------
[Argentinean federation contest rules](https://www.favav.com.ar/img/deportiva/Reglamento%20general%20agosto%202018.pdf) (item 3.7.2.5.c) have a **95% distance rule**. 

* In an AAT task, if a pilot flies more than 95% of the task maximum distance, then the computed task time is the actual flown time, even if lower than the task time.

In other words, if you fly close to the max distance and you are still early, you are not penalized, the actual flown time is used.

Today, pilots have to calculate the 95% distance mentally, and have no cues whether the projected targets are within it.

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------
Feedback is welcome as to which helpers for the 95% rule would be more useful.

* Add option in settings (Configuration > Glide Computer > Route?) so users in Argentina can selectively enable the features.
* **AAT Max distance** infobox show the 95% of max distance in the secondary field
* **AAT distance around target** infobox show which percentaje of the max distance is projected, in the secondary field
* Maybe an arc on the map, showing the 95% of max distance in the last area? (needs more thinking)

I will get started on the first ones while ideas are being refined.

MaxKellermann:
Please upload task files and IGC files describing various outcomes (less than 95%, between 95% and 99%, 100%) - and say what the correct calculation results would be. That way, we can develop the feature and verify the correctness of the implementation.

MaxKellermann:
No reply

andrestotorica:
I somehow missed the last messages, sorry.
I planned to try an implementation as it was suggested by various pilots during a competition last summer, but ultimately I got sidetracked. I can retake working on this if there are no objections.

MaxKellermann:
There are never objections to people who want to improve XCSoar. Your work should contain a unit test so we can reproduce the results and can ensure it continues to work.

lordfolken:
@andrestotorica i see that you did work on this. Can you make a PR?

andrestotorica:
Hi @lordfolken, I did progress on the implementation but then found a deeper problem with AAT task distances that I couldn't solve, and eventually got stuck. I'll explain the issue below. If anyone more experienced can give me a hand with some guidance, I will try to get it completed.

___

### Issue
[...]

PR #1579, opened 2024-10-01
Support for Argentinean 95% distance rule & task maximum distance fix

andrestotorica:
Implementation of feature described in issue #1084 - added comments on two infoboxes as helpers to the AAT target distance being flown, optionally enabled in the profile.

This required refactoring the maximum distance calculation algorithm for re-usability and fixing a minor bug in the distances displayed in the Task Manager ui - see discussion on the issue for more details.

Testing done
-------------
* Added test assertions for the initially calculated min and max distance. Ensured these run ok at every refactor step.
* Added assertions checking min, max & max total distances calculated after flying the task. However these were not committed as test is too heavy to run in the `make check` test suite, and there are few tests presently failing in `make testfast` and `make testslow`. I will fix these and add the test that was left out later (probably inside `test_aat.cpp`)
* Ran AAT tasks in Windows and Android on simulator mode, verified infoboxes run as intended
* On Android, using `adb` and `top`, checked there is no measurable change on CPU usage wrt master due to the added calculations, at all points during the task - using a rather old Moto Z Play phone.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -18,6 +18,7 @@ Version 7.44 - not yet released
   - infoboxen: new setting, allow scaling of title/comment font
   - Radar / Thermalassistant Gauges: Placement to avoid InfoBoxen and Overlap
   - Add new Infobox “Home altitude difference” (Home AltD)
+  - add helpers on infoboxes for Argentinean contests 95% distance rule
 * Android
   - fix crash on startup when loading icons on ldpi screens
   - update 'white list' of USB devices with one more VID/PID pair for SoftRF Academy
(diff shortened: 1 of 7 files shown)


lordfolken (approved):

lordfolken:
So i merged everything except the last commit which you marked as WIP.

andrestotorica:
Please ignore comment in last post, it was wrong. I don't know exactly what happened, but seems my last 2 commits got squished and somehow incorrectly merged with some unintended changes. 

Now I've rebased and updated the branch on this PR with the 2 commits I intended. It should be ready to push now.

lordfolken (approved):

lordfolken:
Thank you.

PR #1580, opened 2024-10-01
Use a shared folder for ccache when running inside Docker

andrestotorica:
With this env variable set, ccache calls from within Docker containers will work seamlessly, sharing a common db on the repo checkout folder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ build/local-config.mk
 .*.swp
 
 .idea
+.ccache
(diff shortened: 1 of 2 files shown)


lordfolken:
Can you please update the documenation in ./ide/docker/readme.md accordingly?

lordfolken (approved):

andrestotorica:
Sure thing, thanks for the suggestion.

PR #1593, opened 2024-10-14
Update map_town.svg

Sundown3867:
Preparation for some map work


lordfolken:
So the usual, please make better git commit messages than "Updated XXX". 
Is this WIP or final?

Sundown3867:
Yes, it is a final version - only a small change for an icon which is AFAIK not used yet in XCS generated maps, but I would like to

PR #1601, opened 2024-10-22
xcsoar.pot: created new with xgettext as a source file for weblate

Joerg-CH:
File created new with xgettext, to reflect the latest commits in XCSoar/master.
To reduce irrelevant strings, the filelist to search from was extracted from the recent xcsoar.pot (2023-12-21).
The output file was then edited to reduce the number of strings to translate from 2832 to 1973 (old xcsoar.pot contained 1956 strings).
I hope that, after merging this new file into XCSoar/master, weblate will find the new source file automatically, and the translation process could resume.


Joerg-CH:
I made a mistake. Instead of only one commit (xcsoar.pot), I made the PR with a lot of internal and test commits.
How can I undo this?

PR #1602, opened 2024-10-23
xcsoar.pot: created new with xgettext as source for weblate

Joerg-CH:
This is the updated file as source for translations in weblate.   
The number of strings is now 1999 (previous was 1956). 

I assume that weblate will automatically merge this file into the weblate-repository

### DIFF
--- a/po/xcsoar.pot
+++ b/po/xcsoar.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-12-21 15:38+0100\n"
+"POT-Creation-Date: 2024-10-23 23:40+0200\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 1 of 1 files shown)


lordfolken (commented):
What is the comand for generating this file? So this can be automated.

Joerg-CH:
afaik there is a file (XCSoar/build/gettext.mk) which could be used for automation of the process. But I did not succeed to run it.
So, I used 'xgettext' with some of the parameters also given in 'gettext.mk' to extract the strings from the *.cpp files in /src and all it's subfolders.
The other part of strings, from /Data/Input/default.xci, I took from the old 'xcsoar.pot', and checked for changed line numbers.
Then I checked the result to be as far as possible similar in form and sequence with the old 'xcsoar.pot', to reduce lines added and lines deleted. This was the most "manual" work, but I found also some errors like double quotes at line end.
Now, writing how I added the strings from 'default.xci', I realized I did not check the file for new entries. (last string was line 1215, but 'default.xci' now has 1290 lines. So I will need to update the 'xcsoar.pot' with another commit (will follow).

Do you know how to make an automation with the above mentioned /build/gettext.mk? Of course, this would make it far easier than what I did, and could perhaps be integrated in regular processes.

lordfolken:
make update-po  seems to generally do this. The question is how to trigger and automate it. -> new issue.

lordfolken:
For the xci there seems to be: 
$(OUT)/po/event.pot: $(GETTEXT_EVENTS) | $(OUT)/po/dirstamp
	@$(NQ)echo "  GEN     $@"
	$(Q)$(PERL) $(topdir)/tools/xci2po.pl $^ >$@.tmp
	$(Q)mv $@.tmp $@

Joerg-CH:
With the workflow "build-translation.yml", the file xcsoar.pot can be updated automatically, and every time in the exact same manner. Therefore this PR is obsolete now, and I will close it. To introduce the workflow, I will create a new PR.

PR #1607, opened 2024-10-30
IGC/IGCParser|Writer: Use WGS84 Altitude for IGC B-Records

### LINKED ISSUE #1605: IGC file GNSS altitude is orthometric/MSL altitude, but should be WGS84 ellipsoid altitude
XcSoar version 7.42 Android.

According to the IGC flight recorder specification, the IGC GNSS altitude needs to be the WGS84 ellipsoid altitude.  XcSoar uses the orthometric altitude instead.

_CHAPTER 3
IGC FLIGHT DATA FILES - PRINCIPLES
Detail on the format of the IGC file is in Appendix A, this chapter contains the main principles
3.1 Earth Model. Lat/long and GNSS altitude figures must be with respect to the WGS 84 ellipsoid (see WGS84 in the
Glossary)._

Given an NMEA GGA sentence:

$GPGGA,152145.00,4537.06717,N,07438.94746,W,1,07,1.25,47.4,M,-33.4,M,,*55
The orthometric/MSL altitude is 47.4 meters, and the Geoid separation is -33.4 meters
Therefore the ellipsoid altitude is 47.4 + -33.4 = 14.0 meters

The IGC record from XcSoar (2024-10-05-XCS-0NB-01.igc) is:
B1608184537038N07438599WA-0030**00049**01512
the GNSS altitude is 49 meters  - this is incorrect, this value is the orthometric/MSL altitude, not the ellipsoid altitude.

https://en.wikipedia.org/wiki/Orthometric_height

h = H + N

ellipsoid height  h
Geoid height N
orthometric height H

Given the above GGA sentence, the ellipsoid altitude should be calculated as:

IGC_GNSS_altitude = 47.4 + -33.4 = 14.0

(NOT as IGC_GNSS_altitude = 47.4)

A review of a many files from PowerFlarm, and LXnav showing the use of the proper ellipsoid altitude confirm this problem.

The IGC record just before take-off for the PowerFlarm (4A5Gxxx1.IGC) is:
B1608254537036N07438610WA-0042**00019**01107
the GNSS altitude is 19 meters, this is the correct ellipsoid altitude.

The IGC record from the LXnav Nano (4A5Vxxx1.igc) is:
B1608394537027N07438780WA-0025**00014**00700109314263
the GNSS altitude is 14 meters, this is the correct ellipsoid altitude.

lordfolken:


nbonniere:
In src/IGC/IGCParser.cpp, the reverse calculation should be done to subtract the Geoid and not add it.

moshe-braner:
Also it would be more efficient to not call IGCParseLocation(buffer + 7, fix.location) twice (in the IGCParser.cpp part of the commit).

moshe-braner:
When the GNSS module does report the geoid height in the GGA sentence, as most of them do, then it may be more accurate and efficient to use that rather than to look up the EGM96 approximation, as is done in this proposed commit.  Who knows how the GNSS module gets that number, maybe its own lookup table.  But what is certain is that adding the MSL altitude and the geoid height from the GGA sentence recreates the exact WGS84 ellipsoid altitude which was computed inside the module from the signals from the GNSS satellites.  And that is what should be output to the IGC file.

moshe-braner:
Proposed a better fix: https://github.com/XCSoar/XCSoar/pull/1610

moshe-braner:
Hi Philipp.  It appears that you're running a build, presumably of my
new PR?  If you give us access to the resulting binary, I and
nbonierre would try and test it.  Thanks.  - Moshe

lordfolken:
Yes you can download the build from the github actions page.

Issue #1603, opened 2024-10-24
autoupdate pot files for translations


lordfolken:
I found this: 
https://github.com/tarantool/doc/commit/1040a4f5c6061ea3dfa426a82746fcc81f67786e

This would autocommit the pot file once a merge from a PR is done. 

@Joerg-CH

lordfolken:
i did some tests in my branch https://github.com/lordfolken/XCSoar/actions/workflows/build-translation.yml. The problem i'm having is that it updates the timestamp even if there is no change in the pot or po file.

Joerg-CH:
I made some modifications in the workflow, but need more testing
https://github.com/Joerg-CH/XCSoar/blob/auto-translate/.github/workflows/build-translation.yml

lordfolken:
The sed you cold do sed 's/sed 's/\"POT-Creation-Date: .*//'  That way the line would be deleted to the end. '11d' is positional and could break when there is a change in the line order or format of the file.  Check the pattern with https://regexp101.com

Joerg-CH:
That's a good hint, thanks!
But what I am more concerned of is whether the 'if .. else .. fi' is working correctly when there is a difference between .pox1 and .pox2.


Von: Philipp Wollschlegel ***@***.***>
Gesendet: Freitag, 25. Oktober 2024 23:27
An: XCSoar/XCSoar ***@***.***>
Cc: Jörg Schuon ***@***.***>; Mention ***@***.***>
Betreff: Re: [XCSoar/XCSoar] autoupdate pot files for translations (Issue #1603)


The sed you cold do sed 's/sed 's/"POT-Creation-Date: .*//' That way the line would be deleted to the end. '11d' is positional and could break when there is a change in the line order or format of the file. Check the pattern with https://regexp101.com<https://regexp101.com/>

-

Joerg-CH:
I did some modifications, and now it seems to work:
- In case of no difference found, the old file (with old timestamp) stays in the po-folder. The workflow ended with success.
- In case of a differece, the xcsoar.pot is updated to the latest version, and a new timestamp set. 

But the process is then completed with an exit code 127, causing the workflow run fail.
(see https://github.com/Joerg-CH/XCSoar/actions/runs/11532607588/job/32104588888):
">Run if diff -q po/xcsoar.pot po/xcsoar.pox; then
Files po/xcsoar.pot and po/xcsoar.pox differ
Differences found, run normal update-po
[...]

PR #1608, opened 2024-11-04
Create build-translation.yml: autocommit of translation sources

Joerg-CH:
This workflow checks if there are new strings to translate. If so, it provides an update of the translation template 'xcsoar.pot' and the language files *.po in the folder XCSoar/po, which then can be pulled from weblate. New translations are later pushed from weblate into the *.po files in XCSoar/po.
This PR will solve the issues #1602 and #1603.


lordfolken (approved):
Lets try.

PR #1610, opened 2024-11-08
Handle ellipsoid and geoid altitudes

moshe-braner:
Proposed corrected and improved fix for issue #1605:
 * add ellipsoid altitude field to NMEAInfo & IGCFix structs
* fill in the ellipsoid altitude based on the GGA sentence
* output the ellipsoid altitude to the IGC file
* convert B-record altitude to AMSL
* only look up EGM96 approximation when necessary

Appears as 3 commits due to limitations of github web interface - feel free to "squash merge".

### DIFF
--- a/src/Replay/IgcReplay.cpp
+++ b/src/Replay/IgcReplay.cpp
@@ -67,6 +67,7 @@ IgcReplay::Update(NMEAInfo &basic)
 
   if (fix.gps_altitude != 0) {
     basic.gps_altitude = fix.gps_altitude;
+    basic.gps_ellipsoid_altitude = fix.gps_ellipsoid_altitude;
     basic.gps_altitude_available.Update(basic.clock);
   } else
     basic.gps_altitude_available.Clear();
--- a/src/Device/Parser.cpp
+++ b/src/Device/Parser.cpp
@@ -601,7 +601,7 @@ NMEAParser::GGA(NMEAInputLine &line, NMEAInfo &info)
     info.gps_altitude_available.Clear();
 
   double geoid_separation;
-  if (ReadAltitude(line, geoid_separation)) {
+  if (ReadAltitude(line, geoid_separation) && (geoid_separation != 0)) {
     // No real need to parse this value,
     // but we do assume that no correction is required in this case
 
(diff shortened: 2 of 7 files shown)


src/Replay/IgcReplay.cpp
```diff
@@ -67,6 +67,7 @@ IgcReplay::Update(NMEAInfo &basic)
 
   if (fix.gps_altitude != 0) {
     basic.gps_altitude = fix.gps_altitude;
+    basic.gps_ellipsoid_altitude = fix.gps_ellipsoid_altitude;
```
lordfolken, line 70: I think we should do this during parsing already.

lordfolken (changes requested):
The unit tests need to be fixed `make check`.
commit messages should follow the general format:  filename: commit message

moshe-braner:
"commit messages should follow the general format: filename: commit message"
- do you want one commit per file?

moshe-braner:
reorganized commits and comments into https://github.com/XCSoar/XCSoar/pull/1611

moshe-braner:
I am puzzled by the build failure for some targets (ubuntu) and not others
(android).  The error message is:

src/IGC/IGCParser.cpp:241: error: undefined reference to
'EGM96::LookupSeparation(GeoPoint const&)'

which seems to be a linker, not compiler error?  I did add this line:

#include "Geo/Geoid.hpp"

to src/IGC/IGCParser.cpp, as was done in other files that call that
function (Device/Parser.cpp) and as you added to IGCParser.cpp in PR
#1607.  Does the order of the #include lines matter here?  Thanks.

lordfolken:
Please dont open a new pull request every time. Just fore push to your branch.

moshe-braner:
Assuming you meant "force push", I don't have a local clone, so could not
do that.  Trying to do this purely online in github.  So I had to create a
new branch from master and re-do the commits.

I see that, again, android build succeeded but ubuntu build failed.  Is
that something in my code?

Meanwhile I have realized (thanks to testing by nbonierre on an Android 10
smartphone of the Android binary created from PR #1610) that my code only
handles the case where the GNSS data is derived from NMEA sentences, but
when using the internal device in, e.g., Android, the data is obtained via
OS API calls, and thus my new ellipsoid altitude field is never filled in.
So need more work.  Could simply adjust the altitude just before writing to
the IGC file, as your original PR did, but that throws away the exact
ellipsoid altitude when known from the GGA sentence.  Also, depending on
the device, the original ellipsoid altitude may be available even when NMEA
is not used, e.g., in AndroidSensors.cpp if the device descriptor
"geoid_altitude" is false.

lordfolken:
The project provides both a docker container and a vagrant file under https://github.com/XCSoar/XCSoar/tree/master/ide These you can use to get a local development environment.

PR #1613, opened 2024-11-19
Android: Activity in front of lockscreen

abzicht:
This PR enables XCSoar to be displayed in front of the Android lockscreen.

The idea of this PR is to remove in-flight distractions during phone unlocking (entering a PIN, drawing a pattern, etc.). Instead, XCSoar is immediately available when powering the phone display. Equally, the lockscreen is easily reached by moving the App to the background (e.g., through an upward gesture or pressing the home button). Note that the feature only applies, if the screen was locked while the App was in foreground.

The PR appends two lines to the Android manifest and, as such, includes the feature by default. In many phones, however, the feature must be manually activated over the system's App settings, as illustrated in the attached screenshot (cf. _Show on Lock screen_).

 _(Image credit: [Victor Brandalise](https://victorbrandalise.com/how-to-show-activity-on-lock-screen-instead-of-notification/))_

It must be noted that, in some phones, the feature could be activated by default without the possibility of manual configuration through the system's App settings. If this is not desired, the `showWhenLocked` property in the manifest would have to be replaced by a call to [`setShowWhenLocked(boolean)`](https://developer.android.com/reference/android/app/Activity#setShowWhenLocked(boolean)) from within the main activity, based on an in-app configuration made by the user. I have not yet dived deep enough into the repository to implement this configuration option straight away and I don't know, if there even is community interest for this. For now, this PR hence only adds the feature through the manifest.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -73,6 +73,7 @@ The following people and organizations have contributed code to XCSoar:
  Winfried Simon <winfried.simon@gmail.com>
  Jörg Schuon <schuongf@hotmail.com>
  Ulisse Perusin <uli.peru@gmail.com>
+ Moritz Finke <abzicht@gmail.com>
 
 Documentation:
 
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -21,6 +21,7 @@ Version 7.44 - not yet released
 * Android
   - fix crash on startup when loading icons on ldpi screens
   - update 'white list' of USB devices with one more VID/PID pair for SoftRF Academy
+  - make xcsoar accessible from lock screen
 * Windows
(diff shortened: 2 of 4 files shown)


lordfolken:
Please update both Android manifests, 
I'm going to have to test this...

lordfolken:
This makes a lot of sense. I like it. 

I tested android 6 / 9 / 11 / 14 
Android 6 it of course doesn't work because the call is only supported from api level 27 (Androdi 6 is 23).
This is supported from API Level 27. Presently the minSDK is set to 21.

lordfolken (approved):

PR #1614, opened 2024-11-24
src/Device/Driver/Condor: implement driver for condor3

lordfolken:


JBlyth:
Thank you. That will be great for all those Condor 3 pilots out there!

PR #1632, opened 2024-12-24
Android:GATT fix and improvement

fastdown:
GATT stuff runs now on the main thread, thus the os does not call close() before connection is fully established. Discovered services and characteristics are fetched from os and handled in xcs code, thus consecutive os calls will not fail.

This will fix long standing bugs for me. Tested on Pixel 6, 9 and Xiaomi Redmi 9S with Android 14, 15 and 13.

### DIFF
--- a/android/src/BluetoothSensor.java
+++ b/android/src/BluetoothSensor.java
@@ -4,7 +4,9 @@
 package org.xcsoar;
 
 import java.util.Queue;
+import java.util.UUID;
 import java.util.LinkedList;
+import java.util.List;
 import java.io.IOException;
 
 import android.bluetooth.BluetoothDevice;
@@ -16,6 +18,8 @@ import android.bluetooth.BluetoothGattService;
 import android.bluetooth.BluetoothProfile;
 import android.content.Context;
 import android.os.Build;
+import android.os.Handler;
+import android.os.Looper;
 
 /**
  * Read Bluetooth LE sensor values and report them to a
@@ -28,7 +32,7 @@ public final class BluetoothSensor
   private final SensorListener listener;
   private final SafeDestruct safeDestruct = new SafeDestruct();
 
-  private final BluetoothGatt gatt;
+  private BluetoothGatt gatt;
 
   private int state = STATE_LIMBO;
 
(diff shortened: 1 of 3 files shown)


android/src/BluetoothSensor.java
```diff
@@ -40,14 +44,25 @@ public final class BluetoothSensor
...
     throws IOException
   {
     this.listener = listener;
 
-    if (Build.VERSION.SDK_INT >= 23)
-      gatt = device.connectGatt(context, true, this, BluetoothDevice.TRANSPORT_LE);
+    if (Build.VERSION.SDK_INT >= 23){
+      // Start GATT stuff on main thread. If not,
+      // newer Android os will call close() before
+      // connection is fully established.
```
lordfolken, line 56: Comment format for multiline please with /* */.

android/src/BluetoothSensor.java
```diff
@@ -270,48 +281,40 @@ public void onServicesDiscovered(BluetoothGatt gatt,
...
-        enableNotification(c);
-
-      c = service.getCharacteristic(BluetoothUuids.FLYTEC_SENSBOX_SYSTEM_CHARACTERISTIC);
-      if (c != null)
-        enableNotification(c);
+    // Check if we know the discovered services & characteristics,
+    // if known, enable notification.
+    // Consecutive calls to getServices() will fail, so we do it just once and
+    // handle the lookups in the loops.
+    List<BluetoothGattService> services = gatt.getServices();
```
lordfolken, line 288: This will continuously trigger on all service uuids , whether they are supported by xcsoar or not.  Please filter in a list of known UUIDs before continuing.

android/src/BluetoothSensor.java
lordfolken: "Stuff" isn't really a good description of GATT, please use different wording.

lordfolken (changes requested):
Thank you. Please see comments above.

android/src/BluetoothSensor.java
fastdown: I assumed (backed by observation), that only newly discovered services trigger onServicesDiscovered(). If this is true, the implementation should be okay, right?

fastdown: Added filters for supported services. Now the scanner does only a quick scan for supported services. I am not very happy with the implementation because now the "SERVICE" and "CHARACTERISTIC" from BluetoothUuids.java are mandatory keywords :(
Any suggestions for improvement?

fastdown: I changed BluetoothUuids.java... now I am way more happy with the implementation. The scanner should look only for supported services and after discovering those, the supported characteristics will get notified. Any suggestions?

lordfolken: No I think this is good. Ready for merge from my point of view. Unless you have reservations.

lordfolken (approved):

PR #1637, opened 2025-01-05
UI: Remove pure attribute of NextChar function

### LINKED ISSUE #1635: XCSoar application crashes during start on Debian

First established on 7.40 (as downloaded from project website). Confirmed up until current master branch.

Did not go back in history to find latest working version, sorry.

System information
------------------

Linux host 6.12.6-amd64 # 1 SMP PREEMPT_DYNAMIC Debian 6.12.6-1 (2024-12-21) x86_64 GNU/Linux

No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux trixie/sid
Release:        n/a
Codename:       trixie

description: Motherboard
product: Z97-PRO GAMER
vendor: ASUSTeK COMPUTER INC.

product: Intel(R) Core(TM) i5-4570 CPU @ 3.20GHz

product: GM204 [GeForce GTX 970]
configuration: driver=nvidia latency=0

Install the downloaded XCSoar package and launch the application from the command line: `xcsoar`.

Expected behavior
-----------------

A running application.

Actual behavior
---------------

/usr/include/c++/12/string_view:276: constexpr void std::basic_string_view<_CharT, _Traits>::remove_prefix(size_type) [with _CharT = char; _Traits = std::char_traits<char>; size_type = long unsigned int]: Assertion 'this->_M_len >= __n' failed.
Aborted (core dumped)

After extensive debugging of the src/ui/canvas/Fonts.cpp:NextChar(tstring_view &s) function eventually the problem was found: the function is not pure. 

Annotating it pure seems to cause the GCC optimizer to disregard the pass-by-reference? In the end the string_view passed into NextChar() contains a \0 character, which causes NextUTF8() to return a nullptr in the returned value pair, which causes the remove_prefix() method to try and remove a ridicules amount of characters from the string_view, hence the assert.

Removing the [[gnu::pure]] annotation from NextChar() no longer confuses the optimizer, hence results in a working program.

I intend to submit a pull request for this.

JaapKeuter:
The NextChar function has the GCC pure attribute annotation while it is not. The function changes the string_view type parameter s. This may cause GCC to generate incorrect code, resulting in the crash of the application. Removing the annotation results in non-crashing code.

Fixes #1635

### DIFF
--- a/src/ui/canvas/freetype/Font.cpp
+++ b/src/ui/canvas/freetype/Font.cpp
@@ -75,7 +75,6 @@ FT_CEIL(FT_Long x) noexcept
   return FT_FLOOR(x + 63);
 }
 
-[[gnu::pure]]
 static unsigned
 NextChar(tstring_view &s) noexcept
 {


lordfolken (approved):
Thank you.

PR #1638, opened 2025-01-06
doc/build: Add missing macOS instructions

Turbo87:


lordfolken (changes requested):
Hi Turbo

Do we need to add these here?
https://github.com/XCSoar/XCSoar/blob/b4d66cbbb17670780b2cd3cf81bf48ed38b8d3c7/ide/provisioning/install-darwin-packages.sh#L17

Also if you could mention the provisioning script.

Turbo87:
> Do we need to add these here?

no idea, I've never seen these scripts before... but probably? 😅

Turbo87:
OTOH if these scripts are used by CI and CI is green then maybe not? it's possible that the GitHub Actions image already contains them

lordfolken (approved):

lordfolken:
https://github.com/XCSoar/XCSoar/pull/1639 - I'll try the DMG build. Currently we are just building a bin file. 
https://github.com/XCSoar/XCSoar/blob/b4d66cbbb17670780b2cd3cf81bf48ed38b8d3c7/.github/workflows/build-native.yml#L270

PR #1639, opened 2025-01-06
Macosdmg

lordfolken:


lordfolken:
@Turbo87 can you test the dmg that is built by this github action? (You can download it in the workflow run https://github.com/XCSoar/XCSoar/actions/runs/12638760988?pr=1639) and tell me if it runs on your hardware? I don't have apple stuff...

Turbo87:
Bildschirmfoto 2025-01-06 um 20 48 16

Turbo87:
when I built the dmg locally, I only ran the binary directly and that worked fine. I haven't yet checked whether the locally-built dmg works or not.

lordfolken:
Could you analyze what's wrong with it? 
I found this shell script for createing dmgs: https://github.com/create-dmg/create-dmg that is brew installable. 

Maybe something here is wrong:
https://github.com/XCSoar/XCSoar/blob/273cc7f332972732a379c0aa4430d0e4f58eff27/build/osx.mk#L27-L40

@yorickreum do you have an idea?

yorickreum:
First impression: macOS usually shows this unspecific error when trying to run any unsigned / not trusted application. This can be overruled by right-click and then choose open (instead of just double clicking). Can you try this, @Turbo87?

In this case, there should also be a report in Preferences -> Security & Privacy, showing that there was an attempt to run untrusted software.

If further investigation is needed, I can look into it within the next week. Currently I'm traveling without my MacBook.

lordfolken:
Can you say which build command you use locally?

Turbo87:
just `make TARGET=MACOS dmg -j12`

yorickreum:
I finally found some time to have a look at that. In my opinion, it really is a safety policy induced by Apple, although the fix is slightly different from what I suggested originally. When downloading from the GitHub artifacts, macOS automatically adds the ```com.apple.Quarantine``` attribute (since the source appears unsafe to Apple). You can see this by executing ```xattr XCSoar.app```. After removing the attribute via ```sudo xattr -c XCSoar.app```, the app can be executed normally.

Sidenote: When executing, it currently throws a symbol not found. But that's an independent issue.
```(base) yorickreum@172-11-0-22 Downloads % XCSoar.app/Contents/MacOS/xcsoar 
dyld[32843]: Symbol not found: __ZN3fmt3v116detail10vformat_toERNS1_6bufferIcEENS0_17basic_string_viewIcEENS0_17basic_format_argsINS0_7contextEEENS1_10locale_refE
  Referenced from: <2FEDA221-BB14-352A-ABE2-D03028C2EB87> /Users/yorickreum/Downloads/XCSoar.app/Contents/MacOS/xcsoar
  Expected in:     <127BF9ED-34CF-3E1A-A2B1-FB5818E8C935> /opt/homebrew/Cellar/fmt/11.0.2/lib/libfmt.11.0.2.dylib
zsh: abort      XCSoar.app/Contents/MacOS/xcsoar```

Turbo87:
indeed, `sudo xattr -c XCSoar.app` appears to work 🎉

yorickreum:
So, imho, this is expected behavior. Can be documented somewhere where it’s described how to download the artifact. But not much else that can be done here.

PR #1648, opened 2025-01-19
 Make Flarm gauge disappear when traffic is farther than 4 km away

### LINKED ISSUE #1641: Small Flarm radar still displayed when targets outside 5km radius

7.44

not tested

System information
------------------

UNIX

replay/ use the attached (following) NMEA Log file

[Test2.log](https://github.com/user-attachments/files/18346999/Test2.log)

see also the following screenshots

IMG_1 shows the situation when a new Flarm- object appears in the 5 km range (local time 14:48:07)
IMG_2 shows the situation at time 14:51:51, the Flarm Object is "far away"
IMG_3 shows the situation at 14:52:11 at the big radar display

IMG_1
IMG_2
IMG_3

Expected behavior
-----------------

when no more Flarm- objects within the 5km range the small radar should disappear. The desired behavior is given when switching back from the big radar page to the map page

Actual behavior
---------------

the small Flarm- radar within the map is still displayed alltough all flarm objects are outside the 5 km range

no

possible solution: within the XCSoar/src/Device/Driver/FLARM/StaticParser.cpp should also be a condition to filter traffic outside the near 5km range.

ubx:
Manual updated for en, de and fr. Others pending.

Closes #1641

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -12,6 +12,7 @@ Version 7.44 - not yet released
   - apply dark mode to large flarm radar
   - add options to avoid infoboxes to flarm radar widget
   - changed location of setting for flarm radar position
+  - Flarm gauge: when traffic is farther than 4 km away, it disappears
   - add new gesture ULDR to access quick menu
   - add quick menu page buttons in case of touch screen
   - add quit button to quick menu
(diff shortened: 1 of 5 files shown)


lordfolken (approved):

lordfolken:
Thank you for the clean commit and great PR. :)

PR #1652, opened 2025-01-26
iOS: Fix resizing and rendering

yorickreum:
Without the PR, the app effectively gets stuck on the initial loading/progress screen.
As discovered by @piopawlu, this is mostly due to OpenGLES on iOS not having the default framebuffer indexed with 0, making the unbind fail (see e.g., [https://stackoverflow.com/questions/56786676/how-to-unbind-glbindframebuffer-in-c-opengles-ios](https://stackoverflow.com/questions/56786676/how-to-unbind-glbindframebuffer-in-c-opengles-ios)).

### DIFF
--- a/src/ui/canvas/opengl/FBO.hpp
+++ b/src/ui/canvas/opengl/FBO.hpp
@@ -5,6 +5,14 @@
 
 #include "ui/opengl/SystemExt.hpp"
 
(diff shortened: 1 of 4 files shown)


lordfolken:
It would be cool if you could prefix your commits with IOS: <fix> that way we can identify them.

lordfolken:
Code LGTM.

lordfolken (approved):

PR #1654, opened 2025-01-29
Add pan functionality to the waypoint details image

ubx:


lordfolken:
Way to go. :) I'll review it tonight. Thank you.

lordfolken (changes requested):
So i could get to function for basic functionallity.  Great work!

It segfault when you use the cursor keys 4 or 5 times. 
For the keyboard/joystick functionality: please have a look at the PAN mode. There curor keys pan the map, while function keys f1/f3 zoom in and out.

ubx:
The problem is unrelated to this PR; it also exists on the master branch. Please open a new issue.

src/Dialogs/Waypoint/dlgWaypointDetails.cpp
```diff
@@ -539,26 +578,34 @@ WaypointDetailsWidget::OnShrinkClicked()
...
-    next_button.SetFocus();
-    NextPage(+1);
-    return true;
-
+    case KEY_DOWN:  OnMagnifyClicked(); break;
+    case KEY_UP:    OnShrinkClicked(); break;
+    case KEY_LEFT:  previous_button.SetFocus(); NextPage(-1); break;
+    case KEY_RIGHT: next_button.SetFocus(); NextPage(+1); break;
+    case KEY_ESCAPE:
+    // allow return from first zoom level
```
lordfolken, line 599: Its unforntunalty inconsistent with the pan capability that i pointed out. 

You can zoom in once at the moment and then pan, but no further zoom is possible. 

KEY_ENTER should be used to enter pan capability
KEY_F1 zoom in 
KEY_F3 zoom out
KEY_DIRECTIONAL pan
KEY_ESC exit pan

lordfolken (changes requested):

src/Dialogs/Waypoint/dlgWaypointDetails.cpp
lordfolken: Maybe replace the GOTO button with PAN on that window.

ubx: Implemented as described in the commit message. KEY_ENTER is not necessary:  
If not zoomed in, there is nothing to pan. If zoomed in, panning is enabled until zoomed out.

lordfolken (changes requested):
Thank you for putting in this effort

I found the following:

The - button isn´t  grayed out:
  * after hitting escape to exit zoom

*  When you use the cursor selection to go to + and hit enter, the curor keys are not activated, and hitting up 2 crashes xcsoar with segfault 
* the waypoint command widget buttons are no long reachable via cursor keys. 

* opening a waypoint without an image, crashes xcsoar now as soon as you try to go right. xcsoar: src/ui/window/Window.cpp:30: void Window::AssertThread() const: Assertion `IsDefined()' failed.

ubx:
So, hopefully, the widget now catches all possible focus issues, navigation events, and segmentation faults. Give it another try!

lordfolken:
Still crashes Wien going right too mich.

ubx:
I cannot reproduce the crash. What is your test case?

lordfolken:
I might have not pulled your branch properly. It works now. 

Thank you for this effort. :+1: 

Danggeschöön!

lordfolken (approved):

PR #1659, opened 2025-02-02
iOS: Compatibility with current App Store requirements

yorickreum:
This PR addresses the errors the Apple App Store threw at me while I tried uploading the current version of XCSoar into their TestFlight program.

Additionally, it will allow access to the files (maps, etc.) that XCSoar uses to the user via the Apple Files app, making it possible for users to directly add their own files (as it is common on Android, etc.).

Since I didn't find much documentation about this issue on the internet (or in general about building iOS apps without Xcode), I want to document it here: When creating a .ipa with Xcode, it will set a bunch of values in the .plist file corresponding to the build system used. Since we are building without XCode, those values were not said, resulting in Apple's App Store Connect API returning the error: 

```
Asset validation failed (90534)

Unsupported SDK or Xcode version. Your app was built with an SDK or version of Xcode that isn't supported. Although you can use beta versions of SDKs and Xcode to build and upload apps to App Store Connect, you need to use the latest Release Candidates (RC) for SDKs and Xcode to submit the app.

```

To fix this, I'm now setting all the values Xcode is setting via the ```ios.mk``` file. Namely, those are, `DTSDKName`, `DTSDKBuild`, `DTXcode`, `DTXcodeBuild`, `DTPlatformName`, `DTPlatformVersion`, `DTPlatformBuild`, `DTCompiler`, and `BuildMachineOSBuild`.

I'm also parsing the app icons from an ```Assets.xcassets``` file, since Apple is requiring it for support of newer iOS versions.

With this PR, I was able to make XCSoar available in Apple TestFlight. I further plan to publish XCSoar in the App Store with the release of the next version, 7.44. However, we still need to address a few issues, primarily related to permissions and rendering. Expect some more PRs for that soon.


Data/iOS/Assets.xcassets/AppIcon.appiconset/Icon-1024.png
lordfolken, line 1: Is this file not generated below in the build/ios.mk? I'd like to avoid adding binary data especially if it duplicate a file that is svg.

build/ios.mk
```diff
@@ -86,15 +89,49 @@ $(IOS_GRAPHICS_DIR)/Icon.png: $(IOS_ICON_SVG) | $(IOS_GRAPHICS_DIR)/dirstamp
...
+	$(Q)rsvg-convert $< -w 120 -h 120 -a -o $@
+
+$(IOS_GRAPHICS_DIR)/Icon-152.png: $(IOS_ICON_SVG) | $(IOS_GRAPHICS_DIR)/dirstamp
+	$(Q)rsvg-convert $< -w 152 -h 152 -a -o $@
+
+$(IOS_GRAPHICS_DIR)/Icon-1024.png: $(IOS_ICON_SVG) | $(IOS_GRAPHICS_DIR)/dirstamp
+	$(Q)rsvg-convert $< -w 1024 -h 1024 -a -b white -o $@
+
+$(IOS_GRAPHICS_DIR)/Assets.car: $(topdir)/Data/iOS/Assets.xcassets
+# will also generate $(IOS_GRAPHICS_DIR)/AppIcon%.png, but can't combine implicit and explicit rules
```
lordfolken, line 108: Do you see a chance for deduplication with the Android build?

lordfolken (approved):

yorickreum:
Thanks for merging. I'll look into cleaning the assets up and also deduplicating with Android at some point.
(Some level of duplication is currently needed for iOS, since iOS versions below and above version 10 expect assets at different places in the .ipa. However, there are multiple reasons to only support iOS >= 10, so we can probably clean that up and only use the Assets.car.)

Issue #1643, opened 2025-01-09
Different Flarm target colours in map and radar display


fraktal-sb:
XCSoar versions having the problem
-------------------------------------------------

7.44

XCSoar versions not having the problem
-------------------------------------------------

not tested

System information
------------------

UNIX

Steps to reproduce the behavior
-------------------------------

use the following NMEA log file and switch to the radar and map page

[Test2.log](https://github.com/user-attachments/files/18347165/Test2.log)

Expected behavior
-----------------

Within map and radar page we should display targets with the same colour

Actual behavior
---------------

Flarm target colour is allways "blue" in the radar page.

see the following screenshots

![IMG_1](https://github.com/user-attachments/assets/219f93ea-0fa5-4026-898a-ac8782eb7188)
![IMG_2](https://github.com/user-attachments/assets/c6a8b43a-52cf-4422-a4ad-61f7551e8721)

Do you have any idea what may have caused this?
-----------------------------------------------

no

Do you have an idea how to solve the issue?
-------------------------------------------

use the same logic for colouring the Flarm targets within map and radar page. Because the Flarm ID text is allways related to the selected target this text should use the same logic for colouring.

fraktal-sb:
Maybe this is more a proposal then a bug report. When switching to the radar view the same plane as in the map view is displayed. Therefore (my opinion or proposal) the colour should be the same and all other planes should appear in neutral colour. Selecting an other plane should show it in the colour according distance/warning logic within the map view (same colours as in the details view). But I can also live with the current solution. We can leave this issue open for a while and when nobody is interested close it.

fraktal-sb:
To make more clear what the requirement should be I analysed the program code.

In ./XCSoar/src/Look/TrafficLook.hpp we find the colours used to display targets in the map window.

We find there

static constexpr Color safe_above_color{0x1d,0x9b,0xc5};
static constexpr Color safe_below_color{0x1d,0xc5,0x10};
static constexpr Color warning_color{0xfe,0x84,0x38};
static constexpr Color warning_in_altitude_range_color{0xff,0x00,0xff};
static constexpr Color alarm_color{0xfb,0x35,0x2f};

Within ./XCSoar/src/Renderer/TrafficRenderer.cpp

they are used in canvas.DrawPolygon to draw the traffic in the map window.

within ./XCSoar/src/Gauge/FlarmTrafficWindow.cpp traffic is rendered in

FlarmTrafficWindow::PaintRadarTarget

[...]

PR #1660, opened 2025-02-02
FLARM Target Colours in Radar Display

fraktal-sb:
File:
This commit is based on description and discussion on https://github.com/XCSoar/XCSoar/issues/1643

Within the map FLARM- targets are colour coded. Blue means "above", green means "below" and violet means "same height".

The targets in the radar display (small and big) become now the same colours.

The original colour "blue" in the radar display indicated the selected target (i.e. by tapping on it on the touchscreen). Also the upper left field with the FlarmID or call-sign was displayed within this blue colour. Now this blue colour is no more used.The selected target is visually marked by a bold border line.

The original radar classified the targets into 2 types (passive- with engines) and active (gliders,para-gliders).

The text beside the targets shows height or sink/ascend- rate. It was originally coded in grey colour. Because the used blue and grey colour colour is no more used the text beside the FLARM- target is no more necessary to render with a shadow (it was necessary in "Dark Mode" because grey on a black background is badly visible).

The border of map targets "below" and "above" is originally programmed in /src/Renderer/TrafficRenderer.cpp and is only 50 Meters above and below. This range was now also used within the radar display. This is a bit narrow and should be changed in future.

### DIFF
--- a/src/Gauge/FlarmTrafficWindow.cpp
+++ b/src/Gauge/FlarmTrafficWindow.cpp
@@ -325,22 +325,21 @@ FlarmTrafficWindow::PaintRadarTarget(Canvas &canvas,
         // unnecessary - prevents "may be used uninitialized" compiler warning
         circle_pen = &look.default_pen;
       }
+      // same colours of FLARM targets as in map display
+      text_color = &look.default_color;
+      target_pen = &look.radar_pen;
+      arrow_brush = &look.default_brush;
 
-      if (!small && static_cast<unsigned> (selection) == i) {
-        text_color = &look.selection_color;
-        target_brush = arrow_brush = &look.selection_brush;
-        target_pen = &look.selection_pen;
+      if (traffic.relative_altitude > (const RoughAltitude)50) {
+        target_brush = &look.safe_above_brush;
+      } else if (traffic.relative_altitude > (const RoughAltitude)-50) {
+        target_brush = &look.warning_in_altitude_range_brush;
       } else {
-        if (traffic.IsPassive()) {
-          text_color = &look.passive_color;
-          target_pen = &look.passive_pen;
-          arrow_brush = &look.passive_brush;
-        } else {
-          text_color = &look.default_color;
-          target_pen = &look.default_pen;
-          arrow_brush = &look.default_brush;
-        }
+        target_brush = &look.safe_below_brush;
       }
+
+      if (!small && static_cast<unsigned> (selection) == i)
+        target_pen = &look.default_pen;
     }
     break;
   }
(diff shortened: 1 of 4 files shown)


src/Gauge/FlarmTrafficWindow.cpp
```diff
@@ -325,22 +325,21 @@ FlarmTrafficWindow::PaintRadarTarget(Canvas &canvas,
...
-      if (!small && static_cast<unsigned> (selection) == i) {
-        text_color = &look.selection_color;
-        target_brush = arrow_brush = &look.selection_brush;
-        target_pen = &look.selection_pen;
+      // same colours of FLARM targets as in map display
+      text_color = &look.default_color;
+      target_pen = &look.radar_pen;
+      arrow_brush = &look.default_brush;
+
+      if (traffic.relative_altitude > (const RoughAltitude)50) {
```
lordfolken, line 333: These 50m are the arbituary, or did you base them on some standard? 

https://www.flarm.com/wp-content/uploads/2024/04/FTD-013-FLARM-Compatible-Certification-Specification.pdf 3.2.3 i unfortunalty didn't speicify any altitudes that show same altitude / above or below.

lordfolken (changes requested):
So i like the improvment. Please have a look at my comment concerning the 50m separation. 

The other thing:
Please reword your commits to the following:
 <a one liner what is changed>
(on a new line) the reasoning behind. 

Also the second one seems to be a fixup commit for the first, please mend it in. 

You can use git rebase -i HEAD~2 to interactivily update the commits, then do a git push --force to this branch.

src/Gauge/FlarmTrafficWindow.cpp
fraktal-sb: > These 50m are the arbituary, or did you base them on some standard?

the 50m came from src/Renderer/TrafficRenderer.cpp with first appearence in commit 57ae9ed

fraktal-sb:
Rebase done as above

fraktal-sb:
Tried to explain this on top of the PR:

 

We are agree we have no problem in white mode. Shadowing text on white background with white makes no sense.

 

In dark mode the target text was shadowed with white (because blue text on black background is badly visible)

 

See the image in the screenshot.

 

The not selected targets became a grey color text which is also badly visible on dark mode.

 

Now we have consistent target colours in the small and big radar. The blue colour is no more used for selected target. The target colour now indicates above, below, same height (colours sky-blue, land-green, pink)

 

The text beside te target in white mode is allways black and in dark mode allways white. To render white text in addition with white shadow blows the text unnecessary.

 

Hope this is now clear.

 

Von: kobedegeest ***@***.***> 
Gesendet: Dienstag, 8. April 2025 09:42
An: XCSoar/XCSoar ***@***.***>
Cc: fraktal-sb ***@***.***>; Author ***@***.***>
Betreff: Re: [XCSoar/XCSoar] FLARM Target Colours in Radar Display (PR #1660)

 

@kobedegeest commented on this pull request.

  _____  

In src/Gauge/FlarmTrafficWindow.cpp <https://github.com/XCSoar/XCSoar/pull/1660#discussion_r2032594062> :

> @@ -427,9 +426,6 @@ FlarmTrafficWindow::PaintRadarTarget(Canvas &canvas,
     sc[i].y - int(sz.height / 2),
   };
 
-  // Draw vertical speed shadow
-  RenderShadowedText(canvas, tmp, tp, false);

why did this get removed?

PR #1666, opened 2025-02-07
test: Cleanup UTF8 test

JaapKeuter:
- Order the test length calculation according to tests
- Identify NextUTF8 test block with comment
- Add additional test on NextUTF8 End-of-String behavior
- Encapsulate UnicodeToUTF8 tests in block also


lordfolken (approved):
Interesting. Thank you.

PR #1668, opened 2025-02-08
ANDROID: check for valid battery values, use scale factor

lordfolken:
Your PR should:
  * no fixup commits


iltis42:
image
Just now we have installed this build on one XCTouchNav V7.2, Android 13, that was affected and now there are no more battery low reports, so confirmed that this works!

lordfolken:
Perfect! Thx for the feedback.

PR #1670, opened 2025-02-10
Clang: enforce policy for clang 12

JaapKeuter:
Commit 2c43d6f8 changed policy from clang 10 to clang 12. However, unlike commit 9c968559 and commit 0c061b47, this change was not permeated into other parts of the repository. This commit tries to make things consistent.

With this change we loose building using clang on bullseye (oldstable).


lordfolken:
Thank you for doing this non flashy, but needed maintenance work. Much appreciated.

lordfolken (approved):

Issue #1681, opened 2025-02-13
Save/ Restore FLARM radar zoom range on page change


fraktal-sb:
XCSoar version
--------------

7.44

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------

Since

https://github.com/XCSoar/XCSoar/pull/1677

the automatic Flarm radar zoom is limited to the warning/alarm case. The zoom in the (big) radar always starts with the 5km range (as described).

In case the user chooses an other zoom range this range is reset to the default (again 5km) when switching to an other page and then back to the radar

See the following screenshot sequence.

![Image](https://github.com/user-attachments/assets/847e50ee-4f96-435b-b54e-53c1e904526b)

The zoom range should be saved so that this range is reloaded when switching back to the radar or when XCSoar is restarted.

See the following screenshot sequence.

![Image](https://github.com/user-attachments/assets/0df132dc-3331-44ff-b5b5-490e7c4da829)

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------

This enhancement is already done and will be available by pull request

As test Data the already known filkes can be used

https://github.com/user-attachments/files/18737409/Test15.txt
https://github.com/user-attachments/files/18735162/Test16.txt

PR #1682, opened 2025-02-13
Save/Restore FLARM radar zoom on page change

fraktal-sb:
To save/restore the zoom range over pages and XCSoar restart it was necessary to add a new paramater in the profile. If this paramter is not known in the profile the default 5km is used.

By this enhancement the change between pages becomes more consitent. See the description in

https://github.com/XCSoar/XCSoar/issues/1681

In addition the zoom range on the big radar is set to 5 km when clicking on the small radar on the map. This makes sense because the small radar appears when an object comes into the 4km range. see the following screenshots.

sc1

### DIFF
--- a/src/Gauge/BigTrafficWidget.cpp
+++ b/src/Gauge/BigTrafficWidget.cpp
@@ -33,7 +33,7 @@ protected:
   bool enable_auto_zoom = true, dragging = false;
   bool init_defaults = false;
   unsigned zoom = 3;
-  unsigned last_zoom = 4;
+  unsigned last_zoom;
   static constexpr unsigned num_zoom_options = 5;
   Angle task_direction = Angle::Degrees(-1);
   GestureManager gestures;
@@ -73,6 +73,8 @@ public:
   }
 
   void SetAutoZoom(bool enabled);
+
+  void SaveZoom(unsigned value);
 
   void ToggleAutoZoom() {
     SetAutoZoom(!GetAutoZoom());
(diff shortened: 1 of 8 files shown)


src/Gauge/BigTrafficWidget.cpp
```diff
@@ -135,6 +137,9 @@ FlarmTrafficControl::OnCreate() noexcept
   Profile::GetEnum(ProfileKeys::FlarmSideData, side_display_type);
   enable_auto_zoom = settings.auto_zoom;
   enable_north_up = settings.north_up;
+  if (settings.radar_zoom)
```
lordfolken, line 140: Please check this not just for existence but also for plausible values.

src/Gauge/BigTrafficWidget.cpp
```diff
@@ -176,6 +181,15 @@ FlarmTrafficControl::SetAutoZoom(bool enabled)
   //auto_zoom->SetState(enabled);
 }
 
+void
+FlarmTrafficControl::SaveZoom(unsigned zoom_value)
+{
+  TrafficSettings &settings = CommonInterface::SetUISettings().traffic;
+  settings.radar_zoom = zoom_value;
+  Profile::Set(ProfileKeys::FlarmRadarZoom, zoom_value);
+  //last zoom value in Flarm radar
```
lordfolken, line 190: If you want to document your function, please do it in the same style as the ZoomIn Function.  This can be picked up by editors and presented while coding.

src/Gauge/BigTrafficWidget.cpp
lordfolken: even a nullptr check would be ok.

src/Gauge/BigTrafficWidget.cpp
fraktal-sb: done

PR #1687, opened 2025-02-20
Add "M" menu, zoom in/out buttons as Tophat

CazYokoyama:
For "M" menu button, use the code for Kobo, move it to bottom right, above the status icon.
For zoom in/out buttons, locate them on bottom left, above map scale. They may overlap GPS status icon and text. Thinner frame rectangle as tophat.
Demonstration video: https://www.youtube.com/watch?v=RfnM_tPuCHg

### DIFF
--- a/build/main.mk
+++ b/build/main.mk
@@ -301,7 +301,7 @@ XCSOAR_SOURCES := \
 	$(SRC)/Menu/Glue.cpp \
 	$(SRC)/Menu/ButtonLabel.cpp \
 	$(SRC)/Menu/ExpandMacros.cpp \
-	$(SRC)/Menu/ShowMenuButton.cpp \
+	$(SRC)/Menu/ShowButton.cpp \
 	$(SRC)/Pan.cpp \
 	$(SRC)/Input/InputConfig.cpp \
 	$(SRC)/Input/InputDefaults.cpp \
--- a/src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp
+++ b/src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp
@@ -40,9 +40,7 @@ enum ControlIndex {
   AppStatusMessageAlignment,
   AppInfoBoxColors,
   AppInfoBoxBorder,
-#ifdef KOBO
   ShowMenuButton,
-#endif
 #ifdef DRAW_MOUSE_CURSOR
   CursorSize,
   CursorInverted,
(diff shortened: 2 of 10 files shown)


lordfolken:
So this should depend on HasTouchScreen();
The duplication of the menu button code isn't ideal. I would prefer if you refactor that into one function for all 3 ui elements.
Color and Pen selection should be in src/Look
Does the IFDEF TOPHAT make sense? I would prefer not opening yet another code / testing path. 
Maybe instead of drawn buttons we could use some bitmaps. @Sundown3867 has been contributing good looking and consistent imagery.

lordfolken:
Thank you for trying.

CazYokoyama:
I am refactoring the patch
- Remove TOPHAT_FLABOR
- show/hide the buttons by settings.XXXX such as settings.show_menu_button
and thinking about 3 buttons into a single file.
Drawing buttons has advantage over bitmaps. It does not hide underneath map. Tophat makes M/+/- symbols 1/4 of outer frame and avoids hiding large area of the underneath map. I.e. bitmap M/+/- symbols hide almost 3 of 1cm x 1cm rectangulars of the map.

CazYokoyama:
Update the demonstration video which activate menu and zoom buttons, Navigator Widget.
https://youtu.be/_m4J8JGlj4E

lordfolken (approved):

Sundown3867:
Today I tested it on my 7.44 testbuild and found, that new buttons transfered also to the FLARM traffic screen. There this functionality is not wellcomed, because if dark mode is selected, new buttons are invisible, but still doing teir function -> confusing. I suggest to limit their appearance to the Map screen. Small issue - M button has nice fat border while +/- buttons have somewhat broken outline. Thank you for your help
Screenshot_20250328-145728
Screenshot (28  3

CazYokoyama:
> that new buttons transferred also to the FLARM traffic screen.
will enable only on the Map screen.

> M button has nice fat border while +/- buttons have somewhat broken outline.
The boarder of the M button is drawn by a thicker pen while the one of +/- is by thinner pen. It was intentional because I want to hide the map as little as possible. What do you think?

CazYokoyama:
Are you able to tell me how to show flarm screen? I remember how on Tophat, but can't find on xcsoar.

Sundown3867:
Actually, I found out that it is not easy - I use a special .xci config gesture for that. But if you need testing screen, Menu Info->Page 3/3 -> Thermal Assistant have the same behavior

CazYokoyama:
I re-produce the symptom on thermal assistant screen. Thx.

CazYokoyama:
Sundown3867, please test the code in https://github.com/XCSoar/XCSoar/pull/1727. I only tested on Thermal Assistant screen.

Sundown3867:
Perfect, also on Flarm traffic screen works as expected

PR #1688, opened 2025-02-20
Sprint time update

### LINKED ISSUE #569: Custom sprint trace time

Scumi:
Partially fixes #569 - a customizable sprint time would be preferable, but is currently not needed as the two sprint contests known to me both have a 2 hour window now.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -1,6 +1,7 @@
 Version 7.44 - not yet released
 * calculations
   - update sprint/league time window to 2 hours (as per OLC and DMSt rules)
+  - update OLC League calculation to latest ruleset
 * tracking
   - xcsoar-cloud-service: rebuild service, new domain cloud.xcsoar.org
 * data files
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -1,4 +1,6 @@
 Version 7.44 - not yet released
+* calculations
+  - update sprint/league time window to 2 hours (as per OLC and DMSt rules)
 * tracking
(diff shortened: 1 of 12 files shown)


lordfolken:
Is there a documentation reference for this change? I mean a rule change or similar?

lordfolken:
Commit Messages please in the following format:
```
File or Area: change description
longer explanation
```

Scumi:
I can either add a link to the news from last year (https://www.onlinecontest.org/olc-3.0/segelflugszene/singlenews.html?news=3439) or directly to the ruleset, which has however no history (https://www.onlinecontest.org/olc-3.0/segelflugszene/cms.html?url=rules_overview/b5_de). Means if they change it again and someone looks into the commit, they will be very confused. 

Any preference?

lordfolken:
Id say add the news item as an extended commit message.

And in the code where the rules are implemented, the link to the document.

lordfolken (approved):

lordfolken:
Thank you. Very nice.

PR #1701, opened 2025-02-25
Input/InputConfig.cpp: force letter upper case before event lookup

lordfolken:
Your PR should:
  * no fixup commits

### DIFF
--- a/src/Input/InputConfig.cpp
+++ b/src/Input/InputConfig.cpp
@@ -2,6 +2,7 @@
 // Copyright The XCSoar Project
 
 #include "InputConfig.hpp"
+#include "LogFile.hpp"
 #include "util/Macros.hpp"
 
 #ifdef ENABLE_SDL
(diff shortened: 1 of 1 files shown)


src/Input/InputConfig.cpp
```diff
@@ -45,8 +46,17 @@ InputConfig::GetKeyEvent(unsigned mode, unsigned key_code) const noexcept
 {
   assert(mode < MAX_MODE);
 
+  // Convert lowercase letters to uppercase before lookup
+  if (key_code >= 'a' && key_code <= 'z') {
```
CazYokoyama, line 50: Is there any reason to use if-ToUpperASCII() instead of just one line of toupper()?

CazYokoyama: change the line 49-54 to
 unsigned key_code_idx = toupper(key_code);
How different the resulting effect is?

lordfolken: Thank you for reviewing my code. 

The main thing is that there are 2 different implementations for
ToUpperASCII(char ch) noexcept

and  for windows systems:
src/util/WCharUtil.hxx:ToUpperASCII(wchar_t ch) noexcept

They already come with a filter, so you are correct in that the filter is unnecessary. 
constexpr char
ToUpperASCII(char ch) noexcept
{
	return ch >= 'a' && ch <= 'z'
		? (ch - ('a' - 'A'))
		: ch;
}

Issue #1700, opened 2025-02-25
Safe relative height difference to Flarm targets by colour coding in map and radar page


fraktal-sb:
XCSoar versions having the problem
-------------------------------------------------

since version 7.32 Flarm targets are colour coded within the map display. Blue means above, green below, pink means the target is on the same height. With version 7.44 the radar targets will be colour coded the same way.

XCSoar versions not having the problem
-------------------------------------------------

below 7.32 this functionality did not exist

System information
------------------

Unix and Android

Steps to reproduce the behavior
-------------------------------

use Test data https://github.com/user-attachments/files/18930179/StratuxTest.txt and FeedNMEA program

Expected behavior
-----------------

Because the colour information is an essential information about the relative height difference this information should be within a safe range. Then the pink colour can be seen as a warning. 100m above and below is considered as a safe minimum height difference.

Actual behavior
---------------

The software codes the targets as expected but the range for targets on the same height is to narrow. 

![Image](https://github.com/user-attachments/assets/80113679-d47f-4d5e-95aa-d4e3a2bb1c19)

Do you have any idea what may have caused this?
-----------------------------------------------

The range is hardcoded with 50 meters above and below within src/Gauge/FlarmTrafficWindow.cpp and src/MapWindow/MapWindowTraffic.cpp

Do you have an idea how to solve the issue?
-------------------------------------------

Change the range from 50m to 100m within the code

The software change will be soon ready for implementation with pull request.

bomilkar:
What is the reason for extending the range to +/-100 m? That will lead to lots of false alarms.
Either keep +/-50 or make it a parameter in the setup. Maybe in Glide_Computer -> Safety_Factors.

fraktal-sb:
no further need of this requirement (see https://github.com/XCSoar/XCSoar/pull/1702)

PR #1702, opened 2025-02-26
Larger range (+-100m) for colour coded flarm targets of same flight l…

fraktal-sb:
…evel; see https://github.com/XCSoar/XCSoar/issues/1700

### DIFF
--- a/src/Gauge/FlarmTrafficWindow.cpp
+++ b/src/Gauge/FlarmTrafficWindow.cpp
@@ -330,9 +330,9 @@ FlarmTrafficWindow::PaintRadarTarget(Canvas &canvas,
       target_pen = &look.radar_pen;
       arrow_brush = &look.default_brush;
 
-      if (traffic.relative_altitude > (const RoughAltitude)50) {
+      if (traffic.relative_altitude > (const RoughAltitude)100) {
         target_brush = &look.safe_above_brush;
-      } else if (traffic.relative_altitude > (const RoughAltitude)-50) {
+      } else if (traffic.relative_altitude > (const RoughAltitude)-100) {
         target_brush = &look.warning_in_altitude_range_brush;
       } else {
         target_brush = &look.safe_below_brush;
--- a/src/Renderer/TrafficRenderer.cpp
+++ b/src/Renderer/TrafficRenderer.cpp
@@ -59,9 +59,9 @@ TrafficRenderer::Draw(Canvas &canvas, const TrafficLook &traffic_look,
       canvas.Select(traffic_look.alarm_brush);
       break;
     case FlarmTraffic::AlarmType::NONE:
-      if (traffic.relative_altitude > (const RoughAltitude)50) {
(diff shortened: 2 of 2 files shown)


lordfolken:
This seems to be in line with other manufacturers implementation:
From LXNav's TrafficView Manual:

"If the glider vertical distance is less than 100m (330ft), then this glider will be painted with
near glider colour. Gliders with vertical distances above that, will be painted with above
setting, and below 100m (330ft), they will be painted with below setting."

bomilkar:
Pointing to LXNav is not a good justification.

bomilkar:
It's not safe if too many targets show up.
Also in the spirit of not changing the behavior version to version I suggest to make it a variable and keep the default at 50m. Then everybody has the choice.

lordfolken:
This has nothing to do with filtering targets or alarming. This just paints the target wheter its below, level or above in relation to you.  So far it was 50+/- so basically level was never used.

fraktal-sb:
And now? What is to do? An additional parameter? Leave it at +-50 meters? (for my taste not good) Blue and green directly indicate OK. Pink means "pay attention" not more.

bomilkar:
> And now? What is to do? An additional parameter? Leave it at +-50 meters? (for my taste not good) Blue and green directly indicate OK. Pink means "pay attention" not more.

I don't want to get in the way. I just wanted to add my view.
When I introduced the color coding years ago I didn't pick the 50 out of the blue. I thought about the use model and the value is based on how pilots enter/leave thermals and how they convert speed to altitude and back.
I can help myself: I just apply one more patch when I build it.

fraktal-sb:
My Opinion: we leave this open an give other pilots the chance to enter their opinion and decide then. To make a parameter is very much effort. Lots of places must be changed within the software an most people would not touch security setup parameters. In the sense of traffic awareness I would prefer the hard coded version because nobody can change it, except those knowing how to.

bomilkar:
> My Opinion: we leave this open an give other pilots the chance to enter their opinion and decide then.

Sounds good. Although the 50m is in place since years and I haven't heard any complaints.

> To make a parameter is very much effort.

True! But worth it if we want to support different use models. For instance paragliders might like a smaller value if they are among themself.

src/Gauge/FlarmTrafficWindow.cpp
```diff
@@ -330,9 +330,9 @@ FlarmTrafficWindow::PaintRadarTarget(Canvas &canvas,
       target_pen = &look.radar_pen;
       arrow_brush = &look.default_brush;
 
-      if (traffic.relative_altitude > (const RoughAltitude)50) {
+      if (traffic.relative_altitude > (const RoughAltitude)100) {
```
CazYokoyama, line 333: Please don't use a magic number. Define macro.

fraktal-sb:
we should close this PR. There seems to be no requirement  to change.
We should open a new Issue/PR with a new specification (if necessary).

PR #1704, opened 2025-03-03
Navigation widget

CazYokoyama:
This PR is the follow-up of https://github.com/XCSoar/XCSoar/pull/1064 as

As my previous comment, I divided threaderic's patch/commit into 13 commits.
https://github.com/CazYokoyama/XCSoar/tree/NavigationWidget
The commits have the following commit messages.

1. Allocate top pane/widget which is enabled by config -> system -> look -> pages -> top area.
2. Add height of Navigator pane/Widget in settings.
3. Configure Navigator pane/widget such as its color.
4. Draw NavigatorWindow on the top pane.
5. Draw 2 rectangular-ish frames by DrawPolygon() in top pane.
6. Show progress bar on bottom of the navigatorwidget/top pane.
7. Show waypoint/turn point icons on far left and right.
8. Draw Waypoint informations: distance, altitude, glide ratio
9. Draw direction arrow to the next waypoint.
10. Draw "Next waypoint's name"
11. show "Task informations: average speed" It appears top far left.
12. Draw Current speed / Current Altitude on right.
13. draw elapsed, start, local, planned, planned arrival time at bottom.
14. Note: Each commit have magic numbers such as rc_height * 40 / 200. I guess they are for locating icons and text in appropriate location in the navigator widget, but I have no idea how they are introduced.

Let me know whether I am OK to forced push to https://github.com/XCSoar/XCSoar(I may not be allowed because I am not a author).
His patch include the code which starts analysis_page by key event, i.e. the code in InputEvents::eventAnalysis(). I dropped it because I suspect it may be a garbage included by mistake.

the demonstration video, https://youtu.be/_m4J8JGlj4E

reverse the swipe gestures for Next Turnpoint and Previous Turnpoint. Swipe to Left should be Next and Swipe to Right should be Previous.

### DIFF
--- a/build/liblook.mk
+++ b/build/liblook.mk
@@ -22,6 +22,7 @@ LOOK_SOURCES := \
 	$(SRC)/Look/CrossSectionLook.cpp \
 	$(SRC)/Look/GestureLook.cpp \
 	$(SRC)/Look/HorizonLook.cpp \
+	$(SRC)/Look/NavigatorLook.cpp \
 	$(SRC)/Look/TaskLook.cpp \
 	$(SRC)/Look/TrafficLook.cpp \
 	$(SRC)/Look/InfoBoxLook.cpp \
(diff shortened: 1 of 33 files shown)


CazYokoyama:
pushed the code for #1064

Issue #1697, opened 2025-02-22
iOS: Clicking/touching the map does not work


yorickreum:
XCSoar versions having the problem
-------------------------------------------------

current from master https://github.com/XCSoar/XCSoar/commit/c293d2cdcab6f6e7ce0f72cd343038ea9d9bb34d

XCSoar versions not having the problem
-------------------------------------------------

unknown to me

System information
------------------

iOS 18.3.1 (and others), real device and simulator

Steps to reproduce the behavior
-------------------------------

Launch XCSoar on an iOS device, start it, go into sim or fly mode.

Expected behavior
-----------------

When one clicks anywhere on the map, the "Map elements at this location" dialog should open.

Actual behavior
---------------

Nothing happens, dialog won't open.

Do you have any idea what may have caused this?
-----------------------------------------------

–

Do you have an idea how to solve the issue?
-------------------------------------------

Will look into it; any help is appreciated.

yorickreum:
Touch gestures are working fine, tho.

lordfolken:
Many of these issues look timing related. Perhaps reading this will yield some clues:
https://xcsoar.readthedocs.io/en/latest/architecture.html#threads-and-locking

yorickreum:
> Many of these issues look timing related. Perhaps reading this will yield some clues: https://xcsoar.readthedocs.io/en/latest/architecture.html#threads-and-locking

Thanks for pointing me in the right direction. 

The problem indeed seems to be that `&SteadyNow()` in `src/ui/event/sdl/Queue.hpp` returns always the same timepoint on iOS (on macOS it is updating properly). When I work my way around the caching and return a fresh `std::chrono::steady_clock::now()` with every call, the timers are invoked as expected, and clicking works fine.

Will figure out what precisely is happening here and how to address that, then provide a PR...

yorickreum:
Fixed in https://github.com/XCSoar/XCSoar/pull/1705.

Issue #1696, opened 2025-02-22
iOS: Progress bars are not working


yorickreum:
XCSoar versions having the problem
-------------------------------------------------

current from master https://github.com/XCSoar/XCSoar/commit/c293d2cdcab6f6e7ce0f72cd343038ea9d9bb34d

XCSoar versions not having the problem
-------------------------------------------------

unknown to me

System information
------------------

iOS 18.3.1 (and others), real device and simulator

Steps to reproduce the behavior
-------------------------------

Launch XCSoar an iOS device, go into sim or fly mode, download map or waypoints through the app or wait for terrain loading.

Expected behavior
-----------------

Show the progress bar proceeding during any download or on the upper left of the map while the terrain is loading.

Actual behavior
---------------

Download progress bar is stuck at 0 %, altough the download is proceeding.
Terrain download progress bar just never shows up.

Do you have any idea what may have caused this?
-----------------------------------------------

–

Do you have an idea how to solve the issue?
-------------------------------------------

Will look into it; any help is appreciated.

yorickreum:
Fixed in https://github.com/XCSoar/XCSoar/pull/1705.

Issue #1693, opened 2025-02-22
iOS: Power consumption is intense


yorickreum:
XCSoar versions having the problem
-------------------------------------------------

current from master https://github.com/XCSoar/XCSoar/commit/c293d2cdcab6f6e7ce0f72cd343038ea9d9bb34d

XCSoar versions not having the problem
-------------------------------------------------

unknown to me

System information
------------------

iOS 18.3.1 (and others), real device

Steps to reproduce the behavior
-------------------------------

Start XCSoar on an iOS device, feel the device getting warm and watch the battery getting drained.

Expected behavior
-----------------

Consume a reasonable amount of power.

Actual behavior
---------------

Too much CPU time wasted.

Do you have any idea what may have caused this?
-----------------------------------------------

Not yet.

Do you have an idea how to solve the issue?
-------------------------------------------

I'll try to attach the Xcode Time Profiler and try to figure out what is draining so much computational power here.

yorickreum:
Xcode Time Profiler of iOS simulator says most of the CPU time goes into ```SDL_WaitEventTimeout``` called in ```UI::EventQueue::Wait(UI::Event&)```, about 30 % a few seconds after startup. Might be normal, not sure? Will investigate again on a real device.

yorickreum:
The main issue here might ultimately be OpenGLES-related. Even on the XCSoar start screen, the CPU load caused by the app is about 120 %. Going through the SDL iOS Demo projects, they all show a default CPU load of about 20-30%—with the exception of the Fireworks project, which also uses OpenGLES; there, the initial load is also > 120 %.

TheMiNuS:
On the power consumption one of the main problem is that the app do not run in background when the phone is locked. When you unlock the phone XCSoar restart completely and ask again fly/sim.

This improvement could be great to have the app running in background.

yorickreum:
> This improvement could be great to have the app running in background and save energy in addition to not restart all the nav.

This would indeed be great; the app restarting after the iPhone went to sleep also bothered me quite a bit when flying with it.

> Am I correct in assuming that something like SDL_WaitEventTimeout(&event.event, 100); won’t help either, because SDL_WaitEvent is a synchronous function that blocks until the next event arrives? 

This is my understanding of things as well. `SDL_WaitEventTimeout()` just makes things worse, as it sets a timeout at which it stops waiting, leading to even more loop iterations.

[...]

PR #1705, opened 2025-03-04
SDL UI::EventQueue::Wait Better performance and iOS compatibility

yorickreum:
I changed the following things in UI::EventQueue::Wait for SDL. This might be a little drastic, hear me out:

-  Moved `FlushClockCaches();` out of the infinite loop to the top of the function.
This is similar to src/ui/event/windows/Queue.cpp and src/ui/android/windows/Queue.cpp and should suffice in my opinion. Maybe I'm missing a case in which it is really necessary to do the reset inside the infinite loop, but I can't think of one right now? Putting the `FlushClockCaches();` before the loop already fixes the issue with timers not firing on iOS, see https://github.com/XCSoar/XCSoar/issues/1697 (and the terrain loading progress bar mentioned in https://github.com/XCSoar/XCSoar/issues/1696, download progress bars are an unrelated issue apparently).
- Removed the `::SDL_PumpEvents()` call.
Unnecessary in my opinion, since `SDL_WaitEventTimeout` and `SDL_WaitEvent` both implicitly call ``SDL_PumpEvents``.
- Removed the whole timeout and `SDL_WaitEventTimeout` part. 
Removing it saves a factor between 5 and 10 in CPU time (!) for me and does not seem to cause any issues (timers are still firing as I'd expect, tested briefly on iOS and macOS). If we need to keep it for some reason I'm missing right now, I'd advocate for a solution with `SDL_Delay()`. As it is, imo it's spinning the loop unnecessarily and keeping the CPU from sleep. This should help a lot with https://github.com/XCSoar/XCSoar/issues/1693, simulator profiler results look promising, will test the power consumption tomorrow on an actual device. 

Which other devices actually use the SDL EventQueue, and which features would need to be tested in particular?

Looking forward to yalls feedback.

### DIFF
--- a/src/ui/event/sdl/Queue.cpp
+++ b/src/ui/event/sdl/Queue.cpp
@@ -50,29 +50,19 @@ EventQueue::Pop(Event &event) noexcept
 bool
 EventQueue::Wait(Event &event) noexcept
 {
-  /* this busy loop is ugly, and I wish we could do better than that,
-     but SDL_WaitEvent() is just as bad; however copying this busy
-     loop allows us to plug in more event sources */
-
   if (quit)
     return false;
+  
+  FlushClockCaches();
 
   while (true) {
     if (Generate(event))
       return true;
-
-    ::SDL_PumpEvents();
-
-    const auto timeout = timers.GetTimeout(SteadyNow());
-
-    int result = timeout.count() >= 0
-      ? SDL_WaitEventTimeout(&event.event,
-                             std::chrono::duration_cast<std::chrono::milliseconds>(timeout).count())
-      : SDL_WaitEvent(&event.event);
+    
+    int result = SDL_WaitEvent(&event.event);
+    
     if (result != 0)
       return result > 0;
-
-    FlushClockCaches();
   }
 }
 


yorickreum:
It does not seem to make a significant difference for the performance on my iPhone if the `SDL_WaitEventTimeout`call is there or not (somewhat surprisingly to me, I'm still not sure why it's there). We can also keep it, but I'd like to understand why it's there...

lordfolken:
I found this commit from 2022: 2b48e8bac9c4fa362683512bc78ce14009508333

Author: Max Kellermann <max.kellermann@gmail.com>
Date:   Fri Sep 2 13:21:42 2022 +0200

    ui/event/sdl/Queue: use SDL_WaitEventTimeout() instead of hard-coded sleep
    
    A hard-coded sleep time and periodically waking up to poll for new
    events is a bad idea.  It caused too many CPU wakeups (100 per
    second), drained the battery, and put a lower bound on the UI latency
    of 10ms everywhere.  Since SDL2, there is SDL_WaitEventTimeout() which
    does just what we need; we just give it the duration until the next
    timer expires.

Also interesting: 1630d381261293d6ba470f85babafdae76f371b9 

where the FlushClockCache was added.

yorickreum:
Thanks, @lordfolken. I see. 
I do still believe we can do better than as it is. I will do some more thorough benchmark tests to be sure (i.e., just check how often the busy loop is really running per second for different approaches) and will follow up on.

yorickreum:
I explored this in more detail now, and I'm satisfied with the requested changes.

Overall, removing the unnecessary timeout logic brings down the loop spinning count from 20-30 per second to 10-20 per second after an event was fired. During idle, the loop is not fired at all. So, in fact, I don't think this is the major source of battery drain observed on iPhones. Will dig deeper here. Moving `FlushClockCaches()` is still required to make timers work properly on iOS.

yorickreum:
Is there anything I can still do to get this merged? 

As the flying season is in full swing currently (and folks keep reaching out to me via e-mail asking about it...) I'd like to provide a version of XCSoar based on the official master on the App Store as soon as possible – and having the `FlushClockCaches();` change is crucial for the app to work properly.

lordfolken (approved):

lordfolken:
Yup. We have sufficient dark arts already in this code base. :mage:

Issue #1690, opened 2025-02-22
Stratux Support


fraktal-sb:
XCSoar version
--------------

7.44

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------

FLARM and ADS-B are common traffic awareness protocols in general aviation.
Stratux is an open source project to deliver those data to various EFB's (Electronic Flight Bag's)
Is is described at http://stratux.me/ and https://github.com/stratux/stratux

Because of the NMEA format XCSoar works "nearly" out of the box with but is not usable. The radar page of XCSoar is overloaded with ADS-B targets. see the following screenshots

![Image](https://github.com/user-attachments/assets/da4337d8-d405-487e-a8f9-0230049f192c)

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------

 A configurable filter is missing (compared to the "safesky app" (see https://www.safesky.app)

Then the traffic radar could be seen as follows

![Image](https://github.com/user-attachments/assets/e6ca0531-7560-47a7-9520-1bfcaca8a8aa)

This could be done by defining a new device "Stratux" in the device list and developing a special driver and configuration user interface.

The Stratux device could be made available at the device list page

![Image](https://github.com/user-attachments/assets/2b29cf66-a9f9-470e-ba40-e95d53e7e1bd)

A typical connection could be made by WiFi as follows.

![Image](https://github.com/user-attachments/assets/9c84dff9-b8f0-425b-b8fd-ed2d1ae5c01c)

Then the special setup of the horizontal and vertical filter range could be made as follows by using "Manage"
button.

![Image](https://github.com/user-attachments/assets/f46e698c-4e54-438d-87d5-3b57760fc41c)

![Image](https://github.com/user-attachments/assets/cae16312-2321-45b2-b34b-56a2835fc502)

Device parameters could be stored in the profile. As standard a horizontal range of 20km and vertical range of 2km are proposed which should be available without configuration or available profile parameter.

The implementation will be soon available by pull request.

Test Data can be found here:

[StratuxTest.txt](https://github.com/user-attachments/files/18930179/StratuxTest.txt)

fraktal-sb:
actually the PR build checks fail with:

/usr/bin/i686-w64-mingw32-ld: /__w/XCSoar/XCSoar/src/Device/Driver/Stratux/Driver.cpp:20: undefined reference to `Profile::Get(std::basic_string_view<char, std::char_traits<char> >, char const*)'
[...]

PR #1710, opened 2025-03-07
Stratux support

fraktal-sb:
see https://github.com/XCSoar/XCSoar/issues/1690

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -39,6 +39,7 @@ Version 7.44 - not yet released
   - Add driver for Condor 3
   - Add driver for LX Navigation Eos / Era variometers
   - Flarm: parse flarm error messages (PFLAE) and display them
+  - LXNAV: parse acceleration/g-load from PLXVF sentence
   - LXNAV: Overload formula, use reference mass instead of dry mass
   - OpenVario: Overload formula, use reference mass instead of dry mass
   - Vaulter: use fraction instead of overload for ballast
--- a/src/Device/Driver/LX/Parser.cpp
+++ b/src/Device/Driver/LX/Parser.cpp
@@ -223,14 +223,24 @@ PLXVC(NMEAInputLine &line, DeviceInfo &device,
  *
  * $PLXVF,time ,AccX,AccY,AccZ,Vario,IAS,PressAlt*CS<CR><LF>
  *
- * Example: $PLXVF,1.00,0.87,-0.12,-0.25,90.2,244.3,*CS<CR><LF>
+ * Example: $PLXVF,,1.00,0.87,-0.12,-0.25,90.2,244.3,*CS<CR><LF>
  *
  * @see http://www.xcsoar.org/trac/raw-attachment/ticket/1666/V7%20dataport%20specification%201.97.pdf
  */
 static bool
 PLXVF(NMEAInputLine &line, NMEAInfo &info)
 {
-  line.Skip(4);
+  line.Skip();
+
+  double a[3];
+  bool a_available = line.ReadChecked(a[0]);
+  if (!line.ReadChecked(a[1]))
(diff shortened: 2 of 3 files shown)


fraktal-sb:
see my comment about build checks on https://github.com/XCSoar/XCSoar/issues/1690

fraktal-sb:
can't get rid of these build check errors. I am at the end with my latin.

yorickreum:
> see my comment about build checks on #1690

Regarding `src/Device/Driver/Stratux/Driver.cpp:17: undefined reference to `Profile::Get(std::basic_string_view<char, std::char_traits<char> >, unsigned short&)'`: by checking `nm -C  output/UNIX/dbg/profile.a | grep 'Profile::Get'`you can see that the function is actually there, with the right signature. The problem occurs during linking, ld does not include profile.a. You can fix this by explicitly requiring `PROFILE` in `DRIVER_DEPENDS` in `build/driver.mk`.

Regarding the `src/Device/Driver/Stratux/Driver.cpp:79:(.text+0x50c): undefined reference to `FlarmId::Parse(char const*, char**)'`, this is a little bit more difficult to resolve. The function is defined in `src/FLARM/Id.cpp`, however this is directly included in `main.mk`, hence we can not depend on it as it is. One option would be to move the FLARM sources into their own library/archive.

fraktal-sb:
got the first issue done; to solve the second is much more then I can do with my knowledge; strange that I can compile and link on my Linux Bookworm and generate and run KOBO and Android versions as well

yorickreum:
> got the first issue done; to solve the second is much more then I can do with my knowledge

I was thinking about somethink like https://github.com/fraktal-sb/XCSoar/compare/std006...yorickreum:XCSoar:std006
This would fix the build issues, see https://github.com/yorickreum/XCSoar/actions/runs/13816428734
Not sure if we want FLARM to be its own library or not ... I don't know what the creteria here should be, and if there maybe is a simpler solution.

> strange that I can compile and link on my Linux Bookworm and generate and run KOBO and Android versions as well
Indeed. I don't know why the build process is different atm. Are they building a fat binary there or something like that?

fraktal-sb:
great job! Without your help I could not finish this!

At my (bookworm) system the generated binarys are located on the directorys XCSoar/output/UNIX/bin, XCSoar/output/ANDROID/bin and XCSoar/output/KOBO. Everything is standard. A fat binary is not the case (I think).

I would suggest we leave this PR open for a while and give other ideas a chance.

Then I would open a new issue where we only do the change to the extra FLARM library. This could be done in pre release phase of v7.45. We would have enough time to test it in real environment. After that we could integrate the Stratux Modules in v7.45 or 7.46

yorickreum:
> great job! Without your help I could not finish this!

My pleasure. :)

> I would suggest we leave this PR open for a while and give other ideas a chance.

I think you could also merge/cherry-pick my commits into this PR to make the checks green here; we can then still decide whether the commits should be merged in or this should be done differently. But I'd be in favor of having everything in one place.

lordfolken:
So this looks pretty good. Please no fixup commits. Please integrate the bug fixes where they are needed and merge those commits.

fraktal-sb:
done; in flight test is planned for Wednesday 03.19; result I will post here;

yorickreum:
> done
There a still some bug fix commits which I think you can either remove or squash away.

> in flight test is planned for Wednesday 03.19; result I will post here;  
Cool, jealous (I'm currently too far away from any planes I can fly...) and very much looking forward to the results of this!

fraktal-sb:
inflight test done with default values (20km/2km) on KOBO Glo HD; traffic is filtered as expected; no other issues

Replay made with NMEA log

Comparison with/ without Stratux device filter

test

PR #1713, opened 2025-03-12
iOS/macOS: Updated build instructions, example Xcode project

yorickreum:
To make it easier for folks to get started with XCSoar development on iOS.

## Summary by CodeRabbit

* **New Features**
  * Added build/clean/sign/install scripts and Xcode workspace/scheme to streamline building, signing, installing, and debugging on iOS and macOS.
  * Introduced iOS app version and build number variables and automatic SDK path configuration for macOS ARM64 builds.

* **Documentation**
  * Reorganized build docs: separate iOS/macOS sections, added simulator support and LLDB-based debugging guidance.

* **Chores**
  * Added Xcode-related .gitignore to exclude user/build artifacts.


darwin/build.sh
```diff
@@ -0,0 +1,50 @@
...
+echo "ios/build.sh: Executing gmake for target $TARGET"
+IPA_TARGET="ipa"
+if [ "$TARGET" = "MACOS" ]; then
+  IPA_TARGET=""
+fi
+export PATH=/opt/homebrew/bin:"$PATH" # for gmake to be in path
+gmake -j"$(nproc)" USE_CCACHE=y V=2 OPTIMIZE='-O0' DEBUG=$DEBUG TARGET=$TARGET $IPA_TARGET
+
+if [ "$TARGET" = "IOS64" ]; then
+  "$(pwd)"/darwin/sign.sh
```
lordfolken, line 49: This script seems to be missing?

darwin/XCSoar.xcodeproj/xcshareddata/xcschemes/XCSoar.xcscheme
```diff
@@ -0,0 +1,86 @@
...
+      <BuildActionEntries>
+         <BuildActionEntry
+            buildForTesting = "YES"
+            buildForRunning = "YES"
+            buildForProfiling = "YES"
+            buildForArchiving = "YES"
+            buildForAnalyzing = "YES">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "5CBFD0922D742E1000F05795"
```
lordfolken, line 27: Where does this number come from? Can this key be left undefined?

darwin/build.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/sh
+set -xe
```
lordfolken, line 2: +1

darwin/build.sh
```diff
@@ -0,0 +1,50 @@
...
+# Note: We could let Xcode pass through the build settings in the environment, but this currently at least breaks host compiling of tools (GenerateSineTable will then be build for target) and maybe has other unwanted side effects ... so we pass individual settings as script arguments.
+
+TARGET_PLATFORM_NAME=$1
+CONFIGURATION=$2
+
+export DEBUG="n"
+if [ "${CONFIGURATION}" = "Debug" ]; then
+  export DEBUG="y"
+fi
+
```
lordfolken, line 13: This should be ideally handled by the build system. Ideally we don't redefine debug.

doc/build.rst
```diff
@@ -274,32 +278,12 @@ To compile for macOS / x86_64, run::
...
-    libicns \
-    cdrtools
-
-Then compile::
-
-  make dmg
+Debugging under iOS and macOS is possible using the LLDB debugger.
+To make this convenient, Xcode or Visual Studio can be used.
+An example Xcode project is provided in `darwin/XCSoar.xcodeproj`. It includes one target for iOS and macOS and will automatically build the XCSoar binary for the selected device target, using the build helper script `darwin/build.sh`.
+For iOS debugging with Visual Studio Code, the `iOS Debug` extension (https://github.com/nisargjhaveri/vscode-ios-debug) can be used. Note that this also requires an Xcode installation.
```
lordfolken, line 287: Linebreaks at 80 Chars please.

lordfolken:
Looks pretty good. See some of my comments.

darwin/build.sh
yorickreum: Yes. I am unsure how to handle this. For me, `sign.sh` consists of one line calling `fastlane` to sign the .ipa:
`fastlane sigh resign "$(pwd)"/output/IOS64/xcsoar.ipa --signing-identity "foo" -p darwin/bar.mobileprovision`

This assumes `fastlane` to be installed, though, which might be a bit too much to assume. It is also possible to sign the .ipa without `fastlane` directly with Apple `codesign`, but it's a bit tedious because one has to take care of the signing identities and re-zipping of the .ipa.

What do you think about not including the file in the repo, eventually leaving it to the developer how to sign?
One can still develop for the iOS simulator without any signing; Apple only requires signing for anything that is installed on a real (non-jailbroken) device.

darwin/clean.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+cd "$WORKSPACE_PATH/../../.." || exit 1
+echo "Executing gmake clean in $(pwd)..."
+export PATH=/opt/homebrew/bin:"$PATH"
```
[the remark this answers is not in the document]
yorickreum: The script is meant to be executed via XCode, in the XCode sandbox `WORKSPACE_PATH` is set to the path of the .xcworkspace file (i.e. in my case `WORKSPACE_PATH=/Users/yorickreum/Coding/XCSoar/darwin/XCSoar.xcodeproj/project.xcworkspace`). So this is fine imo.

I will adjust the hombrew prefix to be obtained programatically via `$(brew --prefix)`.

darwin/XCSoar.xcodeproj/xcshareddata/xcschemes/XCSoar.xcscheme
yorickreum: It's automatically set by XCode and needs to be consistent with [darwin/XCSoar.xcodeproj/project.pbxproj](https://github.com/XCSoar/XCSoar/blob/a6150c97fc282b03f19124ac132d91d283274bd2/darwin/XCSoar.xcodeproj/project.pbxproj). I also don't like that the XCode project indexes every file, but I'm afraid that's just how XCode works.

darwin/build.sh
yorickreum: What do you mean? This just automatically sets `DEBUG` for the make build system depending on whether _Release_ or _Debug_ is set as the build configuration in Xcode. We can also default to one value and ignore the Xcode configuration, if you prefer. I think it just adds a bit of convenience like this.

darwin/build.sh
lordfolken: Eventually I'd like to get to the point where we can just run the pipeline here, fastlane executes for both the apple the devices and android and throws out a new build to all the appstores. (Ideal world, wishful thinking)

Currently we sign the android builds in build-native with a secret hidden away in the github organization. Potentially something like this would be possible on the Apple side?

yorickreum: > Eventually I'd like to get to the point where we can just run the pipeline here, fastlane executes for both the apple the devices and android and throws out a new build to all the appstores. (Ideal world, wishful thinking)
Cool, that's what we are doing in other projects as well, directly pushing from Travis CI to a Google Play alpha track and Apple TestFlight.

Currently we sign the android builds in build-native with a secret hidden away in the github organization. Potentially something like this would be possible on the Apple side?
> It's possible, we'd need to store a certificate and a provisioning profile base64 encoded in the secrets... It's well documented here: https://docs.github.com/en/actions/use-cases-and-examples/deploying/installing-an-apple-certificate-on-macos-runners-for-xcode-development
Do you want me to provide a certificate and a profile?

yorickreum: I am not sure about using fastlane anymore ... in my experience with our CI setup, it frequently caused more problems than it solved, and I now think we might be better off by just scripting what we need natively.

darwin/build.sh
```diff
@@ -0,0 +1,50 @@
...
+# and linking all executables to the same path, so we can use one build scheme
+case "$TARGET_PLATFORM_NAME" in
+  iphonesimulator)
+    export TARGET="IOS64SIM"
+    rm "$(pwd)"/darwin/XCSoarExecutable
+    ln -s "$(pwd)"/output/IOS64SIM/ipa/Payload/XCSoar.app "$(pwd)"/darwin/XCSoarExecutable
+    ;;
+  iphoneos)
+    export TARGET="IOS64"
+    rm -rf "$(pwd)"/output/IOS64/ipa "$(pwd)"/output/IOS64/xcsoar.ipa "$(pwd)"/output/IOS64/Info.plist*
```
severeCavok, line 24: Why are you removing the build artefact that's linked in line 27?

darwin/build.sh
```diff
@@ -0,0 +1,50 @@
...
+    # export IOS_APP_BUNDLE_IDENTIFIER="de.yorickreum.XCSoar"
+    rm "$(pwd)"/darwin/XCSoarExecutable
+    ln -s "$(pwd)"/output/IOS64/xcsoar.ipa "$(pwd)"/darwin/XCSoarExecutable
+    ;;
+  macosx)
+    export TARGET="MACOS"
+    rm "$(pwd)"/darwin/XCSoarExecutable
+    ln -s "$(pwd)"/output/MACOS/bin/xcsoar "$(pwd)"/darwin/XCSoarExecutable
+    ;;
```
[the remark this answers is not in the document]
lordfolken: @yorickreum the ai has reason here.

WolfgangDrescher:
Would it be possible to get this merged? It would significantly improve the developer experience and be very helpful for future debugging of iOS issues like #1815.

It would also make sense to have a section in the docs that briefly explains that the `BASE` scripts should be installed as well as `MACOS`/`IOS`.

lordfolken:
I would merge it. But there are still some shell scripts that have assumptions in there, that depend on local configuration. 

Also there is still an rm in there that isn't coded for safety.

darwin/clean.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+set -xe
+
+cd "$WORKSPACE_PATH/../../.." || exit 1
```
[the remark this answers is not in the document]
yorickreum: I got rid if this `cd` statement and also set `set -u` to fail early on unset variables.

darwin/clean.sh
```diff
@@ -0,0 +1,7 @@
...
+echo "Executing gmake clean in $(pwd)..."
+export PATH="$(brew --prefix)"/bin:"$PATH"
```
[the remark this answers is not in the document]
yorickreum: `brew --prefix` is a bad idea, as `brew` won't be available in the Xcode sandbox before the PATH is altered. The `WORKSPACE_PATH` variable I removed now.

lordfolken:
The BREW_PREFIX comment seems reasonable.

darwin/build.sh
```diff
@@ -0,0 +1,96 @@
...
+    ln -s "$target" "$link_name"
+}
+
+# Set the make TARGET based on selected device
+# and linking all executables to the same path, so we can use one build scheme
+EXECUTABLE_PATH="$(pwd)/darwin/XCSoarExecutable"
+case "$TARGET_PLATFORM_NAME" in
+    iphonesimulator)
+        TARGET="IOS64SIM"
+        create_symlink "$(pwd)/output/IOS64SIM/ipa/Payload/XCSoar.app" "$EXECUTABLE_PATH"
```
lordfolken, line 40: Instead of $pwd you can just use relative paths ./

yorickreum: We could, but I frankly think it's usually easier to understand and debug from logs with `pwd` expanded.

darwin/clean.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -euo pipefail
+
+# This script is meant for execution by Xcode.
+
+# Ensure homebrew gmake is in PATH (by default not the case within Xcode sandbox)
+export PATH="/opt/homebrew/bin:$PATH"
```
lordfolken, line 7: Please use the BREW variable here too.

yorickreum: Please see https://github.com/XCSoar/XCSoar/pull/1713#issuecomment-3094667700

WolfgangDrescher:
Xcode cannot build because `ccache` is missing. We should probably add it to `ide/provisioning/install-darwin-packages.sh`.

lordfolken (approved):

Issue #1692, opened 2025-02-22
iOS: UI rotation fails


yorickreum:
XCSoar versions having the problem
-------------------------------------------------

current from master https://github.com/XCSoar/XCSoar/commit/c293d2cdcab6f6e7ce0f72cd343038ea9d9bb34d

XCSoar versions not having the problem
-------------------------------------------------

unknown to me

System information
------------------

iOS 18.3.1 (and others), real device and simulator

Steps to reproduce the behavior
-------------------------------

Start XCSoar on an iOS device, go into fly or sim mode, and rotate the phone. The UI will be weirdly sized, in a way too small drawing rectangle, the remainder of the screen is black.

Expected behavior
-----------------

Resize correctly when rotating.

Actual behavior
---------------

The drawing canvas size doesn't correspond to the screen size.

Do you have any idea what may have caused this?
-----------------------------------------------

Something inconsistent in the OpenGLES/SDL part? Something with the high dpi properties of Apple devices? See https://wiki.libsdl.org/SDL2/README/ios

Do you have an idea how to solve the issue?
-------------------------------------------

Both landscape and portrait modes work correctly when the app is started in the orientation. So it has to be something that's missed/done incorrectly when the rotation (= resizing) event is fired. I checked a few things but couldn't figure out what's off so far. Resizing on macOS works correctly. 
I will investigate this matter further when I find the time, but any support is appreciated.
I'm not sure how resizing and rotation is supposed to be handled in all cases. 
In my understanding, "rotation" is not an important operation on current devices, only size matters.

yorickreum:
The UI currently also always shows black strips at the top and the bottom, this probably stems from ```screen->GetSize()``` returning the notch and round-corner safe area. Could be optimzed, but is imo also acceptable as is. 
How are notes / safe area rendering dealt with on Android?

PR #1714, opened 2025-03-12
iOS: Cleaned up assets, full screen support

yorickreum:
This PR deals with the black strip issue mentioned in https://github.com/XCSoar/XCSoar/issues/1692.
This stems from no launch screen being available for the correct display aspect ratio, which will somehow cause the SDL main window not to fill the whole screen.
The PR cleans up the assets that are converted during the build, removes the old mostly unused ones  (as requested in https://github.com/XCSoar/XCSoar/pull/1659#discussion_r1938521129), and introduces a storyboard for the launch screen instead.

This draft currently still has two issues that need to be addressed:
- The app now runs in true full screen. That also means the InfoBoxes at the top and the bottom are partially hidden underneath the status bar / rounded corners / the dynamic island. This needs to be addressed somehow. Obtaining the safe area via UIKit was not too successful so far, especially struggles with the dynamic island. SDL3 has a `SDL_GetWindowSafeArea`-function for this purpose, but this was not down-ported to SDL2. I might look into the implementation or build some workaround.
- The dialog windows are somehow higher than the MainWindow, making the progress bar leak out at the bottom. Could not figure out yet why this is the case. Some further debugging to do...

### DIFF
--- a/Data/iOS/LaunchScreen.storyboard
+++ b/Data/iOS/LaunchScreen.storyboard
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="23504" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" launchScreen="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="01J-lp-oVM">
+    <device id="retina5_9" orientation="portrait" appearance="light"/>
+    <dependencies>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="23506"/>
+        <capability name="Safe area layout guides" minToolsVersion="9.0"/>
+        <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
+    </dependencies>
+    <scenes>
+        <!--View Controller-->
+        <scene sceneID="EHf-IW-A2E">
+            <objects>
+                <viewController extendedLayoutIncludesOpaqueBars="YES" id="01J-lp-oVM" sceneMemberID="viewController">
+                    <view key="view" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="Ze5-6b-2t3">
+                        <rect key="frame" x="0.0" y="0.0" width="375" height="812"/>
+                        <subviews>
+                            <imageView clipsSubviews="YES" userInteractionEnabled="NO" contentMode="scaleAspectFill" horizontalHuggingPriority="251" verticalHuggingPriority="251" misplaced="YES" insetsLayoutMarginsFromSafeArea="NO" image="AppIcon" translatesAutoresizingMaskIntoConstraints="NO" id="Mqb-Gf-gPS">
+                                <rect key="frame" x="0.0" y="50" width="375" height="762"/>
+                            </imageView>
+                        </subviews>
+                        <viewLayoutGuide key="safeArea" id="6Tk-OE-BBY"/>
+                        <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                        <constraints>
+                            <constraint firstItem="Mqb-Gf-gPS" firstAttribute="top" secondItem="6Tk-OE-BBY" secondAttribute="top" constant="-50" id="JSl-Hj-uuT" userLabel="LaunchImage.top = Safe Area.top"/>
+                            <constraint firstItem="Mqb-Gf-gPS" firstAttribute="trailing" secondItem="6Tk-OE-BBY" secondAttribute="trailing" id="QSp-U6-3aa"/>
+                            <constraint firstItem="Mqb-Gf-gPS" firstAttribute="leading" secondItem="6Tk-OE-BBY" secondAttribute="leading" id="ikz-1b-760"/>
+                            <constraint firstItem="Mqb-Gf-gPS" firstAttribute="bottom" secondItem="Ze5-6b-2t3" secondAttribute="bottom" id="uEd-jY-ZkF"/>
+                        </constraints>
+                    </view>
+                </viewController>
+                <placeholder placeholderIdentifier="IBFirstResponder" id="iYj-Kq-Ea1" userLabel="First Responder" sceneMemberID="firstResponder"/>
+            </objects>
+            <point key="canvasLocation" x="50.399999999999999" y="373.15270935960592"/>
+        </scene>
+    </scenes>
+    <resources>
+        <image name="AppIcon" width="128" height="128"/>
+    </resources>
+</document>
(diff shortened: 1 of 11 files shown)


yorickreum:
The PR now takes care of the safe are (insets caused by notch, round screen corners and dynamic island, etc.) properly. This adds back black borders around the app where needed. Not neccessarily pretty, but makes it safe to have all the UI available on all devices.

I added an iOS workaround for the progress bar positioning bug. The original code there has a few magic numbers in there that make it hard for me to understand how it is supposed to work.

yorickreum:
The failing CI build might stem from an incompatibility with current versions of CMake. It compiles fine locally with CMake version 3.30.2, but the CI runner uses 4.0.1. Will investigate.

yorickreum:
> It compiles fine locally with CMake version 3.30.2, but the CI runner uses 4.0.1. Will investigate.

This really seems to be the issue. Especially libgeotiff and proj were not compatible with CMake 4 just because of their ```CMAKE_MINIMUM_REQUIRED(VERSION 3.X.X)``` configuration. Compatibility with versions of CMake older than 3.5 is removed in CMake 4.0. I updated both libraries and the patch files as needed to make it work. The patch files are not pretty now (I fixed them manually and just concatenated my additional changes at the end), there are still some offsets that can be resolved automatically. Lmk if you want me to `rediff` the patch files.

src/ProgressWindow.cpp
```diff
@@ -68,6 +69,20 @@ ProgressWindow::UpdateLayout(PixelRect rc) noexcept
 
   bottom_position = rc;
   bottom_position.top = bottom_position.bottom - progress_border_height;
+  // For some reason (inconsistency with points and native pixels?)
+  // progress bar would be drawn below the window on iOS without correcting here
+  // ToDo: find real issue and fix there
```
lordfolken, line 74: can you add your github handle to the Todo?

yorickreum: Sure.

yorickreum:
Anything that still needs to be done here?

darwin/.gitignore
```diff
@@ -0,0 +1,6 @@
+# Xcode
+xcuserdata
+DerivedData/
+build/
+# build.sh
+XCSoarExecutable
```
lordfolken: Newiline

darwin/clean.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+cd "$WORKSPACE_PATH/../../.." || exit 1
+echo "Executing gmake clean in $(pwd)..."
+export PATH=/opt/homebrew/bin:$PATH
```
lordfolken: How "Universal" is this? Isn't this the shell's job to configure that on the host?

src/ui/window/sdl/TopWindow.cpp
```diff
@@ -43,7 +51,7 @@ MakeSDLFlags([[maybe_unused]] bool full_screen, bool resizable) noexcept
 
 #if defined(__IPHONEOS__) && __IPHONEOS__
   /* Hide status bar on iOS devices */
-  flags |= SDL_WINDOW_BORDERLESS;
+// flags |= SDL_WINDOW_BORDERLESS;
```
lordfolken: Please no commented out code.

lordfolken: Or is this related to the Progressbar bug? (Then please with comment.)

darwin/clean.sh
yorickreum: Imo it's the right way to approach this. The script is there to be executed by Xcode inside its build sandbox -- which has a very minimum shell configuation on its own. If we want to use binaries from homebrew inside the sandbox, we will need to add it into the path inside the sandbox.

However, I will remove this from this PR since it actually belongs to https://github.com/XCSoar/XCSoar/pull/1713.
(sorry for the confusion ... I had my changes from the other pull request merged in to be able to develop properly ... having all the different PRs open currently makes it a bit tedious, looking forward to just be able to base all development on the current master).

darwin/.gitignore
yorickreum: Will address in https://github.com/XCSoar/XCSoar/pull/1713.

src/ui/window/sdl/TopWindow.cpp
yorickreum: Fair. I now removed the code within an atomic commit.

src/ProgressWindow.cpp
```diff
@@ -68,6 +69,20 @@
...
+  if (rc.GetHeight() > rc.GetWidth()) {
+    bottom_position.top -= 3 * progress_height;
+    bottom_position.bottom -= 3 * progress_height;
+  } else {
+    message_position.left -= 3 * progress_height;
+    message_position.right -= 3 * progress_height;
+    bottom_position.left -= 3 * progress_height;
+    bottom_position.right -= 3 * progress_height;
+  }
+#endif
```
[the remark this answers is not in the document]
yorickreum: Outdated. I replaced the workaround with a proper fix.

src/ProgressWindow.cpp
```diff
@@ -29,11 +29,12 @@ ProgressWindow::ProgressWindow(ContainerWindow &parent) noexcept
 #ifndef USE_WINUSER
   font.Load(FontDescription(Layout::FontScale(10)));
   text_height = font.GetHeight();
+#elif defined(__APPLE__) && TARGET_OS_IPHONE
+  text_height = 3 * text_height;
```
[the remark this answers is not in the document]
yorickreum: Outdated. This change was actually not required.

Data/iOS/LaunchScreen.storyboard
```diff
@@ -0,0 +1,39 @@
...
+                            </imageView>
+                        </subviews>
+                        <viewLayoutGuide key="safeArea" id="6Tk-OE-BBY"/>
+                        <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                        <constraints>
+                            <constraint firstItem="Mqb-Gf-gPS" firstAttribute="top" secondItem="6Tk-OE-BBY" secondAttribute="top" constant="-50" id="JSl-Hj-uuT" userLabel="LaunchImage.top = Safe Area.top"/>
+                            <constraint firstItem="Mqb-Gf-gPS" firstAttribute="trailing" secondItem="6Tk-OE-BBY" secondAttribute="trailing" id="QSp-U6-3aa"/>
+                            <constraint firstItem="Mqb-Gf-gPS" firstAttribute="leading" secondItem="6Tk-OE-BBY" secondAttribute="leading" id="ikz-1b-760"/>
+                            <constraint firstItem="Mqb-Gf-gPS" firstAttribute="bottom" secondItem="Ze5-6b-2t3" secondAttribute="bottom" id="uEd-jY-ZkF"/>
+                        </constraints>
```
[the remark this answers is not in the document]
yorickreum: Fixed.

yorickreum:
I just put some more serious work into making this nice, replaced the progress bar workaround with a proper fix. 
In my opinion this can now be merged with a very clear conscience.

lordfolken (approved):

Issue #1647, opened 2025-01-15
Configurable FLARM traffic filter


fraktal-sb:
XCSoar version
--------------

7.43

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------

Newer airtraffic devices also receive ADS-B signals, which means also airline traffic can be monitored. Because this the traffic in Flarm radar is becoming confusing. 

Example:

![Bildschirmfoto zu 2025-01-15 10-48-30](https://github.com/user-attachments/assets/5f71dd68-f1bf-4b66-9b2a-10c33ba335a1)

Other applications therefore have a configurable filter.

Example: Safesky app

![grafik](https://github.com/user-attachments/assets/383947f6-c339-4795-a6ee-918d96a4de25)

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------

I analysed the program code in 

XCSoar/src/Device/Driver/FLARM/StaticParser.cpp

Within the part

ParsePFLAU(NMEAInputLine &line, FlarmStatus &flarm, TimeStamp clock)

we have all the data of incomming traffic to apply a filter

traffic.relative_north
traffic.relative_east
traffic.relative_altitude

Applying a filter here would prevent storing/processing/displaying non interesting traffic.

The filter could be made configurable in the following setup page:

![2025-01-15 12_20_53-Clipboard](https://github.com/user-attachments/assets/bb3ca970-dbfd-4cbf-8174-ed1d918b09bd)

As default filter values we could take 20 km distance and +-2000m height difference

lordfolken:
This is a FLARM setting and belongs into your Flarm. 
Filtering separately in XCSoar is only adding additional confusion and errors. 

You can already set the range in the FlarmConfigWidget:
https://github.com/XCSoar/XCSoar/blob/273cc7f332972732a379c0aa4430d0e4f58eff27/src/Dialogs/Device/FLARM/ConfigWidget.cpp#L108

This probably would have to be updated to the current PowerFlarm Standard:
https://www.flarm.com/wp-content/uploads/2024/04/FTD-014-FLARM-Configuration-Specification-1.18.pdf

There you will find additional setting separate for ADSB / PCAS for example:
![image](https://github.com/user-attachments/assets/4c5d8349-4aa4-4e2a-ba52-b33ee3a15196)

lordfolken:
[...]

PR #1715, opened 2025-03-13
FLARM hardware read-out and additional config options

Scumi:
This is how I would do the hardware read-out and optional config items based on it, but I'm not sure, if it is the best way. I think there are more possibilities to make stuff re-usable and more maintainable. 
Code is tested and based on official FLARM interface documents.

Ref: #1647

### DIFF
--- a/src/Device/Driver/FLARM/Settings.cpp
+++ b/src/Device/Driver/FLARM/Settings.cpp
@@ -25,6 +25,26 @@ FlarmDevice::SendSetting(const char *name, const char *value,
   Send(buffer, env);
 }
 
+bool
+FlarmDevice::RequestAllSettings(const char* const* settings, 
+                                OperationEnvironment &env)
+{
+  try {
+    for (auto i = settings; *i != NULL; ++i)
+      FlarmDevice::RequestSetting(*i, env);
+
+    for (auto i = settings; *i != NULL; ++i)
+      FlarmDevice::WaitForSetting(*i, 500);
+  } catch (OperationCancelled) {
+    return false;
+  } catch (...) {
+    env.SetError(std::current_exception());
+    return false;
+  }
+
+  return true;
+}
+
 void
 FlarmDevice::RequestSetting(const char *name, OperationEnvironment &env)
 {
--- a/src/Device/Driver/FLARM/Settings.cpp
+++ b/src/Device/Driver/FLARM/Settings.cpp
@@ -2,6 +2,7 @@
 // Copyright The XCSoar Project
 
 #include "Device.hpp"
+#include "system/Sleep.h"
 
 #include <stdio.h>
 
@@ -30,6 +31,25 @@ FlarmDevice::RequestSetting(const char *name, OperationEnvironment &env)
   char buffer[64];
   sprintf(buffer, "PFLAC,R,%s", name);
   Send(buffer, env);
+}
+
+bool
+FlarmDevice::WaitForSetting(const char *name, unsigned timeout_ms)
+{
+  for (unsigned i = 0; i < timeout_ms / 100; ++i) {
+    if (FlarmDevice::SettingExists(name))
+      return true;
+    Sleep(100);
+  }
+
+  return false;
+}
+
+[[gnu::pure]]
+bool
+FlarmDevice::SettingExists(const char *name) noexcept
+{
+  return (bool)FlarmDevice::GetSetting(name);
 }
 
 std::optional<std::string>
--- a/src/Device/Driver/FLARM/Settings.cpp
+++ b/src/Device/Driver/FLARM/Settings.cpp
@@ -82,3 +82,16 @@ FlarmDevice::GetSetting(const char *name) const noexcept
 
   return *i;
 }
+
+unsigned
+FlarmDevice::GetUnsignedValue(const char *name, unsigned default_value)
+{
+  if (const auto x = FlarmDevice::GetSetting(name)) {
+    char *endptr;
+    unsigned long y = strtoul(x->c_str(), &endptr, 10);
+    if (endptr > x->c_str() && *endptr == 0)
+      return (unsigned)y;
+  }
+
+  return default_value;
+}
--- a/src/Dialogs/Device/FLARM/ConfigWidget.cpp
+++ b/src/Dialogs/Device/FLARM/ConfigWidget.cpp
@@ -34,7 +34,8 @@ FLARMConfigWidget::Prepare([[maybe_unused]] ContainerWindow &parent,
   device.RequestAllSettings(flarm_setting_names, env);
 
   baud = device.GetUnsignedValue("BAUD", 2);
-  thre = device.GetUnsignedValue("THRE", 1);
+  unsigned thre_default = hardware.isPowerFlarm() ? 255 : 1;
+  thre = device.GetUnsignedValue("THRE", thre_default);
   acft = device.GetUnsignedValue("ACFT", 0);
   log_int = device.GetUnsignedValue("LOGINT", 2);
   priv = device.GetUnsignedValue("PRIV", 0) == 1;
--- a/src/Dialogs/Device/FLARM/ConfigWidget.cpp
+++ b/src/Dialogs/Device/FLARM/ConfigWidget.cpp
@@ -22,15 +22,15 @@ static const char *const flarm_setting_names[] = {
 };
 
 static bool
-RequestAllSettings(FlarmDevice &device)
+RequestAllSettings(FlarmDevice &device, const char* const* settings)
 {
   PopupOperationEnvironment env;
 
   try {
-    for (auto i = flarm_setting_names; *i != NULL; ++i)
+    for (auto i = settings; *i != NULL; ++i)
       device.RequestSetting(*i, env);
 
-    for (auto i = flarm_setting_names; *i != NULL; ++i)
+    for (auto i = settings; *i != NULL; ++i)
       device.WaitForSetting(*i, 500);
   } catch (OperationCancelled) {
     return false;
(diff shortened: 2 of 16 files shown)


src/Device/Driver/FLARM/Settings.cpp
```diff
@@ -32,6 +53,25 @@ FlarmDevice::RequestSetting(const char *name, OperationEnvironment &env)
   Send(buffer, env);
 }
 
+bool
+FlarmDevice::WaitForSetting(const char *name, unsigned timeout_ms)
+{
+  for (unsigned i = 0; i < timeout_ms / 100; ++i) {
+    if (FlarmDevice::SettingExists(name))
+      return true;
+    Sleep(100);
```
lordfolken, line 62: Please use std::chrono::milliseconds()

Scumi: Well...`Sleep()` is not taking `std::chrono` as input, just `unsigned`. Only `env.Sleep()` can take `std::chrono`, unless I miss something. The entire codebase shows this.

src/Dialogs/Device/FLARM/ConfigWidget.cpp
```diff
@@ -83,14 +44,14 @@ void
...
+  PopupOperationEnvironment env; 
+  device.RequestAllSettings(flarm_setting_names, env);
 
   baud = GetUnsignedValue(device, "BAUD", 2);
-  priv = GetUnsignedValue(device, "PRIV", 0) == 1;
   thre = GetUnsignedValue(device, "THRE", 2);
-  range = GetUnsignedValue(device, "RANGE", 3000);
   acft = GetUnsignedValue(device, "ACFT", 0);
   log_int = GetUnsignedValue(device, "LOGINT", 2);
+  priv = GetUnsignedValue(device, "PRIV", 0) == 1;
```
lordfolken, line 54: Why do you move the priv? its adding noise to the commit.

src/Dialogs/Device/FLARM/ConfigWidget.cpp
```diff
@@ -146,24 +111,12 @@ try {
     changed = true;
   }
 
-  if (SaveValue(Priv, priv)) {
```
lordfolken, line 149: Again why, move it?

src/FLARM/Hardware.hpp
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * The FLARM hardware read-out from PFLAC config sentences.
+ */
+struct FlarmHardware {
+  Validity available;
+
+  NarrowString<64> device_type;
+  NarrowString<64> capabilities;
+
+  bool isPowerFlarm() noexcept {
```
lordfolken, line 20: PowerFlarm supports additional baud rates, and has 2 data ports:

lordfolken (changes requested):
In general, pretty cool. :) 
Can you merge the fixup commits into the regular ones please?

src/FLARM/Hardware.hpp
Scumi: Yes, true. It seems to work with just one baud input as well (sets both ports at once). If you want, I can include the other one as well. There are even more PowerFlarm (and ADSB) related config items as well as possible read-outs of the FLARM ID and others, but I wanted to get the basics through first.

Scumi:
Yes, I'll have a look. I'm not happy with the capabilities-display in the ConfigWidget because it gets cut-off. Do you have a proposal to solve this elegantly?
image

src/Dialogs/Device/FLARM/ConfigWidget.cpp
lordfolken: I just wanted to hear your argument. :) Thank you.

Scumi: I will add my reasoning to the commit message, so it will be clear for a future reader.

lordfolken:
So I was able to test your PR thanks due to a loaner (development) device from FLARM. 

I think the only 2 things really missing, is displaying the 
* Flarm hex id 
* Obstacle DB
* Second Port Baud (potentially)
And making the FLARM Id editable.

Then i think we have solid powerflarm support. :)

lordfolken:
The only thing i'm a little concerned about is is the threshold value. 

So according to flarm this should be:

```
PowerFLARM-based devices also support an automatic setting (value 255) where
the threshold is based on the configured aircraft type (ACFT setting), see the table
below for values.
```

on classic flarms this needs to be set manually. 

The 255 automatic value is not selectedable in the list

lordfolken:
I also noticed that all the settings are written, not just the ones that have changed. This is dangerous.

lordfolken:
I was able to trigger an assert when setting the range on a Classic Flarm:
`xcsoar: ./output/src/boost_1_87_0/boost/container/vector.hpp:1757: boost::container::vector<T, Allocator, Options>::const_reference boost::container::vector<T, Allocator, Options>::operator[](size_type) const [with T = RowFormWidget::Row; A = boost::container::dtl::static_storage_allocator<RowFormWidget::Row, 32, 0, true>; Options = void; const_reference = const RowFormWidget::Row&; size_type = long unsigned int]: Assertion `this->m_holder.m_size > n' failed.`

src/Dialogs/Device/FLARM/RangeConfigWidget.cpp
```diff
@@ -0,0 +1,131 @@
...
+    device.RequestAllSettings(pf_setting_names, env);
+  if (hardware.hasADSB())
+    device.RequestAllSettings(adsb_setting_names, env);
+
+  range = GetUnsignedValue(device, "RANGE", 3000);
+  vrange = GetUnsignedValue(device, "VRANGE", 500);
+  pcas_range = GetUnsignedValue(device, "PCASRANGE", 7408);
+  pcas_vrange = GetUnsignedValue(device, "PCASVRANGE", 610);
+  adsb_range = GetUnsignedValue(device, "ADSBRANGE", 65535);
+  adsb_vrange = GetUnsignedValue(device, "ADSBVRANGE", 65535);
```
lordfolken, line 66: I think this needs to move into the hasADSB() resp. isPowerFlarm functions.

src/Dialogs/Device/FLARM/RangeConfigWidget.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+  AddInteger(_("Range"), NULL, _T("%d m"), _T("%d"), 2000, 25500, 250, range);
```
lordfolken, line 68: Beware from FTD-014-Flarm-Configuration Document:
```

For PowerFLARM-based devices, unsigned decimal values between 2000 and
65535 (unlimited) are allowed. For Classic FLARM-based devices, unsigned decimal
values between 2000 and 25500 are allowed.
Default value:
For PowerFLARM-based devices, the default value is 65535. For Classic FLARM-
based devices, the default value is 25500.
```

src/Dialogs/Device/FLARM/RangeConfigWidget.cpp
```diff
@@ -0,0 +1,131 @@
...
+    buffer.UnsafeFormat("%u", adsb_range);
+    device.SendSetting("ADSBRANGE", buffer, env);
+    changed = true;
+  }
+
+  if (SaveValueInteger(ADSBVrange, adsb_vrange)) {
+    buffer.UnsafeFormat("%u", adsb_vrange);
+    device.SendSetting("ADSBVRANGE", buffer, env);
+    changed = true;
+  }
```
lordfolken, line 122: These need to be gated by isPowerFlarm() and hasADSB(); Otherwise xcsoar will trigger the mentioned assertion when saving the range on a classic flarm.

Scumi:
Thanks for the extensive review and testing. I try to look into this next week. I would like to merge the basic integration and then do stuff like Flarm ID in a follow-up PR because there are more changes required on how data is read-out and processed by the FLARM driver.

src/Dialogs/Device/FLARM/RangeConfigWidget.cpp
Scumi: done

Scumi:
> The only thing i'm a little concerned about is is the threshold value.
> 
> So according to flarm this should be:
> 
> ```
> PowerFLARM-based devices also support an automatic setting (value 255) where
> the threshold is based on the configured aircraft type (ACFT setting), see the table
> below for values.
> ```
> 
> on classic flarms this needs to be set manually.
> 
> The 255 automatic value is not selectedable in the list

Do you have an idea how to do this? According to FTD-014, 0-20 are the allowed values. `AddInteger()` can not set 0-20 plus 255 in addition at the end. I could use `AddEnum()` and create a list that uses 0-20 and 255 (which then can even be named "automatic").

src/Dialogs/Device/FLARM/RangeConfigWidget.cpp
Scumi: done

Scumi:
Okay, I fixed the general issues and bugs. I suggest we open an issue with a wish list for the follow-up PR. Flarm ID being the most important one, other settings like enable/disable ADSB or PCAS can follow. The list of available config options in FTD-014 is long, we should see what are the more important ones that users might want to play with.

src/Device/Driver/FLARM/Settings.cpp
```diff
@@ -42,3 +82,16 @@ FlarmDevice::GetSetting(const char *name) const noexcept
...
+{
+  if (const auto x = FlarmDevice::GetSetting(name)) {
+    char *endptr;
+    unsigned long y = strtoul(x->c_str(), &endptr, 10);
+    if (endptr > x->c_str() && *endptr == 0)
+      return (unsigned)y;
+  }
+
+  return default_value;
+}
```
lordfolken, line 97: missing newline

Scumi: done

lordfolken (approved):

lordfolken:
Thank you for your work. :-)

PR #1721, opened 2025-03-21
Stratux support (rebased)

fraktal-sb:
This is a code rebase of https://github.com/XCSoar/XCSoar/pull/1710


lordfolken:
Please don't open new pull request. The pull request only reference a branch name, regardless of how you update and change them.

lordfolken:
Ok there is a few things here:
 The heart of this change seems to be to filter on received traffic from the stratux device. 
 * Is the stratux device not capable of configuring this range?
 * Is this not dynamically configuredable like in flarm with PFLAC? (See: https://github.com/XCSoar/XCSoar/pull/1715 for examples)

Code:
 * This is basically a copy of the flarm driver. Is there any delta other than your filter? If so couldn't  this be imported from FLARM and then processed? Or even integrated into the flarm driver instead? Otherwise we have to maintain 2 almost identical drivers. 

CodeOrganization:
 * This is a bit all in one file, I think the topolicial seperation like in other drivers would seem approriate (Parser from Configuration etc)
 * Commitwise its better to see all the commits than one big commit. then we can separate and merge whats good.

fraktal-sb:
The Stratux configuration can be changed/modifyed by web interface. There are 2 parameters where you can enter a horizontal and vertical range but they are not used to filter the output data. It is only a display filter within the web interface. So the device receiving Stratux data must care himself about filtering. A robust method by using the serial interface communication to access internal settings (as Flarm does) is not foreseen (to be sure I contacted also the main developer of stratux-europe).

Yes it is a copy of the the ParsePFLAA part of the FLARM/StaticParser.cpp. The intention behind was to not touch the Flarm Part of XCSoar to keep the reliability and integrety of the Flarm data processing. Of couse it would be more simple to maintain only one driver then 2. But the complexity will rise if special information should be extracted as for example the callsign information which is different stored in Stratux and Flarm. Stratux uses "!" to separate the HexId from the callsign and puts both in the name field which is a non valid information in Flarm- sense.

We also could forget about the current development and simply define an overall filter for "non Flarm Devices" This could be easyly integrated in the current Flarm driver. To let receive the flarm driver all data and later on filter I would not recommend. Sometimes Stratux receives more then the 24 targets (i think) XCSoar is able to maintain and then the displayed information is by chance. The best place to filter traffic data is within this driver part.

From my opinion it is "psychologicly" (for user and programmer) better to have an extra device then puting more functionality into the existing flarm driver. The part of parsing the PFLAA records will be the same in 100 years. So the necessary adaptions or changes over the time will be very much limited and if it is allways separated from Flarm.

Concerning the code organisation: yes it could be made different and with more modularity. My approach is more to have the related logic together. Users will not notice this and the complexity of the catual implementation is not that high that nobody can understand what's going on in this part of the software. If in future more and more pilots will use stratux and we need more functionality we can think about a code revision. But for now it is a state we can start with. It is simple, robust, well tested and reliable.

So at this point it's only a decision wether we continue or not or do something else.

lordfolken:
I like this approach better. 
Now we should make it 3 commits as i think there are 3 topics here:
* rearange the flarm files into its own build.mk
* extent Flarm's PFLAA witn rangefilter
* add the stratux driver

lordfolken:
Also please have a look at git amend  and force push, that way you don't need to add commits, but can replace / alter them.

fraktal-sb:
is it possible to reopen this PR because of some command it was automatically closed?

fraktal-sb:
hope this is ok now. thanks for having patience.

src/Device/Driver/FLARM/StaticParser.cpp
```diff
@@ -28,8 +28,7 @@ ParsePFLAE(NMEAInputLine &line, FlarmError &error, TimeStamp clock) noexcept
   StringFormatUnsafe(buffer, _T("%s - %s"),
                      FlarmError::ToString(error.severity),
                      FlarmError::ToString(error.code));
-  if (error.severity != FlarmError::Severity::NO_ERROR)
-    Message::AddMessage(_T("FLARM: "), buffer);
+  Message::AddMessage(_T("FLARM: "), buffer);
 
   error.available.Update(clock);
```
lordfolken: Could this be from on old commit?

fraktal-sb: changes from previous PR (see below) were not included. Will submit an update based on current head

Bildschirmfoto zu 2025-06-06 09-22-48

lordfolken: Please do git pull --rebase

fraktal-sb:
Done

lordfolken:
continuing in: Std006 #1801

fraktal-sb:
last commit obsolete because https://github.com/XCSoar/XCSoar/pull/1818 does it; PR closed

PR #1727, opened 2025-03-30
[tophat_flabor] "M" appears only on map screen, e.g. not in the Thermal Assistant screen.

CazYokoyama:
Fix the bug which is reported by Sundown3867 in https://github.com/XCSoar/XCSoar/pull/1687.


lordfolken (approved):

CazYokoyama:
Thank you for merging.

lordfolken:
サンキュー for sticking with it. :bowing_man:

Issue #1730, opened 2025-03-31
FLARM Traffic Details Dialog - Issues and Improvements


lordfolken:
(This may have to be spitted into several issues)

![Image](https://github.com/user-attachments/assets/71479a4c-171a-43b3-9884-ceb76c363a58)

Problems:
 * When one opens a user defined id, the subsequent input filed is blank instead of showing the existing record
 * The Title of the input editor says Competition ID, but the button Change Callsign was pressed. 
 * Blanking the competition id leaves an empty record in xcsoar-flarm.txt:
        39103C=f8
        4B4C6B=
 * there is no way of removing a record from xcsoar-flarm.txt not even by blanking it.
 * The plane field should be named Plane Type
 * The only thing gathered from xcsoar-flarm.txt is the flarmid=callsign. All other data is not editable and is still being fetched from the DDB configured. 
 * When using the FLARMNET db, the Airport field contains the Aircraft REG.
 * Traffic type should be indicated (Flarm,ADSB,OGN,Skysight,XCSoar Cloud...)

Scumi:
> When using the FLARMNET db, the Airport field contains the Aircraft REG.

I looked into this and I think its a fault (or change) within the data.fln file. I parsed some of the content of a freshly downloaded data.fln.

- The pilot-field is always empty
- The airfield-field is alway registration now (registration shows up twice, again at the place where it used to be after plane and before callsign)

I could not find any documentation for the flarmnet datafile.

PR #1732, opened 2025-04-03
Improvements for flarm details dialog

Scumi:
Fixes parts of #1730

### DIFF
--- a/src/FLARM/NameDatabase.cpp
+++ b/src/FLARM/NameDatabase.cpp
@@ -71,8 +71,13 @@ FlarmNameDatabase::Set(FlarmId id, const TCHAR *name) noexcept
 
   int i = Find(id);
   if (i >= 0) {
-    /* update existing record */
-    data[i].name = name;
+    if (!StringIsEqual(name, _T("")))
+      /* update existing record */
+      data[i].name = name;
+    else
+      /* remove record if empty */
+      data.remove(i);
+
     return true;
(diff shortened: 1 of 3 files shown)


lordfolken (changes requested):
The implementation is fine. 

However we are clearly working around here something, that is simply wrong in the external format. Please drop this patch, and hope someone at flarmnet fixes their exports.

src/FLARM/NameDatabase.cpp
```diff
@@ -87,3 +87,17 @@ FlarmNameDatabase::Set(FlarmId id, const TCHAR *name) noexcept
...
+  assert(id.IsDefined());
+
+  int i = Find(id);
+  if (i >= 0) {
+    data.remove(i);
+    return true;
+  }
+
+  return false;
+}
```
lordfolken, line 103: Please fix newline

Scumi: fixed

Scumi:
> The only one i don´t want to merge is [00f1c7e](https://github.com/XCSoar/XCSoar/commit/00f1c7ea99115257e69aa7c9631584c2ef7b2bdd) as I believe its not good to fix an external format in XCSoar's display code. I really think the flarmnet export should be fixed here instead.

Fair enough, I removed it. Someone tell the Flarm team? :-)

lordfolken (approved):

PR #1735, opened 2025-04-06
Show Flarm ID in config dialog

Scumi:
Small amendment to #1715.

@lordfolken can you check with a classic FLARM device? Also, using line.Rest() is a bigger change which puts more responsibility to the caller. Haven't found a regression for existing implementations though.


lordfolken:
So tested with both classic and powerflarm. LGTM.

lordfolken (approved):

Issue #1546, opened 2024-08-23
Startup Wizard - Display Gesture Help


lordfolken:
When starting XCSoar for the first time, display a Popup that introduces you to the gestures that are possible. 

![Image](https://github.com/user-attachments/assets/57154e08-cb30-410f-bdc7-743cc77c5afe)

WolfgangDrescher:
I'm working on a small "onboarding" dialog:

```cpp
void
dlgOnboardingShowModal()
{
  const DialogLook &look = UIGlobals::GetDialogLook();

  WidgetDialog dialog(WidgetDialog::Full{}, UIGlobals::GetMainWindow(),
                      look, _("Onboarding"));

  auto text_widget = std::make_unique<LargeTextWidget>(
    look,
    _("Welcome to XCSoar!\nPlease follow the instructions.")
  );

  dialog.FinishPreliminary(std::move(text_widget));
  dialog.AddButton(_("Close"), mrCancel);
  dialog.ShowModal();
}
```

And in `Startup.cpp`:

```cpp
  // Show onboarding dialog
  if (Profile::GetPath(ProfileKeys::MapFile) == nullptr) {
    dlgOnboardingShowModal();
  }
```

However, I have a few questions, as I don't have an overview of the available widgets.

1. To determine whether XCSoar has already been set up, I use `Profile::GetPath(ProfileKeys::MapFile)`. Is this appropriate?
2. What is the best way to create a list with e.g. multiple `LargeTextWidget` items that provide the user with tips on what to do (setup “Site Files,” airspaces, InfoBox pages, etc.)? Is `ListWidget` useful for this?
3. How can I use a button `dialog.AddButton(_(“Settings”), [](){});` to refer to `Config` -> `System`?
4. Should I create a new `GridWidget` for the icons of the gestures with the corresponding descriptions, or is there already something useful for this?
5. Can multiple widgets be combined? (`LargeTextWidget` + `GridWidget`)

lordfolken:
1. Please use your own ProfileKey:: e.g. GestureHelpShown. Write that one into the Profile if displayed once, unless there a check box  "show again on startup" is selected. It also should only be shown if HasTouchScreen() is true.
2. LargeTextWidget was filling the screen typically. There is also the issue of scrolling and overflow espeically when using the keyboard to operate. Maybe have a look at dlgCredits.cpp there multiple LTWs are in use. 
3. That one is a little ... tricky. The menu calls SystemConfiguration();
[...]

PR #1739, opened 2025-04-10
[menu zoom button] large buttons on raspberry pi which has touch screen

CazYokoyama:
As the comment below for HasTouchScreen() in Asset.hpp, we need bigger icons for touch screen. 

a touch screen may require bigger areas.

defined(RASPBERRY_PI) does not guarantee whether touch screen is attached. sysfs has an entry for the touch screen as below.
$ cat /sys/class/input/event?/device/name 
.....
1-0014 Goodix Capacitive TouchScreen
There are 2 ways to determine whether a touch screen is attached, C code and
`grep -i TouchScreen /sys/class/input/event?/device/name`
On either cases, we need fairly large amount of code. Let me know your preference, this simple code, C code or run grep.

### DIFF
--- a/src/Asset.cpp
+++ b/src/Asset.cpp
@@ -2,6 +2,7 @@
 // Copyright The XCSoar Project
 
 #include "Asset.hpp"
+#include <fcntl.h>
 
 #if defined(USE_CONSOLE) || defined(USE_WAYLAND)
 #include "ui/event/Globals.hpp"
@@ -20,10 +21,56 @@ HasPointer() noexcept
 
 #if defined(USE_LIBINPUT) || defined(USE_WAYLAND)
 
+#define MAX_EVENT_DEVICES 20
+#define EVENT_SYSFS_PREFIX "/sys/class/input/event"
+#define EVENT_SYSFS_POSTFIX "device/name"
+#define MAX_FNAME 40
+#define MAX_DNAME 80
+#define TS_STRING "TouchScreen"
+
 bool
 HasTouchScreen() noexcept
 {
-  return UI::event_queue->HasTouchScreen();
+  static enum {
+    not_detected,
+    not_attached,
+    attached
+  } touchscreen = not_detected;
+
+  if (touchscreen != not_detected)
+    return touchscreen == attached;
+  if (UI::event_queue->HasTouchScreen()) {
+    touchscreen = attached;
+    return true;
+  }
+  touchscreen = not_attached;
+  for (int i = 0; i <  MAX_EVENT_DEVICES; i++) {
+    char fname[MAX_FNAME], dname[MAX_DNAME];
+    int eventf;
+    ssize_t count;
+
+    snprintf(fname, sizeof(fname),
+	     "%s%d/%s",
+	     EVENT_SYSFS_PREFIX, i, EVENT_SYSFS_POSTFIX);
+    eventf = open(fname, O_RDONLY);
+    if (eventf < 0)
+      break;
+    count = read(eventf, dname, sizeof(dname));
+    if (count <= 0) {
+      close(eventf);
+      break;
+    }
+    if (count >= (ssize_t)sizeof(dname))
+      count = sizeof(dname) - 1;
+    dname[count] = '\0';
+    if (strstr(dname, TS_STRING)) {
+      touchscreen = attached;
+      close(eventf);
+      return true;
+    }
+    close(eventf);
+  }
+  return false;
 }
 
 bool


CazYokoyama:
Please hold. Afterward test shows the code does not work. Will update.

CazYokoyama:
Can I assume the person who wants menu and zoom buttons has touch screen? Those buttons are clicked by mouse and touch screen. Larger icons cover large area of the map, but who care? You can disable them by configuration.

lordfolken:
We should have a  real touchscreen detection for Raspberry PI. A lot of the code depends on this and is imo good philosophy. 

I for example don't have a touch screen with mine, because i don't want fingermarks all over it and also rather have the increased contrast than a filter in front of the screen. Also personally i find button presses i can do without looking at the screen, far superior in a cockpit setting. (But i might be alone with that opinion.) 

About activating this by default:

These buttons are now the 4th way of interacting with xcsoar: (for example map zoom)
* Nav->Display-> Zoom in/out
* Gesture (up / down)
* Volume Rocker on a Phone / Keyboard / Joystick on others
* OnScreen Buttons

None of these (except the on-screen buttons) are self-explanatory and intuitive or build on familiarity of typical smartphone apps. For exmaple the first thing people try when i give xcsoar to them is pinch for zoom. 

Its also incredibly inconsistent. On the PAN page you can´t use gestures to zoom, and then there is onscreen buttons (others than the mainpage). 

I think we need a clear design guideline when it comes to interaction with xcsoar.

CazYokoyama:
Add touchscreen checking. See its commit message for detail.

4 ways of interacting with xcsoar
-------------------
I did not know Gesture (up / down). We may not need zoom in/out button because Gesture is much easier, no need to point a specific area.
I was surprised when I push Volume up button on a Phone. It zooms in instead of volume up when I find warning sound is too weak. I have to get out xcsoar, volume up and invoke xcsoar again. Not good.

CazYokoyama:
Codacy static code analysis reports

- #include <fcntl.h> - is not need because stdlib. But fail Raspbian(rpi). fcntl.h is included many other places.
- count = read(eventf, dname, sizeof(dname)); - need boundary check. No idea how to include boundary check.

I will be glad if you give me an advice.

lordfolken:
> Add touchscreen checking. See its commit message for detail.
> 
> ## 4 ways of interacting with xcsoar
> I did not know Gesture (up / down). We may not need zoom in/out button because Gesture is much easier, no need to point a specific area. 

I know we have a dscoverability Problem #1546 

> I was surprised when I push Volume up button on a Phone. It zooms in instead of volume up when I find warning sound is too weak. I have to get out xcsoar, volume up and invoke xcsoar again. Not good.

I agree with you, we should change the volume buttons.     Zooming without panning ability, doesnt make sense here.

lordfolken:
https://github.com/XCSoar/XCSoar/pull/1867 @CazYokoyama what do you think?

CazYokoyama:
> #1867 @CazYokoyama what do you think?

I enables menu and zoom buttons and use a lot even I know I can do it by gestures because I am Tophat user for a long time. I am glad if you keep those buttons with enabling/disabling capability.

lordfolken (commented):
## Review notes

Thanks for tackling Raspberry Pi touch detection when running without X — the sysfs fallback is a reasonable approach when `UI::event_queue->HasTouchScreen()` stays false.

### `[[gnu::pure]]` vs. cached state

`HasTouchScreen()` is still declared `[[gnu::pure]]` in `Asset.hpp`. A `pure` function must not modify global/static state, but the new implementation updates a `static` cache. That mismatch can confuse the optimizer and contradicts the attribute contract.

**Suggestion:** Drop `[[gnu::pure]]` on `HasTouchScreen()` for this configuration, or avoid mutable cache (with the usual tradeoff of repeated sysfs walks).

### Includes

The `.cpp` uses `open`, `read`, `close`, `snprintf`, and `strstr`. Elsewhere in the tree these typically get explicit includes, e.g. `<unistd.h>` for POSIX read/close, `<cstdio>` for `snprintf`, `<cstring>` for `strstr`, plus `<fcntl.h>` as you already have.

### Heuristic

Matching the substring `"TouchScreen"` in `/sys/class/input/event*/device/name` is hardware/driver-specific. Consider a one-line comment naming which Pi touch setup or kernel naming this targets so future readers know when the fallback applies or needs extending.

### Minor nits

- Extra space in `i <  MAX_EVENT_DEVICES`.
- Alignment of the `snprintf` continuation with the surrounding style (tabs/spaces) for consistency.

Happy to re-review after tweaks.

lordfolken (changes requested):
Thanks for tackling Raspberry Pi touch detection when running without X — the sysfs fallback is a reasonable approach when `UI::event_queue->HasTouchScreen()` stays false.

Below is an updated review against current `master`, with the project PR template and checkboxes.

---

## Pre-Submission Checklist

### Git & Commit History

#### Branch & Rebase
- [ ] Use `git rebase -i` to clean up commit history before PR submission

#### Commit Format & Messages
- [ ] All commits follow the format: `<Component>: <Summary>` (no `src/` prefix)
- [ ] Use present tense in commit messages ("Fix" not "Fixed", "Add" not "Added")
- [x] Commit message body explains *why* the change was made (console without X, sysfs naming, cache)
- [ ] Commit subject uses project format (current: `[menu zoom button] add checking…`)

#### Commit Structure
- [x] Single file, one logical change
- [x] Self-contained (touch detection only)
- [x] No unrelated drive-by changes
- [ ] Each commit builds on current `master` after rebase

---

- [ ] I'm ready to merge

**Note:** CI was green on the PR branch (Apr 2025), but the branch predates `HasTouchScreenHardware()` + `CommandLine::ApplyTouchInputOverride()` on `master`. Rebase and re-run CI.

---

## What's already good

- [x] **Surgical scope** — only `Asset.cpp`; no unrelated changes
- [x] **Real problem** — `HasTouchScreen()` drives layout/zoom hit areas; false on Pi console breaks UX
- [x] **Sysfs cache** — one-time scan, then static state; no per-frame cost
- [x] **Tries event queue first** — sysfs only when `UI::event_queue->HasTouchScreen()` is false
- [x] **Heuristic matches reported hardware** — `"TouchScreen"` fits Goodix (`Goodix Capacitive TouchScreen`)
- [x] **Commit body documents intent** (hid-input naming, performance)

---

## Blockers (must fix before merge)

- [ ] **Rebase onto current `master`** — touch detection is now split:

  - `HasTouchScreenHardware()` — device detection
  - `HasTouchScreen()` — wraps with `CommandLine::ApplyTouchInputOverride()`

  Put the sysfs fallback in **`HasTouchScreenHardware()`**, not by replacing `HasTouchScreen()` wholesale, so `-touchscreen` / `-notouchscreen` keep working.

- [ ] **`open()` failure must `continue`, not `break`** — if `event0` is missing but `event2` exists, `break` stops the loop and never checks later nodes. Gaps in `/sys/class/input/eventN` are common.

- [ ] **`[[gnu::pure]]` vs static cache** — `HasTouchScreen()` is still declared `[[gnu::pure]]` in `Asset.hpp`, but this implementation mutates a `static` cache. Drop `pure` for this configuration, or avoid mutable static state.

---

## Should fix

- [ ] **Scope** — code runs for all `USE_LIBINPUT || USE_WAYLAND` builds, not only Raspberry Pi. Either guard with `#ifdef RASPBERRY_PI` or document sysfs as a general Linux console fallback.
- [ ] **Includes** — add explicit headers: `<unistd.h>` (`read`, `close`), `<cstdio>` (`snprintf`), `<cstring>` (`strstr`); keep `<fcntl.h>`.
- [ ] **Prefer `constexpr` over `#define`** for limits/paths where practical (project style).
- [ ] **Heuristic** — `strstr(..., "TouchScreen")` is case-sensitive; consider case-insensitive match or additional substrings.
- [ ] **Enumeration** — scanning `event0`…`event19` is fragile; consider `opendir("/sys/class/input")` on `event*`, or document why sequential scan is sufficient on Pi.
- [ ] **Comment** — one line naming target hardware/kernel (e.g. Goodix Pi panel) so future readers know when to extend the match.
- [ ] **`NEWS.txt`** — brief UI/embedded note if this fixes Pi zoom/menu touch sizing.

---

## Minor nits

- [ ] Extra space: `i <  MAX_EVENT_DEVICES`
- [ ] `snprintf` continuation indent (tabs/spaces) for consistency with surrounding code

---

## Design question

Default Pi builds use `USE_LIBINPUT=y`. If libinput/udev can be made to report touch reliably on console, that may be preferable long term — but sysfs fallback is still reasonable when enumeration fails or is late.

---

Happy to re-review after rebase and the blockers above.

PR #1742, opened 2025-04-13
fix track line scaling on high res android devices

DavisChappins:
With 7.44 (unsure if other versions) the track line on high resolution android devices is very small. It seems to be scaled down and is the same 3 pixels wide and 400 pixels long but that makes it super tiny on android devices such that its very short and barely visible due to being 3 pixels wide.

This fixes it making it scale with the resolution so it looks the same as it does on other platforms.

Tested on LG G7 thinQ at 1440 x 3120 pixels (picture from same android device)
image


lordfolken (changes requested):
Cool this was something that bothered me.

Please remove the white space on line 57.
Please remove the comment about the include.
Change the commit message to Renderer/Trackline: dynamic scaling for high res devices

DavisChappins:
done

PR #1744, opened 2025-04-18
PEV Start times with second resolution

andrestotorica:
Ref #1099 - added support for Start Open/Close times represented from a time of day, with seconds resolution.

* The internal representation of times in `OrderedTaskSettings::start_constraints` is upgraded to time of day based on seconds (class `RoughTime` is generalized to support better resolution, keeping all stuff like midnight wraparound logic unmodified).

* When a PEV is launched, the start window is now set precisely. Also fixed a problem where the start window was being calculated from `BrokenDateTime::NowUTC()`, which is based on the system clock and thus can be different from GPS time, which is used for logging and calculations. 

* Infoboxes now show precise time remaining. Other UI (Task Manager) and saving to .tsk drops the seconds, so behavior and file format is unchanged - user can only see and set round-minute times on the task. 
_(There is the minor caveat of setting the start open & close via a PEV and then saving the task, seconds will be truncated. However this use case is not realistic and not worth a change the .tsk file format)_

### Testing done
* Unit tests added/updated.
* Linux: tested general UI input for task values. Launched PEV and checked infobox countdown. Checked midnight wraparound, when start or open times are on the next UTC day.
* Android: with Condor simulator as GPS fix source, tested with system clock different than GPS time, verified start window works as expected.

### DIFF
--- a/src/time/RoughTimeDecl.hpp
+++ b/src/time/RoughTimeDecl.hpp
@@ -1,0 +1,10 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The XCSoar Project
+
+#pragma once
+
+/**
+ * This data type stores a time of day with minute granularity.
+ */
+class TimeSinceMidnight;
+using RoughTime = TimeSinceMidnight;
--- a/src/time/RoughTimeDecl.hpp
+++ b/src/time/RoughTimeDecl.hpp
@@ -3,8 +3,13 @@
 
 #pragma once
 
+#include <chrono>
+
+template <typename Duration>
+class TimeSinceMidnight;
+
 /**
  * This data type stores a time of day with minute granularity.
  */
-class TimeSinceMidnight;
-using RoughTime = TimeSinceMidnight;
+using UnsignedMinutes = std::chrono::duration<uint16_t,std::chrono::minutes::period>;
+using RoughTime = TimeSinceMidnight<UnsignedMinutes>;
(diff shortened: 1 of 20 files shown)


lordfolken:
Overall good PR. I will have a look at the details.

src/time/RoughTimeDecl.hpp
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The XCSoar Project
+
+#pragma once
+
+/**
+ * This data type stores a time of day with minute granularity.
+ */
+class TimeSinceMidnight;
+using RoughTime = TimeSinceMidnight;
```
lordfolken: Newline

src/time/RoughTimeDecl.hpp
```diff
@@ -3,8 +3,13 @@
...
+class TimeSinceMidnight;
+
 /**
  * This data type stores a time of day with minute granularity.
  */
-class TimeSinceMidnight;
-using RoughTime = TimeSinceMidnight;
\ No newline at end of file
+using UnsignedMinutes = std::chrono::duration<uint16_t,std::chrono::minutes::period>;
+using RoughTime = TimeSinceMidnight<UnsignedMinutes>;
```
lordfolken: newline

src/time/RoughTimeDecl.hpp
andrestotorica: added newline

src/time/RoughTimeDecl.hpp
andrestotorica: added newline

andrestotorica:
Rebased to master

lordfolken:
Thank you for the detailed work.

PR #1745, opened 2025-04-24
Revert "NMEAChecksum: initializer list is ambigous on OSx"

yorickreum:
See https://github.com/XCSoar/XCSoar/issues/1365.


yorickreum:
The CI build failure showing up here is due to the CMake 4.0 compatibility issue addressed in https://github.com/XCSoar/XCSoar/pull/1714#issuecomment-2828760665.

lordfolken (approved):

PR #1748, opened 2025-05-03
iOS: Use barometric sensor

yorickreum:
Adds support for the internal barometric sensor of iPhones. Mainly @piopawlu's work, I'm just making it pull-ready. I take the liberty to add both of us to the `AUTHORS` file.


src/Device/AndroidSensors.cpp
lordfolken, line 1: If you are renaming a file. please use 
```
git mv <old> <new>
```
That way the history is continuous. Now it seems like the file has been deleted and re added.

yorickreum: Sure. Here, I did not rename a file. I added `src/Device/MobileSensors.cpp` and moved some code from `src/Device/AndroidSensors.cpp` there.

yorickreum:
@coderabbitai review

src/Device/SensorListener.hpp
```diff
@@ -78,4 +80,8 @@ class SensorListener {
   virtual void OnSensorStateChanged() noexcept = 0;
   virtual void OnSensorError(const char *msg) noexcept = 0;
 #endif // ANDROID
+#if defined(ANDROID) || (defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE)
+  virtual void OnBarometricPressureSensor(float pressure,
+                      float sensor_noise_variance) noexcept = 0;
+#endif // ANDROID or iPhone
 };
```
lordfolken, line 87: This would limit it to only android/ios. Is this really true?

src/Apple/InternalSensors.cpp
```diff
@@ -130,6 +130,69 @@ void InternalSensors::Init()
...
+                NSLog(@"Error querying motion activities: %@", error);
+                query_error = YES;
+                return;
+            }
+        }];
+        
+        // Exit if there was an error during the permission check
+        if (query_error) {
+          return;
+        }
```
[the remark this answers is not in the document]
yorickreum: Fair, did update.

src/Apple/InternalSensors.cpp
```diff
@@ -138,4 +201,9 @@
 void InternalSensors::Deinit()
 {
   [location_manager stopUpdatingLocation];
+  #if defined(TARGET_OS_IOS) && TARGET_OS_IOS
+  if (altimeter != nullptr) {
+    [altimeter stopRelativeAltitudeUpdates];
+  }
+  #endif
```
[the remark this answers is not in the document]
yorickreum: Fair, did update.

src/Device/SmartDeviceSensors.cpp
```diff
@@ -0,0 +1,124 @@
...
+    basic.track = Angle::Degrees(bearing);
+    basic.track_available.Update(basic.clock);
+  } else
+    basic.track_available.Clear();
+
+  if (hasSpeed) {
+    basic.ground_speed = ground_speed;
+    basic.ground_speed_available.Update(basic.clock);
+  }
+
```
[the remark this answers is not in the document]
yorickreum: Not sure about this one. Was like this already before I touched it... @lordfolken, do you have an opinion on whether to `.Clear()` here or not?

src/Device/SensorListener.hpp
yorickreum: Mhhh. It was limited to Android before this PR. Now it's limited to Android and iPhone...

lordfolken (approved):

PR #1751, opened 2025-05-05
Airspace/AirspaceVisibilty.cpp: Also check type

### LINKED ISSUE #1746: Airspace entry warnings on invisible (unconfigured) airspaces

7.44

all other versions before the AY (airspace type) was introduced

System information
------------------

UNIX

use airspace filter (uncomment/delete warning and display to airspace class E)

use France airspace and the following test data with the replay function

[Test24.txt](https://github.com/user-attachments/files/19930395/Test24.txt)

the plane enters airspace class E/TMA Sarre

Image

Expected behavior
-----------------

the uncommented airspace (class) "E" should not issue a warning

Actual behavior
---------------

visible and audio warning is issued. The reason is the AY record with TMA where "display and warning" is active

Bug

Disable TMA warnings is not the solution; they exist also in combination with airspace class D

kobedegeest:
With the extended openair format airspace have at class and type We need to check both of these to know weather or not to draw the airspace

Renamed the boolean to reflect that it now checks both.

closes #1746

### DIFF
--- a/src/Airspace/AirspaceVisibility.cpp
+++ b/src/Airspace/AirspaceVisibility.cpp
@@ -7,10 +7,10 @@
 #include "Renderer/AirspaceRendererSettings.hpp"
 
 bool
-IsAirspaceTypeVisible(const AbstractAirspace &airspace,
+IsAirspaceTypeOrClassVisible(const AbstractAirspace &airspace,
                       const AirspaceRendererSettings &renderer_settings)
 {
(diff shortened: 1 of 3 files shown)


bomilkar:
Just to be clear: the issue is not only drawing. Equally important is that it can generate Warnings when approaching these objects.

lordfolken (approved):

PR #1757, opened 2025-05-08
XCVario:  S2F parsing docs & XCVario Tests for Driver

lordfolken:
Your PR should:
  * no fixup commits


lordfolken:
So the commit was based on wrong information. Nevertheless i"d like the test for the driver and will correct the documentation.

iltis42:
> So the commit was based on wrong information. Nevertheless i"d like the test for the driver and will correct the documentation.

Yes, a test is okay, the fix in docu as well. The code commit not.

Issue #1370, opened 2024-03-06
Release Capability


lordfolken:
The project isn't in the position to make a release and push it to end users right now.

Current Situation:

On F-droid there is a release of xcsoar, but its basically a 3rd party compilation, that seems to be triggered by releases on github. However currently its stuck on V7.41

Then there is download.xcsoar.org/releases for all current releases. (Kobo,Linux,Android etc)

There is a play store release by Max tied to his personal account.

At this moment the steps to create a release aren't defined or documented.

zinuzoid:
Regards Play store, I manage my fork (XCSoar JET) in Play store so let me know if I can help. 

First step is probably to create a company publisher account (XCSoar) them we can transfer the app from personal account to the new one.

lordfolken:
@zinuzoid your help would be very welcome. 

I tried applying for an org account on the google play store. This would need a D-U-N-S number. This number seems to be available only to legal entities.

zinuzoid:
@lordfolken I can try to create the org account. I just want a confirmation that we're committed on this route before I proceed and pay the account fee? (otherwise I would pay for nothing)

This route means: we'll use the new play store developer account going forward by either
- Migrate the current app from personal account to the new org account
- Or start from scratch (probably lose all stats)

lordfolken:
> @lordfolken I can try to create the org account. I just want a confirmation that we're committed on this route before I proceed and pay the account fee? (otherwise I would pay for nothing)
I would have preferred so, but in lack of a DUNS number this doesn't seem possible. Do you have access to one?

> This route means: we'll use the new play store developer account going forward by either
> 
>     * Migrate the current app from personal account to the new org account
Max has left the project and removed his play store entry for xcsoar. So this might already be too bad. 
I don't have the signing keys for his release, so I'm afraid that its going to be option 2:

>     * Or start from scratch (probably lose all stats)

zinuzoid:
I've got DUNS number and paid the fee, but now they needed a hard document for verification. One of any
- Certificate of incorporation
- VAT registration certificate
- Certificate of registration of charity
🤔

Another less bad option is I can turn this account (newly created) back to individual and we share the credential, not the best but still an option.

lordfolken:
I think the shared credential would be the way to go. I would add this as a github secret, and we can then use it in a pipeline that hopefully pushes to firebase/play store.

lordfolken:
[...]

PR #1760, opened 2025-05-15
Add Android App Bundle build capability

andrestotorica:
Added support for building Android App Bundles for the Google Play Store. For background see issue #1370 .

- Using `TARGET=ANDROIDFAT`, Makefile now creates an App Bundle -suitable for distribution to the Play Store- and derives a "fat" APK from it -for other distribution channels-.
  - Other `ANDROID*` targets create single arch APKs as before, but they are derived from partial bundles as well.
  - For testing and transitioning, this behavior is currently enabled by adding the argument `ANDROID_BUNDLE_BUILD=y` to the make command.

- Added a new Github Actions job for the Bundle. Currently it only builds and exports the bundle artifact to the workflow. This can be used for manually uploading into Play Store and can be extended to automate this step in the future.

**Testing done**: downloaded APK from both PR and Release GH workflows and installed correctly to my Android phone. Tested the bundle locally with bundletool, however I couldn't test Play Store upload as that required paying the Google dev account (which I don't really have another use for). @lordfolken / @zinuzoid if you could test uploading the .aab to a test track on the project's account that'd be helpful.


lordfolken:
This looks good. Will test with the play account.

lordfolken (approved):

PR #1763, opened 2025-05-16
No netcoupe anymore

mlep:
Since Netcoupe ended up in 2024 (see https://www.netcoupe.net/Main_archives.aspx), the French Gliding Federation moved to Weglide. So, there is no need to keep netcoupe in XCSoar.

Note: I am doing this PR from GitHub web interface, and I did not find how to squash the various commits


lordfolken:
Every commit must compile and be atomically complete.

mlep:
I did not use VS. What you are seeing is what I could do using the Github web interface (default commit message from .. Github!). I am afraid this is the most I can do (and, as stated initially, I did not find a way to squash the commits...). Anyway, if this PR is of no use because of its defects, it will be easy for somebody more skilled than me to redo it.

lordfolken:
Thank you for your work.

PR #1773, opened 2025-05-24
Allow users to select a default setting for "Arm start manually"

cjwoolley:
This update adds a new option "Arm start manually" to the list of default task rules.

It can be set in Config->System->Task Defaults->Task Rules->Arm start manually to either "On" of "Off"

Previously the default setting when creating a new task was "On" for AAT tasks and "Off" for all other task types.


cjwoolley:
I'll resubmit the request as I've just noticed that "Arm start manually" does not display when "FAI badges/records" task type is selected.

PR #1776, opened 2025-05-25
Set AAT task type "Arm start manually" default to "Off"

cjwoolley:
This change makes the "Arm start manually" default consistent across all task types.


lordfolken (approved):

lordfolken:
Lgtm. Thanks.

PR #1779, opened 2025-05-27
default.xci: add Weglide button

kobedegeest:
This opens file picker to select igc file to upload to weglide

Your PR should:
  * no fixup commits

### DIFF
--- a/Data/Input/default.xci
+++ b/Data/Input/default.xci
@@ -624,6 +624,14 @@ location=5
 
 mode=Config2
 type=key
+data=7
+event=UploadIGCFile
+event=Mode default
+label=Weglide Upload
+location=6
+
+mode=Config2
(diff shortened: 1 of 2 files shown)


Data/Input/default.xci
```diff
@@ -622,6 +622,14 @@ event=Mode default
 label=Profiles
 location=5
 
+mode=Config2
+type=key
+data=7
+event=UploadIGCFile
+event=Mode default
+label=Weglide
```
lordfolken, line 630: So thank you for creating the change. 

Do you think the label is sufficient with "Weglide"?

lordfolken (commented):
How would you inform users of this change? Do you think a changelog entry is susficient, or do we need some documentation in the manual?

PR #1785, opened 2025-06-01
increase font size and waypoint landable size

freevariode:
font size and waypoint landable size is scalable up to 300%. Is necessary for some displays with a high resolution at Android boards


lordfolken:
i believe this now to be addressed in https://github.com/XCSoar/XCSoar/pull/2395/changes/01e91dff44cbd7fabd52f35173c1e1aa3a725136

PR #1791, opened 2025-06-08
improve parsing of OpenAIP files

tschulte:
See https://github.com/naviter/seeyou_file_formats/blob/main/OpenAir_File_Format_Support.md

AC is "UNC", not "UNCLASSIFIED" in OpenAIP files.
Add possible AY values.


lordfolken:
Thank you for pointing this out and providing a PR.

This parser was written before the OpenAIr 2.0 spec was out. It was based on an OpenAir extended format based on the docs formerly on winpilot.com. And also based on what OpenAIP did. 

Winpilot seems to have passed on the spec to @naviter, which is doing a good job of now maintaining the standard. 

For the PR:
- [ ]  Please replace the old UNCLASSIFED with the new UNC (also the other types)
- [ ]  For new Airspace types they need to be added:
          * src/Dialogs/Airspace/AirspaceList.cpp  (Dialog to show when displaying airspace info
          * src/Engine/Airspace/AirspaceClass.hpp (General engine definition)
          * src/Renderer/AirspaceRendererSettings.cpp (Display coloring)

PR #1794, opened 2025-06-09
Openair2

lordfolken:
Todo:
- [x] Station Name field stays even if there is no string
- [x] Display Transponder Code for airspace
- [x] Ability to send transponder code to transponder
- [x] Station Name should be transmitted to the radio if present
- [ ] Airspace opening hours implementation missing


lordfolken:
- [ ] Station Name field stays even if there is no string
- [ ] TransponderCode always is 0
- [ ] Station Name should be transmitted to the radio if present
- [ ] Airspace opening hours implementation missing

src/Airspace/AirspaceParser.cpp
```diff
@@ -62,8 +62,7 @@ static constexpr AirspaceClassStringCouple airspace_class_strings[] = {
   { "G", CLASSG },
   { "RMZ", RMZ },
   { "MATZ", MATZ },
-  { "GSEC", WAVE },
-  { "UNCLASSIFIED", UNCLASSIFIED },
```
kobedegeest, line 66: were these removals intended?

lordfolken: According to 
https://github.com/naviter/seeyou_file_formats/blob/main/OpenAir_File_Format_Support.md#ay-airspace-type 
* unclassified is now UNC. 
* GSEC is now Gliding Sector.

kobedegeest: I understand but this change breaks existing files from working 

and the reason i asked if it was intended is because the addition of "UNC" is a separate commit from the deletion of "UNCLASSIFIED"

PR #1796, opened 2025-06-10
reformat AirspaceGlue.cpp, SiteConfigPanel.cpp, DataField.cpp, Startu…

Eldalie:
reformat AirspaceGlue.cpp, SiteConfigPanel.cpp, DataField.cpp, StartupDialog.cpp, Keys.hpp , Map.hpp, PathValue.cpp, Profile.cpp, ProfileRowFormWidget.cpp, RowFormWidget.hpp

I accidentally reformatted these files in my previous PR, so I prefer to make a separate PR for formatting only.

Would you like me to reformat the whole project as well ?

### DIFF
--- a/src/Airspace/AirspaceGlue.cpp
+++ b/src/Airspace/AirspaceGlue.cpp
@@ -3,22 +3,22 @@
 
 #include "Airspace/AirspaceGlue.hpp"
 #include "Airspace/AirspaceParser.hpp"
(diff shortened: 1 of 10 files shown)


lordfolken:
General reformatting isn't something we want. The reason being that changes the git history of a file, making it hard to track down bugs through git history identifying a breaking commit. 
If there is no functional change its therefore not helpful of reformatting a file.

PR #1812, opened 2025-07-08
AirspaceBuilder.cpp: ignore height filter

### LINKED ISSUE #1755: Map elements at this location to list also invisible airspaces
XCSoar version
--------------

7.43

When clicking on the map you get a list of "Map elements at this location".
The airspaces listed here are limited to those visible at this location.
I'd like to also see the airspaces listed which are **not** visible on the map.

The airspace Filter may turn off (inadvertently) the Display attribute of certain airspace Class or Type.
This feature would improve the situational awareness.

kobedegeest:
closes #1755 

## Summary by CodeRabbit

* **Refactor**
  * Simplified airspace visibility handling for map items, enhancing consistency in how airspaces are displayed without altering the user interface.


lordfolken:
I'm not against such a change. But it would have to be clearly visible in the list that these airspace where filtered. I really don't want to introduce an ui inconsistency, where the user has to render part of the application in his head.

PR #1813, opened 2025-07-11
Improvement to Nano flight download

### LINKED ISSUE #1252: LXNav devices  - Bluetooth flight download / declaration intermittent

Current version 7.35, run on Pixel 4a. Tested on 2 pcs Nano3 and one Nano4 loggers

Expected behavior
-----------------

XCSoar should be able to download series of flight/declare without issues

Actual behavior
---------------

Currently I can sually download first flight without troubles, but if I select to download next flight it usualy ends with failed download. After several failed atttempts it can start to work again. Sometimes download fails during the data transmit/task declaration. Sometimes it fails even with first flight. LXNav NanoConfig sw, running on the same phone can download series of flights rock solid

Connect LX Nano3/4 via BT and try to download series of flight records from the logger

No

No

kobedegeest:
Closes #1252

- [x] General improvements
- [x] Fix first download attempt due to longer lookup
- [x] Slow down reconnect such that it actually works
- [x] Store download progress in uniek file so we can resume failed downloads
- [x] Update splash screen to show we resumed download
- [x] Exit flight list after failed download (since comm is down with device need to go to device list to reconnect)
- [x] Clean up xcsoar log entries
- [ ] Clean up commit history, can still merge the three SlowReconnect commits to one.

## Summary by CodeRabbit

* **New Features**
  * Flight downloads resume from partial files, show resume/progress messages, and finalize partial files on success.
  * Devices gain a delayed reconnect flow; reconnect UI disables while a reopen is pending.

* **Bug Fixes**
  * More resilient flight downloads: larger per-request batches, periodic flushes, safer retry/timeout handling, and early exit on fatal download errors.

* **Tests**
  * Added fake log inputs to test and replay suites.

### DIFF
--- a/src/Device/Driver/LX/NanoLogger.cpp
+++ b/src/Device/Driver/LX/NanoLogger.cpp
@@ -283,8 +283,8 @@ DownloadFlightInner(Port &port, const char *filename, BufferedOutputStream &os,
   unsigned row_count = 0, i = 1;
 
   while (true) {
-    /* read up to 32 lines at a time */
-    unsigned nrequest = row_count == 0 ? 1 : 32;
+    /* read up to 50 lines at a time */
+    unsigned nrequest = row_count == 0 ? 1 : 50;
     if (row_count > 0) {
       assert(i <= row_count);
       const unsigned remaining = row_count - i + 1;
--- a/src/Device/Driver/LX/NanoLogger.cpp
+++ b/src/Device/Driver/LX/NanoLogger.cpp
@@ -307,7 +307,11 @@ DownloadFlightInner(Port &port, const char *filename, BufferedOutputStream &os,
       }
 
       TimeoutClock timeout(std::chrono::seconds(row_count == 0 ? 20 : 2)); // using row_count to detect first request
-      const char *line = reader.ExpectLine("PLXVC,FLIGHT,A,", timeout);
+      const char *line = nullptr;
+      try {
+        line = reader.ExpectLine("PLXVC,FLIGHT,A,", timeout);
+      } catch (...) {
+      }
       if (line == nullptr || !HandleFlightLine(line, os, i, row_count)) {
         if (request_retry_count > 5)
           return false;
--- a/src/Device/Driver/LX/NanoLogger.cpp
+++ b/src/Device/Driver/LX/NanoLogger.cpp
@@ -322,8 +322,9 @@ DownloadFlightInner(Port &port, const char *filename, BufferedOutputStream &os,
         LogError(std::current_exception(), "Download failing");
       }
       if (line == nullptr || !HandleFlightLine(line, os, i, row_count)) {
-        if (request_retry_count > 5)
-          return false;
+        if (request_retry_count > 5) {
+          throw std::runtime_error("Flight download failed: maximum retries exceeded");
+        }
 
         /* Discard data which might still be in-transit, e.g. buffered
            inside a bluetooth dongle */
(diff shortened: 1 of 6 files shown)


Sundown3867:
I have tried all sugested improvemnts including the delay in first download attempt and my observation are that downloads are indeed more robust but first attempt is still failing

Sundown3867:
Hey AI, is there any other reason why src/Device/Driver/LX/NanoLogger.cpp is still failing the firs attempt

lordfolken:
So I would be all for the buffer size increase to 50.
The missing flush statement should be a separate commit.

Im not a fan of increasing timeouts, as that hides underlying problems like demonstrated here. However I do agree that the 2 second one is too short, especially if the device has a lot of flights stored on it.

Maybe cap the timeout around 20seconds or so.

src/Device/Driver/LX/NanoLogger.cpp
```diff
@@ -309,10 +309,10 @@ DownloadFlightInner(Port &port, const char *filename, BufferedOutputStream &os,
         request_retry_count++;
       }
 
-      TimeoutClock timeout(std::chrono::seconds(2));
+      TimeoutClock timeout(std::chrono::seconds(2*request_retry_count));
       const char *line = reader.ExpectLine("PLXVC,FLIGHT,A,", timeout);
       if (line == nullptr || !HandleFlightLine(line, os, i, row_count)) {
```
kobedegeest, line 353: @coderabbitai if this if statement is never true how can the download return a failed message to the user? where does it fail?

src/Device/Driver/LX/NanoLogger.cpp
```diff
@@ -309,8 +310,13 @@ DownloadFlightInner(Port &port, const char *filename, BufferedOutputStream &os,
         request_retry_count++;
       }
 
-      TimeoutClock timeout(std::chrono::seconds(2));
-      const char *line = reader.ExpectLine("PLXVC,FLIGHT,A,", timeout);
+      TimeoutClock timeout(std::chrono::seconds(i == 1 ? 20 : 2));
+      const char *line = nullptr;
+      try {
+        line = reader.ExpectLine("PLXVC,FLIGHT,A,", timeout);
```
kobedegeest, line 344: @coderabbitai if reader.ExpectLine times out how to properly reset the connection with the device

lordfolken:
Can you attach and capture a debug log?

src/Device/Driver/LX/NanoLogger.cpp
```diff
@@ -309,8 +310,14 @@ DownloadFlightInner(Port &port, const char *filename, BufferedOutputStream &os,
...
+      } catch (...) {
+        LogFormat("Communication with logger timedout, tries: %d, line: %d", request_retry_count, i);
```
kobedegeest: @coderabbitai why is reference to LogFormat() not found? Should be defined in LogFile.hpp which is included ...

kobedegeest: @lordfolken could you have a check why this doesn't work? 
I can just leave these out but seems good to write the failure to the logfile

lordfolken: @kobedegeest This is the unit test that is failing. (make check TARGET=UNIX -j8) The logFile.hpp needs to be included somewhere near TestDriver.cpp

kobedegeest:
@coderabbitai this loses the value of resume_row

kobedegeest:
@coderabbitai trying to borrow the device again has 100% failure rate, please look into why and propose fix

PR #1814, opened 2025-07-12
lib/lua: Remove manual patch for iOS

yorickreum:
Imo Lua handles this acceptably well if `LUA_USE_IOS` is set, so no reason for the manual patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for iOS builds by adding a specific compiler flag during the build process.

* **Refactor**
  * Removed the `os_execute` function from Lua on iOS to enhance platform security and compliance.


yorickreum:
@coderabbitai Hey, AI, you realize this PR deletes the .diff file, so we are not doing anything to `os_execute` anymore...

PR #1833, opened 2025-07-29
iOS: build/python/build/libs.py: Enable SDL_TIMERS

yorickreum:
I finally figured out the root cause for https://github.com/XCSoar/XCSoar/issues/1693. Without the `SDL_TIMERS` module, SDL delays just won't work on iOS (see my remark here: https://github.com/XCSoar/XCSoar/issues/1693#issuecomment-3134062145), hence the main event loop was completely busy no matter how we implemented it... Enabling the `SDL_TIMERS` module makes SDL_Delay work as expected and immediately brings down the CPU load (around 20% now in the iOS simulator when clicking around in the app, < 5% for an empty dummy event loop, compared to always > 100% before).

## Summary by CodeRabbit

* **Chores**
  * Updated SDL2 library build configuration to enable the timers subsystem, ensuring SDL_Delay() works correctly on iOS.


yorickreum:
From a quick investigation, it seems to have been originally disabled here https://github.com/XCSoar/XCSoar/commit/c27049b265eb45b8b60107c31027ee4ddfb1eeb8, I don't know why.

lordfolken (approved):

Issue #1834, opened 2025-07-30
iOS: Prevent app restart XCSoar was running in background


WolfgangDrescher:
XCSoar versions having the problem
-------------------------------------------------

Unreleased XCSoar 7.44

XCSoar versions not having the problem
-------------------------------------------------

unknown

System information
------------------

iPhone with iOS 18.5

Steps to reproduce the behavior
-------------------------------

1. Open XCSoar and interact with the map
2. Go back to the iPhone home screen (XCSoar is running in the background now)
3. Wait until the iPhone locks the screen (set it to 30s for testing) and maybe wait another minute or so
4. Reopen XCSoar

Expected behavior
-----------------

XCSoar should continue running where you left it.

Actual behavior
---------------

XCSoar crashes, and you have to restart from the fly/sim screen.

Do you have any idea what may have caused this?
-----------------------------------------------

See https://github.com/XCSoar/XCSoar/issues/1693#issuecomment-3066702973

Do you have an idea how to solve the issue?
-------------------------------------------

<details>
  <summary>Click to see my suggestions in `EventLoop::Dispatch`</summary>

```cpp
void
EventLoop::Dispatch(const Event &_event)
{
  const SDL_Event &event = _event.event;

  switch (event.type) {
  case SDL_APP_WILLENTERBACKGROUND:
    LogFormat("[DEBUG] SDL_APP_WILLENTERBACKGROUND");
    glFlush();
    glFinish();
    // appIsInBackground = true;
    break;

  case SDL_APP_DIDENTERFOREGROUND:
    // appIsInBackground = false;
    LogFormat("[DEBUG] SDL_APP_DIDENTERFOREGROUND");
    break;

  case EVENT_CALLBACK: {
    Callback callback = (Callback)event.user.data1;
    callback(event.user.data2);
  } break;

  default:
    if (top_window != nullptr) {
      if (top_window->OnEvent(event) && _event.IsUserInput()) ResetUserIdle();
    }
    break;
  }
}
```

Unfortunately, that had no effect. The `SDL_APP_WILLENTERBACKGROUND` and `SDL_APP_DIDENTERFOREGROUND` events are being properly detected and logged.

</details>

yorickreum:
When the app crashes, do we see anything in the console? Like a context loss / failed to make context error?
Can't find anything in the XCSoar log on my device. Might try checking that with the simulator or my phone attached to my Mac...

WolfgangDrescher:
<details>
[...]

PR #1835, opened 2025-08-01
iOS: ui/canvas/sdl/TopCanvas.cpp: Adds check if iOS app is in foreground

yorickreum:
This pull request implements the `TopCanvas::IsReady()` method for iOS. Fixes https://github.com/XCSoar/XCSoar/issues/1834.

Previously, `TopCanvas::IsReady()` always returned `true` on iOS, unlike Android which properly checks EGL surface availability. This caused the app to attempt rendering operations even when backgrounded, leading to crashes and wasting resources.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved app stability on iOS by ensuring drawing only occurs when the app is active and in the foreground, reducing the risk of crashes when the app is inactive or in the background.


yorickreum:
@coderabbitai review

yorickreum:
Thanks for testing. This is quite tedious to debug.

The version with the full PR also crashed on me again.

Without https://github.com/XCSoar/XCSoar/commit/ffc2869c2ff7040ac58ed7f29f9dd273aad90a9c and with the Location Setting on _Allow always_ it didn't crash on my so far today. Not sure if that really makes a difference. Will keep observing it for a while.

PR #1839, opened 2025-08-11
Fix for custom DPI value not being applied.

kicune:
Check on custom DPI value added to the condition for main window re-initialization.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fonts now refresh immediately when a custom DPI value is applied, ensuring consistent sizing.
  * UI reinitializes font rendering on DPI changes, complementing existing scaling options.
  * Improves readability and visual consistency on high-DPI displays by updating both general and title fonts after changing custom DPI settings.

### DIFF
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -219,7 +219,7 @@ MainWindow::InitialiseConfigured()
 {
   const UISettings &ui_settings = CommonInterface::GetUISettings();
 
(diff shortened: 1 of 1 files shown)


lordfolken:
@kicune makes sense, also as we are initializing custom dpi in Initialize(). Could you please change your commit message: 
MainWindow: force Initialise when custom dpi is set or changed

lordfolken (approved):
Thanks!

PR #1850, opened 2025-08-23
iOS: Improve audio support

WolfgangDrescher:
This PR improves the audio playback on iOS.

```objc++
[session setCategory:AVAudioSessionCategoryPlayback
         withOptions:AVAudioSessionCategoryOptionMixWithOthers
               error:&error];
```

The `AVAudioSessionCategoryPlayback` category allows sounds to play even when the iPhone is muted. The `AVAudioSessionCategoryOptionMixWithOthers` option enables mixing the audio with playback from other apps.

Instead of using `SoundUtil::RawToWav`, we could alternatively modify `resource.mk` so that, for iOS, not the "raw" files are copied as resources (currently done with `$(Q)sox -V1 $< --bits 16 --rate 44100 --channels 1 $@`). Instead, the original WAV files would be copied directly.

## Summary by CodeRabbit

* **New Features**
  * Added Apple platform audio playback for in-app sounds, using the system audio session with support for mixing with other audio.
  * Integrated Apple-specific startup/shutdown handling to initialize and clean up audio services.

* **Chores**
  * Updated macOS build configuration to include required Apple frameworks and sources.
  * Added a new contributor to the credits.


WolfgangDrescher:
@yorickreum Have you had time to take a closer look at this PR yet? Also there is still the question of whether `SoundUtil::RawToWav` makes sense at all, or whether it would be better to add the WAV (not raw) files for iOS directly as data in `resource.mk`.

WolfgangDrescher:
Currently, the build fails with this change (`'AVAudioSession' is unavailable: not available on macOS`). However, the macOS build doesn’t work for me anyway, since the cursor isn’t visible. Is this build intended to be used with keyboard shortcuts only? But I didn’t test it either…

Should I wrap it in `#if TARGET_OS_IPHONE` for now?

yorickreum:
> > However, the macOS build doesn’t work for me anyway, since the cursor isn’t visible.
> 
> Never mind, when I build in Xcode (for macOS), I do see the cursor in XCSoar, but the app seems quite broken, especially the graphics. However, the build from the Mac App Store doesn't work at all.

The version in the app store is actually the iOS app running via Catalyst ... not functional, pretty useless atm.
Ideally we should fix the native macOS app at some point to an extent where we can actually publish it on the AppStore.
Currently it's set to use software rendering instead of OpenGL, that's why the graphics look so broken ... I guess that would be the first thing to address here. But tbh this is quite far down on my priority list currently, as I don't think many people will use the app directly on macOS. Might be useful for us to debug and develop without the iOS simualtor being involved, though, maybe...

> I added some `TARGET_OS_IPHONE` conditions, and now the macOS build works again and sounds are played using `PCMResourcePlayer` (I tested it and it works). The newly added conditions are a bit verbose, but I prepared them with the idea that we might switch from `PCMResourcePlayer` to `Apple/SoundUtil` in the future, and I didn't want to keep `SoundUtil` iOS-only. Any thoughts on this?

Naive question: What is the new implementation via `AVFoundation` doing better than the `PCMResourcePlayer` then?

For now we can definitely keep the new `SoundUtil` iOS-only imo. We just should add a comment then at the beginning of the file. (That's something I often miss in the XCSoar codebase -- just some comments that explain why a file/function is there and what its scope is...).

I put some more inspiration here: https://github.com/WolfgangDrescher/XCSoar/pull/3

WolfgangDrescher:
@lordfolken, do you have any change requests for this PR? As far as I'm concerned, it's ready to merge.

lordfolken (approved):
Well probably will have to find a way out of tje ifdef hell at some point. Lgtm

PR #1864, opened 2025-09-02
Allow per-page InfoBox geometry

### LINKED ISSUE #1863: Allow per-page InfoBox geometry
XCSoar version
--------------

current master (949e8f1)

It would be very convenient if one could not only set the geometry globally, but also define a separate layout for each page (`InfoBox Set`).

I would like to have some info boxes displayed on the AUX-1 page, for which I no longer have space on the other sets (Circling, Cruise, FinalGlide). I like to keep the three standard sets clean with only 8 InfoBoxes, but on the AUX-1 page, I would like a layout with e.g. 12 InfoBoxes. This is currently not possible.

WolfgangDrescher:
This PR targets issue https://github.com/XCSoar/XCSoar/issues/1863. I created this draft PR to get some feedback. I will clean up and squash the commits later.

Some comments and questions:

- [x] I think it would be better not to add `InfoBoxSettings::Geometry::INHERIT`, but instead create a getter that returns the global page geometry when none is set for a specific page. This way, none of the switch cases need to handle it, and we guarantee that a valid `InfoBoxSettings::Geometry` is always passed.
- [ ] Is `InfoBoxSettings -> Panel -> geometry` the correct place to store this new option?
- [x] How can I persist `AddEnum(_("Page geometry"),` in the user config in `dlgConfigInfoboxes.cpp`? Currently, this setting is lost when the app is restarted.
- [x] Currently, when a different `Page geometry` is selected in `Config -> System -> Look -> InfoBox Sets -> AUX 1`, the InfoBox preview does not update to reflect the new layout.
- [x] Add to `NEWS.txt`

## Summary by CodeRabbit

* **New Features**
  * InfoBox panels can now use individual page geometry settings or inherit the global layout.
  * Geometry changes update the layout, controls, panel count, and previews immediately.
  * Panel-specific geometry choices are preserved in profiles and restored when settings are loaded.

* **Bug Fixes**
  * Active layouts now consistently use the selected panel’s InfoBox geometry.
  * Resetting an InfoBox panel correctly restores global geometry inheritance.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -26,6 +26,8 @@ Version 7.45 - not yet released
   - snail trail: Catmull-Rom smoothing between fixes; vario colouring samples
     high-rate netto from the merge thread between GPS points #2447
 * ui
+  - InfoBox sets can now define their own page geometry, overriding the default
+    InfoBox geometry configured in the screen layout settings #1864
   - infoboxen: refresh titles after changing the interface language #2314
   - infoboxen: add "Home" InfoBox (waypoint name, arrival height at home,
     distance; tap to change home waypoint) #2320
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -445,6 +445,8 @@ Version 7.45 - not yet released
   - Fix intermittent white-map hang on startup for non-OpenGL builds when
     drawing began before map bounds were ready #2734
 * ui
+  - InfoBox sets can now define their own geometry, overriding the default
+    geometry configured in the screen layout settings #1863
   - infoboxen: refresh titles after changing the interface language #2314
   - infoboxen: add "Home" InfoBox (waypoint name, arrival height at home,
     distance; tap to change home waypoint) #2320
--- a/src/ActionInterface.cpp
+++ b/src/ActionInterface.cpp
@@ -337,6 +337,11 @@ ActionInterface::UpdateDisplayMode() noexcept
   state.panel_name = gettext(panel.name);
 
   UpdateMapScalePageInfo(state);
+
+  /* the active panel may have changed, and a panel may override the
+     global InfoBox geometry */
+  if (main_window != nullptr)
+    main_window->CheckInfoBoxGeometry();
 }
 
 void
--- a/src/Dialogs/Settings/Panels/InfoBoxesConfigPanel.cpp
+++ b/src/Dialogs/Settings/Panels/InfoBoxesConfigPanel.cpp
@@ -14,6 +14,8 @@
 #include "Language/Language.hpp"
 #include "UIGlobals.hpp"
 #include "Look/Look.hpp"
+#include "MainWindow.hpp"
+#include "UIState.hpp"
 
 class InfoBoxesConfigPanel final
   : public RowFormWidget {
@@ -40,12 +42,18 @@ InfoBoxesConfigPanel::OnAction(int id) noexcept
     dlgConfigInfoboxesShowModal(UIGlobals::GetMainWindow(),
                                 UIGlobals::GetDialogLook(),
                                 UIGlobals::GetLook().info_box,
-                                InfoBoxManager::layout.geometry, data,
+                                settings.ResolveGeometry(data),
+                                data,
                                 i >= InfoBoxSettings::PREASSIGNED_PANELS);
   if (changed) {
     Profile::Save(Profile::map, data, i);
     Profile::Save();
     ((Button &)GetRow(i)).SetCaption(gettext(data.name));
+
+    // If the edited panel is currently active, rebuild layout so
+    // the new geometry apply immediately
+    if (CommonInterface::GetUIState().panel_index == i)
+      CommonInterface::main_window->ReinitialiseLayout();
   }
 }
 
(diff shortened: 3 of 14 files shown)


WolfgangDrescher:
I have now managed to resolve most of the open issues and questions. Feedback on this PR would be great, so I know in which direction to develop it further before I remove the draft flag.

WolfgangDrescher:
This PR also randomly and unintentionally solves the problem that when the `ThermalAssistant` is open, switching from a page with InfoBoxes to a page without InfoBoxes (map only), the `ThermalAssistant` was not displayed at the very bottom, but above the non-existent infoboxes, thus covering more of the central area of the map than necessary.

WolfgangDrescher:
Are there any thoughts or suggestions regarding this PR @lordfolken?

lordfolken:
I like the general direction this is going.
This way one can for example display the vario when in circling but otherwise not.

lordfolken:
The main things I think we need to look at: how does this behave when adding/modifying/deleting and how do we store the profilekeys, At the moment these are string matches, but that may be ok.

WolfgangDrescher:
I thoroughly tested the InfoBoxes yesterday and didn't notice any issues when adding, editing, or deleting `InfoBoxPanel` items in any scenario. Was there something specific you had in mind?

Currently, the profile key for the panel geometry is stored, for example, as e.g. `InfoBoxPanel3Geometry` where `255` means that the layout is inherited (`INHERIT_GEOMETRY = 0xff`). In this case, the panel falls back to the `InfoBoxGeometry` value.

`InfoBoxPanels` cannot be deleted, and pages only reference panels. So I don't see any problems with this.
```
InfoBoxGeometry="14"
InfoBoxPanel3Box0="14"
InfoBoxPanel3Box1="45"
InfoBoxPanel3Box10="0"
InfoBoxPanel3Box11="0"
InfoBoxPanel3Box12="0"
InfoBoxPanel3Box13="0"
InfoBoxPanel3Box14="0"
InfoBoxPanel3Box15="0"
InfoBoxPanel3Box16="0"
InfoBoxPanel3Box17="0"
InfoBoxPanel3Box18="0"
InfoBoxPanel3Box19="0"
InfoBoxPanel3Box2="11"
InfoBoxPanel3Box20="0"
InfoBoxPanel3Box21="0"
InfoBoxPanel3Box22="0"
InfoBoxPanel3Box23="0"
InfoBoxPanel3Box3="42"
InfoBoxPanel3Box4="18"
InfoBoxPanel3Box5="41"
InfoBoxPanel3Box6="39"
InfoBoxPanel3Box7="36"
InfoBoxPanel3Box8="61"
InfoBoxPanel3Box9="0"
InfoBoxPanel3Geometry="14"
InfoBoxPanel3Name="AUX-1"
```

```
Page0Bottom="0"
Page0InfoBoxMode="0"
Page0InfoBoxPanel="3"
Page0Layout="2"
Page0Main="0"
```

WolfgangDrescher:
Is there any feedback on this so I know how to finalise this PR?

DanD222:
I've just had a look - I like that the `InfoBox geometry` changes are reflected immediately in the `InfoBox Sets` assignment page for the given Circling/Cruise pages (comparing this to a fairly recently compiled version where I needed to exit the settings for the `InfoBox geometry` changes to propagate to the `InfoBox Sets` assignment page - I don't know if that's something you've fixed or if that's something that's been fixed in the meantime).

What's a little strange is when, for example, assigning a Landscape layout while in Portrait mode, and vice versa. 
In that case, some of the infoboxes are hidden. Not sure whether to restrict the offer of layouts only to the currently selected layout, or to let the user figure out why some of the infoboxes on the `InfoBox Sets` assignment page are missing.

DanD222:
The behaviour in the "Inherit from global settings" mode is the same as in 7.44, so raised https://github.com/XCSoar/XCSoar/issues/2596 / _Settings: InfoBox geometry changes not propagated to Look > InfoBox Sets without exit from Settings_

WolfgangDrescher:
I rebased onto the current master and did a quick test. It seems to work.

@DanD222 or @B-apt, do you have time to test whether everything works as expected? I also haven't had the chance yet to check whether @B-apt's approach in #2724 is better than mine, but mine seems to be a bit more minimal.

@DanD222, if I remember correctly, you mentioned that my PR fixes #2596. I hope my rebase didn't break @lordfolken's implementation in #2816.

DanD222:
I've just had a quick look, and it looks good
- Changes to layouts work
- Copy/paste seems to work OK
- The fix for #2596 is present, i.e. when I change the default page geometry, the change successfully propagates to layouts that have "Inherit from global settings" set, even without exit from the settings

@WolfgangDrescher  do you maybe want to convert this from Draft and run it past The Rabbit?

lordfolken:
@coderabbitai full review

WolfgangDrescher:
Reworked after review. Main fix: the per-set geometry only applied on an explicit page change, so switching between circling and cruise did nothing. The active panel changes through `ActionInterface::UpdateDisplayMode()`, which never rebuilt the layout. `MainWindow::CheckInfoBoxGeometry()` now reinitialises only when the geometry actually differs.

Also: dropped the uninitialised `PageLayout::InfoBoxConfig::geometry`, `StringFormat()` instead of `sprintf()` for the profile key, previews and the "Content" field now follow a geometry change in the editor, plus include and wording cleanups.

PR #1867, opened 2025-09-03
Add onboarding dialogue for gestures

WolfgangDrescher:
See https://github.com/XCSoar/XCSoar/issues/1546.

IMG_2520
IMG_2521

Here are some questions and todos before I remove the draft flag from this PR.

1. Someone needs to review `resource.mk`. It works on iOS, but I did not check if it works for other builds (Windows, Android, Kobo).
2. I'm not sure about the filename convention in XCSoar. Should `Onboarding.cpp` be changed to `dlgOnboarding.cpp`?
3. Is there a script to automatically generate new translations? If not, is it relevant to add the filename and line numbers such as `#: src/Dialogs/dlgCredits.cpp:145`?
4. Adding a new `ProfileKeys::*` is not done yet and `Startup.cpp` need to be changed accordingly.
5. How should we handle responsive rendering issues? See the 2nd screenshot.

Also see https://github.com/XCSoar/XCSoar/issues/1546#issuecomment-3246896687 for a discussion about what we want to achieve with the onboarding dialogue.


WolfgangDrescher:
IMG_D098A65B7CB6-1

build/resource.mk
```diff
@@ -196,6 +196,23 @@ else
 RESOURCE_FILES += $(PNG_BITMAPS)
 endif
 
+
+
+GESTURES = du.png up.png down.png left.png right.png urdl.png dr.png rd.png dl.png ud.png ldr.png urd.png ldrdl.png
+GESTURES_DST = $(addprefix $(DATA)/graphics2/gesture_,$(GESTURES))
+
+# sicherstellen, dass das Zielverzeichnis existiert
```
lordfolken: Eventhough I do speak german, for the project we generally stick to english, as there are many international contributers.

build/resource.mk
```diff
@@ -196,6 +196,23 @@ else
 RESOURCE_FILES += $(PNG_BITMAPS)
 endif
 
+
+
```
lordfolken, line 235: Please no excessive return lines

src/Startup.cpp
```diff
@@ -377,6 +378,11 @@ Startup(UI::Display &display)
   // Read the terrain file
   main_window->LoadTerrain();
 
+  // Show onboarding dialog
+  if (true || Profile::GetPath(ProfileKeys::MapFile) == nullptr) {
+    dlgOnboardingShowModal();
+  }
+
```
lordfolken: Please use a own ProfileKey. Depending on something else is in transparent. We already have a lot of "hidden" behaviors  in XCSoar. Lets try to not to add.

src/Dialogs/Onboarding.cpp
```diff
@@ -0,0 +1,244 @@
...
+// Copyright The XCSoar Project
+
+#include "Dialogs/Dialogs.h"
+#include "Dialogs/WidgetDialog.hpp"
+#include "Widget/LargeTextWidget.hpp"
+#include "Look/DialogLook.hpp"
+#include "UIGlobals.hpp"
+#include "Language/Language.hpp"
+
+#include "Screen/Layout.hpp"
```
lordfolken: I think you need to add Onboarding.hpp here. That will fix the compilation failure.

lordfolken (changes requested):

lordfolken:
image

This is target WAYLAND, with 165% fontsize.

A few things: 
 -[ ] xcsoar support resolutions from a minimal 320x240 xcsoar -320x240 or -small
 -[ ] xcsoar supports landscape, portrait and square layouts xcsoar -square xcsoar 
 -[ ] the largetextwindow needs a s scrollbar. its implemented somewhere but i haven't figuired out how to display it. 
 -[ ] keyboard up and down should scroll the dialog

lordfolken:
@coderabbit full review

lordfolken:
For Fonts please use src/Look/DialogLook.hpp and src/Look/DefaultFonts.cpp

lordfolken:
I know this is a draft, so the comments are just here to help.

build/resource.mk
WolfgangDrescher: Sorry about that. These are leftovers from ChatGPT, since I had never written a Makefile before and needed some help. I'll delete the comments.

Data/resources.txt
```diff
@@ -115,6 +115,20 @@ bitmap_icon_scaled IDB_REPORTING_POINT "map_reporting_point"
...
+bitmap_graphic IDB_GESTURE_RIGHT "gesture_right"
+bitmap_graphic IDB_GESTURE_URDL "gesture_urdl"
+bitmap_graphic IDB_GESTURE_DR "gesture_dr"
+bitmap_graphic IDB_GESTURE_RD "gesture_rd"
+bitmap_graphic IDB_GESTURE_DL "gesture_dl"
+bitmap_graphic IDB_GESTURE_UD "gesture_ud"
+bitmap_graphic IDB_GESTURE_LDR "gesture_ldr"
+bitmap_graphic IDB_GESTURE_URD "gesture_urd"
+bitmap_graphic IDB_GESTURE_LDRDL "gesture_ldrdl"
+
```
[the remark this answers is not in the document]
WolfgangDrescher: I'm not sure the coderabbitai is correct about this.

build/resource.mk
```diff
@@ -196,6 +196,23 @@ else
...
+	@$(NQ)echo "  MKDIR   $(DATA)/graphics2/"
+	$(Q)mkdir -p $(DATA)/graphics2
+	@touch $@
+
+# Kopiere PNGs aus doc/manual/en/figures nach graphics2/ mit Prefix "gesture_"
+$(GESTURES_DST): $(DATA)/graphics2/gesture_%.png: doc/manual/en/figures/%.png | $(DATA)/graphics2/dirstamp
+	$(Q)cp $< $@
+
+RESOURCE_FILES += $(GESTURES_DST)
+
```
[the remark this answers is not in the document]
WolfgangDrescher: @lordfolken can you review this coderabbitai code for `build/resource.mk`? I'm not familiar with the Makefile logic there.

Data/Input/default.xci
```diff
@@ -908,6 +908,13 @@ event=RepeatStatusMessage
 label=Message Repeat
 location=7
 
+mode=Info3
+type=key
+data=9
```
WolfgangDrescher, line 975: Is is correct to use `data=9` here?

src/Dialogs/Onboarding.cpp
```diff
@@ -0,0 +1,263 @@
...
+  std::unique_ptr<GestureHelpWindow> window;
+
+public:
+  PixelSize GetMinimumSize() const noexcept override {
+    return { Layout::FastScale(200), Layout::FastScale(200) };
+  }
+  
+  PixelSize GetMaximumSize() const noexcept override {
+    return { Layout::FastScale(300), Layout::FastScale(450) };
+  }
```
WolfgangDrescher: I'm not sure if im setting `GetMinimumSize` and `GetMaximumSize` correct here. Do I need `Layout::FastScale`?

WolfgangDrescher:
IMG_2529
IMG_2530

WolfgangDrescher:
Personally, I think the beginning of the line should be marked with a dot. That way, you have to trace the line backwards from the dot to find the beginning.

yorickreum:
> Personally, I think the beginning of the line should be marked with a dot. That way, you have to trace the line backwards from the dot to find the beginning.

We could do it as in the Aresti symbols, with a dot and a dash. 👀

WolfgangDrescher:
IMG_2533
IMG_2534

Here is the new screen for what I consider to be the most important configurations for XCSoar before you can use it effectively. Any feedback?

I plan to add three more pages:

## Welcome

To get the most out of XCSoar and to learn about its many functions in detail, it is highly recommended to read the Quick Guide or the complete documentation. The manuals explain step by step how to use XCSoar efficiently and cover both basic and advanced features.

Documentation is available in several languages, including English, German, French, and Portuguese.
You can always access the latest versions online:

https://xcsoar.org/discover/manual.html

The XCSoar documentation is actively maintained, but some topics may still be missing or outdated. If you find mistakes or have ideas for improvement, please contribute — XCSoar is open source and thrives on community input:

https://github.com/XCSoar/XCSoar/

## Pre-flight

* Setup checklist
* Setup Aircraft / Polar
* Setup Flight
* Setup Wind
* Setup Task?

## Post-flight

* Download Flight Log
* Flight Analysis
* Flight Status
* Upload Flight (WeGlide)

src/Dialogs/Onboarding.cpp
```diff
@@ -0,0 +1,590 @@
...
+
+class ConfigurationWidget final : public WindowWidget {
+public:
+  PixelSize GetMinimumSize() const noexcept override {
+    return { Layout::FastScale(200), Layout::FastScale(200) };
+  }
+  
+  PixelSize GetMaximumSize() const noexcept override {
+    return { Layout::FastScale(300), Layout::FastScale(1000) };
+  }
```
WolfgangDrescher: Currently, the widget size is fixed. I'm not sure about the best way to ensure that the scrollbar is always large enough. I'm using `ConfigurationWindow::OnPaint` together with `canvas.DrawFormattedText`, so the window height can vary across different devices. Do you have any suggestions?

src/Dialogs/Onboarding.cpp
```diff
@@ -0,0 +1,590 @@
...
+  PixelRect l3_rc{x_text, y, int(canvas.GetWidth()) - margin, int(canvas.GetHeight())};
+  unsigned l3_height = canvas.DrawFormattedText(l3_rc, l3, DT_LEFT);
+  y += int(l3_height) + margin;
+
+  // Aircraft polar
+  canvas.Select(fontMono);
+  const char *c4 = Profile::Get(ProfileKeys::Polar); // TODO does not work
+  canvas.DrawText({x, y + icon_offset}, (c4 == nullptr || StringIsEmpty(c4)) ? undone : done);
+  canvas.Select(fontDefault);
+  const TCHAR *t4 = _("Add your aircraft and, most importantly, select the corresponding polar curve and activate the aircraft, so that the flight computer can calculate everything correctly.");
```
WolfgangDrescher: I experimented a bit but couldn't find a way to check whether an aircraft with a polar is set up and activated. What should I use here?

src/Dialogs/QuickGuide/ConfigurationWidget.cpp
```diff
@@ -0,0 +1,249 @@
...
+#include "ui/canvas/Font.hpp"
+#include "Look/FontDescription.hpp"
+#include "Language/Language.hpp"
+#include "Profile/Profile.hpp"
+#include "system/Path.hpp"
+#include "util/StringCompare.hxx"
+#include "util/OpenLink.hpp"
+#include "util/ConvertString.hpp"
+
+#include <winuser.h>
```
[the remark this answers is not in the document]
WolfgangDrescher: Codacy Static Code Analysis does also complain about this. If I remove this I cannot compile anymore because `DT_LEFT` is used. It actually imports `src/unix/winuser.h`. How can I fix this for the Codacy Static Code Analysis?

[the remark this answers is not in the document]
WolfgangDrescher: Other files in XCSoar also include `winuser.h`, just like I did. For example, see `src/Renderer/TextRenderer.cpp`. So why should I not do it like this here? If I use `#include "unix/winuser.h"`, will this cause problems on Windows where these constants already exist?

src/Dialogs/QuickGuide/ConfigurationWidget.cpp
```diff
@@ -0,0 +1,249 @@
...
+  canvas.Select(fontSmall);
+  const TCHAR *t98 = _("Maps, waypoints, etc. downloaded using the download feature can be updated in the file manager (Config → Config → File manager).");
+  PixelRect t98_rc{x, y, int(canvas.GetWidth()) - margin, int(canvas.GetHeight())};
+  unsigned t98_height = canvas.DrawFormattedText(t98_rc, t98, DT_LEFT);
+  y += int(t98_height) + margin;
+
+  canvas.Select(fontSmall);
+  const TCHAR *t99 = _("Files are stored in the operating system of the mobile device and can also be replaced and supplemented there, e.g., to add custom waypoints for a competition. On iOS, these files are located here: Files app → On my iPhone → XCSoar → XCSoarData. On Android, these files are located here: Android → media → org.xcsoar.");
+  PixelRect t99_rc{x, y, int(canvas.GetWidth()) - margin, int(canvas.GetHeight())};
+  canvas.DrawFormattedText(t99_rc, t99, DT_LEFT);
```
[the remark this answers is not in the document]
WolfgangDrescher: It is scrollable in src/Dialogs/Onboarding/dlgOnboarding.cpp with `VScrollWidget`.

src/util/OpenLink.cpp
```diff
@@ -0,0 +1,24 @@
...
+bool OpenLink(const char *url) noexcept {
+#ifdef TARGET_OS_IOS
+  NSString *ns_url = [NSString stringWithUTF8String:url];
+  NSURL *nsu = [NSURL URLWithString:ns_url];
+  if ([[UIApplication sharedApplication] canOpenURL:nsu]) {
+    [[UIApplication sharedApplication] openURL:nsu options:@{} completionHandler:nil];
+    return true;
+  }
+#endif
+  return false;
```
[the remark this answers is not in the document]
WolfgangDrescher: @yorickreum What do you think, should I change something here? I changed to `TARGET_OS_IPHONE` instead of `TARGET_OS_IOS`.

src/Dialogs/Onboarding/ConfigurationWidget.cpp
```diff
@@ -0,0 +1,248 @@
...
+  y += int(l9b_height) + margin / 2;
+  canvas.SetTextColor(COLOR_BLUE);
+  const TCHAR *l9c = _("https://youtube.com/user/M24Tom/playlists"); // https://youtube.com/playlist?list=PLb36zVafnfctzNG7yJoNXc2x8HnX5J9j7
+  PixelRect l9c_rc{x_text, y, int(canvas.GetWidth()) - margin, int(canvas.GetHeight())};
+  unsigned l9c_height = canvas.DrawFormattedText(l9c_rc, l9c, DT_LEFT | DT_UNDERLINE);
+  youtube_link_rect = {x_text, y, int(canvas.GetWidth()) - margin, y + int(l9c_height)};
+  canvas.SetTextColor(COLOR_BLACK);
+  y += int(l9c_height) + margin;
+  
+  // NMEA devices
```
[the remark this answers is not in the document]
WolfgangDrescher: No, it's better the way it is. The screen cuts off the full link, so https://youtube.com/user/M24Tom/playlists is more readable. But when you click it, you'll still be redirected to the correct playlist.

lordfolken:
I completely agree that this is needed.
* But at presen't it doesn't compile for me. (UNIX/WAYLAND)
* There are quite a few comments by the ai and static code analysis that needs to be addressed. 
* On the UNIX target the dialog doesn't show up at startup
* it does on wayland. 
* the vscrollbar is intermittently present or not. 
* the scrollheight isn't calculated correctly in landscape so the guesture dialogs don't scroll to the bottom
* the logo's size isn´t calculated correctly. 
* there are no translations

please test with UNIX target and -small and -square and landscape resolutions

image

src/util/OpenLink.cpp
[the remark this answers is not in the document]
WolfgangDrescher: Fixed in c4b6965

yorickreum:
I've put fixes for the remaining platforms (Windows, KOBO) into https://github.com/WolfgangDrescher/XCSoar/pull/4.

yorickreum:
> The Up/Down arrow keys already cycle through the buttons (next page / previous page / close). Using Up/Down to cycle through the pages of ArrowPagerWidget would conflict with that, so I don't think this is a good idea. Maybe we could consider handling OnMouseWheel instead. Is this relevant for any devices I'm not aware of, or can we ignore it for now?

Would be relevant i.e. for the XCRemote. But not sure how we handle this. (We could use PageUp/PageDown or +/- or j/k but thats all kinda ugly.) Maybe we can also ignore it for now, as the guide is not supposed to be used during flight, only before and after.

kobedegeest:
> The Up/Down arrow keys already cycle through the buttons (next page / previous page / close). Using Up/Down to cycle through the pages of ArrowPagerWidget would conflict with that, so I don't think this is a good idea. Maybe we could consider handling OnMouseWheel instead. Is this relevant for any devices I'm not aware of, or can we ignore it for now?

The scroll bar seems to have up down buttons at top and bottom so maybe have the Up/Down key cylce include those as well so you cycle through next page/ previous page/ close/ up / down

WolfgangDrescher:
I noticed, that `Set your timezone` (`Profile::Get(ProfileKeys::UTCOffsetSigned, utc_offset)`) is set correctly `+1h` on my Mac by default, but the checkbox does not properly check it until I switch to something else and then back to `+1h`.

Does XCSoar automatically switch between daylight saving time and standard time? If not should we keep this in mind for this widget or add a flash message when XCSoar is opened the first time after a time change (in another PR)?

yorickreum:
Something I realized while letting a (non-pilot) friend test the app: they almost immediately clicked the **Close** button after launching it and missed all the other pages of the guide.

Maybe it would be better to move the **Close** button to the left and place the navigation arrows on the right as the "standard" action?

src/Widget/VScrollWidget.cpp
yorickreum, line 1: Not okay imo. Makes widgets unnecessarily high and causes issues e.g. with the Terrain view in `Configuration → Map Display → Terrain`.

WolfgangDrescher: Interesting, thanks for pointing this out. However reverting 897a81a will remove the scrollbars of e.g. `pager->Add(std::make_unique<VScrollWidget>(std::make_unique<GestureHelpWidget>(), look));`. I'm not sure how this should be resolved. Is it possible to combine my widgets with `VScrollWidget` somehow (without the need for 897a81a)?

WolfgangDrescher: I rebased master (so #2034 is included) and reverted 897a81a but the scrollbars are still missing. I'm not sure what `HAVE_CLIPPING` is doing and why this is related to `VScrollWidget`.

WolfgangDrescher: I have reverted this change and resolved the missing scrollbars in 
1cbf7bf.

src/Dialogs/QuickGuide/ConfigurationWidget.cpp
```diff
@@ -0,0 +1,404 @@
...
+  y += int(DrawLinkLine(LinkAction::SITE_FILES_3, l3)) + margin;
+
+  // Aircraft polar
+  if (canvas != nullptr) {
+    const bool has_plane_path = Profile::GetPath("PlanePath") != nullptr;
+    const bool has_polar = has_plane_path &&
+      CommonInterface::GetComputerSettings().plane.polar_shape.IsValid();
+    canvas->Select(fontMono);
+    canvas->DrawText({x, y + icon_offset}, !has_polar ? _T("[ ]") : _T("[X]"));
+  }
```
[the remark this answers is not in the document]
WolfgangDrescher: I don't think this is necessary. `PlanePath` seems to be calculated and stored differently from the other values in ProfileKeys.

src/Dialogs/QuickGuide/WelcomeWidget.cpp
```diff
@@ -0,0 +1,224 @@
...
+  }
+  const PixelRect t5_rc{x, y, rc.right - margin, rc.bottom};
+  const unsigned t5_height = canvas != nullptr
+    ? canvas->DrawFormattedText(t5_rc, t5, DT_LEFT | DT_UNDERLINE)
+    : renderer.GetHeight(look.text_font, text_width, t5);
+  const PixelRect github_link_rect{x, y, rc.right - margin, y + int(t5_height)};
+  if (canvas != nullptr)
+    canvas->SetTextColor(COLOR_BLACK);
+  y += int(t5_height) + margin;
+
```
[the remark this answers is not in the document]
WolfgangDrescher: What is the correct way to use a non-translatable string that also build on `PC` and `WIN64`? Is something like this correct:

```
const TCHAR *t5 = _T("https://github.com/XCSoar/XCSoar");
```

src/Dialogs/QuickGuide/ConfigurationWidget.cpp
```diff
@@ -0,0 +1,402 @@
...
+    dialog.ShowModal();
+    if (dialog.GetChanged())
+      Profile::Save();
+    return true;
+  }
+
+  case LinkAction::YOUTUBE_TUTORIAL:
+    OpenLink("https://youtube.com/playlist?list="
+             "PLb36zVafnfctzNG7yJoNXc2x8HnX5J9j7");
+    return true;
```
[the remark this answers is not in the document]
WolfgangDrescher: This is expected for now. Can you suggest how to implement support for Android in `src/util/OpenLink.cpp`?

PR #1869, opened 2025-09-05
Add ScrollableLargeTextWidget and use it in dlgCredits

WolfgangDrescher:
Add a `ScrollableLargeTextWidget` that automatically calculates the height of the `WindowWidget` based on the number of lines of the text. This can be wrapped by `VScrollWidget`. E.g.:

```cpp
auto pager = std::make_unique<ArrowPagerWidget>(look.button, dialog.MakeModalResultCallback(mrOK));
auto authorInner = std::make_unique<ScrollableLargeTextWidget>(look, authors2);
auto authorScroll = std::make_unique<VScrollWidget>(std::move(authorInner), look);
pager->Add(std::move(authorScroll));
```

`VScrollWidget` is bugged and needs to be fixed before this PR can be merged. It is fixed in 1d2f16e (see #1867). Please tell me if I should cerry-pick it into this PR.

### DIFF
--- a/src/Widget/ScrollableLargeTextWidget.cpp
+++ b/src/Widget/ScrollableLargeTextWidget.cpp
@@ -1,0 +1,31 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The XCSoar Project
+
+#include "Widget/ScrollableLargeTextWidget.hpp"
+#include "Screen/Layout.hpp"
+
+ScrollableLargeTextWidget::ScrollableLargeTextWidget(const DialogLook &dlgLook,
+                                                     const TCHAR *t)
+    : LargeTextWidget(dlgLook, t), look(dlgLook), text(t)
+{
+}
+
+PixelSize
+ScrollableLargeTextWidget::GetMinimumSize() const noexcept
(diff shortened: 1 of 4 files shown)


WolfgangDrescher:
There is a bug when rendering https://github.com/XCSoar/XCSoar/blob/master/NEWS.txt with around 4000 lines.

IMG_2526

src/Widget/ScrollableLargeTextWidget.cpp
```diff
@@ -0,0 +1,33 @@
...
+  for (auto ch : text) {
+    if (ch == _T('\n')) ++lines;
+  }
+
+  unsigned line_height = look.text_font.GetHeight();
+  unsigned total_height = lines * line_height + Layout::FastScale(10);
+  
+  const unsigned max_height = Layout::FastScale(6500);
+
+  return {static_cast<unsigned>(Layout::FastScale(300)), std::min(total_height, max_height)};
```
WolfgangDrescher, line 32: This is probably not the best solution, but it works for now. At least it provides a lot more text than before where you could just see one screen size of text.

PR #1900, opened 2025-09-24
FlightStatusPanel: update share button, use geo URI and iOS support

### LINKED ISSUE #1896: Status Dialog: location share

 * the button should be renamed to share location
 * the output on all platforms should be a uri according to geo_uri_scheme
 * the output should be decimal in notation.

WolfgangDrescher:
Closes https://github.com/XCSoar/XCSoar/issues/1896.

I also added support for the native iOS share popup. @yorickreum Could you take a look at `Apple/Share.cpp`? ChatGPT helped me with this class, but I'm not entirely sure if the code is correct.

I haven't tested the new geo URI scheme on Android yet.

What exactly do we want to achieve with links like `geo:47.95101864,7.95627518`? My understanding is that the share popup should automatically recognize this as a location link and suggest appropriate apps.

However, on iOS this doesn't improve the suggested apps for sharing (you need to convert it to `NSURL` first). The best workaround I've found is using `https://www.google.com/maps/search/?api=1&query=LAT,LON`, but that's probably not what we want…

## Summary by CodeRabbit

* **New Features**
  * Added a Share Location button on iOS and Android that uses each platform’s native share sheet.
  * Added plain decimal degrees as a new coordinate display option for lat/lon.
  * Enabled general text sharing on iOS via the system share sheet.

* **Notes**
  * Plain decimal is view-only; input/editing for this format is not yet supported.

### DIFF
--- a/src/Apple/Share.cpp
+++ b/src/Apple/Share.cpp
@@ -1,0 +1,128 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The XCSoar Project
+
+// iOS implementation for sharing text using UIActivityViewController
+
+#include "Apple/Share.hpp"
+#include "util/ConvertString.hpp"
+
+#if defined(__APPLE__) && TARGET_OS_IPHONE
+#import <UIKit/UIKit.h>
+#import <dispatch/dispatch.h>
+
+static UIViewController *
+GetTopMostController(UIViewController *controller) noexcept
+{
+  if (controller == nil)
+    return nil;
+
+  while (true) {
+    if ([controller isKindOfClass:[UINavigationController class]]) {
+      controller = ((UINavigationController *)controller).visibleViewController;
+    } else if ([controller isKindOfClass:[UITabBarController class]]) {
+      controller = ((UITabBarController *)controller).selectedViewController;
+    } else if (controller.presentedViewController != nil) {
+      controller = controller.presentedViewController;
+    } else {
+      break;
+    }
+
+    if (controller == nil)
+      break;
+  }
+
+  return controller;
+}
+
+static UIViewController *
+FindActiveController() noexcept
+{
+  UIWindow *activeWindow = nil;
+
+  if (@available(iOS 13.0, *)) {
+    for (UIScene *scene in UIApplication.sharedApplication.connectedScenes) {
+      if (scene.activationState != UISceneActivationStateForegroundActive)
+        continue;
+
+      if (![scene isKindOfClass:[UIWindowScene class]])
+        continue;
+
+      UIWindowScene *windowScene = (UIWindowScene *)scene;
+      for (UIWindow *window in windowScene.windows) {
+        if (window.isHidden)
+          continue;
+
+        if (window.isKeyWindow) {
+          activeWindow = window;
+          break;
+        }
+
+        if (activeWindow == nil)
+          activeWindow = window;
+      }
+
+      if (activeWindow != nil)
+        break;
+    }
+  }
+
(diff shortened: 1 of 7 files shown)


src/Dialogs/StatusPanels/FlightStatusPanel.cpp
```diff
@@ -14,18 +15,50 @@
...
   Location,
   Altitude,
   MaxHeightGain,
   Near,
   Bearing,
   Distance,
-#ifdef ANDROID
-  ShareButton,
+#if defined(ANDROID) || (defined(__APPLE__) && TARGET_OS_IPHONE)
+  ShareLocationButton,
```
lordfolken: I don't think its needed to limit this to android/ipphone. xdg-open on linux platforms supports geo:.

src/Dialogs/StatusPanels/FlightStatusPanel.cpp
```diff
@@ -74,11 +107,21 @@ FlightStatusPanel::Prepare([[maybe_unused]] ContainerWindow &parent,
...
+    const auto uri = BuildGeoURI(basic.location);
+    if (uri.empty())
+      return;
 #ifdef ANDROID
-  AddButton(_("Share"), [this](){
-    native_view->ShareText(Java::GetEnv(), GetText(Location));
+    native_view->ShareText(Java::GetEnv(), uri.c_str());
+#elif defined(__APPLE__) && TARGET_OS_IPHONE
+    ShareTextIOS(uri.c_str());
+#endif
```
lordfolken: I think a call to src/system/Runfile would be good here for neutral platforms.

src/Apple/Share.cpp
```diff
@@ -0,0 +1,128 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The XCSoar Project
+
+// iOS implementation for sharing text using UIActivityViewController
+
+#include "Apple/Share.hpp"
+#include "util/ConvertString.hpp"
+
+#if defined(__APPLE__) && TARGET_OS_IPHONE
```
lordfolken, line 9: Do we really need to if def files again internally when they are already only built on certain targets?

WolfgangDrescher: I added `$(SRC)/Apple/Share.cpp` to `build/main.mk`. Is that correct? If I remove these conditions, it would no longer compile for macOS, for example. `#import <UIKit/UIKit.h>` is iOS only.

src/Dialogs/StatusPanels/FlightStatusPanel.cpp
WolfgangDrescher:
```cpp
#ifdef ANDROID
    native_view->ShareText(Java::GetEnv(), uri.c_str());
#elif defined(__APPLE__) && TARGET_OS_IPHONE
    ShareTextIOS(uri.c_str());
#else
    if (RunFile(uri.c_str()) == false) {
      Message::AddMessage(_("Failed to open location link"));
    }
#endif
```

Any idea why `RunFile` always returns falsy?

https://github.com/XCSoar/XCSoar/blob/2be79c15a50f14079a25d7a1d549bb0a360f6877/src/system/Process.cpp#L53-L64

src/Dialogs/StatusPanels/FlightStatusPanel.cpp
WolfgangDrescher: macOS actually does not support links with the geo URI scheme. But it is probably irrelevant anyway. So it's certainly a good idea to add a fallback with `RunFile`.

Issue #1895, opened 2025-09-23
UI: create a selectable list dialog


lordfolken:
What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------

To aid with file management, checklist and similar list items, xcsoar needs a list dialog that contains items that are de/-selectable.  

General layout:
 - [ ] item1
 - [x] item2
 - [x] item3 
 - [ ] item4

The return should then be an array of the selected items. 

This could be used for the file selection when selecting waypoints & airspaces.

PR #1903, opened 2025-10-02
Add Multiple File Waypoints

Eldalie:
Add multiple files support in Waypoints UI

## Summary by CodeRabbit

- **New Features**
  - UI supports selecting and managing multiple waypoint files, watched waypoint files, and multiple airspace files; file loading uses pattern filters.
  - Waypoint filter simplified: a single “File” option replaces separate file-source choices.

- **Chores**
  - Existing profiles are migrated automatically to the new multi-file format.

- **Documentation**
  - Configuration manual updated: waypoint entries presented as lists; watched-waypoints settings consolidated and clarified with examples.

### DIFF
--- a/src/Profile/PathValue.cpp
+++ b/src/Profile/PathValue.cpp
@@ -9,6 +9,10 @@
 #include "util/StringCompare.hxx"
 #include "util/StringPointer.hxx"
 
+#ifdef HAVE_POSIX
+#include <fnmatch.h>
+#endif
+
 #ifdef _UNICODE
 #include "util/AllocatedString.hxx"
 #endif
@@ -33,7 +37,7 @@ ProfileMap::GetPath(std::string_view key) const noexcept
 }
 
 std::vector<AllocatedPath>
-ProfileMap::GetMultiplePaths(std::string_view key) const
+ProfileMap::GetMultiplePaths(std::string_view key, const TCHAR *patterns) const
 {
 
   std::vector<AllocatedPath> paths;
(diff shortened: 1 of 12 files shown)


WolfgangDrescher:
This would be a good PR to consider integrating https://github.com/XCSoar/XCSoar/issues/1895.

Eldalie:
> This would be a good PR to consider integrating #1895.

@WolfgangDrescher Yes, but I will keep making small MRs into small MRs!

The next one will be AirfieldFileList!

src/Profile/PathValue.cpp
```diff
@@ -51,11 +51,6 @@ ProfileMap::GetMultiplePaths(std::string_view key) const
 
     Path path(file_string.c_str());
 
-    if (!(path.EndsWithIgnoreCase(_T(".txt")) ||
-          path.EndsWithIgnoreCase(_T(".air")) ||
-          path.EndsWithIgnoreCase(_T(".sua"))))
-      continue;
-
     paths.push_back(ExpandLocalPath(AllocatedPath(path)));
```
lordfolken: This function should probably be extended by an argument to pass a file filter.

There is:
Waypoint/Patterns.hpp:#define WAYPOINT_FILE_PATTERNS _T("*.dat\0*.xcw\0*.cup\0*.wpz\0*.wpt\0")

Eldalie: Yes, I can do it, but I don’t understand why we need to filter files with the wrong extension here. In what cases could it happen that someone writes files with the wrong extension in the configuration?

lordfolken: I was a little scant on the details. Sorry. 

So the parameter to pass files should be best analog to FilePicker:
Dialogs/FilePicker.cpp:FilePicker(const TCHAR *caption, const TCHAR *patterns)

Usage example:
Dialogs/Device/ManageCAI302Dialog.cpp:  const auto path = FilePicker(_("Waypoints"), WAYPOINT_FILE_PATTERNS);

What I mean with the Pattern example, is that we would have a global somewhere which lists all the extensions for Waypoints, or Checklists, or Maps, or Airspaces. That way we would reference that everywhere in the code.

Eldalie: Done

Issue #1808, opened 2025-06-30
IOS: .XCI file seems not to be recognized


XCNav:
XCSoar versions having the problem
-------------------------------------------------

XCsoar IOS

XCSoar versions not having the problem
-------------------------------------------------

System information
------------------

IOS 18.5 iPhone XR

Steps to reproduce the behavior
-------------------------------

Download the glb-xci-xcremote-xcnav.xci file in The events tab and restart xcsoar. Install a keyboard and press "M"

Expected behavior
-----------------

With the .xci file installed the xcvario Menu should pop up

Actual behavior
---------------

Nothing happens

Do you have any idea what may have caused this?
-----------------------------------------------

Do you have an idea how to solve the issue?
-------------------------------------------

yorickreum:
Thanks for sending me the test device, @XCNav!

Our current evaluation concluded that .xci files probably are indeed working correctly and the underlying problem is that on iOS the app doesn't correctly recognize that a keyboard is connected.

E.g., the quick menu and MC ready / Vario menus can be opened correctly via the XCRemote. However, it does not show the currently selected menu button in blue, as it does under Android if a keyboard input device is connected.

I will investigate this further.

yorickreum:
We'd need to (at least) implement `HasCursorKeys()` for iOS in https://github.com/XCSoar/XCSoar/blob/506de14dc0a59aa82f3be256ab7ee3ec17f1387e/src/Asset.hpp#L177C1-L196C2

On Android, this is done but triggering on the first press of any cursor key, see https://github.com/XCSoar/XCSoar/blob/506de14dc0a59aa82f3be256ab7ee3ec17f1387e/src/Android/EventBridge.cpp#L53C1-L68C2

yorickreum:
Thanks a lot! I will compile a new version for TestFlight in the next days.

PR #1926, opened 2025-10-21
iOS: src/Input/InputConfig.cpp: Make key codes uppercase...

yorickreum:
… before event lookup

Addresses https://github.com/XCSoar/XCSoar/issues/1808.

On iOS/SDL2, at least the simulator sends lowercase key codes (e.g., 'c' instead of 'C') even when Shift is pressed, but the mapping and generated code only use uppercase key codes. This caused key events for letters to be missed. Now, key codes are converted to uppercase before event lookup, making the mapping case-insensitive for ASCII letters and fixing the issue where lowercase key presses would not trigger mapped events.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard input handling on iOS by implementing proper normalization for letter keys.

### DIFF
--- a/src/Input/InputConfig.cpp
+++ b/src/Input/InputConfig.cpp
@@ -54,6 +54,13 @@ InputConfig::GetKeyEvent(unsigned mode, unsigned key_code) const noexcept
   }
 #endif
 
+#if defined(__APPLE__) && TARGET_OS_IPHONE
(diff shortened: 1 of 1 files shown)


yorickreum:
I did not test this with the virtual keyboard, only with my MacBook keyboard connected to the iOS simulator as a hardware keyboard. @coderabbitai

PR #1955, opened 2025-11-18
src/XCSoar,sdl/Queue: Implements proper quiting for iOS

yorickreum:
SDL_WaitEvent() on iOS doesn't return when quit flag is set, causing the event loop to hang. Push a dummy SDL_USEREVENT to wake it up, allowing proper cleanup instead of forcing exit().

Similar approach to Android's event injection pattern.

## Summary by CodeRabbit

* **Bug Fixes**
  * iOS now reliably exits after closing the app to avoid a stuck/black screen.

* **Refactor**
  * Event queue updated for safer concurrent handling and to ensure quit signals wake the event loop promptly.

* **Chores**
  * Translation files refreshed and source references consolidated (metadata only; no text changes).

### DIFF
--- a/po/bg.po
+++ b/po/bg.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: xcsoar\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2025-11-07 05:19+0100\n"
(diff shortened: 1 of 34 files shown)


yorickreum:
Apple apparently actually heavily discourages using `exit(),` so I will drop this for now... https://stackoverflow.com/questions/355168/proper-way-to-exit-iphone-application

Issue #1969, opened 2025-12-06
Stratux GY-91 Support (Three-axis gyroscope + triaxial accelerometer + triaxial magnetic field + pressure)


fraktal-sb:
XCSoar version
--------------

7.44

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------

Since version 7.44 XCSoar supports Stratux, a DIY and low cost ADS-B/FLARM/OGN receiver. Equipped with an additiononal GY-91 board it can also deliver AHRS data to XCSoar.

see the following information about Stratux and GY-91 and AHRS

https://stratux.me/
https://www.amazon.de/TECNOIOT-GY-91-Acceleration-Kompass-Sensor-Modul-Beschleunigung/dp/B07HMQZ7N9
https://www.gs-avionik.de/gs-datenblaetter/Produktdatenblatt%20GS-Stratux-AHRS.pdf

If the Stratux device is enabled and Stratux contains the GY-91 board the XCSoar artificial horizon window should be feeded.

<img width="864" height="1536" alt="Image" src="https://github.com/user-attachments/assets/0865866c-949c-4428-8c11-ed374fb12ecb" />

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------

The existing Stratux driver should be adapted to process $RPYL and $APENV1 records from the data stream. This will be done by using functionality from the already implemented LevilAHRS driver

PR #1970, opened 2025-12-06
Stratux AHRS Support using LevilAHRS components

fraktal-sb:
Stratux AHRS Support see https://github.com/XCSoar/XCSoar/issues/1969

## Summary by CodeRabbit

* **New Features**
  * Added support for RPYL and APENV1 NMEA sentence types from compatible devices.

* **Refactor**
  * Restructured internal parsing functions to improve code modularity and reusability.

* **Chores**
  * Minor code formatting and whitespace cleanup.

### DIFF
--- a/src/Device/Driver/LevilAHRS_G.cpp
+++ b/src/Device/Driver/LevilAHRS_G.cpp
@@ -24,7 +24,7 @@ ErrorMessage([[maybe_unused]] unsigned code)
   Message::AddMessage(_T("Levil AHRS: hardware error !"));
 }
 
-static bool
+bool
 ParseRPYL(NMEAInputLine &line, NMEAInfo &info)
 {
   // $RPYL,Roll,Pitch,MagnHeading,SideSlip,YawRate,G,errorcode,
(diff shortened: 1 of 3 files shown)


lordfolken:
It would probably make sense to recognize the Levil sentences on any configured driver.

We are doing this with the FLARM Traffic (like) sentences currently in src/Device/Parser.cpp

fraktal-sb:
What we could do is the following. But I am not so happy with that. I would more prefer the approach to process data individually with the specific driver. The effort should be limited to do this analog to Stratux on other drivers.

Bildschirmfoto zu 2025-12-09 05-58-01

lordfolken:
Yes, what you state would be the approach to implement this in the parser. 
However as you state, this puts a lot more code into the (well tested) critical path.

(I wasn´t  aware that the levil protocol does not come with checksums, and now i'm annoyed at it. ;-) )

For the record NMEALine does support both sentences with and without Checksums, but it doesn´t do error checking.

So doing this in the Stratux driver does sound like the safer approach.

lordfolken:
LGTM. Thank you.

lordfolken (approved):

PR #1977, opened 2025-12-08
Send exchange frequency command directly to devices

Scumi:
Old: #594 (finally all devices support this)

## Summary by CodeRabbit

* **New Features**
  * Central "exchange frequencies" action now forwards requests to connected devices and merges returned device data.
  * Several device drivers added/extended so devices can perform an exchange-frequency action (ATR833, KRT2, XCOM760, AirControlDisplay and others).

* **Bug Fixes**
  * Central control no longer directly mutates internal radio frequency state; frequency updates now come from device responses.

### DIFF
--- a/src/Device/Driver/ATR833/Device.cpp
+++ b/src/Device/Driver/ATR833/Device.cpp
@@ -160,6 +160,15 @@ ATR833Device::PutStandbyFrequency(RadioFrequency frequency,
   return true;
 }
 
+bool
+ATR833Device::ExchangeRadioFrequencies(OperationEnvironment &env,
+                                       [[maybe_unused]] NMEAInfo &info)
+{
+  ATRBuffer buffer(EXCHANGE);
+  port.FullWrite(buffer.Finish(), env, std::chrono::seconds{2});
+  return true;
+} 
+
 void
 ATR833Device::LinkTimeout()
 {
--- a/src/Device/Driver/ATR833/Device.hpp
+++ b/src/Device/Driver/ATR833/Device.hpp
@@ -27,6 +27,7 @@ public:
   bool PutStandbyFrequency(RadioFrequency frequency,
                            const TCHAR *name,
                            OperationEnvironment &env) override;
+  bool ExchangeRadioFrequencies(OperationEnvironment &env, NMEAInfo &info) override;
   bool EnableNMEA(OperationEnvironment &env) override;
   void OnSysTicker() override;
   void LinkTimeout() override;


src/Device/Driver/KRT2.cpp
```diff
@@ -164,6 +164,11 @@ class KRT2Device final : public AbstractDevice {
   virtual bool PutStandbyFrequency(RadioFrequency frequency,
                                    const TCHAR *name,
                                    OperationEnvironment &env) override;
+  /**
+   * Exchanges active and standby frequencies on the radio.
+   */
+  virtual bool ExchangeRadioFrequencies(OperationEnvironment &env) override;
+
```
lordfolken, line 174: Yep, please address this.

Scumi: That's the easy part and I have made the update already.
The main issue is that the KRT2 only sends the information about its frequencies once (during start-up). After that its on XCSoar to keep device and itself in sync. Weird design, will take a while to address this.

lordfolken: .1 Remote Connection Confirmation
Starting with initial power-up, the radio module pings an ASCII “S” (0x53) over the RS232
connection and repeats every few seconds. If the remote module answers this query with any
ASCII character within 60ms, the radio module recognizes the connection as active and
activates a yellow “r” flag on the display to mark the remote connection as established.
This is immediately followed by sending the current device status of the main module (see 2.3
Set New Active Frequency and Channel Name, 2.6 Set Receiver Volume, Receiver Squelch,
and VOX squelch, MIC selection, 2.7 Configure PTT Buttons, 2.8 Set Intercom Volume, 2.9
Set External Audio Input Volume, 2.10 Set Sidetone Level).
Sending of radio module device status and settings will happen only once immediately after
confirming the ping “S” query for synchronization. If the connection is lost, the radio module
tries again to re-establish the connection.
The same procedure aforementioned may also be used in the other direction wherein the
remote queries the radio module every few seconds for trying to establish a connection. If this
becomes success the remote connection is marked active with the "r" symbol in the display
too. Local changes at the radio module will be send to the remote module.

This is from https://aeroplaying.uk/I/+/KRT2/+/KRT2_Series_RS232_Protocol_Rev.002.pdf

This would suggest that you can abort this connection (by not sending this ping) and reestablish it from both ends. Then the device should send the configs again. 

I thought about tracking the information locally, but yeah then synchronization is always the issue.

Scumi: Yes, I thought about that too, but you have to wait for a fair while until the connection gets aborted. I haven't checked if the device actually sends the data again afterwards. The very delayed update might cause a very unpleasant user experience.

I changed the implementation a bit so the KRT2 driver updates the NMEAInfo separately (unlike all other drivers). It also solves the issue that pressing the "swap"-button on the device did not update the values in XCSoar. Please specifically check if the implementation in Device/MultipleDevices are acceptable or if those could cause issues.

Tested with KRT2 and with an AirControlDisplay in SIM mode.

src/Device/Driver/KRT2.cpp
```diff
@@ -406,6 +415,35 @@ KRT2Device::PutStandbyFrequency(RadioFrequency frequency,
...
 }
 
+bool
+KRT2Device::ExchangeRadioFrequencies(OperationEnvironment &env,
+                                     NMEAInfo &info)
+{
+  const std::byte msg[] = {STX, EXCHANGE_FREQUENCIES};
+  if (Send(msg, env)) {
+    UpdateRadioFrequencies(info);
+
```
lordfolken: Please remove whitespace. :)

src/ActionInterface.cpp
```diff
@@ -345,15 +345,10 @@ ActionInterface::OffsetStandbyFrequency(double offset_khz,
...
-    const auto old_active_freq = radio_settings.active_frequency;
-    const auto old_active_freq_name = radio_settings.active_name;
-
-    ActionInterface::SetActiveFrequency(radio_settings.standby_frequency, radio_settings.standby_name, to_devices);
-    ActionInterface::SetStandbyFrequency(old_active_freq, old_active_freq_name, to_devices);
+  /* send to external devices */
+  if (to_devices) {
+    MessageOperationEnvironment env;
+    backend_components->devices->ExchangeRadioFrequencies(env);
   }
```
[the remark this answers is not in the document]
lordfolken: yes please.

src/Device/Driver/KRT2.cpp
```diff
@@ -319,7 +328,7 @@ KRT2Device::HandleSTX(std::span<const std::byte> src, NMEAInfo &info) noexcept
     return src.size() < 6 ? 0 : 6;
 
   case EXCHANGE_FREQUENCIES:
-    info.settings.swap_frequencies.Update(info.clock);
```
lordfolken, line 322: Add a comment here along the lines:
    /* Optimistic update: KRT2 doesn't send frequencies back after
        exchange command, so we update state immediately. If the device
        sends an unsolicited EXCHANGE_FREQUENCIES message later, it will
        be handled by HandleSTX which also calls UpdateRadioFrequencies. */

Issue #1725, opened 2025-03-27
FLARM messaging interface PFLAM


lordfolken:
### Discussed in https://github.com/XCSoar/XCSoar/discussions/1724

<div type='discussions-op-text'>

<sup>Originally posted by **Scumi** March 26, 2025</sup>
Flarm has implemented a messaging interface, which can provide aircraft information like type, registration, pilot name and others without the need of a dedicated database like FlarmNet or OGN.

I have started to extend the FLARM driver to read-out those PFLAM messages. 
The question is, which data should XCSoar prefer: FlarmNet records or PFLAM data? I would tend to the latter because I expect FlarmNet records to be not as up-to-date as Flarm internal data. Furthermore, FlarmNet records might not be maintained anymore by everyone now that Flarm internal data is available to be read-out. Disadvantage of using PFLAM data is that in high traffic situations, the bandwidth for PFLAM messages might be throttled in favor of actual collision data. I personally do not see this as an issue though as in those situations you have other things to do than looking up a pilot name or whatever.
I thought about merging the data, but with a FlarmNet record potentially not maintained, the result can be very wrong and confusing.

What do you think about this?

Documentation:
https://www.flarm.com/wp-content/uploads/2024/12/FTD-109-FLARM-Messaging-Interface-Control-Document-v1.00.pdf</div>

lyusupov:
> The question is, which data should XCSoar prefer: FlarmNet records or PFLAM data? I would tend to the latter because ...

Aircraft data that comes from MESSAGING INTERFACE is periodic. It is supplied at low priority rate 1-2 times per minute or less.

![Image](https://github.com/user-attachments/assets/11b0d6cd-3eba-41d9-9c06-832009cde693)

Data from OGN or FlarmNet file can be accessed momentarily.

My opinion that XCSoar traffic display could resolve aircraft ID from a fast data resource first _(such as a database)_ .
When MESSAGING INTERFACE will provide it's own data _(a dozen of seconds or minutes later)_  - XCSoar could update the traffic information based on this data.

> I thought about merging the data, but with a FlarmNet record potentially not maintained

Even though FlarmNet may not be maintained very well -  but the OGN database is updated quite fast, they grow at rate of approx. 1000-1500 new records every year and they have 25000+ records accumulated at this time.

Scumi:
> Aircraft data that comes from MESSAGING INTERFACE is periodic. It is supplied at low priority rate 1-2 times per minute or less.

[...]

PR #1995, opened 2025-12-20
Integration of Flarm Messaging protocol

Scumi:
My proposal to integrate the Flarm Messaging protocol parsed from PFLAM NMEA sentences.

Architecture:
- PFLAM messages are parsed and stored as records in a messaging database.
- Each messaging record tracks which fields have been received per FLARM target and can therefore determine when a complete record cycle (registration, pilot, aircraft type, and callsign) has been received. This is required to drop stale values, especially when a target is received again in a subsequent flight with different field contents.
- Database records are written to a file every 5 minutes if a new record is available or an existing record has changed. The database is also written on shutdown.
- The priority order of the three databases in use is: user names => FLARM messaging => FLARMNet. FLARM/Details have been updated to provide a unified output independent of different string types (TCHAR, std::string, StaticString).
- If a message of type VHF is received, the frequency is stored temporarily in the record but is not persisted in the database.
- All dialogs and renderers that use FLARM records have been updated to resolve data according to the database priority defined above.

@lordfolken I have not yet rebased onto your flarmup-branch. I saw some parts that might need work and would have a look after we finished this integration.

Ref: #1724,  #1725

## Summary by CodeRabbit

* **New Features**
  * FLARM messaging store with CSV import/export, periodic and immediate persistence, and a new messaging test harness.
  * Parser support for new FLARM messaging sentences to capture AREG/PNAME/ATYPE/ACALL/VHF fields.

* **Improvements**
  * UI: traffic lists and details show resolved pilot/plane/registration/callsign/frequency and a read-only "Data source" label.
  * Thread-safe messaging backend with change detection to reduce unnecessary saves.

* **Tests**
  * Expanded tests covering messaging I/O, update cycles, parsing, and thread-safety.

### DIFF
--- a/src/FLARM/Glue.cpp
+++ b/src/FLARM/Glue.cpp
@@ -86,11 +86,14 @@ LoadFlarmDatabases() noexcept
 void
 ReloadFlarmDatabases() noexcept
 {
-  traffic_databases = new TrafficDatabases();
-
   /* the MergeThread must be suspended, because it reads the FLARM
      databases */
   backend_components->merge_thread->Suspend();
+
+  delete traffic_databases;
+  traffic_databases = nullptr;
+
+  traffic_databases = new TrafficDatabases();
 
   LoadSecondary(traffic_databases->flarm_names);
   LoadFlarmMessagingData(traffic_databases->flarm_messages);


src/Device/Driver/FLARM/StaticParser.cpp
```diff
@@ -174,3 +179,46 @@ ParsePFLAA(NMEAInputLine &line, TrafficList &flarm, TimeStamp clock, RangeFilter
...
+      } else if (msg_type == "PNAME"sv) {
+        record.pilot = ParseHexString(hex_value);
+      } else if (msg_type == "ATYPE"sv) {
+        record.plane_type = ParseHexString(hex_value);
+      } else if (msg_type == "ACALL"sv) {
+        record.callsign = ParseHexString(hex_value);
+      } else if (msg_type == "VHF"sv) {
+        // VHF message contains comma-separated frequencies (not hex-encoded)
+        // Only parse the first frequency from the comma-separated list
+        const auto comma = std::find(hex_value.begin(), hex_value.end(), ',');
```
lordfolken: Can you please enforce a size limit for parsing these? according to the standard its 17 chars so 34 bytes. 
VHF should also be a predictable length.

src/Device/Driver/FLARM/StaticParser.cpp
```diff
@@ -174,3 +179,46 @@ ParsePFLAA(NMEAInputLine &line, TrafficList &flarm, TimeStamp clock, RangeFilter
 
   flarm_slot->Update(traffic);
 }
+
+void
+ParsePFLAM(NMEAInputLine &line) noexcept
+{
```
lordfolken, line 185: can you add some documentation for the new function?

src/Dialogs/Traffic/FlarmTrafficDetails.cpp
```diff
@@ -269,6 +250,8 @@ FlarmTrafficDetailsWidget::Update()
     value = _T("--");
   SetText(CALLSIGN, value);
 
+  SetText(SOURCE, FlarmDetails::ToString(info.source));
```
lordfolken, line 253: The FlarmDetails::ToString should be wrapped in gettext(), since the source strings are marked as translateable.

lordfolken (changes requested):
ReloadFlarmDatabases() technically not part of your pr, but could you fix the leak by adding a destroy traffic_databases?

please use nullptr instead of NULL. 

Some comments in the code.  Thank you.

src/FLARM/MessagingFile.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "io/BufferedCsvReader.hpp"
+#include "util/StringStrip.hxx"
+#include "util/HexString.hpp"
+
+unsigned
+LoadFlarmMessagingFile(BufferedReader &reader, FlarmMessagingDatabase &db)
+{
+  // Read and ignore header
+  std::array<std::string_view, 5> header;
+  ReadCsvRecord(reader, header);
```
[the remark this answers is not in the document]
Scumi: I personally don't think this is needed, noone should really fiddle with this file.

lordfolken (approved):

PR #2002, opened 2025-12-26
unix/tchar: Use snprintf/vsnprintf instead of sprintf/vsprintf

yorickreum:
Replace deprecated sprintf/vsprintf with bounds-checked snprintf/vsnprintf in _stprintf/_vstprintf macros to eliminate deprecation warnings on macOS and improve buffer overflow protection.


yorickreum:
Superseded by #2377.

Issue #2005, opened 2025-12-28
XCSoar Testing crash on GetClientRect()


DanD222:
Installing the latest testing version of 2025-12-28 01:09 from https://download.xcsoar.org/testing/ANDROID/, XCSoar crashes on startup.

The device is a Marvsmart, RK3566 Android device.

The crash seems to occur in 
`12-28 15:20:00.382  5207  5244 F libc    : ./src/ui/window/Window.hpp:682: virtual const PixelRect Window::GetClientRect() const: assertion "IsDefined()" failed)`

adb logcat output:
[XCS_crash_20251228_0109.txt](https://github.com/user-attachments/files/24360468/XCS_crash_20251228_0109.txt)

DanD222:
Sure; XCSoar.log only has the following:

_[2025-12-28T15:57:51Z] Starting XCSoar v7.44-Android
[2025-12-28T15:57:51Z] EGL vendor: Android
[2025-12-28T15:57:51Z] EGL version: 1.4 Android META-EGL
[2025-12-28T15:57:51Z] EGL extensions: EGL_KHR_get_all_proc_addresses EGL_ANDROID_presentation_time EGL_KHR_swap_buffers_with_damage EGL_ANDROID_get_native_client_buffer EGL_ANDROID_front_buffer_auto_refresh EGL_ANDROID_get_frame_timestamps EGL_EXT_surface_SMPTE2086_metadata EGL_EXT_surface_CTA861_3_metadata EGL_KHR_image EGL_KHR_image_base EGL_EXT_image_gl_colorspace EGL_KHR_gl_colorspace EGL_KHR_gl_texture_2D_image EGL_KHR_gl_texture_cubemap_image EGL_KHR_gl_renderbuffer_image EGL_KHR_fence_sync EGL_KHR_create_context EGL_KHR_config_attribs EGL_KHR_surfaceless_context EGL_EXT_create_context_robustness EGL_ANDROID_image_native_buffer EGL_KHR_wait_sync EGL_ANDROID_recordable EGL_KHR_partial_update EGL_EXT_pixel_format_float EGL_KHR_mutable_render_buffer EGL_EXT_yuv_surface EGL_IMG_context_priority EGL_KHR_no_config_context 
[2025-12-28T15:57:51Z] EGL config: RGB=8/8/8 alpha=8 depth=24 stencil=8
[2025-12-28T15:57:51Z] GL vendor: ARM
[...]

PR #2008, opened 2025-12-29
src/ui/window/custom/TopWindow: guard macOS HiDPI native size

yorickreum:
Limit GetNativeSize() to macOS so Android/EGL doesn't see a zero size before the surface is ready, which triggers `Window::GetClientRect()` assertions on startup (issue #2005).

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined OpenGL drawable size handling on macOS to ensure correct HiDPI scaling and more reliable graphics rendering, preserving existing behavior on other platforms.

### DIFF
--- a/src/ui/window/custom/TopWindow.cpp
+++ b/src/ui/window/custom/TopWindow.cpp
@@ -9,6 +9,10 @@
 #include "ui/event/Globals.hpp"
 #include "Hardware/CPU.hpp"
 
+#ifdef __APPLE__
+#include <TargetConditionals.h>
+#endif
+
 #ifdef ANDROID
 #include "Android/Main.hpp"
 #include "Android/NativeView.hpp"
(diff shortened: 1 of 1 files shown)


lordfolken:
Lets fix it for now as its broken in the test build, and people would like to use them. 

I would love to avoid apple specific ifdefs whenever possible. 


This would have been a prefeered approach:
  size = screen->GetSize();```

#elif defined(ENABLE_OPENGL)
  // On HiDPI displays, the drawable size may differ from window size
  PixelSize native_size = screen->GetNativeSize();
  // On Android, surface might not be ready yet, so GetNativeSize() may return 0x0
  // In that case, use the size passed to Create() (which should have a fallback)
  if (native_size.width > 0 && native_size.height > 0)
    size = native_size;
```

lordfolken (approved):

yorickreum:
> I would love to avoid apple specific ifdefs whenever possible. 

That's fair. The many #ifdefs in the codebase really make it more annoying to reason about the codeflow. Can provide a different fix maybe also just checking with the `IsValid()` function that would later trigger the assertion as a guard.

lordfolken:
@yorickreum i put some cpu cycles into an an analysis, which i added as an issue. We can improve it at a later date. Right now the android part of 7.44 is my focus. 

 (and the ifdef hell is in no way apple specific. We should see what we can abstract out. The automatic analysis has yielded some concentrated files such as LocalPath.cpp)

PR #2030, opened 2026-01-06
Form/VScrollPanel, ui/control/ScrollBar: Add focusable scroll bar arrows

yorickreum:
Your PR should:
  * no fixup commits

## Summary by CodeRabbit

* **New Features**
  * Enhanced vertical scrollbar with interactive up/down arrow buttons supporting keyboard and mouse navigation.
  * Improved visual feedback for arrow states (focused, pressed, disabled) with dynamic foreground coloring.
  * Step-based scrolling with clamped bounds and synchronized arrow visibility for more precise, consistent scrolling.

### DIFF
--- a/src/Form/VScrollPanel.cpp
+++ b/src/Form/VScrollPanel.cpp
@@ -4,10 +4,98 @@
 #include "VScrollPanel.hpp"
 #include "Look/DialogLook.hpp"
 #include "ui/canvas/Canvas.hpp"
+#include "ui/event/KeyCode.hpp"
 
 #ifdef ENABLE_OPENGL
 #include "ui/canvas/opengl/Scissor.hpp"
 #endif
+
+class VScrollPanel::ScrollBarButton final : public PaintWindow {
+  VScrollPanel &panel;
+  int direction;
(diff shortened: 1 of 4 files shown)


lordfolken:
Currently when we are inside the content part of the page, using up and down on keyboard scrolls the content up and down should your next field be outside of the visual range. 

What is the advantage of moving the focus to the scrollbar? (Not cirtique, i'm tring to understand the rationale.)

yorickreum:
> What is the advantage of moving the focus to the scrollbar? (Not cirtique, i'm tring to understand the rationale.) 

Critique is always very welcome, as it might lead to a better solution. :)

I wrote this due to comment here https://github.com/XCSoar/XCSoar/pull/1867#issuecomment-3714635308
In the Quick Guide, I see no other intuitive way to move the focus from the next/previous buttons to inside the scroll area (and then outside again)... As both, moving the focus and scrolling is done via the Up/Down button?
If I remember correctly I also had the issue for some parts of the menu, is was not able to fully navigate it via the keyboard...
Maybe I'm missing something?

PR #2034, opened 2026-01-08
Form/Panel: Fix OpenGL panels not clearing background

yorickreum:
On OpenGL builds, PanelControl used ContainerWindow and didn’t clear the background, so scroll panels could show map content bleeding through, at least on iOS. Use SolidContainerWindow for OpenGL to clear the dialog background before painting children.

This fixes an issue where the map in **Configuration → Map Display → Terrain** leaked into the safe area due to an unnecessary scrollbar being rendered, as reported by @XCNav.

## Release Notes

* **Bug Fixes**
  * Improved rendering support for scenarios using OpenGL or hardware clipping features, ensuring consistent background handling across both display modes.

### DIFF
--- a/src/Form/Panel.cpp
+++ b/src/Form/Panel.cpp
@@ -9,7 +9,7 @@ PanelControl::Create(ContainerWindow &parent, [[maybe_unused]] const DialogLook 
                      const PixelRect &rc,
                      const WindowStyle style)
 {
-#ifdef HAVE_CLIPPING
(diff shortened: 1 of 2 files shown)


yorickreum:
Not quite sure what `HAVE_CLIPPING` is used for and if selecting on `ENABLE_OPENGL` is the right thing to do here ... it solves the issue on iOS but appears a bit too unselective to me. 

@coderabbitai?

PR #2043, opened 2026-01-10
Detect network state on linux, windows and ios and show network state on map

Scumi:
## Summary by CodeRabbit

* **New Features**
  * Linux network monitoring: app now reports network state (Connected / Disconnected / Unknown).
  * Map system-status indicator: network status icon shown on the map overlay alongside GPS info.

* **UI Assets**
  * Added network-connected and network-disconnected icons.

* **Platform**
  * Network status surfaced for Linux builds and integrated into the map UI.


lordfolken:
I like this feature. 
We should probably handle the downloads with this. GetnetState is currently sparingly used.

Scumi:
Windows implementation added.

lordfolken (approved):

src/MapWindow/GlueMapWindowOverlays.cpp
```diff
@@ -141,35 +142,56 @@ GlueMapWindow::DrawPanInfo(Canvas &canvas) const noexcept
 }
 
 void
-GlueMapWindow::DrawGPSStatus(Canvas &canvas, const PixelRect &rc,
-                             const NMEAInfo &info) const noexcept
+GlueMapWindow::DrawSystemStatus(Canvas &canvas, const PixelRect &rc) const noexcept
 {
-  const char *txt;
-  const MaskedIcon *icon;
```
lordfolken, line 148: Top-left placement note: DrawSystemStatus() starts at (left+FastScale(4), top+FastScale(4)), while DrawThermalBand() uses tb_rect left=rc.left..left+Scale(25), top=Scale(2)..; so overlap is likely on small screens / larger scaling. Might need coordination (offset/margin) between these overlays.

lordfolken:
Plain-text additional UX note:

- Suggestion: for these system-status overlays (network/GPS/FLARM), consider indicating only fault/degraded states (e.g. disconnected / no-fix / unavailable) and stay silent in the normal/OK case. This reduces clutter and keeps attention on actionable problems.

lordfolken (changes requested):

lordfolken:
I'm open to the general idea. But this mixes two things; interface state and connectivity. 

By implementing a period ping, you could have a platform neutral connectivity check. This would also resolve the issue you have with Fault-Only display.

Scumi:
I agree, that makes sense. If you want you can still already merge the net/State backend. Connectivity class would be separate.
I also think about not using an on-map icon, but transfer the status to an info box. Would you agree with that?

DanD222:
> I also think about not using an on-map icon, but transfer the status to an info box. 

My personal 2 cents here would be that for me, info boxes are precious screen real estate that I probably wouldn't dedicate to show network state. That's not to say someone else wouldn't, so I like the idea of having both an on-map icon, and an info box (for those that want the info in an info box).

Scumi:
Added a class that probes for actual internet reachability and displays the on-map icon based on this.

lordfolken:
I think on map is fine. This is something you might want to know just like gps/flarm failure.

lordfolken:
closed in favor of #2449

Issue #2052, opened 2026-01-12
Status message location


kobedegeest:
Status message location for Top left got changed to top center but setting is still named TopLeft.
Also the message avoids infoboxes but also when on page with no infoboxes 

XCSoar versions having the problem
-------------------------------------------------

Master

XCSoar versions not having the problem
-------------------------------------------------

I think it might be changed in cf9901ce62d2a83713c7abc5a79e30cdf22a9dfe (but not tested)

System information
------------------

Steps to reproduce the behavior
-------------------------------

Select in Config -> Look -> Screen Layout -> Message display: Topleft

get a message on the map screen with infoboxes and without infoboxes 

Expected behavior
-----------------

Setting name to match behavior and to not be rendered under invisible infoboxes but top of screen 

Actual behavior
---------------

Setting name top left render top center and on same place with or without infoboxes

Do you have any idea what may have caused this?
-----------------------------------------------

cf9901ce62d2a83713c7abc5a79e30cdf22a9dfe

Do you have an idea how to solve the issue?
-------------------------------------------

PR #2054, opened 2026-01-12
Adjust map window calculations to display PopupMessage and bottom offset margin properly

Scumi:
Based on findings of #2043 and reported issue #2052

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored map bottom margins reliably when info dialogs open or close to prevent content overlap.
  * Improved bottom-margin calculation in non-default/pan modes to better protect map controls across aspect ratios.
  * Ensured popup layouts are safely refreshed during fullscreen and layout changes.
  * Adjusted icon drawing to respect bottom margin; removed an obsolete comment in message sizing.

### DIFF
--- a/src/MapWindow/GlueMapWindowOverlays.cpp
+++ b/src/MapWindow/GlueMapWindowOverlays.cpp
@@ -279,6 +279,11 @@ GlueMapWindow::SetBottomMargin(unsigned margin) noexcept
 void
 GlueMapWindow::SetBottomMarginFactor(unsigned margin_factor) noexcept
 {
+  if (Layout::landscape && follow_mode == FOLLOW_PAN) {
+    SetBottomMargin(0);
+    return;
+  }
+
   if (margin_factor == 0) {
     SetBottomMargin(0);
     return;


src/Input/InputEvents.cpp
```diff
@@ -210,9 +210,14 @@ InputEvents::drawButtons(Mode mode, bool full) noexcept
...
        * when in Pan mode. */
-      map->SetBottomMarginFactor(menubar_height_scale_factor);
+      PixelRect screen_rect = map->GetParentClientRect();
+      unsigned factor = (screen_rect.GetHeight() > screen_rect.GetWidth())
+        ? menubar_height_scale_factor
+        : 5;
+      
+      map->SetBottomMarginFactor(factor);
     } else {
       map->SetBottomMarginFactor(0);
```
lordfolken, line 222: Hmm, this doesn´t work in landscape mode. 
image

Scumi: Hm, it seems to assume the bottom buttons. Didn't test it in pan mode. I'll have a look.

lordfolken: Use  a Layout::landscape condition  in GlueMapWindowOverlays.cpp directly, don´t go via input.

Scumi: Well in that case (5 bottom menu buttons + landscape) it works. So its essentially dependent on the amount of buttons shown. 
grafik
The more important case is pan mode however.

Issue #2027, opened 2026-01-04
Next waypoint direction arrow should be consistent in map and radar page


fraktal-sb:
XCSoar version
--------------

7.44

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------

The "next waypoint direction arrow" within the map and the radar display have different appearance. They should be consistent over the pages. In addition the arrow in the actual version is badly visible on most displays. Within the radar screen colours should only be reserved for targets. The arrow should therefore be in a neutral colour (grey)

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------

Here a proposal how this could be done (old/new)

<img width="602" height="832" alt="Image" src="https://github.com/user-attachments/assets/bb8348c6-30f1-45d7-bedb-9d1e4c4f1919" />

![Image](https://github.com/user-attachments/assets/551b0510-1e2d-4ca9-822c-781a38fa1ec8)

lordfolken:
I agree on the need forconsistency. 

I wanted to use color to do so, see:
https://github.com/XCSoar/XCSoar/issues/1733

fraktal-sb:
I would recommend not to use the magenta colour for this arrow. This conflicts with the maganta colour of an aircraft on the same height level (+-50m hardcoded). It could be confusing. Therfore I took the neutral grey colour. I think it is only a minor information in the flarm radar as for example the north indicator. Therfore the grey colour.

lordfolken:
**Open PR:** #2057 — BigTrafficWidget: next waypoint direction arrow shape aligned with map window.

lordfolken:
Fixed in #2057 (merged 2026-06-01): BigTrafficWidget task direction arrow uses the shared best-cruise outline between range rings, matching the map. Documented in NEWS.txt for v7.45.

PR #2057, opened 2026-01-14
BigTrafficWidget: Next waypoint direction arrow shape as in map window

fraktal-sb:
Your PR should:
  * no fixup commits

Related to #2027

## Summary by CodeRabbit

* **Style**
  * Updated the task direction arrow in the traffic widget to be rendered as an outline between range rings, matching the best-cruise shape on the map for improved visual consistency.

### DIFF
--- a/src/Gauge/BigTrafficWidget.cpp
+++ b/src/Gauge/BigTrafficWidget.cpp
@@ -21,9 +21,13 @@
 #include "UIUtil/GestureManager.hpp"
 #include "Formatter/UserUnits.hpp"
 #include "Renderer/UnitSymbolRenderer.hpp"
+#include "Renderer/BestCruiseArrowRenderer.hpp"
 #include "Input/InputEvents.hpp"
 #include "Interface.hpp"
 #include "Asset.hpp"
+#include "util/Macros.hpp"
+
+#include <algorithm>
 
 /**
  * A Window which renders FLARM traffic, with user interaction.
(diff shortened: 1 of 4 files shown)


fraktal-sb:
My opinion about leaving the commit as it is:

- The arrow is small
- It’s used in only two places
- already working, readable code

In cases like this, leaving it as-is is often the best engineering decision.

If later on:

- a third arrow appears, or
- the shape needs to change in multiple places, or
- rotation logic diverges

…then it might be worth revisiting with more context and less churn.

Until then: ship the code, keep it readable, move on

src/Gauge/BigTrafficWidget.cpp
```diff
@@ -287,23 +288,29 @@ FlarmTrafficControl::PaintTaskDirection(Canvas &canvas) const
...
-  triangle[1].x = 7;
-  triangle[1].y = triangle[0].y + 30;
-  triangle[2].x = -triangle[1].x;
-  triangle[2].y = triangle[1].y;
-
-  PolygonRotateShift(triangle, radar_renderer.GetCenter(),
-                     task_direction - (enable_north_up ?
-                                       Angle::Zero() : heading),
-                     Layout::FastScale(100u));
+  const Color color = COLOR_GRAY;
```
lordfolken: Please use the look.radar_pen or similar. if we want to change something then this allows it to do centrally. Then you also don't need the color.hpp include.

src/Gauge/BigTrafficWidget.cpp
```diff
@@ -287,23 +288,29 @@ FlarmTrafficControl::PaintTaskDirection(Canvas &canvas) const
...
+  Brush brush(color);
+  canvas.Select(brush);
+
+  int offset = -(int)radar_renderer.GetRadius()/Layout::FastScale(1)/4;
+
```
lordfolken: The scaling factor behavior is not the same as the rings. Resulting in somewhat strange behavior:

image

lordfolken (changes requested):

lordfolken:
image

please test with -dpi=400 or similar, to simulate smartphones etc.

lordfolken:
Rebased onto current `master` and applied the scaling fix from review:

- `PolygonRotateShift` now uses `Layout::FastScale(100u)` instead of `radar_renderer.GetRadius()`, matching `BestCruiseArrowRenderer` on the map so the arrow size stays consistent with the range rings when zooming.
- Commit subject typo fixed (`BigTrafficWidget`).

lordfolken (approved):
Scaling, annulus placement, and shared BestCruiseArrowRenderer helpers address the earlier review points.

PR #2060, opened 2026-01-15
Dialogs/StatusPanels/TaskStatusPanel.cpp: refresh on calculated update

### LINKED ISSUE #556: Status:Task Set MacCready value is not updated as soon as you change the value

XCSoar v7.0-Android , stable & develop

Expected behavior
-----------------

Calculation on the Status: task page are updated as soon as you change the value Set MacCready and close the selection dialog

Actual behavior
---------------

Calculated values remain the same, you have to change to other tab and then back to Task tab to update the values

kobedegeest:
This closes #556 

Calling refresh immediately after changing MC value is to fast,
calculation has no time to finish hence values are not updated after MC
change.

Now the panel is updated OnCalculatedUpdate().
Addition benefit, panel also updates during flying not just after MC
change.
Your PR should:
  * no fixup commits

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed redundant immediate refreshes to reduce UI flicker and unnecessary work.
  * Hiding the task status panel cancels pending refreshes and unregisters update listeners to prevent stray updates.

* **Improvements**
  * Showing the panel now triggers a refresh and registers for rate-limited calculated updates.
  * Panel refreshes in response to recalculation events, improving update accuracy.

### DIFF
--- a/src/Dialogs/StatusPanels/TaskStatusPanel.cpp
+++ b/src/Dialogs/StatusPanels/TaskStatusPanel.cpp
@@ -36,7 +36,6 @@ TaskStatusPanel::OnModified(DataField &df) noexcept
     const DataFieldFloat &dff = (const DataFieldFloat &)df;
     auto mc = Units::ToSysVSpeed(dff.GetValue());
     ActionInterface::SetManualMacCready(mc);
-    Refresh();
   }
 }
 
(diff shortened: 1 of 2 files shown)


lordfolken (changes requested):
I don´t like the timer approach.
Can you use the BlackboardListener pattern like SystemStatusPanel?

PR #2066, opened 2026-01-18
Deprecate TCHAR: Guidelines and initial modernization

yorickreum:
## Summary
This change establishes a project-wide framework for deprecating `TCHAR` in favor of UTF-8 (`std::string`) as the standard string type. It updates development guidelines and demonstrates the migration pattern by refactoring a small set of modules.

## Motivation
Current `TCHAR` usage creates a dual-path architecture (`wchar_t` on Windows, `char` on Unix), which leads to:

- Code complexity from `#ifdef _UNICODE` conditionals
- Maintenance burden of parallel code paths
- API inconsistency from mixing `TCHAR` and `std::string`

See also: https://github.com/XCSoar/XCSoar/discussions/1800.

### Target approach
Use **UTF-8 everywhere** internally and convert to wide strings **only at Windows API boundaries**.

- Internal code: `std::string` / `std::string_view` containing UTF-8
- Windows boundary: convert using `UTF8ToWideConverter` at the call site
- Avoid conversions in mid-layer utilities; push them outward to the platform edge

### Migration strategy
This migration is intended to be gradual. We will incrementally switch functions and interfaces from `TCHAR` to `std::string` while keeping legacy interfaces working via adapters.

During the transition, the number of `UTF8ToWideConverter` uses may temporarily increase. As more functions are converted to UTF-8, conversions should naturally move outward and eventually collapse to the API boundary.

## Development Guidelines
Updated `.cursor/rules/xcsoar-project-rules.mdc`:

- Mandate UTF-8 for all new code (`std::string` / `std::string_view`)
- Discourage new `TCHAR`, `tstring`, `_T()`, and `#ifdef _UNICODE` usage
- Document the `UTF8ToWideConverter` boundary-conversion pattern for Win32 calls
- Provide guidance for incremental migration (UTF-8 core + legacy wrappers)

## Refactors
- Refactored selected modules to return and operate on UTF-8 (`std::string`) and to perform wide conversions only at Windows API boundaries, improving consistency across platforms.

## Breaking Changes
None. This change maintains backward compatibility via conversion utilities and adapter layers.

### DIFF
--- a/src/Airspace/AirspaceParser.cpp
+++ b/src/Airspace/AirspaceParser.cpp
@@ -762,10 +762,8 @@ ParseLine(Airspaces &airspace_database, unsigned line_number,
     case 'X':
     case 'x':
       if (input.SkipWhitespace()) {
-        tstring tempString = tstring(
-            string_converter.Convert(input.c_str())); // Convert to tstring
         temp_area.transponder_code =
-            TransponderCode::Parse(tempString.c_str());
+            TransponderCode::Parse(std::string_view{input.c_str()});
       }
       break;
     }
--- a/src/Device/Driver/AirControlDisplay.cpp
+++ b/src/Device/Driver/AirControlDisplay.cpp
@@ -20,6 +20,8 @@
 #include "Operation/Operation.hpp"
 #include "time/PeriodClock.hpp"
 
+#include <cstdio>
+
 using std::string_view_literals::operator""sv;
 
 static bool
@@ -99,9 +101,10 @@ ParsePAAVS(NMEAInputLine &line, NMEAInfo &info)
      */
     unsigned code_value;
     if (line.ReadChecked(code_value)) {
-      StaticString<16> buffer;
-      buffer.Format(_T("%04u"), code_value);
-      TransponderCode parsed_code = TransponderCode::Parse(buffer);
+      char code_buffer[8];
+      std::snprintf(code_buffer, sizeof(code_buffer), "%04u", code_value);
+      TransponderCode parsed_code =
+          TransponderCode::Parse(std::string_view{code_buffer});
 
       if (!parsed_code.IsDefined())
         return false;
--- a/src/Dialogs/Airspace/dlgAirspaceDetails.cpp
+++ b/src/Dialogs/Airspace/dlgAirspaceDetails.cpp
@@ -51,12 +51,11 @@ AirspaceDetailsWidget::Prepare([[maybe_unused]] ContainerWindow &parent,
   AddMultiLine(airspace->GetName());
 
   const TransponderCode transponderCode = airspace->GetTransponderCode();
-  TCHAR buffer2[5];
-
-  transponderCode.Format(buffer2, sizeof(buffer2));
-
   if (transponderCode.IsDefined()) {
-    AddReadOnly(_("Squawk code"), nullptr, buffer2);
+    const auto code_str = transponderCode.Format();
+    const UTF8ToWideConverter code_wide(code_str.c_str());
+    if (code_wide.IsValid())
+      AddReadOnly(_("Squawk code"), nullptr, code_wide.c_str());
     AddButton(_("Set Squawk Code"), [transponderCode]() {
       ActionInterface::SetTransponderCode(transponderCode);
     });
--- a/src/InfoBoxes/Content/Radio.cpp
+++ b/src/InfoBoxes/Content/Radio.cpp
@@ -42,14 +42,19 @@ UpdateInfoBoxTransponderCode(InfoBoxData &data,
                              TransponderMode mode) noexcept
 {
   if(code.IsDefined()) {
-    code.Format(data.value.data(), data.value.capacity());
-
-    if (data.value.equals(_T("7500")) ||
-        data.value.equals(_T("7600")) ||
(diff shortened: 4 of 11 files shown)


src/Dialogs/Waypoint/WaypointInfoWidget.cpp
```diff
@@ -88,10 +89,14 @@ WaypointInfoWidget::Prepare(ContainerWindow &parent,
...
-    buffer += _T(" MHz");
-    AddReadOnly(_("Radio frequency"), nullptr, buffer);
+  const auto freq_str = waypoint->radio_frequency.Format();
+  if (!freq_str.empty()) {
+    const UTF8ToWideConverter radio_wide(freq_str.c_str());
+    if (radio_wide.IsValid()) {
+      buffer = radio_wide.c_str();
+      buffer += _T(" MHz");
+      AddReadOnly(_("Radio frequency"), nullptr, buffer);
+    }
```
Scumi, line 99: I cannot comment on the need to deprecate TCHAR, and I appreciate the effort if this change is agreed to be necessary. From a coding and readability perspective, however, this appears to be a regression to me (I’ve highlighted just one example for discussion - others are similar). I would also need to look at areas beyond Radio and Transponder - classes in which I’ve worked on in the past - to understand whether the situation is similar there.

Regarding this specific change, couldn’t the Format() function format directly into the wide string type (from UTF-8)? That would improve readability in the calling code and would also ensure the conversion cannot be forgotten, which is something I would currently be concerned about.

yorickreum: > Regarding this specific change, couldn’t the Format() function format directly into the wide string type (from UTF-8)? That would improve readability in the calling code 

As discussed in the description, the `UTF8ToWideConverter` here is only a temporary compatability hack until all functions are updated to consume standard string types instead of `TCHAR`. Once this is done, almost all the wide converters will vanish from the codebase, together with the `_T` macro. `Format()` should not return wides, this would be counterproductive, it should only take and return `utf8` strings.

This draft PR is mainly for reasoning about a pattern that makes it easily to incrementaly move away from having TCHAR macro function signatures. I don't think we can get away without any compatability bridge to legacy signatures, if we don't want to change all the code base at once (which would maybe also be an option ... just sit down for one day, update all signatures ... merge it in as one big, fat commit ...).

> and would also ensure the conversion cannot be forgotten, which is something I would currently be concerned about.

The compiler will make sure of this here (as long as it is needed, in the long run it won't be needed anymore...).

yorickreum: Alternatively, we add explicit UTF8 overloads for all the UI functions... `AddReadOnlyUTF8`, etc. ...

yorickreum: I am starting to think a bulk replace that get rids of it everywhere all at once will be the better choice. Will think about a clean way to do so and test it.

August2111: Hi Yorick (@yorickreum), I completely agree – or rather, I already implemented this in my Folk OpenSoar almost two years ago ;-) (see https://GitHub.com/OpenSoaring/OpenSoar, commits edbb0fca..8ff973f7d). Since then, my folk has been running very reliably even without UNICODE, TCHAR, etc. Since I primarily develop under Windows (Visual Studio), I also have a good feeling about the functionality of XCSoar/OpenSoar under Windows. After the modification, the WideStrings are only present in the GDI in the ui folder, nowhere else. I was discussing with Philipp (@lordfolken  last week that I should submit this modification as a pull request for XCSoar as soon as possible (unfortunately, I was sick last week, so I haven't gotten around to it yet). So, before you start tackling the problem twice, I'd like to present my existing version of the 'bulk replace' as a pull request – hopefully this week!

August2111: That would also address @Scumi's concerns, since there would no longer be any WideString conversions and back outside the UI... If it also works out to integrate @groundhogxc's pull request, we could even remove the entire GDI from the project in the end ;-)

groundhogxc: I can not comment on the discussion about what is the best route to go forward to enable special characters everywhere and in a consistent manner. I do not know enough about the history and future of text encodings.

It does feel like it might be some work to rebase my Win OpenGL PR #2063 on after a big UTF8 migration gets merged, but other than that these two things are quite separate, so that is fine, it is probably better to keep that separate.

It is more a question of what we do when, i.e. for which release. So do we do the UTF8 change right now or later?  And do we introduce the OpenGL Windows targets immediately or later? And then of course do we remove the GDI stuff right then or later? - I had assumed that, if that PR ever gets merged, we would keep GDI around for a bit longer in case some problem comes up (since we have no clue what devices people might sill use :) )

lordfolken: I'd really like to get [v7.44](https://github.com/XCSoar/XCSoar/milestone/17) out the door first. The last blocker I have there is the android play store.

groundhogxc: Agree 100%.

PR #2067, opened 2026-01-19
Fix android service issues

lordfolken:
Your PR should:
  * no fixup commits

## Summary by CodeRabbit

* **New Features**
  * Unified, user-friendly permission flow and cloud-enable dialog; simulator detection added.
  * GPS-required dialog to guide enabling location services.

* **Bug Fixes**
  * More accurate touch coordinates; improved fullscreen and safe-area handling.
  * Better text wrapping for long words/paths.
  * More reliable resize and event handling.

* **Improvements**
  * More robust foreground service and notification behavior.
  * Android target/tooling updated to API 35; adds notification and foreground-location permissions.

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -1,0 +1,128 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          package="org.xcsoar"
+          android:installLocation="auto"
+          android:versionCode="166"
+          android:versionName="7.44">
+  <application
+      android:requestLegacyExternalStorage="true"
+      android:preserveLegacyExternalStorage="true"
+      android:hasFragileUserData="true"
+      android:appCategory="maps"
+      android:icon="@drawable/icon"
+      android:label="@string/app_name"
+      android:theme="@style/XCSoarTheme">
+    <activity android:name=".XCSoar"
+              android:showOnLockScreen="true"
+              android:showWhenLocked="true"
+              android:exported="true"
+              android:icon="@drawable/icon"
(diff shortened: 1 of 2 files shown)


groundhogxc:
OK, I tested this branch, found no new issues. Specifically to the GPS permission handling, it worked well. Two observations:
 - when you touch the initial "requesting your permission" messages, the font color changes from light grey to a much darker grey. It goes back to the previous color as soon as you stop the touch. No big deal, just weird.
 - when you give the permissionto access location, but not background location, GPS does not connect even in the foreground. Might be on purpose though. I think it could be better this way (force people to enable background - because it is more likely that this happened on accident, and then it can create serious trouble on subsequent flights).

 Otherwise the permission process worked fine, I tried a few times; and two devices (Android 11 and 12 -- therefore 30 and 31 API, I think)

groundhogxc:
Actually, now that I think about it, I am slightly puzzled - I believe was never asked for a bluetooth permission on the Android 11 device. 
Is that correct?
Just tried, uninstall and install again (so need to give GPS permission again), I click on the port and I immediately see my bluetooth devices, no permission dialog. 

On the Android 12, I get asked for blutooth permission when I tap on the port type selection.

groundhogxc:
@coderabbitai Can you find the reason for the display bug, potentially as a screen buffer issue with the system displaying a stale buffered image or something along those lines?

groundhogxc:
> Would you like me to help draft a potential fix, or would you prefer to discuss this with @lordfolken first?

Sometimes these AI things are becoming a little spooky...

groundhogxc:
@coderabbitai Why is the Android build failing?

PR #2071, opened 2026-01-19
Infobox single action single click

### LINKED ISSUE #1442: When Pressing on Infoboxes that have only one option -> open the dialog directly
For example WP AltD Infobox
Short press opens Details of the Waypoint directly.
Long Press opens the current menu that allows also to change the infobox

kobedegeest:
closes #1442 
This PR updates all infoboxes that only have one action but use a panel to use HandleClick() instead reducing the pilots interaction time.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added defensive null checks to several info‑box handlers to prevent crashes when data or backend components are missing.

* **New Features**
  * Many info‑box items now open their dialogs/modals immediately on click and return success/failure feedback.

* **Refactor**
  * Simplified dialog handling by removing panel-registration indirection and migrating many info‑box entries to a unified creation/click pathway.

### DIFF
--- a/src/InfoBoxes/Content/Trace.cpp
+++ b/src/InfoBoxes/Content/Trace.cpp
@@ -144,12 +144,15 @@ InfoBoxContentBarogram::OnCustomPaint(Canvas &canvas, const PixelRect &rc) noexc
                        backend_components->protected_task_manager.get());
 }
 
-static void
+static bool
 ShowAnalysisBarograph() noexcept
 {
-  if (!backend_components || !data_components || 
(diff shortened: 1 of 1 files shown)


kobedegeest:
Outside the scope for this PR but i see no reason to keep the Switch infobox button on the remaining panels (this is also not consistent for the moment see MC and altitude vs Radio Frequency), reason being that long press works just fine to switch the infobox regardless whether there is a panel or not. Hence further streamlining the UI. @lordfolken you agree?

PR #2100, opened 2026-01-26
FlarmTrafficWindow: Traffic target information in radar page

### LINKED ISSUE #1719: Enhanced Traffic Radar Page
XCSoar version
--------------

7.44

Currently the radar page has 2 views. They are accessible by gesture (see documentation) and show targets with either climb/sink- rate or the height difference. 

These are shown in the following screenshots:

Image

View "A" shows the targets with climb/sink. View "B" shows the targets with their height difference in height unit (meter or feet)

Because a lot more of information came out of current traffic awareness systems (Stratux, Flarm...) an enhanced version of the radar page **(only "B"**) is proposed. View "A" should still contain the reduced information.

Actually a test version (with incoming data from the Stratux receiver) is implemented showing the possibilitys

Image

Here an explanation of the displayed data:

1) callsign from Mode-S Transponder
2) instead of displaying in full accuracy the height difference is only shown in "hundreds" of the height unit (as in the small radar
3) Mode-S targets with no position info are displayed in flight direction with their distance and height difference. To distinguish these targets from the ones with relative east/north they are displayed as a circle with exclamation mark. It means that the targets can be within every direction but with the received height and distance. the existing colour coding (green,pink,blue) is not affected.
4) the callsign instead of the ICAO-Hexcode is displayed
5) the north direction is no more displayed as a circle (see 4). It is shown only as a "N" and outside the last distance ring

If discussed/accepted these information can be implemented by single feature request for each of the points above or as a whole.

If possible we should coordinate this with https://github.com/XCSoar/XCSoar/issues/1700 because the colour change in units of 50m is not coincident with the displayed height difference in hundreds of the height difference. 

The small traffic radar on the map page should also contain "no position traffic" as a circle element

In addition we also should have more space for callsign information (6) and therefore we should change the position of the "+" and "-" zoom buttons to the right border of the radar

Image

In addition we could think about a short form of the displayed callsign (last 2 characters) as in the following screenshot

Image

fraktal-sb:
Thank you for contributing to XCSoar!

We truly appreciate your time and effort in making XCSoar better. We know
that contributing to an open-source project can be challenging, and we're
grateful that you've chosen to help.

This checklist is here to help guide you through the submission process.
Don't worry if you're not sure about something—feel free to ask questions
or submit your PR, and we'll work together to get it ready.

For coding, style guide, architecture information please see our
[development guide](https://xcsoar.readthedocs.io/en/latest/index.html).

For Git tips and tricks, including interactive rebase, fixup commits, and
common workflows, see the
[Git Tips](https://xcsoar.readthedocs.io/en/latest/git_tips.html)
documentation.

For testing and debugging utilities, see the
[Test & Debug
Utilities](https://xcsoar.readthedocs.io/en/latest/test_debug_utilities.html)
documentation. 

## Pre-Submission Checklist

#### Branch & Rebase
- [ ] Use `git rebase -i` to clean up commit history before PR
  submission

#### Commit Format & Messages
- [ ] All commits follow the format: `<Component>: <Summary>` (no
  `src/` prefix)
- [ ] Use present tense in commit messages ("Fix" not "Fixed", "Add"
  not "Added")
- [ ] Commit messages explain *why* the change was made, not just
  *what* changed
- [ ] Commit message body (if needed) provides detailed reasoning and
  context

#### Commit Structure
- [ ] Each commit is atomic and builds successfully (every commit
  must compile)
- [ ] One commit per logical change (don't mix refactoring with
  feature changes)
- [ ] Self-contained commits (each commit changes one thing)
- [ ] No fixup commits (squashed into parent commits using
  `git rebase -i`)
- [ ] No duplicate commits (check with `git log --oneline`)
- [ ] No "WIP" or "testing" commits (clean up before PR)

---

- [ ] I'm ready to merge

## Summary by CodeRabbit

* **Bug Fixes**
  * Consistent default colors/brushes/pens applied to radar targets to prevent missing visuals.
  * Text rendering always uses correct sizing and placement.

* **New Features**
  * Optional callsign shown beside traffic targets when relative-altitude side display is active.

* **Improvements**
  * Unified display rules for climb rate and relative altitude; climb rate shown only when significant and available.
  * Streamlined positioning and sizing for more consistent traffic labels.

[Review Change Stack](https://app.coderabbit.ai/change-stack/XCSoar/XCSoar/pull/2100?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/Gauge/FlarmTrafficWindow.cpp
+++ b/src/Gauge/FlarmTrafficWindow.cpp
@@ -422,64 +422,78 @@ FlarmTrafficWindow::PaintRadarTarget(Canvas &canvas,
                      Layout::Scale(100u));
 
   // Select pen and brush
+  if (target_brush == nullptr) {
+    target_brush = &look.passive_brush;
+    target_pen = &look.passive_pen;
+    text_color = &look.passive_color;
+  }
   canvas.Select(*target_pen);
-  if (target_brush == nullptr)
-    canvas.SelectHollowBrush();
-  else
-    canvas.Select(*target_brush);
+  canvas.Select(*target_brush);
+  canvas.SetTextColor(*text_color);
 
+  // Select font; prepare object sizes and distances by text height as reference
+  canvas.Select(look.label_font);
+  const PixelSize sx = canvas.CalcTextSize("X");
+
+  canvas.SetBackgroundTransparent();
   if (!traffic.relative_east) {
-    // Select font; prepare object sizes and distances by text height as reference
-    canvas.SetBackgroundTransparent();
-    canvas.Select(look.label_font);
-    const PixelSize sx = canvas.CalcTextSize("X");
-     // No position targets - Paint the dot
+    // No position targets - Paint the dot
     PaintNoPositionTarget(canvas, sc[i], radar_mid, scale, small, sx, target_pen, text_color);
-   } else
-     // All other targets - Draw the polygon
-     canvas.DrawPolygon(Arrow, 4);
+  } else
+    // All other targets - Draw the polygon
+    canvas.DrawPolygon(Arrow, 4);
 
   if (small) {
     if (!WarningMode() || traffic.HasAlarm())
       PaintTargetInfoSmall(canvas, traffic, i, *text_color, *arrow_brush);
-
     return;
   }
 
-  // if warning exists -> don't draw vertical speeds
+  // if warning exists -> don't draw side labels on other targets
   if (WarningMode())
     return;
 
-  // if vertical speed to small or negative -> skip this one
-  if (side_display_type == SideInfoType::VARIO &&
-      (!traffic.climb_rate_avg30s_available ||
-       traffic.climb_rate_avg30s < 0.5 ||
-       traffic.IsPowered()))
-      return;
-
-  // Select font
-  canvas.SetBackgroundTransparent();
   canvas.Select(look.side_info_font);
-
-  // Format string
-  char tmp[10];
-
-  if (side_display_type == SideInfoType::VARIO)
-    FormatUserVerticalSpeed(traffic.climb_rate_avg30s, tmp, false);
-  else
-    FormatRelativeUserAltitude(traffic.relative_altitude, tmp, true);
-
-  PixelSize sz = canvas.CalcTextSize(tmp);
-  const PixelPoint tp{
-    sc[i].x + int(Layout::FastScale(11u)),
-    sc[i].y - int(sz.height / 2),
-  };
-
-  // Select color
   canvas.SetTextColor(*text_color);
 
-  // Draw vertical speed
-  canvas.DrawText(tp, tmp);
+  // Draw callsign only in combination with relative altitude
+  if (traffic.HasName() &&
+      side_display_type == SideInfoType::RELATIVE_ALTITUDE) {
+    const PixelPoint ts{
+      sc[i].x + int(sx.height),
+      sc[i].y - int(sx.height * 3 / 2),
+    };
+
+    const size_t len = traffic.name.length();
+    const std::string_view text = (len > 2)
+      ? std::string_view(traffic.name.c_str() + (len - 2), 2)
+      : std::string_view(traffic.name);
+
+    canvas.DrawText(ts, text);
+  }
+
+  StaticString<10> side_text;
+
+  if (side_display_type == SideInfoType::VARIO) {
+    if (traffic.climb_rate_avg30s_available &&
+        traffic.climb_rate_avg30s > 0.5 &&
+        !traffic.IsPowered())
+      FormatUserVerticalSpeed(traffic.climb_rate_avg30s,
+                              side_text.data(), false);
+  } else {
+    const int relalt =
+      iround(Units::ToUserAltitude(traffic.relative_altitude) / 100);
+    if (relalt != 0)
+      side_text.Format("%+d", relalt);
+  }
+
+  const PixelPoint tp{
+    sc[i].x + int(sx.height),
+    sc[i].y - int(sx.height / 2),
+  };
+
+  if (!side_text.empty())
+    canvas.DrawText(tp, side_text);
 }
 
 void
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -42,6 +42,9 @@ Version 7.45 - not yet released
     recently-visited waypoints again; it had been silently aliased to the
     unrelated "Map file" filter since the multi-file rework (#1994) due to
     a positional dropdown/enum mismatch
+  - large FLARM radar: optional callsign and relative altitude (in hundreds,
+    as on the small gauge) beside targets when side data is in altitude mode;
+    passive targets use default colours when no altitude brush is set
   - waypoint search: Type filter now lists every CUP waypoint type
     (Mountain Top/Pass, Bridge, Tunnel, Tower, Power Plant, Obstacle,
     Thermal Hotspot, Marker, VOR, NDB, Dam, Castle, Intersection,
--- a/src/MainWindow.cpp
+++ b/src/MainWindow.cpp
@@ -1006,6 +1006,15 @@ MainWindow::SetFullScreen(bool _full_screen) noexcept
   if (widget != nullptr)
     widget->Move(GetMainRect());
 
+  /* Overlapped gauges (FLARM, thermal assistant) use GetMainRect() for
+     "avoid InfoBoxes" corners; re-layout when fullscreen changes. */
+  const PixelRect rc = GetClientRect();
+  const InfoBoxLayout::Layout ib_layout =
+    InfoBoxLayout::Calculate(rc,
+                             CommonInterface::GetUISettings().info_boxes.geometry);
+  ReinitialiseLayout_flarm(rc, ib_layout);
+  ReinitialiseLayoutTA(rc, ib_layout);
+
   if (map != nullptr)
     map->FastMove(GetMainRect());
 


src/Gauge/FlarmTrafficWindow.cpp
```diff
@@ -422,64 +422,77 @@ FlarmTrafficWindow::PaintRadarTarget(Canvas &canvas,
...
+    text_color = &look.passive_color;
+  }
   canvas.Select(*target_pen);
-  if (target_brush == nullptr)
-    canvas.SelectHollowBrush();
-  else
-    canvas.Select(*target_brush);
+  canvas.Select(*target_brush);
 
+  canvas.Select(look.label_font);
```
lordfolken, line 435: The old code used `look.side_info_font` for the vario/altitude side info text. This now uses `look.label_font` for everything. These two fonts likely have different sizes, so this changes the rendering. Was this intentional? If so it should be mentioned in the commit message.

src/Gauge/FlarmTrafficWindow.cpp
```diff
@@ -422,64 +422,77 @@ FlarmTrafficWindow::PaintRadarTarget(Canvas &canvas,
...
-    const PixelSize sx = canvas.CalcTextSize(_T("X"));
-     // No position targets - Paint the dot
+    // No position targets - Paint the dot
     PaintNoPositionTarget(canvas, sc[i], radar_mid, scale, small, sx, target_pen, text_color);
-   } else
-     // All other targets - Draw the polygon
-     canvas.DrawPolygon(Arrow, 4);
+  } else
+    // All other targets - Draw the polygon
+    canvas.DrawPolygon(Arrow, 4);
```
lordfolken, line 444: `canvas.SetBackgroundTransparent()` is only called inside the `!traffic.relative_east` branch (line 437). For positioned targets (the `else` branch that draws the polygon), it is never called before the subsequent callsign and vario/altitude text drawing further below. This could cause text to render with an opaque background.

Consider moving `canvas.SetBackgroundTransparent()` before the `if` so it applies to both paths.

src/Gauge/FlarmTrafficWindow.cpp
```diff
@@ -422,64 +422,77 @@ FlarmTrafficWindow::PaintRadarTarget(Canvas &canvas,
...
-  // Select font
-  canvas.SetBackgroundTransparent();
-  canvas.Select(look.side_info_font);
+    size_t len = traffic.name.length();
 
-  // Format string
-  TCHAR tmp[10];
+    tstring_view text = (len > 3)
+        ? tstring_view(traffic.name.c_str() + (len - 2), 2)
+        : tstring_view(traffic.name);
```
lordfolken: The truncation logic takes the last 2 characters when the name is longer than 3 characters. This seems arbitrary and could be confusing for the user — e.g. a callsign `D-KABC` would display as `BC`, `N12345` would display as `45`. Could you add a comment explaining the rationale (competition number suffix? registration tail?).

src/Gauge/FlarmTrafficWindow.cpp
```diff
@@ -422,64 +422,77 @@ FlarmTrafficWindow::PaintRadarTarget(Canvas &canvas,
...
+  if (side_display_type == SideInfoType::VARIO) {
+    if (traffic.climb_rate_avg30s_available &&
+        traffic.climb_rate_avg30s > 0.5 &&
+        !traffic.IsPowered()) {
+      FormatUserVerticalSpeed(traffic.climb_rate_avg30s,txtbuf,false);
+    }
+  } else {
+    const short relalt = iround(Units::ToUserAltitude(traffic.relative_altitude) / 100
+    );
+    _stprintf(txtbuf, _T("   %+d"), static_cast<int>(relalt));
```
lordfolken: This replaces the old `FormatRelativeUserAltitude(traffic.relative_altitude, tmp, true)` with a manual `_stprintf` and a hardcoded division by 100. A few concerns:

1. **Display change**: The old code used `FormatRelativeUserAltitude` which produced a properly formatted relative altitude string with unit suffix (`include_unit=true`). The new code divides by 100 after unit conversion, so it displays altitude in *hundreds* of feet/meters (e.g. `+3` instead of `+300ft`). Is this the intended behavior?

2. **Type narrowing**: `iround()` returns `int`, but the result is stored in a `short` which could overflow for large altitude differences. Just use `int`.

3. **Hardcoded padding**: The 3-space prefix `"   %+d"` is a fragile way to offset the text from the target arrow. Consider using proper positioning via the `PixelPoint` calculation instead.

4. **Style**: `if (!traffic.HasAlarm()) _stprintf(...)` puts the statement on the same line as the `if`. XCSoar style puts the body on a separate line.

src/Gauge/FlarmTrafficWindow.cpp
```diff
@@ -422,64 +422,77 @@ FlarmTrafficWindow::PaintRadarTarget(Canvas &canvas,
...
+
+  // Draw climb/sink rate or relative height, skip if vertical speed to small or negative
+  TCHAR txtbuf[10];
+  txtbuf[0]='\0';
+
+  if (side_display_type == SideInfoType::VARIO) {
+    if (traffic.climb_rate_avg30s_available &&
+        traffic.climb_rate_avg30s > 0.5 &&
+        !traffic.IsPowered()) {
+      FormatUserVerticalSpeed(traffic.climb_rate_avg30s,txtbuf,false);
```
lordfolken: Minor style nits:
- Line 472: `txtbuf[0]='\0';` — missing spaces around `=`.
- Line 478: `FormatUserVerticalSpeed(traffic.climb_rate_avg30s,txtbuf,false);` — missing spaces after commas.

src/Gauge/FlarmTrafficWindow.cpp
```diff
@@ -422,64 +422,77 @@ FlarmTrafficWindow::PaintRadarTarget(Canvas &canvas,
...
+    sc[i].y - int(sx.height / 2),
   };
 
   // Select color
   canvas.SetTextColor(*text_color);
 
-  // Draw vertical speed
-  canvas.DrawText(tp, tmp);
+  // Draw vertical speed or relative altitude
+  canvas.DrawText(tp, txtbuf);
```
lordfolken: When in VARIO mode and the conditions aren't met (climb rate too low, unavailable, or powered), `txtbuf` remains empty. `canvas.DrawText(tp, txtbuf)` is still called, which is a no-op but wasteful. Consider guarding this with `if (txtbuf[0] != '\0')`.

lordfolken (changes requested):
Thanks for working on improving the radar target info display. There are a few issues that should be addressed before merging.

src/Gauge/FlarmTrafficWindow.cpp
fraktal-sb: will move this outside

src/Gauge/FlarmTrafficWindow.cpp
fraktal-sb: In communication with ATC the last 2 or 3 characters of the tail number as abbreviation is widely used. Within the example of https://github.com/user-attachments/files/19562368/2025-03-29_14-10.nmea.log this abbreviation (call-sign) is stored in traffic.name, which we all ways have together with the Hex-Id. All other information in traffic.name from other sources as Flarm should not be displayed in full length. I assume the standard tail number format as in Germany or France. There only the last 2 characters are used in communication. With this "construction we have the 2 or 3 short code used by gliders and the 2 short code in general aviation, so my thoughts. But I am completely unemotional. We also could always display 3 chars but then we have more information to the pilot he must sort out.

src/Gauge/FlarmTrafficWindow.cpp
fraktal-sb: ad 1) yes, we already have this in the small radar. so this is only a consistency adaption. Who needs the precise information +283 Meters or feed height difference? The rough information is more then enough. What we could do is to add the unit "hm" or "hf" but we already don't have this information in the small radar and if we add this the the display is more clear but at the end we have a lot of redundant information. I would not recommend this.

ad 2) will do this

ad 3) will do this

ad 4) will do this

In addition this mode is only available if chosen by the LR gesture according handbook. The default to display the climb rate only is untouched

src/Gauge/FlarmTrafficWindow.cpp
fraktal-sb: will do this

src/Gauge/FlarmTrafficWindow.cpp
fraktal-sb: will do this

lordfolken (commented):
## Summary

Focused improvement to **large FLARM radar** target labels: fixes missing pen/brush when `target_brush` is null (notably passive targets in warning focus), shows **relative altitude in hundreds** (matching the small radar / gauge style), and adds a **callsign** (last two characters) when side info is in relative-altitude mode. Single file, CI green on all platforms.

**Verdict:** Good direction and the null-brush fix is valuable. **Request changes** on one likely warning-mode regression, a few consistency nits, and release-notes / checklist items before merge.

---

## What works well

- **Null `target_brush` handling:** Defaulting to `passive_brush` / `passive_pen` / `passive_color` before draw avoids hollow/unstyled targets in warning-focus mode (previously `SelectHollowBrush()` when brush was unset).
- **Relative altitude format:** `/100` display aligns large radar with `PaintTargetInfoSmall` and the commit description (e.g. +245 m → `+2`).
- **Callsign only with `RELATIVE_ALTITUDE`:** Keeps vario mode uncluttered.
- **VARIO gating:** Climb text only when rate is available, > 0.5 m/s, and not powered — same thresholds as before, but without skipping the whole function when vario is weak.
- **Title/commit component** matches the file (`FlarmTrafficWindow:`).

---

## Requested changes

### 1. Warning-mode side labels (likely regression)

**Before:** `if (WarningMode()) return;` — when a warning target is focused, **no** traffic got vario/relative-alt labels on the large radar.

**After:** `if (traffic.HasAlarm()) return;` — only alarm-class targets are skipped; **other** traffic still get labels while warning mode is active.

Please confirm this is intentional. If the goal is only to skip labels on the alarming aircraft itself, consider `if (WarningMode() && traffic.HasAlarm())` (or equivalent) so unrelated targets stay hidden during warning focus.

### 2. Skip `+0` relative altitude (match small radar)

`PaintTargetInfoSmall` returns early when `relalt == 0`. The new path always `sprintf(..., "%+d", relalt)` and draws, so **+0** can appear. Skip drawing when `relalt == 0` for consistency.

### 3. Prefer existing formatters / buffers

- Relative alt: reuse the same pattern as `PaintTargetInfoSmall` (`StaticString` + `buffer.Format`) instead of `sprintf`.
- Project style prefers `fmt` / `StringFormat` over raw `sprintf` in touched code.

### 4. Callsign truncation and UTF-8

`traffic.name` is `StaticString<10>`; taking `c_str() + (len - 2)` is **byte-based**. Prefer a safe last-*character* slice (or document ASCII-only FLARM names). Also verify last-two-chars is the intended UX (vs full short name).

### 5. Font consistency

Large radar previously used `look.side_info_font` for side numeric text; this uses `look.label_font` for both callsign and numbers. Fine if deliberate — please confirm readability on Kobo / small layouts.

### 6. `NEWS.txt`

User-visible radar labelling change should get a bullet under the unreleased version (e.g. `* gauge` or `* ui`).

### 7. PR checklist

Template is present but **all boxes are unchecked**. Please tick what applies after rebase (commit format/body are mostly OK for a single-commit PR).

---

## Suggestions (non-blocking)

- Commit body typos: "addidion", "comming".
- Draw callsign with `side_info_font` if it should match numeric side text size on dense radar.
- Manual test: RL gesture toggle vario ↔ relative altitude; warning focus on alarm target; passive traffic in warning mode; ADS-B/FLARM name lengths 1–10.

---

## Test plan

- [ ] Large radar, relative-alt mode: hundreds display, callsign when name present
- [ ] Large radar, vario mode: climb only when > 0.5 m/s; no callsign
- [ ] Warning focus: confirm whether non-alarm traffic should show side labels
- [ ] Warning-mode passive target: filled/outline correct (brush fix)
- [ ] Kobo / memory canvas: labels readable

PR #2164, opened 2026-02-09
ui/canvas/opengl/Init: Fix projection matrix for software-rotated displays

### LINKED ISSUE #2163: Display scaling broken on OpenVario (Cubie2)

Current HEAD of 4.77 as of Feb 9 or earlier.

Current HEAD of 4.77 as of Feb 2.
So it broke between Feb 2 and Feb 9.

System information
------------------

OpenVario box (Cubie2)
`uname -a`
`Linux ov-DKXRN 6.6.8+ #1 SMP Wed Dec 20 16:02:07 UTC 2023 armv7l GNU/Linux`

Configure screen in Portrait orientation. Then start xcsoar.
Image

Expected behavior
-----------------

Should look like this:
Image

Actual behavior
---------------

In Portrait orientation the screen is wider than the graphics, and the screen is not high enough for the graphics.

The issue doesn’t appear on a “normal” Linux machine (Debian Trixie). So debugging on the target seems the only way.

It works OK in Landscape orientation!

I looked at the list of commits between Feb 2 and Feb 9.
The only one which sounded a bit suspect is:
5fc1edb1d9 ui/window: Enforce minimum window size (320x240)

So I reverted that, but it didn’t help.

lordfolken:
## Summary

- Fix portrait display scaling broken on OpenVario (and all `SOFTWARE_ROTATE_DISPLAY` platforms) caused by commit 1e4b8fee4f moving the `projection_matrix` computation before `OrientationSwap()`
- The ortho projection must use the swapped (virtual screen) dimensions, not the native EGL surface dimensions

## Root Cause

In `OpenGL::SetupViewport()`, the `projection_matrix` was moved before `OrientationSwap()` by commit 1e4b8fee4f. This meant the orthographic projection used the native surface dimensions (e.g. 480×800 for portrait) instead of the virtual screen dimensions (800×480 after swap). The rotation matrix then operated on the wrong base projection, causing the rendering to be distorted and misaligned.

Android was unaffected because it does not define `SOFTWARE_ROTATE_DISPLAY`. All other OpenGL platforms (Linux/MESA_KMS, X11, Wayland) with portrait orientation were broken.

On the OpenVario specifically, `DetectInitialOrientation()` reads `/sys/class/graphics/fbcon/rotate` and feeds externally-rotated framebuffer orientations into the software rotation pipeline, so even external rotation triggered the bug.

## Test plan

- [ ] @bomilkar: Please test on OpenVario Cubie2 in portrait orientation (external framebuffer rotation) and verify the display renders correctly
- [ ] Test with XCSoar "Display orientation" set to Portrait (with landscape external rotation)
- [ ] Verify landscape orientation still works correctly

Fixes #2163

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the order of display viewport initialization to ensure screen orientation adjustments are properly applied.

### DIFF
--- a/src/ui/canvas/opengl/Init.cpp
+++ b/src/ui/canvas/opengl/Init.cpp
@@ -221,11 +221,12 @@ OpenGL::SetupViewport(UnsignedPoint2D size) noexcept
   window_size = size;
 
   glViewport(0, 0, size.x, size.y);
-  projection_matrix = glm::ortho<float>(0, size.x, size.y, 0, -1, 1);
 
 #ifdef SOFTWARE_ROTATE_DISPLAY
   OrientationSwap(size, display_orientation);
 #endif
+
+  projection_matrix = glm::ortho<float>(0, size.x, size.y, 0, -1, 1);
 
 #ifdef SOFTWARE_ROTATE_DISPLAY
   glm::mat4 rot_matrix = glm::rotate(
@@ -235,7 +236,6 @@ OpenGL::SetupViewport(UnsignedPoint2D size) noexcept
   projection_matrix = rot_matrix * projection_matrix;
 #endif
 
-  /* viewport_size is the EGL surface size (which is already the safe area in non-fullscreen) */
   viewport_size = size;
 
   UpdateShaderProjectionMatrix();


bomilkar:
External orientation: Portrait
Display orientation: Portrait
p11

bomilkar:
External orientation: Portrait
Display orientation: Landscape
p12

PR #2183, opened 2026-02-13
Dialogs/WidgetDialog: re-run button layout after widget.Show()

### LINKED ISSUE #2176: "Select None" Button wraps to cut off text in Multi File Select Dialog (minor issue)
Version: Current Master, UNIX, X11

Reproduce: For example in the waypoint selector: Click at least one file, I see this:

Image

Scumi:
ShowModal() already lays out buttons before showing the widget, but with scaled fonts some text metrics are only fully available after show, causing initial button captions to clip until resize. Run a second buttons.UpdateLayout() immediately after widget.Show() so initial button widths are correct on first draw.

fixes #2176 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dialog button layout so captions render correctly after dialogs are shown, addressing clipping issues under certain scaling or font settings.

### DIFF
--- a/src/Dialogs/WidgetDialog.cpp
+++ b/src/Dialogs/WidgetDialog.cpp
@@ -176,6 +176,13 @@ WidgetDialog::ShowModal()
     widget.Move(buttons.UpdateLayout());
 
   widget.Show();
+  if (!auto_size) {
+    /* Ensure button layout is recalculated with any metrics that may have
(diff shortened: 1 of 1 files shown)


lordfolken (approved):
its a good work around. lets keep it for now.

lordfolken (approved):

PR #2200, opened 2026-02-15
Add configurable display cutout avoidance for Android

johnthomas22:
## Summary
- Add a new **"Avoid display cutout"** setting under Settings > Look > Layout (Android only)
- When enabled in fullscreen mode, UI elements (InfoBoxes, menus, etc.) are positioned within the display's safe area, avoiding rounded corners and camera notches
- Default is **off** (upstream behavior — fullscreen uses the entire screen)

## Implementation
Follows the same pattern as the existing iOS safe-area handling:
- Java passes real display cutout insets to C++ via `resizedNative()` (the parameters were already in the JNI signature but previously ignored)
- C++ `TopWindow` stores the insets and overrides `GetClientRect()` to apply them when the setting is enabled
- Safe-area handling is now in one place on both platforms (TopWindow)

## Files changed
- `src/ui/window/TopWindow.hpp` — Inset storage, `GetClientRect()` override, `SetAvoidDisplayCutout()`/`SetDisplayCutoutInsets()` methods
- `src/Android/Main.cpp` — Pass inset params from `resizedNative()` to TopWindow
- `android/src/XCSoar.java` — Compute and pass real display cutout insets via `resizedNative()`
- `android/src/NativeView.java` — Pass display cutout insets in `surfaceChanged()`
- `src/MainWindow.cpp` — Full-screen clear on Android to cover inset edges
- `src/Dialogs/Settings/Panels/LayoutConfigPanel.cpp` — UI control and save logic
- `src/DisplaySettings.hpp/cpp` — `avoid_display_cutout` field (default: false)
- `src/Profile/Keys.hpp`, `src/Profile/UIProfile.cpp` — Profile persistence
- `src/Startup.cpp` — Apply setting at startup

## Test plan
- [ ] Default (setting off): fullscreen uses entire screen, no inset margins
- [ ] Setting on: UI elements move inside safe area, map renders full screen, black bars at edges
- [ ] Toggle setting in Layout panel: layout updates immediately
- [ ] Setting persists across restarts
- [ ] Non-fullscreen mode: unchanged behavior

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added Android "Avoid display cutout" setting to control layout around notches and screen cutouts; persists across launches.

* **Bug Fixes**
  * Improved resizing and layout on Android by correctly accounting for display cutouts when rendering and notifying the app.
  * Cleared framebuffer to eliminate visual artifacts in areas affected by cutouts, improving fullscreen rendering on notched devices.

### DIFF
--- a/src/Look/Colors.hpp
+++ b/src/Look/Colors.hpp
@@ -51,4 +51,10 @@ static constexpr Color COLOR_ADMONITION_IMPORTANT_DARK =
 static constexpr Color COLOR_ADMONITION_TIP =
   Color(0x00, 0x80, 0x00);
 
+/**
+ * A muted green readable on light backgrounds.
+ * Standard COLOR_GREEN (0,255,0) is too bright on white.
+ */
+static constexpr Color COLOR_LIGHT_GREEN = Color(0x00, 0xc0, 0x00);
+
 static constexpr uint8_t ALPHA_OVERLAY = 0xA0;
--- a/src/Look/InfoBoxLook.cpp
+++ b/src/Look/InfoBoxLook.cpp
@@ -43,7 +43,7 @@ InfoBoxLook::Initialise(bool _inverse, bool use_colors,
   if (HasColors() && use_colors) {
     colors[1] = inverse ? COLOR_INVERSE_RED : COLOR_RED;
     colors[2] = inverse ? COLOR_INVERSE_BLUE : COLOR_BLUE;
-    colors[3] = inverse ? COLOR_INVERSE_GREEN : Color(0, 192, 0);
+    colors[3] = inverse ? COLOR_INVERSE_GREEN : COLOR_LIGHT_GREEN;
     colors[4] = inverse ? COLOR_INVERSE_YELLOW : COLOR_YELLOW;
     colors[5] = inverse ? COLOR_INVERSE_MAGENTA : COLOR_MAGENTA;
   } else


johnthomas22:
Thanks for you comments. I found the rest of the text OK, just the map
labels difficult to read.

Will pick this up later and, as you suggest, split the changes. (Can't do
this in daytime, or probably until tomorrow night in fact.)

Regards,

John

Rethink, Refuse, Reduce, Reuse, Repair, Regift, Recycle

johnthomas22:
> ## Code Review Feedback
> 
> Nice work — the core implementation is solid and aligns well with the project's architecture (centralizing safe-area handling in `TopWindow`, mirroring iOS). A few items to consider:
> ### 1. Unconditional `canvas.Clear(COLOR_BLACK)` on every paint
> 
> The `OnPaint` change in `MainWindow.cpp` clears the entire framebuffer on _every_ Android frame, even when `avoid_display_cutout` is `false` and there are no inset margins. Consider guarding it:
> 
> ```c++
> if (apply_display_cutout_insets.load(std::memory_order_relaxed))
>   canvas.Clear(COLOR_BLACK);
> ```
> 
> Or at minimum benchmark the unconditional clear on lower-end devices.
> ### 2. `GetSize()` hides the base class method
> 
> The new `GetSize()` in `TopWindow.hpp` is not `override` — it shadows `ContainerWindow::GetSize()`. Callers holding a `ContainerWindow*` or `Window*` will still get the un-insetted size. If this is intentional, it deserves a comment. If not, it's a bug.
> ### 3. Mutex on the render hot path
> 
> `GetClientRect()` acquires `paused_mutex` on every call when cutout avoidance is on. Since inset changes are rare (only on resize/rotation), a lighter approach could avoid lock contention — e.g. packing left/top/right/bottom as 16-bit values into a `std::atomic<uint64_t>`, removing the mutex from the read path entirely.
> ### 4. `const` on by-value returns
> 
> `GetClientRect()` returns `const PixelRect` and `GetSize()` returns `const PixelSize`. The `const` qualifier on by-value returns is meaningless (caller gets a copy) and can inhibit move semantics. Should be plain `PixelRect` / `PixelSize`.
> ### 5. Setting visibility when not in fullscreen
> 
> The "Avoid display cutout" checkbox is always visible even when fullscreen mode is off, but the setting only applies in fullscreen. Consider graying it out when fullscreen is disabled, or adding a note to the help text.
> ### 6. Minor: whitespace-only changes
> 
> A few whitespace-only changes (trailing spaces in `XCSoar.java`, reformatted `if` in `MainWindow.cpp`) are mixed in. Separating these into their own commit keeps the diff focused.

Thanks for your comments. Here's our responses:

# | Issue | Action | Rationale
-- | -- | -- | --
1 | Unconditional canvas.Clear() | No change | Single glClear per frame — negligible cost. Making it conditional adds coupling to TopWindow internals for no practical gain.
2 | GetSize() hides base class | Added comment | By design — matches the existing Windows (line 344) and iOS (line 355) pattern. Added the same clarifying comment as the Windows version.
3 | Mutex on render hot path | No change | The atomic check gates the mutex, and the critical section is 4 integer copies. Lock contention is effectively zero. Packing into atomic<uint64_t> adds complexity for no measurable benefit.
4 | const on by-value returns | No change | Matches existing codebase style — all three platform overrides (Windows, iOS, Android) use const PixelRect/const PixelSize. A style cleanup should be a separate PR.
5 | Setting visibility when not fullscreen | No change needed | The help text already says "in full screen mode".
6 | Whitespace-only changes | None found | Verified — no trailing whitespace or reformatting-only changes in the diff.

I hope that is all acceptable to you.

johnthomas22:
Can bring it back if you want, but for me, Full Screen Off under
Config->System->Look ->Screen Layout does pretty much the same thing?


Regards,

John

Rethink, Refuse, Reduce, Reuse, Repair, Regift, Recycle

PR #2202, opened 2026-02-16
Remove unicode step3

August2111:
# Last Step(s) Removal Unicode

This addresses the missing UNICODE and modifies the associated conversion macros (e.g., `cahr` instead of `TCHAR`, removal of `_T()`) and functions (e.g., `_stprintf(...)`, `_putts(...)`, etc.). All unnecessary files and header includes are also removed. Step by step, the code is cleaned of these superfluous and bloated elements.

This step is carried out in several independent sub-steps:

1.) Fixing three (minor) issues that arise from or hinder the changes.

2.) Removal of the conversion files (WString..., ..) that are no longer needed after Step 2 – these will no longer be used after removing the UNICODE macro!

3.) Remove all TCHAR macros from XCSoar, replace them with the simple 'char'.

4.) The documentation has also been changed to reflect the removal of TCHAR, as well as the removal of the corresponding labels (e.g., PRODUCT_NAME_T) which are no longer needed.

5.) Removal of the no longer necessary functions that refer to a definition of TCHAR (e.g. _stprintf(..), _putts(..),_tfopen(..) ) and replacement with the corresponding 'char' string functions.

6.) Removal of the conversion macro '_T(..)', which is no longer needed yet.

7.) This also allows the include file 'tchar.h' to be removed from XCSoar. Afterwards, any incorrect use of TCHAR functions and macros will be displayed with an error message!

8.) Removal of unnecessary NarrowString and NarrowWeatherName

9.) Finally, the removal of the various conversion functions between UTF8 and WideString, which only unnecessarily bloat the code!

It is very important that this pull request is implemented as soon as possible, because every other change to the code must be checked against this removal of TCHAR, _T(), etc. - otherwise, every rebase will require a lot of extra work.


lordfolken:
merged.

PR #2230, opened 2026-02-20
Build System: Add -MP flags to both compilers

groundhogxc:
I stumbled upon this one as well when I looked into the -Wp thing. This should get rid of the "missing header file needed by xxx"-errors of the make command when a file is removed from the tree (due to actual removal or a checkout / branch switch)

This should get rid of 95% of all the "make clean"s I need to do and save tons of time.

However, I am a bit unsure whether there might be some problem with the flag, and this is why it was not introduced years ago? I mean, if this really works as advertised, this would have removed so much pain - there must be some reason for not putting it?

## Summary by CodeRabbit

* **Chores**
  * Optimized build system dependency handling to improve build efficiency and reduce redundant compilation steps.

### DIFF
--- a/build/compile.mk
+++ b/build/compile.mk
@@ -64,7 +64,7 @@ SRC_TO_OBJ = $(subst /./,/,$(patsubst %.cpp,%$(OBJ_SUFFIX),$(patsubst %.cxx,%$(O
 ####### dependency handling
 
 DEPFILE = $(@:$(OBJ_SUFFIX)=.d)
-DEPFLAGS = -MD -MF $(DEPFILE) -MT $@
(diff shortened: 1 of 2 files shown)


lordfolken:
I'm a bit confused. 90% of the time i do not need to clean things out. I can just recompile. Changes if i remove a file or similar, then i do rm -rf output/UNIX

with ccache compiling is fast:

`make -j$(nproc) TARGET=UNIX USE_CCACHE=y WERROR=y`

lordfolken (approved):

lordfolken:
The flag was added by @jwharington in 2008. :-)

groundhogxc:
Oh, historic! :)

And it was later removed?

Issue #2259, opened 2026-03-01
Terrain Rendering: Resolution and Orientation issue


groundhogxc:
While working on the terrain render cod for #2234, I came upon a pretty significant issue (or pair of issues) that I think need discussion. 

On OpenGL, the terrain is rendered beyond the visible screen, to an area that is approximately 1.5 times as wide and high as the screen. This was introduced in commit 469b8e277a3f56120d8fe6ca5cb70f127f2fe985, already 13 years ago now.

According to the PR text, the goal of that commit was to 
1) allow reusing the computed texture when moving (especially useful for panning), and 
2) leverage the GPU for rotation of the bitmaps, so the terrain would only be rendered once at a fixed angle (north up), and then rotated to the correct orientation for e.g. track-up mode

So for this, the currently visible area is multiplied in length/width by 1.5 to compute the area that is then rendered as a terrain bitmap, which is then drawn to the screen as a texture.

Now, there are two significant issues with this change:
1) The factor 1.5 is applied to the geographic bounds rendered, but not to the dimensions of the bitmap. Therefore, the effective resolution is now worse - one terrain pixel is 1.5x1.5 times larger than before this PR (and therefore also 1.5x larger than on non-GL builds)
2) This PR changed the standard "fixed" illumination of the terrain effectively from "fixed" (illumination from top left of the screen regardless of orientation) to "northwest" (sun always in the northwest). 

Both are is still the case today. I am very surprised that the last point did not immediately create issues with people using track-up / heading-up / task-up mode. I tried it for myself with the simulator, it would drive me absolutely crazy trying to read the map if the illumination is not fixed relative to the screen. I just can't do it, the mountains always "invert" in my brain when I fly south in track-up mode.

It is just that I never noticed because I always fly in north-up map orientation, where this does not apply.

Of course, both can be changed (by rotating the illumination with the map orientation angle, and by scaling up the rendered bitmap size with the same factor). I am testing this, it works. 

But since it has been like this for such a long time I wonder if this would create issues - particularly, the number of rendered pixels is twice as large, and the terrain actually has to be re-rendered on every update (like what the non-GL path seems to do), so this is significantly more load on the system, particularly for the track-up crowd.

Incidentally, this also finally explains why the terrain rendering is different on OpenGL / non-GL, even though it is the same code. This has always puzzled me, but it is effectively because all slopes were exaggerated by a factor 1.5 on OpenGL vs. non-GL.

[...]

PR #2266, opened 2026-03-02
Terrain/RasterRenderer: Keep OpenGL terrain sampling density

lordfolken:
## Summary
- keep the OpenGL off-screen terrain bounds expansion (`1.5x`) for rotation/panning cache behavior
- scale the sampled terrain bitmap dimensions by the same factor, so OpenGL texel density matches non-OpenGL builds
- preserve existing shading/orientation behavior while removing the GL-vs-non-GL resolution mismatch

Related issue: #2259

## Test plan
- [x] `make -j$(nproc) TARGET=UNIX USE_CCACHE=y output/UNIX/dbg/src/Terrain/RasterRenderer.o output/UNIX/dbg/src/Terrain/TerrainRenderer.o`
- [ ] Run full `make -j$(nproc) TARGET=UNIX USE_CCACHE=y check` (currently blocked in this environment by disk space)

## Summary by CodeRabbit

* **Refactor**
  * Optimized terrain rendering calculations for improved performance and consistency.

Related to #2259


groundhogxc:
@lordfolken I am a bit confused about your formula here - particularly the +1, and why it breaks up the fraction. How did you arrive at that?

PR #2270, opened 2026-03-03
Android/USB: Use package name for USB permission action

groundhogxc:
This is to avoid a TESTING package and a regular package installed on the same Android device to register their permissions with the same package name.

I admit that I do not really know how Android permissions work, but seems like the right thing to do to avoid this?
Or is it better to have the permissions under a common package name?

## Summary by CodeRabbit

* **Refactor**
  * Improved USB permission handling implementation with enhanced instance-level control.

### DIFF
--- a/android/src/UsbSerialHelper.java
+++ b/android/src/UsbSerialHelper.java
@@ -29,7 +29,7 @@ import com.felhr.usbserial.UsbSerialDevice;
 public final class UsbSerialHelper extends BroadcastReceiver {
 
   private static final String TAG = "UsbSerialHelper";
-  private static final String ACTION_USB_PERMISSION = "org.xcsoar.otg.action.USB_PERMISSION";
+  private final String ACTION_USB_PERMISSION;
 
   private final Context context;
   private final UsbManager usbmanager;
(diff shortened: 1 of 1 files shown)


android/src/UsbSerialHelper.java
```diff
@@ -29,7 +29,7 @@
 public final class UsbSerialHelper extends BroadcastReceiver {
 
   private static final String TAG = "UsbSerialHelper";
-  private static final String ACTION_USB_PERMISSION = "org.xcsoar.otg.action.USB_PERMISSION";
```
lordfolken, line 32: Is it enough to remove the name here, or do we need to paramterize it on package name?

lordfolken: @coderabbitai can you investigate?

lordfolken: @groundhogxc can you add the setpackage name?

groundhogxc: > @groundhogxc can you add the setpackage name?

I can do that, yes.

groundhogxc: @coderabbitai Why is the SDK version conditional necessary in your proposed code? Is it necessary?

groundhogxc: OK, we keep it for now.

groundhogxc: > @groundhogxc can you add the setpackage name?

OK, @lordfolken this is done now in #2275. Also see #2276 that ended up being triggered by this.

PR #2346, opened 2026-03-29
NOTAM support: fetch, map, settings, and airspace UI

lordfolken:
## Summary
Adds NOTAM fetching, filtering, map/airspace integration, and configuration UI, with follow-up fixes for UTF-8 text paths, Windows GDI list rendering, and airspace warning acknowledgement when NOTAM geometry is refreshed.

## Commits
- `src/NOTAM,build/libnotam.mk`: Implement fetching, filtering and displaying of NOTAMs (includes UTF-8 `char` cleanup for NOTAM UI paths and NTOAM→NOTAM filter label fix)
- `src/Dialogs/dlgAirspaceWarnings`: Add Details button
- `Renderer/AirspacePreviewRenderer`: Restore list text color after GDI preview fill (dark mode)
- `Engine/Airspace/AirspaceWarningManager`: Preserve NOTAM day-ack across refresh (stable key by NOTAM number)

## Testing
- `make TARGET=UNIX` and `TARGET=WIN64` built successfully locally.

Supersedes / relates to prior NOTAM PR discussion as applicable.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added NOTAM (Notice to Airmen) support with dedicated configuration panel and list display dialog.
  * Implemented NOTAM filtering by radius, effective time, IFR status, and Q-codes.
  * Enabled NOTAM caching and automatic refresh with configurable intervals.
  * Integrated NOTAMs into airspace display and warning system.

* **Bug Fixes**
  * Fixed typo in NOTAM airspace label display.
  * Fixed text color preservation during airspace rendering.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -169,6 +169,7 @@ include $(topdir)/build/libio.mk
 include $(topdir)/build/shapelib.mk
 include $(topdir)/build/libwaypoint.mk
 include $(topdir)/build/libairspace.mk
+include $(topdir)/build/libnotam.mk
 include $(topdir)/build/libtask.mk
 include $(topdir)/build/libxml.mk
 include $(topdir)/build/libcupfile.mk
--- a/build/libnotam.mk
+++ b/build/libnotam.mk
@@ -1,0 +1,11 @@
+# Build rules for the NOTAM library
+
+LIBNOTAM_DEPENDS = LIBHTTP LIBBOOST LIBJSON LIBCOROUTINES LIBFMT
+
+LIBNOTAM_SOURCES = \
+	$(SRC)/NOTAM/Client.cpp \
+	$(SRC)/NOTAM/Converter.cpp \
+	$(SRC)/NOTAM/Filter.cpp \
+	$(SRC)/NOTAM/NOTAMGlue.cpp
+
+$(eval $(call link-library,libnotam,LIBNOTAM))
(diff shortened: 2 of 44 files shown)


yorickreum:
> Hello, just a remark on this PR for the source of the NOTAM: I saw in the description of PR #2018 that the source of the NOTAM will be an FAA API.
> 
> I think that for Europe, the authoritative API should be Autorouter, a subsidiary of Eurocontrol EAD which is in charge of Europe sky: https://www.autorouter.aero/wiki/api/notams/ / Eurocontrol: https://en.wikipedia.org/wiki/European_AIS_Database
> 
> I don't know if it will be easy to access the API of Autorouter, but I think it should be tried. So in this PR, the possibility to choose your source for NOTAM should be made, or prepared as future extension?
> 
> Thanks for the work 👍

Hey @B-apt! I totally agree with you that it would be much better to have a Eurocontrol source instead of the FAA, at least for European airspace. I actually reached out to Eurocontrol EAD last summer, and they basically told me that direct access to their API unfortunately comes with prohibitive fees:

> Although you work for a non-commercial open-source project, the EAD service is not a free-cost service. As per your email, we are assuming that you are not going to commercialise nor resell the EAD data, therefore we can consider you as a Type 2 client, which has a minimum cost of 3.600€/year. Besides, all the EAD clients are required to have a third-party liability insurance as per regulation.  

A third party like Autorouter sounds interesting to me. Autorouter itself, though, seems to allow querying only by airport identifier, not by arbitrary location. I’ll take a more thorough look at this and other options, and if it turns out to be feasible, I’d be more than happy to update the NOTAM proxy: https://github.com/yorickreum/xcsoar-notam-proxy

lordfolken:
Superceeded by #2018

PR #2420, opened 2026-04-22
Touch text entry: cursor-key navigation and resize

### LINKED ISSUE #1819: The ability to navigate the virtual keyboard with the arrow keys to search for waypoints and save task
XCSoar version 7.43
--------------

The virtual keyboard allows you to navigate using just the mouse arrow or a touchscreen.

It would be convenient, functional, and fast to be able to navigate using the up, down, and left/right arrow keys, as in the F1 menu, so you can write in the box the first letter of the wpt or save task name

Image

lordfolken:
Fixes #1819

## Summary
- `WndForm`: optional client layout callback after `OnResize` so dialogs can relayout children when the window resizes.
- `TouchTextEntry`: shared layout computation; repositions editor, action row, on-screen backspace, and on-screen keyboard on client resize; keyboard navigation is delegated to `KeyboardWidget::KeyPress`.
- `KeyboardWidget`: hardware cursor key routing between on-screen keys (OK ↔ Space, letter/number rows, on-screen backspace) with geometry-based movement; `Up` from Space skips the shift key so the path follows QWERTY rows.

**Label:** needs-work (WIP / feedback welcome).

## Notes
Unstaged `po/*.po` / `xcsoar.pot` in the working tree were left out of this branch; no new translatable user strings in these commits.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced cursor-key navigation for on-screen keyboards with improved focus routing
  * Better dialog layout responsiveness when window is resized
  * Refined keyboard navigation logic for text entry dialogs

### DIFF
--- a/src/Dialogs/TouchTextEntry.cpp
+++ b/src/Dialogs/TouchTextEntry.cpp
@@ -7,17 +7,98 @@
 #include "Form/Button.hpp"
 #include "Form/Edit.hpp"
 #include "Widget/KeyboardWidget.hpp"
+#include "Asset.hpp"
 #include "Screen/Layout.hpp"
 #include "ui/event/KeyCode.hpp"
 #include "UIGlobals.hpp"
 #include "Language/Language.hpp"
 #include "util/StringCompare.hxx"
 #include "util/TruncateString.hpp"
+#include "ui/window/Window.hpp"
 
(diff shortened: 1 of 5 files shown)


lordfolken:
superceeded by #2423

PR #2427, opened 2026-04-25
Add LX Navigation LX160 device driver

### LINKED ISSUE #2424: Feature request: dedicated LX160 device driver (bidirectional: read LXWPx, write GPGGA/GPRMC/GPRMB)
## Motivation

The LX Navigation LX160 is a legacy but still widely flown vario/final-glide computer in club fleets. It has no internal GPS and requires an external source feeding it `$GPGGA` + `$GPRMC` on its NMEA IN line, plus optional `$GPRMB` for waypoint bearing/distance/arrival. It outputs its own telemetry on NMEA OUT as `$LXWP0`/`$LXWP1`/`$LXWP2`/`$LXWP3` (vario, TAS, pressure altitude, MC, ballast, bugs, polar, SC/vario mode, filter constants).

Today XCSoar has no single driver that covers both directions on one port, which is the only port the LX160 exposes:

- The **Cambridge L-Nav** (`cai_lnav`) driver emits `$GPRMC` + `$GPRMB` + `$PCAIB`. The GPRMB is fine for the LX160 (fields 10/11/13 are what the LX160 reads), but fields 5–9 (waypoint name, destination lat/lon) are left empty, so the LX160 cannot display the waypoint name. No `$GPGGA` is emitted, and `$PCAIB` is meaningless to the LX160. It also does not parse LXWP.
- The **LX** driver parses LXWP, but does not emit `$GPRMC`/`$GPRMB`/`$GPGGA` — it targets the modern LXNAV vario family (LX9000/Eos/Era etc.) and uses a very different command set.

Result: with the LX160 you have to choose one direction or the other, which defeats the point of having a bidirectional NMEA port.

## Proposal

Add a dedicated `LX160` device driver that combines both directions on a single port:

### Read side (LX160 → XCSoar)
- Parse `$LXWP0` (TAS, altitude, vario/netto/avg)
- Parse `$LXWP1` (device name, serial, sw/hw version)
- Parse `$LXWP2` (MacCready, load factor / ballast, bugs, polar a/b/c, volume)
- Parse `$LXWP3` (alt offset, SC mode, filter constants, etc.)

The existing `LX` driver already contains parsers for these sentences (`src/Device/Driver/LX/LXWP*.cpp` or similar). Those can be reused / factored out.

### Write side (XCSoar → LX160)
- On `OnCalculatedUpdate()` emit `$GPGGA` + `$GPRMC` + `$GPRMB` at ~1 Hz:
  - `FormatGPGGA()` — already in `src/Formatter/NMEAFormatter.hpp` (prior art: `AirControlDisplay.cpp:252`)
  - `FormatGPRMC()` — already used by `cai_lnav`
  - `FormatGPRMB()` — already in `cai_lnav`, but ideally extended (see below) to populate fields 5–9

### GPRMB field enrichment
The current `FormatGPRMB()` in `cai_lnav` (`src/Device/Driver/CaiLNav.cpp:37-54`) populates only fields 1 (validity), 10 (range), 11 (bearing), 13 (arrival). For the LX160 the missing waypoint name (field 5) and destination lat/lon (fields 6–9) would let the LX160 show the active turnpoint name on its display. Small enhancement; could be a separate PR if preferred.

### Configuration
One device option: **"Emit position sentences (GPGGA/GPRMC)"**, default ON.
- Default ON covers the common case: LX160 paired with XCSoar running on a phone/tablet as the GPS source.
- Off for installations where another GPS source (e.g. a FLARM or dedicated GPS module) is already feeding the LX160's NMEA IN on the same bus — in that case XCSoar should emit only `$GPRMB` so it doesn't collide with the upstream GPS feed.

### Explicitly out of scope
- No task declaration upload — LX160 doesn't accept it over NMEA.
- No MC/ballast/bugs writeback — LX160's MC, ballast (BAL switch), and bugs are physical panel switches, not software-settable values. Read-only from XCSoar's side is correct.
- No polar sync — same reason.

## References

- LX160 user manual (v0212), ch. 9.13 "Enabling NMEA reception" and ch. 11 "Wiring extension for WinPilot": confirms LX160 NMEA input is "only GPGGA, GPRMB and GPRMC".
- LX160 manual ch. 8.13–8.16: confirms MC/BAL/BUGS are hardware switch settings; no NMEA input sentence accepts them.
- XCSoar source: `src/Device/Driver/CaiLNav.cpp`, `src/Device/Driver/LX/`, `src/Formatter/NMEAFormatter.{hpp,cpp}`, `src/Device/Driver/AirControlDisplay.cpp`.

## Estimated effort / willingness to contribute

Small — roughly one new file, ~150 lines of glue reusing existing formatters and LXWP parsers. Happy to submit a PR if the approach is acceptable. Would appreciate feedback on:

1. Preferred driver name in the device picker (`LX160`? `LX160 (LX Navigation)`? other convention?).
2. Whether the GPRMB field-5–9 enrichment should be in this PR or split out.
3. Whether the "emit position sentences" option belongs on the driver itself or as a device-config flag.

Thanks.

rikvdb229:
## Summary

New `LX160` device driver for the LX Navigation LX160 — a legacy vario / final-glide computer (no internal GPS) widely flown in club fleets.

Until now LX160 owners had to choose:

- the `cai_lnav` driver — writes `$GPRMC` / `$GPRMB` but doesn't parse `$LXWP`, or
- the `LX` driver — parses `$LXWP` but doesn't write position sentences.

Both leave one direction of the LX160's bidirectional NMEA port idle. This driver combines both: reuses the LXWP0/2/3 parsers from the LX driver (exposed via a new `LX/Parsers.hpp`) and emits `$GPGGA` + `$GPRMC` + `$GPRMB` on each calculation tick using shared `NMEAFormatter` helpers.

## Commits

1. **`Formatter/NMEAFormatter: Lift FormatGPRMB out of CaiLNav`** — moves the file-static helper to the shared formatter so multiple drivers can share it. Adds an optional `destination_name` parameter (default empty) for future field-5–9 enrichment. CaiLNav byte-for-byte unchanged.

2. **`Device/Config: Add SEND_POSITION capability and send_position toggle`** — new `DeviceRegister::SEND_POSITION` capability flag and matching `DeviceConfig::send_position` bool (default ON), wired through profile load/save and an "Emit GPGGA/GPRMC" expert checkbox in the Edit Device dialog. The row is hidden for drivers that don't advertise the flag, so existing drivers are unaffected.

3. **`Device/Driver/LX: Expose LXWP0/LXWP2/LXWP3 via Parsers.hpp`** — drops `static` from the three helpers in `LX/Parser.cpp` and adds a lightweight forward-declaration header. Sibling drivers can now reuse the LXWP parsing without including the heavy `Internal.hpp` surface that carries the full `LXDevice` declaration / logger / protocol state. No change to the LX driver itself.

4. **`Device/Driver/LX160: New driver for LX Navigation LX160`** — the driver itself.

## Capabilities

| Flag | Reason |
|---|---|
| `RECEIVE_SETTINGS` | LXWP2 carries MC, ballast, bugs read into XCSoar |
| `SEND_POSITION` | gates the new "Emit GPGGA/GPRMC" toggle |
| `PASS_THROUGH` | enables pairing with a second driver (typically FLARM) on the same transport via the existing "Use second device" UI |

`DECLARE`, `LOGGER`, `MANAGE`, `SEND_SETTINGS` are intentionally **not** advertised:

- The LX160 manual (v0212, ch. 9.13 + ch. 11) lists only `$GPGGA` / `$GPRMB` / `$GPRMC` as accepted input sentences — there is no NMEA path for task declaration or logger access.
- Bench-testing `$PFLX2` (the LX1600/MiniMap MC-write form) confirmed the LX160 silently ignores it. Ballast and bugs are physical front-panel switches anyway. So write-side settings sync would be no-ops; better to not advertise them.

## Echo handling

The LX160 echoes any `$GPGGA` / `$GPRMC` it receives back onto its NMEA OUT line (verified on bench). The driver handles both modes of operation:

- **`send_position = ON`** (XCSoar is the GPS source): swallow the echoes silently — they're our own data bouncing back, not a second GPS fix.
- **`send_position = OFF`** (external GPS feeds the LX160 directly): pass the incoming `$GPGGA` / `$GPRMC` through to XCSoar's generic parser — those echoes are then the only GPS path into XCSoar.

This makes one toggle do double duty: it gates both the outbound emission and the inbound swallow, with consistent semantics ("XCSoar provides position").

## Tested

Bench-tested against a real LX160 (firmware sw 2.10) at 4800 bps RS232:

- LXWP0 vario / TAS / pressure altitude → XCSoar info boxes track
- LXWP2 MC, ballast → XCSoar settings track the front-panel state
- LXWP2 bugs → confirmed **not** propagating on LX160 firmware sw 2.10. The existing `LX/Parser.cpp` decoder uses the documented `bugs = 2 − f3` rule for the LX160 quirk, but on sw 2.10 f3 stays hardcoded at 1.0 and the actual bugs value is encoded into the polar `a` coefficient (f4) instead. Not introduced by this PR — same behavior as the existing LX driver. Worth a follow-up to teach the parser to fall back to the polar-`a` shift on this firmware.
- LXWP1 device identity → populated in `info.device`
- `$GPGGA` + `$GPRMC` → LX160 displays position
- `$GPRMB` → LX160 displays distance / bearing / arrival flag for the active task leg
- `send_position = OFF` → echoed `$GPGGA` / `$GPRMC` reach XCSoar's generic parser, no double-source warnings

## Open questions for review

These were flagged in the linked feature-request doc and we'd appreciate maintainer guidance:

1. **Driver name** — chosen here as `LX160` / display `LX160` to mirror the device's marketing name. Other LX drivers use `LX` / display `LXNAV`. Happy to align.
2. **GPRMB field 5–9 enrichment** — the new `FormatGPRMB` already takes an optional `destination_name` parameter so a future patch can fill the active waypoint name + destination lat/lon without further formatter changes. We left the population for a follow-up because reaching the active waypoint name from `OnCalculatedUpdate` requires touching `ElementStat` / `TaskStats` (which currently carry only the location, not the name). Happy to roll that in if preferred, or split.
3. **PASS_THROUGH semantics** — we set this on the LX160 driver to enable the existing "Use second device" UI for muxed transports (LX160 + FLARM on one BLE/serial channel, mirroring the hardware-daisy-chain pattern). The driver doesn't override `EnablePassThrough()` since the muxing happens at the transport layer, not in the device. Let us know if you'd prefer a separate capability flag for the UI gate.

## Summary by CodeRabbit

* **New Features**
  * LX160 device driver is now available for selection.
  * New expert setting "Emit GPGGA/GPRMC" lets compatible drivers emit periodic GPS position output; visibility is gated to drivers that support it.
  * Device profiles now persist the "Emit GPGGA/GPRMC" per-port setting.

* **Improvements**
  * Improved waypoint/navigation sentence formatting (GPRMB) with better range/bearing and destination naming.
  * LX160 suppresses echoed device GPS sentences when position emission is enabled to avoid duplicate position sources.

### DIFF
--- a/build/driver.mk
+++ b/build/driver.mk
@@ -140,6 +140,7 @@ DRIVER_SOURCES = \
 	$(DRIVER_SRC_DIR)/AirControlDisplay.cpp \
 	$(DRIVER_SRC_DIR)/Larus.cpp \
 	$(DRIVER_SRC_DIR)/LoEFGREN.cpp \
+	$(DRIVER_SRC_DIR)/LX160.cpp \
 	$(DRIVER_SRC_DIR)/ATR833/Device.cpp \
 	$(DRIVER_SRC_DIR)/ATR833/Register.cpp
 
--- a/src/Device/Config.cpp
+++ b/src/Device/Config.cpp
@@ -182,6 +182,7 @@ DeviceConfig::Clear() noexcept
   enabled = true;
   sync_from_device = true;
   sync_to_device = true;
+  send_position = true;
   k6bt = false;
   polar_sync = PolarSync::OFF;
   engine_type = EngineType::NONE;
--- a/src/Device/Config.hpp
+++ b/src/Device/Config.hpp
@@ -278,6 +278,15 @@ struct DeviceConfig {
   bool sync_from_device;
 
   /**
+   * Should XCSoar send its current GPS position to the device as
+   * $GPGGA / $GPRMC sentences?  Only honored by drivers that advertise
+   * #DeviceRegister::SEND_POSITION (e.g. LX160).  When this is false the
+   * driver still emits navigation context such as $GPRMB.  Defaults to
+   * true; turn off when an upstream GPS source already feeds the device.
+   */
+  bool send_position;
+
+  /**
    * Polar synchronization direction (off, receive, or send).
    */
   PolarSync polar_sync;
--- a/src/Device/Driver.hpp
+++ b/src/Device/Driver.hpp
@@ -426,6 +426,13 @@ struct DeviceRegister {
      * EnablePassThrough() is implemented.
      */
     PASS_THROUGH = 0x200,
+
+    /**
+     * Does this driver emit GPGGA/GPRMC position sentences to the
+     * device?  When set, the user can suppress that emission via
+     * #DeviceConfig::send_position.
+     */
+    SEND_POSITION = 0x400,
   };
 
   /**
@@ -522,5 +529,12 @@ struct DeviceRegister {
    */
   bool HasPassThrough() const {
     return (flags & PASS_THROUGH) != 0;
+  }
+
+  /**
+   * Does this driver emit GPGGA/GPRMC position sentences?
+   */
+  bool CanSendPosition() const {
+    return (flags & SEND_POSITION) != 0;
   }
 };
--- a/src/Device/Driver/CaiLNav.cpp
+++ b/src/Device/Driver/CaiLNav.cpp
@@ -8,7 +8,7 @@
 #include "Operation/Operation.hpp"
 #include "NMEA/MoreData.hpp"
 #include "NMEA/Derived.hpp"
-#include "Units/Units.hpp"
+#include "Geo/GeoPoint.hpp"
 #include "Formatter/NMEAFormatter.hpp"
 #include "util/SpanCast.hxx"
 
@@ -20,37 +20,6 @@ PortWriteNMEANoChecksum(Port &port, const char *line,
   constexpr auto timeout = std::chrono::seconds(1);
 
   port.FullWrite(AsBytes(std::string_view{line}), env, timeout);
-}
-
-/*
-$GPRMB,<1>,,,,<5>,,,,,<10>,<11>,,<13>*hh<CR><LF>
-
-<1>  Position Valid (A = valid, V = invalid)
-<5>  Destination waypoint identifier, three digits
-     (leading zeros will be transmitted)
-<10> Range from present position to distination waypoint, format XXXX.X,
-     nautical miles (leading zeros will be transmitted)
-<11> Bearing from present position to destination waypoint, format XXX.X,
-     degrees true (leading zeros will be transmitted)
-<13> Arrival flag <A = arrival, V = not arrival)
-*/
-static bool
-FormatGPRMB(char *buffer, size_t buffer_size, const GeoPoint& here,
-            const AGeoPoint &destination)
-{
-  if (!here.IsValid() || !destination.IsValid())
-    return false;
-
-  const GeoVector vector(here, destination);
-  const bool has_arrived = vector.distance < 1000; // < 1km ?
-
-  snprintf(buffer, buffer_size, "GPRMB,%c,,,,,,,,,%06.1f,%04.1f,%c",
-           here.IsValid() ? 'A' : 'V',
-           (double)Units::ToUserUnit(vector.distance, Unit::NAUTICAL_MILES),
-           (double)vector.bearing.Degrees(),
-           has_arrived ? 'A' : 'V');
-
-  return true;
 }
 
 /*
--- a/src/Device/Driver/LX/Internal.hpp
+++ b/src/Device/Driver/LX/Internal.hpp
@@ -452,9 +452,6 @@ public:
 
   bool ReadLXGPSBaudrate(unsigned &baudrate, OperationEnvironment &env);
 
-  // These methods are reused by the LX Eos driver
-  static void LXWP1(NMEAInputLine &line, DeviceInfo &device);
-
 public:
   /* virtual methods from class Device */
   void LinkTimeout() override;
--- a/src/Device/Driver/LX/Parser.cpp
+++ b/src/Device/Driver/LX/Parser.cpp
@@ -2,6 +2,7 @@
 // Copyright The XCSoar Project
 
 #include "Internal.hpp"
+#include "Parsers.hpp"
 #include "NMEA/Checksum.hpp"
 #include "NMEA/InputLine.hpp"
 #include "NMEA/Info.hpp"
@@ -53,7 +54,9 @@ ShouldSwitchHostBaudForNinc(const DeviceInfo &device_info) noexcept
           device_info.hardware_version.equals("8"));
 }
 
-static bool
+namespace LX {
+
+bool
 LXWP0(NMEAInputLine &line, NMEAInfo &info)
 {
   /*
@@ -99,6 +102,8 @@ LXWP0(NMEAInputLine &line, NMEAInfo &info)
   return true;
 }
 
+} // namespace LX
+
 void
 LXDevice::LXWP1(NMEAInputLine &line, DeviceInfo &device)
 {
@@ -118,7 +123,9 @@ LXDevice::LXWP1(NMEAInputLine &line, DeviceInfo &device)
   device.license = line.ReadView();
 }
 
-static bool
+namespace LX {
+
+bool
 LXWP2(NMEAInputLine &line, NMEAInfo &info)
 {
   /*
@@ -165,7 +172,7 @@ LXWP2(NMEAInputLine &line, NMEAInfo &info)
   return true;
 }
 
-static bool
+bool
 LXWP3(NMEAInputLine &line, NMEAInfo &info)
 {
   /*
@@ -196,6 +203,8 @@ LXWP3(NMEAInputLine &line, NMEAInfo &info)
 
   return true;
 }
+
+} // namespace LX
 
 /**
  * Parse double from a string_view value field.
@@ -692,7 +701,7 @@ LXDevice::ParseNMEA(const char *String, NMEAInfo &info)
 
   const auto type = line.ReadView();
   if (type == "$LXWP0"sv)
-    return LXWP0(line, info);
+    return LX::LXWP0(line, info);
 
   if (type == "$LXWP1"sv) {
     DeviceInfo &device_info = mode == Mode::PASS_THROUGH
@@ -704,10 +713,10 @@ LXDevice::ParseNMEA(const char *String, NMEAInfo &info)
   }
 
   if (type == "$LXWP2"sv)
-    return LXWP2(line, info);
+    return LX::LXWP2(line, info);
 
   if (type == "$LXWP3"sv)
-    return LXWP3(line, info);
+    return LX::LXWP3(line, info);
 
   if (type == "$PLXV0"sv) {
     is_colibri = false;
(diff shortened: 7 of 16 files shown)


rikvdb229:
Thanks for the review. Pushed `a373f05` addressing the three findings on `Formatter/NMEAFormatter`:

1. **Magic arrival threshold** — hoisted `1000` into a named `constexpr ARRIVAL_RADIUS_M = 1000.0` (metres) at the top of `NMEAFormatter.cpp` and used it where `has_arrived` is computed. `Units::ToUserUnit`/`vector.distance` references unchanged.

2. **Field 12 vs field 13 for the arrival flag** — kept the wire format as-is and updated the header docstring instead. The placement in field 12 is deliberate: the existing `cai_lnav` driver has emitted the arrival flag in field 12 (with no closing-velocity field 12) for years, the LX160 (and other `cai_lnav` consumers) ignore everything past field 11, and the goal of `d83780d` was a pure byte-for-byte lift of `CaiLNav`'s helper. Inserting an extra empty field would silently break those consumers without any benefit. The `NMEAFormatter.hpp` sentence sketch and field list now match the actual emitted output and call out the spec deviation explicitly.

3. **UTF-8-safe truncation + comma sanitisation in `name_buf`** — after the byte-wise copy, the new code calls `CropIncompleteUTF8(name_buf)` (from `util/UTF8.hpp`) to drop any partial trailing multi-byte sequence, then walks the buffer replacing embedded `,` with `_` so a stray comma in a waypoint name cannot forge extra NMEA fields. (Note: today both in-tree callers — `LX160` and `CaiLNav` — pass an empty `destination_name`, so this path is dormant; it's exposed via the public API though, hence the hardening.)

Verified the file still compiles cleanly under `-Werror -Wall -Wextra` on the WIN64 target.

lordfolken (commented):
## Pass-through / `PASS_THROUGH` / second device

`DeviceRegister::PASS_THROUGH` in XCSoar is not only a label: it allows the “use second device” expert option and, when that is enabled, a second `CreateOnPort` and a second parser on the **same** `Port` (see `DeviceDescriptor::SetConfig` / `Open`). That is separate from a full Lxnvario `EnablePassThrough` / baud hand-off in the `LX` driver.

For a **standalone LX160** (one bidirectional NMEA port, manual I/O as described in the PR), is a two-driver-on-one-cable / mixed stream setup actually a supported real-world topology? If the vario is only ever a single NMEA end-point on that link, consider **dropping `PASS_THROUGH`** from `lx160_driver` so the UI does not offer a second device and the descriptor does not build a `second_device` the hardware cannot use.

Thanks for the thorough PR description and hardware testing.

lordfolken:
**Git / history:** For final merge, please avoid leaving a long chain of small `fixup!` / WIP-style commits; squash and clean the history per usual XCSoar practice (one logical change per commit, self-contained messages) so reviewers and `git log` stay readable.

lordfolken:
**Co-authored-by:** The commits still carry `Co-Authored-By: Claude ...` (and similar) in the message trailers. That reads as product advertising in the mainline history, which we try to avoid.

I’m happy for AI-assisted work in the diff; please **rebase or amend** so the final commit messages **do not** include those co-author lines (no change to the code required from this request—only the metadata).

lordfolken:
**CODEOWNERS:** `.github/CODEOWNERS` only has an extra rule for `src/Device/Driver/LX_EOS/*` under “LX Navigation Devices.” The new `LX160` driver (`src/Device/Driver/LX160.*`) and the shared `LX` parser header changes fall under the default `*` → `@XCSoar/code-reviewers` entry only, not that LX_EOS line.

If the maintainers want future LX driver / Lxnvario-touched PRs to ping the same team automatically, consider extending `CODEOWNERS` (e.g. add `src/Device/Driver/LX160*` and/or `src/Device/Driver/LX/*` to the appropriate team) in a follow-up on `master`—not necessarily blocking this PR.

rikvdb229:
Thanks @lordfolken — pushed `5ec38f3` addressing all four points.

**`PASS_THROUGH`**: dropped. You're right — for a standalone LX160 there's no real-world use for a second driver on the same NMEA cable, and exposing it would only invite users to wire something the hardware can't actually mux. My own bridge use case doesn't need it either (I expose two separate TCP endpoints upstream). Capability flags on `lx160_driver` are now `RECEIVE_SETTINGS | SEND_POSITION` only.

**Git history**: squashed from 5 commits down to 4 logical, self-contained changes (no `fixup!` / WIP commits). The previous review-feedback fix-up commit on `NMEAFormatter` is folded into the original "Lift FormatGPRMB" commit where it belongs:

  - `Formatter/NMEAFormatter: Lift FormatGPRMB out of CaiLNav`
  - `Device/Config: Add SEND_POSITION capability and send_position toggle`
  - `Device/Driver/LX: Expose LXWP0/LXWP2/LXWP3 via Parsers.hpp`
  - `Device/Driver/LX160: New driver for LX Navigation LX160`

**`Co-Authored-By`**: stripped from every commit message. Understood re: mainline history.

**CODEOWNERS**: noted, no action on this PR — happy to open a follow-up on `master` once this lands if maintainers want it.

While rebasing I also folded in the two remaining @coderabbitai nitpicks from the earlier pass:
  - `LXWP0/2/3` are now scoped under a new `LX::` namespace (in `LX/Parsers.hpp` + `LX/Parser.cpp`) instead of polluting the global namespace; call sites in `LXDevice::ParseNMEA` and `LX160Device::ParseNMEA` updated accordingly.
  - One-line comment on `LX160Device::send_position` documenting that it's sampled at port-open time and a reconnect is required to change it.

Local WIN64 build is clean (`-Werror -Wall -Wextra`).

lordfolken:
looks better now. Can you address the build failures on the other platforms?

rikvdb229:
Pushed a rebased history (`62c7972`):

- **Build failures (Android / macOS / iOS):** root cause was `class GeoPoint;` forward declaration in `NMEAFormatter.hpp` while `GeoPoint` is actually a `struct`; clang's `-Wmismatched-tags -Werror` rejected it. Fixed and verified locally with clang 18.
- **CodeRabbit nitpicks:** `LXWP1` hoisted to the shared `LX::` namespace (resolves the duplicate body in `LX160.cpp` and the two `namespace LX` blocks in `Parser.cpp` in one go). Help-text on "Emit GPGGA/GPRMC" now notes that the change takes effect on reconnect.
- **History:** each fix folded into its originating commit via `--autosquash`; branch stays at 4 self-contained commits, no AI co-author trailers.

rikvdb229:
Addressed CodeRabbit's `static constexpr` nit in `13dd907` (folded into the `FormatGPRMB` commit so history stays at 4 self-contained commits).

lordfolken (approved):

lordfolken:
Thank you.

PR #749, opened 2021-10-14
Dialogs/Waypoint: Return to waypoint list after closing waypoint details

HeroicLemon:
Brief summary of the changes
----------------------------
Adds a new waypoint list dialog that remains open after closing waypoint details.
New dialog is accessible through the input events. Add the following to a .xci file to override the defaults:
____________________________
mode=default
type=gesture
data=DR
event=WaypointDetailsPersistent

mode=Nav1
type=key
data=9
event=WaypointDetailsPersistent
event=Mode default
label=Waypoint\nList$(CheckWaypointFile)
location=8
____________________________
Related issues and discussions
------------------------------
Closes #620

PR #2429, opened 2026-04-26
Dialogs, Task/AlternatesList, Input: Waypoint details nesting and browse lists

### LINKED ISSUE #620: Back to waypoints list after watching waypoint detail
Hello,

XCSoar version
--------------

I'm using XCSoar 7.6

After seeing waypoints list to watch distance of several waypoints, I want to have more information about some specific waypoints (such as required glide ratio).
So I can choose one and look for data such as required glide ratio or Alt diff MC safety.

Unfortunately when clicking on "close" button... it's getting back to map page... not waypoint lists which is not very convenient.... especially when considering several waypoint options.

Similar problem also occurs when using Nav / Alternates window

Changing in this context "Close" button to a "Back" button... which turn back to waypoints list will be a nice feature to have

Waypoint details page should store the context of the "previous" page which was used to display waypoint detail.
If it's waypoints list or alternates, close button should be turned to a back button

lordfolken:
## Summary

- **Persistent waypoint search** (`ShowWaypointListPersistentDialog`): plumbs optional `WaypointDetailsNesting` so waypoint details can report a *committed* action; the outer list is dismissed on commit (e.g. GoTo, pan) and not when the user only closes details without changing task/navigation.
- **Waypoint details / commands** share the nesting contract in `dlgWaypointDetails` and `WaypointCommandsWidget`.
- **Alternates list** now matches the same *browse* pattern: details opens on top of the list (no separate modal result that finished the list first). On a committed state change from details, the list closes with `SetModalResult(mrOK)`.
- **Input:** documents `eventWaypointDetailsPersistent` as the alias for the persistent list entry (see `InputEvents` / defaults).
- **.cursor:** `search-dialog-ux.mdc` updated to describe this stack.

Supersedes the long-standing #749 (nested list + details flow); see comment there.

Closes #620.

## How to test

- Open the persistent waypoint list, open details, use Close only → you should return to the list. Use GoTo (or a flow that sets `state_change_committed`) → the list should close and the map should reflect the change.
- Open alternates (e.g. from the InfoBox), open **Details** on a row, close details without a commit → you should be back on the alternates list; with a commit → list should close as for the persistent list.

## Summary by CodeRabbit

* **New Features**
  * Waypoint search now stays open until Esc or a successful navigation action (GoTo, task, home, or pan); pan closes the search so the map isn’t obscured. Waypoint Details open without forcing the list to close and only dismiss the list when a details action commits a state change. A persistent waypoint search/list flow is now available via input bindings.

* **Documentation**
  * Updated v7.45 notes to document the new waypoint search lifecycle; minor v7.21 formatting fix.

### DIFF
--- a/.cursor/rules/search-dialog-ux.mdc
+++ b/.cursor/rules/search-dialog-ux.mdc
@@ -1,0 +1,73 @@
+---
+description: Stacked search dialogs (filter + list + optional details) — lifecycle and commit contract
+globs:
+  - "src/Dialogs/**/*.cpp"
+  - "src/Dialogs/**/*.hpp"
+  - "src/Input/InputEvents*.cpp"
+  - "src/Input/InputEvents*.hpp"
+alwaysApply: false
+---
+
+# Search dialog UX (general)
+
+Applies to **modal search UIs**: filter controls + result list, optionally opening a **second modal** (details, preview, editor) on top. Not every list dialog uses this pattern—see **Pick-only** vs **Browse** below.
+
+## Pick-only vs browse (persistent outer dialog)
+
+1. **Pick-only** — User chooses a row and the **outer** dialog closes immediately with a return value (or callback). No nested “return to list after closing child” loop.  
+   *Example:* `ShowWaypointListDialog` for Goto lookup or task relocate.
+
+2. **Browse** — Outer search **stays open** while the user opens and closes a **child** dialog on top. The user can inspect several rows without restarting the search.  
+   *Example:* `ShowWaypointListPersistentDialog` + waypoint details.
+
+Do not mix these casually: pick-only callers expect a single modal lifetime; browse mode requires an explicit **commit** signal to close the outer dialog.
+
+## When the outer search should dismiss
+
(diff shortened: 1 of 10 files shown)


lordfolken:
**CI: Android (and other) red runs — checkout, not the NDK build**

Run [24953367022](https://github.com/XCSoar/XCSoar/actions/runs/24953367022) failed the **Android** matrix leg (and other legs) with:

`fatal: couldn't find remote ref refs/pull/2429/merge`

So the job never reached `gmake TARGET=ANDROID` (or the Docker build step for that target). This is the same class of issue as a missing/stale `refs/pull/<n>/merge` ref on GitHub (e.g. timing around merge/close/branch updates), not a Java/Kotlin/NDK or `android.mk` failure.

Earlier runs on the same branch (e.g. [24953332775](https://github.com/XCSoar/XCSoar/actions/runs/24953332775)) completed successfully, which is consistent with a transient or race on the merge ref rather than a regression in the PR’s Android sources.

*Investigation only — no code or workflow change from this note.*

PR #2431, opened 2026-04-26
Device/Driver/LX_EOS: Enable Declare and Logger functionality

### LINKED ISSUE #2433: Unable to declare to Lx Eos
Declaration and flight download is not enabled for the Lx Eos driver, even though it used to work in previous XCSoar versions.

7.44

#2431

#2431

jirimaier:
The Lx Eos / Era driver is missing the `DeviceRegister::DECLARE | DeviceRegister::LOGGER` flags, even though the declare and download functionality is implemented and functional. This fixes it.


lordfolken (approved):
Thanks for the maintenance.

PR #2432, opened 2026-04-26
Navigator widget (rebased; supersedes #1064)

lordfolken:
Continues the work from #1064 (original author: @threaderic). This branch is **rebased** onto current `master` and split/cleaned so the history is buildable and follows XCSoar conventions (see discussion in #1064).

**Supersedes #1064** because the previous head branch on `threaderic/XCSoar:NavigationWidget` is protected in a way that blocks force-pushes (merge commits in `master` history), so the updated history is published from this fork.

### Summary
- Navigator strip / `NavigatorRenderer`, `NavigatorWidget`, page layout, settings, input, docs, etc. as in the original PR scope.
- Rebased and commit-order fixes for per-step `make` runs; include/layering cleanups.

### Please test
- Flight or sim: navigator strip, top/bottom panes, settings, Nav input modes, task / waypoint display.

Replaces: #1064

Related to #1064

### DIFF
--- a/Data/Input/default.xci
+++ b/Data/Input/default.xci
@@ -530,6 +530,7 @@ event=ArmAdvance toggle
 
 # -------------
 # mode=Nav1
+# (Status / Analysis / Target on keys 2--4.)
 # -------------
 
 mode=Nav1
@@ -538,6 +539,30 @@ data=APP1
 event=Mode Nav2
 label=Nav\nPage 2/2
 location=1
+
+mode=Nav1
+type=key
+data=2
+event=Status all
+event=Mode default
+label=Status
+location=2
+
+mode=Nav1
+type=key
+data=3
+event=Analysis
+event=Mode default
+label=Analysis
+location=3
+
+mode=Nav1
+type=key
+data=4
+event=Setup Target
+event=Mode default
+label=Target Show$(CheckTask)$(CheckTaskResumed)
+location=4
 
 mode=Nav1
 type=key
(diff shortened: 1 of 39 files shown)


joecapra:
Really looking forward this this feature!

efoertsch:
Is this feature no longer going to be added to XCSoar? It would be a very nice option to have.

lordfolken:
Hmm this was an accidental close. I'm happy though if anyone wants to take it over. Especially testing is time consuming.

threaderic:
i'm going to test it within next weeks!

DanD222:
I'm thinking that it could also be really useful to display the wind arrow in the widget (and also the Instant wind if it's available, once https://github.com/XCSoar/XCSoar/pull/2954 is merged)?

threaderic:
> I'm thinking that it could also be really useful to display the wind arrow in the widget (and also the Instant wind if it's available, once #2954 is merged)?

I think it's a good idea to keep in mind. It has to be optional cause it could be difficult to  read --> adding another arrow above the current ring + arrow could be challenging to read!

image

My first idea is to  test the current PR to make a release possible within next weeks / months. After that I will add  this feature.

DanD222:
I was thinking it could be next to the navigation arrow, but:
- I might as well use an infobox for that
- Probably best to not complicate things further at this time

So feel free to disregard at this point :)

WolfgangDrescher:
Have a look at https://github.com/XCSoar/XCSoar/issues/3034. With the `Merge` InfoBox proposed there, you could actually reserve a whole row (or at least 75%) of the InfoBoxes for such a feature. This could be a pretty good solution for portrait devices. In landscape, though, we would probably also need to allow a vertical arrangement of the Navigator widget's contents.

Just as input and food for thought.

PR #2435, opened 2026-04-26
FLARM: send active VHF to device via PFLAM

lordfolken:
When a FLARM port has **Sync to device** enabled, the active COM frequency is emitted as NMEA `$PFLAM,S,VHF,<MHz>,,,` per FTD-109 (single mandatory channel; optional B/C/D empty). `ExchangeRadioFrequencies` sends the new active after swap.

- New `Pflam.cpp` with `PutActiveFrequency` / `ExchangeRadioFrequencies`
- `EnableNMEA` before send so the link is not left in binary mode
- `NEWS.txt` under 7.45 *devices*

Related to #1725

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -85,6 +85,9 @@ Version 7.45 - not yet released
   - openair: map AY ASRA to aerial sporting/recreational airspace type #1827
 * devices
   - fix native crash on device reconnect (stale delayed reopen timer) #2382
+  - FLARM: when "Sync to device" is enabled, the active VHF frequency is sent to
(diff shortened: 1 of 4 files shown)


lordfolken:
cc @Scumi for information — you authored the incoming `$PFLAM` handling in `StaticParser` ([commit bff3fb9](https://github.com/XCSoar/XCSoar/commit/bff3fb9165191b6392523d0886f2f538bceaf3a8)). This PR adds the complementary path that **sends** the active VHF to the device via `$PFLAM,S,VHF,…` when “Sync to device” is enabled. No response expected; just FYI in case you want to weigh in on messaging / FTD-109 behaviour.

PR #2445, opened 2026-04-29
doc: Improve install script for macOS

### LINKED ISSUE #2442: dither missing compiling for Mac?
I'm trying to compile XCSoar for Arm64 Mac following instructions here:
https://xcsoar.readthedocs.io/en/latest/build.html

But I don't have the dither tool, and can't find the correct one. 
I did find this: https://crates.io/crates/dither but this doesn't have the -t option, so not sure it's the correct tool?

Output:
`
...
  CXX     output/MACOS/dbg/src/ui/event/shared/TimerQueue.o
  CXX     output/MACOS/dbg/src/ui/event/sdl/Loop.o
  CXX     output/MACOS/dbg/src/ui/event/sdl/Queue.o
  AR      output/MACOS/dbg/libevent.a
make: dither: No such file or directory
make: *** [output/data/bitmaps/aboveterrain.png] Error 1
`

And when the found rust dither tool is installed it fails with:
`
error: Found argument '-t' which wasn't expected, or isn't valid in this context

USAGE:
    dither [FLAGS] [OPTIONS] <input> [output]
`

WolfgangDrescher:
Closes https://github.com/XCSoar/XCSoar/issues/2442.

## Summary by CodeRabbit

* **Documentation**
  * Clarified macOS build provisioning requirements to specify that BASE packages must be installed alongside target-specific packages.
  * Added explicit guidance clarifying that the BASE provisioning step is a mandatory prerequisite for both iOS and macOS build configurations.

### DIFF
--- a/doc/build.rst
+++ b/doc/build.rst
@@ -297,7 +297,9 @@ Compiling for macOS (with Homebrew)
 
 Install the required Homebrew packages via the provisioning script::
 
-  ./ide/provisioning/install-darwin-packages.sh MACOS
+  ./ide/provisioning/install-darwin-packages.sh BASE MACOS
+
+Note that you always need to install the BASE packages and the specific IOS or MACOS packages.
 
 To compile for macOS / ARM64, run::
 


lordfolken (approved):
Thanks

PR #2447, opened 2026-04-30
Renderer/TrailRenderer: Smooth snail trail and merge-rate vario colour

lordfolken:
## Summary

Improve the map snail trail with **Catmull–Rom smoothing** in screen space and **high-rate netto vario** for colour between GPS trace points.

### 1. Catmull–Rom spline (trail geometry + linear scalar blend)
- Smooth segments between stored trace points; zoom-dependent base segment count (6–12).
- Sharper bends increase subdivision (up to 20) using turn angle at each vertex (`Math/Point2D` `DotProduct` / `Magnitude`).
- Vario/altitude colour along the spline uses linear interpolation between trace endpoints (altitude mode unchanged logically).
- Existing trail modes (dots-only, dots+lines, etc.) preserved.

### 2. Merge-rate netto for vario trail colour
- `MergeThread` records netto vario while flying (after releasing `DeviceBlackboard` mutex) into `TraceComputer` ring buffer.
- `LockedCopySnapshot` copies filtered trace + samples under one lock for rendering.
- Piecewise-linear netto in **time** between consecutive trace timestamps when drawing smoothed **vario** trails (`BuildVarioKnots` / `LookupVarioFromKnots`).
- `CopyMergeVarioSamplesUnlocked` deduplicates snapshot fill; `GlideComputer` / `TaskComputer` expose non-const `GetTraceComputer()` for `Protection` wiring.

## Testing
- `make TARGET=UNIX` (local).

## Commits
1. `Renderer/TrailRenderer: Smooth snail trail with Catmull-Rom spline`
2. `Renderer/TrailRenderer: Colour trail from merge-rate netto`

## Summary by CodeRabbit

* **New Features**
  * High-rate vertical-motion sampling added to provide continuous vario-driven coloring between GPS trace points for smoother climb/sink visualization.
  * Trail rendering rewritten to batch points, use adaptive spline interpolation and per-segment vario interpolation for higher-quality, more accurate trail visuals and consistent current-position coloring.
* **Refinements**
  * Dot/line visibility and midpoint dot placement improved for clearer, more stable trail display.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -14,6 +14,8 @@ Version 7.45 - not yet released
   - obstacles hidden at wider map scales than other non-landables (5000 vs 10000)
   - draw up to 1024 waypoints at once (was 256) #2327
   - terrain: fix fluctuating hill-shading strength #2262
+  - snail trail: Catmull-Rom smoothing between fixes; vario colouring samples
+    high-rate netto from the merge thread between GPS points #2447
 * ui
   - infoboxen: refresh titles after changing the interface language #2314
   - infoboxen: add "Home" InfoBox (waypoint name, arrival height at home,
--- a/src/Computer/GlideComputer.hpp
+++ b/src/Computer/GlideComputer.hpp
@@ -121,6 +121,10 @@ public:
     return task_computer.GetTraceComputer();
   }
 
+  TraceComputer &GetTraceComputer() noexcept {
+    return task_computer.GetTraceComputer();
+  }
+
   const ProtectedTaskManager &GetProtectedTaskManager() const {
     return task_computer.GetProtectedTaskManager();
   }
--- a/src/Computer/TaskComputer.hpp
+++ b/src/Computer/TaskComputer.hpp
@@ -52,6 +52,10 @@ public:
     return trace;
   }
 
+  TraceComputer &GetTraceComputer() noexcept {
+    return trace;
+  }
+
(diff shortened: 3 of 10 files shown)


groundhogxc:
I did not find an issue related tot his, so asking here: What does this do with the "thinned out" part snail trail that is created during a long flight? XCSoar reduces the point density dynamically after some time, leading to points which are quite far apart on long flights.

Towards the end of a long flight, when trying to retrace a path flown, say 20-30 Minutes earlier, the trail is often already quite strongly reduced in density and the line therefore inaccurate, and, what is worse, it can be difficult to see what is "real" data and what is the line between data points. The only way to see what the actual data points are is to find the corners in the line, or, where the color changes (but that works only if the lift changed on that point). I am afraid that is no longer possible at all with spline interpolation, so then it becomes impossible to guess where the actual trajectory was. So, for now I disabled this in my own device, but I am generally a sceptic when it comes to interpolation of data. 

Not sure what a good solution is (beyond not interpolating). We could plot additional markers where the anchor points of the spline are, when plotting a thinned-out snail trail? 

Of course, a far preferred solution, with or without interpolation, would be: Not to thin out the trail any more (or much less) when for zoomed in display, and only use a thinned version when zoomed out? This of course requires changing how the data is stored  so much ore involved than a display change (store multiple detail levels of the path, plus a KD Tree for retrieval or something along those lines).

But maybe there is something less invasive and more elegant.

Issue #933, opened 2022-05-24
replay +10' not working


fronzee:
XCSoar versions having and not having the problem
-------------------------------------------------
all versions down to 6.6 not working (PC versions) and 7.23 android not working. 

Expected behavior
-----------------

when opening replay and pressing +10' it should advance log file 10min ahead. 

Actual behavior
---------------

Does nothing. 

Steps to reproduce the behavior
-------------------------------

File recorded during actual flight with android 7.23 version xcsoar.
choose file to replay - press +10'

Do you have any idea what may have caused this?
-----------------------------------------------
- something changed in igc file format during the years that causes it?

Do you have an idea how to solve the issue?
-------------------------------------------

lordfolken:
Can you attach the igc file?

fronzee:
[2022-05-23-XCS-XYA-01.zip](https://github.com/XCSoar/XCSoar/files/8765851/2022-05-23-XCS-XYA-01.zip)
igc itself not allowed to upload, so compressed attached

lordfolken:
Seems to happen with any igc file i tried.

fronzee:
In addition to +10' I would say -10' would be good to have too, in case you go 10' too much in a 5hours flight... There would be a nice chance to just go back 10' instead of doing it all again from the beginning.

lordfolken:
Yeah the function needs work. There is also a demo mode if you leave the filename empty and press play. Very intuitive. 🙃

kobedegeest:
@coderabbitai have a look and propose a fix

lordfolken:
**Open PR:** #2472 — [mockup] Replay: flight-minute seek (experimental).

WolfgangDrescher:
Actually, this `+10'` button does not skip 10 minutes, but instead enters a fast-forward mode where the flight is played faster for probably 10 flight minutes (I did not wait for the full cycle, so I can't say for sure).

Can we agree to change this behavior to skip 10 minutes, as we already have a rate feature to play back a flight faster?

I'm working on seek-to-next-cruise/circling buttons which could be a more intuitive way to navigate through a flight than jumping ahead by 10 minutes. See https://github.com/WolfgangDrescher/XCSoar/tree/replay-seek

PR #2472, opened 2026-05-05
[mockup] Replay: flight-minute seek (sync fixes)

lordfolken:
**Mockup / experimental** — for discussion; not release-ready as-is.

Adds replay dialog control + Lua `seek_to_flight_minutes` to restart the recording and apply every fix synchronously (MergeThread / CalculationThread) up to N minutes after the first fix.

NEWS entry marked mockup.

Related to #933

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -1,4 +1,7 @@
 Version 7.45 - not yet released
+* replay (mockup / experimental — not release-ready)
+  - Replay dialog: seek to a chosen number of minutes after the first fix,
+    restarting the file and applying every merged fix (dense snail trail)
 * devices
   - FLARM $PGRMZ (when FLARM is detected) feeds igc_pressure_altitude for the
     ISA logger datum; cleared when strong pressure replaces weak pressure or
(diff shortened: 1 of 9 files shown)


WolfgangDrescher:
What needs to be changed in this PR in order to make it mergeable? I'm working on seek-to-next-cruise/circling buttons which make debugging in XCSoar quite a bit easier.

See https://github.com/WolfgangDrescher/XCSoar/tree/replay-seek

WolfgangDrescher:
Could I get some feedback on what is left to do here, as well as on my new "seek-to-next-cruise/circling" feature? It would make developing certain features easier by allowing faster navigation through a replay file. Should I go ahead and open a draft PR in the meantime? Also note my comments in https://github.com/XCSoar/XCSoar/issues/933:

> Actually, this +10' button does not skip 10 minutes, but instead enters a fast-forward mode where the flight is played faster for probably 10 flight minutes (I did not wait for the full cycle, so I can't say for sure).
> 
> Can we agree to change this behavior to skip 10 minutes, as we already have a rate feature to play back a flight faster?
> 
> I'm working on seek-to-next-cruise/circling buttons which could be a more intuitive way to navigate through a flight than jumping ahead by 10 minutes. See https://github.com/WolfgangDrescher/XCSoar/tree/replay-seek

lordfolken:
Mockups are exploratory for me. They just are a whatif without regards to compatibility or testing. I dont really intend to merge these.

PR #2475, opened 2026-05-06
Airspace/AirspaceWarning: Suppress warning triangle for ACK'ed airspaces

groundhogxc:
This PR fixes a bug where acknowledging an airspace would up to now suppress the warning sound, the message, and the filling - but not the little yellow warning icon at the expected location of incursion.

This is an old bug, so I am not 100% sure whether it is maybe by design. But I do not really see a particular reason why.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed airspace warning behavior to correctly record solution location only when acknowledgment has expired, preventing improper data recording in other warning scenarios.

### DIFF
--- a/src/Airspace/AirspaceWarningCopy.hpp
+++ b/src/Airspace/AirspaceWarningCopy.hpp
@@ -27,7 +27,8 @@ public:
       ids_inside.checked_append(&as.GetAirspace());
     } else if (as.IsWarning()) {
       ids_warning.checked_append(&as.GetAirspace());
-      locations.checked_append(as.GetSolution().location);
+      if (as.IsAckExpired())
+        locations.checked_append(as.GetSolution().location);
     }
 
     if (!as.IsAckExpired())


lordfolken:
I understand the motivation. But would be possible to use another icon when its acknowledged? I can of still want to be aware when interecting an airspace.

groundhogxc:
It is possible, sure - although even with that I find this behavior a bit confusing, personally. If you want to be made aware of a crossing, then should there not just be no ACK in the first place?

My personal interpretation: The ACK state's role is to suppress notifications. Also, that triangle shows only for "NEAR" type warnings, not for "INSIDE" type warnings, so even with a different symbol it feels very inconsistent.

The outline of the airspace is still there of course, so when you look at the map you see there is a boundary. Only the padding and the display of warning messages (and with this commit the triangle symbol) is suppressed.

Did you really use this "feature" in this way so far? I did not, TBH, until recently I had not even fully and completely understood when and why they appear without warnings.

lordfolken:
Ok. I studied it a bit now and i come to the same conclusion as you. That icon should be gone for a dismissed airspace. Also the outline makes it actually clear. 

Btw.  i think we should rename ack to "dismiss". This is more according your description and also other gliding apps/computers describe it as such.

lordfolken (approved):

PR #2487, opened 2026-05-11
Waypoint search: Type filter cleanup and full CUP type coverage

### LINKED ISSUE #1376: Waypoint Selection Dialog allows *.XCM file to be chosen?

XCSoar v7.42-Linux~git#010af8a728
No information - above version is only one I have used

System information
------------------

Dell Precision M6700 laptop, 16GB, 500GB, Debian 12 (Bookworm)

Not sure how the .XCM file got on the list for 'TYPE', but it is selectable

Expected behavior
-----------------

Only files containing valid waypoints should show on the list

Actual behavior
---------------

'US_FL_CENTRAL.xcm' file available as one of the items on the list that displays when 'Type' is selected
see screenshot
Screenshot from 2024-03-09 19-31-48

Not filtering the File1/File2 candidates for valid waypoint file extensions?

### LINKED ISSUE #2485: Waypoint search dialog: Type filter is missing most CUP waypoint types
XCSoar version
--------------

master (commit dcfbdb61d4 at time of writing) and earlier; this is a long-standing gap rather than a regression.

The **Type** filter dropdown in the Waypoint Selection dialog (`Dialogs/Waypoint/WaypointList`) currently exposes only a small set of categorical / task-related filters:

- All (`*`)
- Airport
- Landable
- Turnpoint
- Start / Finish
- Left / Right FAI Triangle
- Custom (= USER-origin)
- (dynamic per-file entries)
- Map file
- Recently Used

These map to `enum class TypeFilter` in `Waypoint/WaypointFilter.hpp`. However the underlying `enum class Waypoint::Type` (`Engine/Waypoint/Waypoint.hpp`) carries 20 distinct values, and SeeYou `.cup` files use most of them. The following types **cannot** be filtered for in the dialog at all today:

| `Waypoint::Type` value | User-visible name (already translated in `GetWaypointTypeName`) |
|---|---|
| `MOUNTAIN_TOP` | Mountain Top |
| `MOUNTAIN_PASS` | Mountain Pass |
| `BRIDGE` | Bridge |
| `TUNNEL` | Tunnel |
| `TOWER` | Tower |
| `POWERPLANT` | Power Plant |
| `OBSTACLE` | Transmitter Mast |
| `THERMAL_HOTSPOT` | Thermal Hotspot |
| `MARKER` | Marker |
| `VOR` | VOR |
| `NDB` | NDB |
| `DAM` | Dam |
| `CASTLE` | Castle |
| `INTERSECTION` | Intersection |
| `REPORTING_POINT` | Reporting Point |
| `PGTAKEOFF` | Paragliding Takeoff |
| `PGLANDING` | Paragliding Landing |

Some of these were introduced as part of the CUP 2022 format support (commit `7c3e545bba` "waypoint: new waypoint types for cup 2022", and `8fd07bb243` "WaypointEditor: support more waypoint types") which extended the parser and the editor but did not extend the search-dialog filter.

Proposed change: extend the Type filter dropdown so the user can narrow the list to a single specific waypoint type (e.g. only Mountain Passes, only Reporting Points, only Paragliding Takeoffs). The labels can be sourced verbatim from the existing `GetWaypointTypeName()` helper in `Dialogs/Waypoint/WaypointInfoWidget.cpp` so the strings stay in sync and translations are reused.

Implementation notes (for whoever picks this up):

- Add new entries to `enum class TypeFilter` for each `Waypoint::Type` value that isn't already covered by an existing categorical filter (`AIRPORT` / `LANDABLE` / `TURNPOINT` already cover `AIRFIELD` / `OUTLANDING` / `NORMAL`).
- Extend `WaypointFilter::CompareType()` (`Waypoint/WaypointFilter.cpp`) with a case per new value.
- Make sure the static array `type_filter_items` in `Dialogs/Waypoint/WaypointList.cpp` stays in lockstep with the enum (the current array uses positional `addEnumTexts()` which is fragile; switching to explicit `addEnumText(text, id)` per enum value would also fix the existing label/index misalignment around `MAP` / `LAST_USED`).
- The new `TypeFilter::*` IDs must remain `< _DYNAMIC_FILE_ID_START` (= 100) -- there is already a `static_assert` enforcing this.

Add a `COUNT` sentinel to `TypeFilter` and `static_assert` arrays match (per project conventions for multi-place enums).

Locating a specific waypoint by type when you do not remember its name. Examples that come up in real flight planning:

- "Show me all the **Mountain Passes** in the loaded database" -- useful for ridge/wave route planning.
- "Show me all the **Paragliding Takeoffs**" -- for PG pilots flying with a database that mixes airfields, turnpoints and PG sites.
- "Show me only **Reporting Points**" -- common in regional databases (FR, CH) where reporting points are mixed with airfields and turnpoints.
- "Show me all **Thermal Hotspots**" -- preflight/in-flight when planning a leg over unfamiliar terrain.

Today the user has to scroll through the entire database (often thousands of waypoints) or rely on the name to contain a recognisable substring -- which is precisely the case the substring search work in #2483 also addresses, but type-filtering would let the user narrow down without typing anything at all.

lordfolken:
## Summary

Four small commits that fix two related issues in the Waypoint search dialog's **Type** filter dropdown. Each commit compiles standalone and is internally self-contained.

1. **Hide empty map archive filter** -- `WaypointGlue::LoadWaypoints` only loads the embedded `.cup` / `.xcw` from a configured `.xcm` map when no other waypoint file produced any result. The Type dropdown was nevertheless advertising a *Map file* (or the `.xcm` filename) entry regardless of whether any MAP-origin waypoints had actually been loaded -- so users could pick a filter that always yielded zero matches. The entry is now only shown when at least one waypoint with `WaypointOrigin::MAP` is loaded.

2. **Refactor Type dropdown to explicit IDs** -- the dropdown was populated from a positional `const char *const[]` with the dropdown index implicitly assumed to equal the `TypeFilter` enum value. This was off by two: the static array had no entry for `TypeFilter::FILE`, so the *Map file* entry was actually sitting at the FILE slot, and *Recently Used* was at the MAP slot, where it got overwritten with the `.xcm` filename when one was loaded. `TypeFilter::LAST_USED` had no working dropdown entry at all and the existing `FillLastUsedList` code path was therefore unreachable from the UI. Switch to `addEnumText(text, id)` populated from a structured `TypeFilterChoice` table so dropdown layout and enum values can evolve independently. The MAP gating from commit 1 folds into the loop.

3. **Add type filters for all CUP waypoint types** -- 17 new `TypeFilter` values covering Mountain Top/Pass, Bridge, Tunnel, Tower, Power Plant, Obstacle (Transmitter Mast), Thermal Hotspot, Marker, VOR, NDB, Dam, Castle, Intersection, Reporting Point, PG Take Off and PG Landing Zone. These types have existed in `Waypoint::Type` since the cup-2022 format support landed but the search dialog never exposed them, so users could not filter by them. Labels match the strings returned by `GetWaypointTypeName()` in `WaypointInfoWidget` so the dropdown reads the same as the *Type* row in the waypoint details dialog and existing po translations are reused. A `TypeFilter::COUNT` sentinel + `static_assert` on the dropdown choices array catches future enum/dropdown drift at compile time.

4. **`.cursor/rules`: Add waypoint-types checklist** -- adding a new `Waypoint::Type` silently breaks CUP round-trip (`WaypointReaderSeeYou` ↔ `CupWriter`), the editor dropdown, the details label, the icon chain (`WaypointLook` + `resources.txt` + `WaypointIconRenderer`), the search filter, and (rarely) the visibility threshold -- the compiler does not catch the misses (most sites have neither `default:` nor a `COUNT` sentinel). The new rule documents every site so the checklist surfaces automatically when working in any of the relevant directories.

## Issues addressed

- Closes #1376 -- Waypoint Selection Dialog allows `*.XCM` file to be chosen even when its embedded waypoints are not loaded.
- Closes #2485 -- Waypoint search dialog: Type filter is missing most CUP waypoint types.

## Test plan

- [x] Full unit test suite passes (\`Files=82, Tests=9235\`).
- [x] \`make -j\$(nproc) TARGET=UNIX USE_CCACHE=y\` builds clean after each individual commit.
- [x] Manual: open Waypoint search Type dropdown with no \`.xcm\` configured -- *Map file* entry is absent.
- [x] Manual: configure an \`.xcm\` whose embedded waypoints are loaded -- entry shows the actual \`.xcm\` filename.
- [x] Manual: configure an \`.xcm\` alongside a populated WPFileList -- *Map file* entry is absent (no zero-match filter).
- [x] Manual: select *Recently Used* from the dropdown -- recently-used waypoints appear (was previously dead from this UI path).
- [x] Manual: load a CUP file with various non-airfield waypoint types and filter by each new entry (Mountain Top, VOR, Intersection, PG Take Off, etc.).

## Summary by CodeRabbit

* **New Features**
  * Waypoint search type filter now includes comprehensive filtering for all CUP waypoint types, including mountain tops/passes, bridges, tunnels, towers, power plants, obstacles, thermal hotspots, VOR/NDB stations, dams, castles, intersections, reporting points, and PG takeoff/landing sites.

[Review Change Stack](https://app.coderabbit.ai/change-stack/XCSoar/XCSoar/pull/2487)

### DIFF
--- a/.cursor/rules/waypoint-types.mdc
+++ b/.cursor/rules/waypoint-types.mdc
@@ -1,0 +1,44 @@
+---
+description: Where to register a new Waypoint::Type or TypeFilter value across the codebase
+globs:
+  - "src/Engine/Waypoint/Waypoint.hpp"
+  - "src/Waypoint/**"
+  - "src/Renderer/WaypointIconRenderer.cpp"
+  - "src/Look/Waypoint*"
+  - "src/Dialogs/Waypoint/**"
+  - "Data/resources.txt"
+alwaysApply: false
+---
+
+# Waypoint Types
+
+Waypoint type information is duplicated across many sites; the compiler does **not** flag missing cases (the enums have no `default:` fallthrough catch in some sites and most arrays carry no size sentinel). When adding or changing values, walk this checklist or you will silently break round-trip, rendering, or UI.
+
+## Adding a new `Waypoint::Type`
+
+The enum is in `src/Engine/Waypoint/Waypoint.hpp`. Every new value needs:
+
+1. **CUP round-trip** — both directions must agree:
+   - Reader: `src/Waypoint/WaypointReaderSeeYou.cpp` (`case <cup-id>:` → `Waypoint::Type::X`)
+   - Writer: `src/Waypoint/CupWriter.cpp` (`case Waypoint::Type::X:` → `"<cup-id>"sv`)
+2. **Editor dropdown** — `waypoint_types[]` in `src/Dialogs/Waypoint/dlgWaypointEdit.cpp` (`N_("...")` label).
+3. **Details label** — `GetWaypointTypeName()` in `src/Dialogs/Waypoint/WaypointInfoWidget.cpp` (`_("...")` label; **must match** the editor dropdown so po entries are reused).
+4. **Map icon** — only if visually distinct:
+   - `src/Look/WaypointLook.hpp` (declare `Icon foo_icon`)
+   - `src/Look/WaypointLook.cpp` (`foo_icon.LoadResource(IDB_FOO_ALL)`)
+   - `Data/resources.txt` (register `IDB_FOO`)
+   - `src/Renderer/WaypointIconRenderer.cpp` (`case Waypoint::Type::X: return look.foo_icon;`)
+5. **Search filter** — see "Adding a `TypeFilter`" below; the dropdown label must match step 3.
+6. **Visibility threshold** *(rare)* — `src/Projection/MapWindowProjection.cpp` only if the type warrants a non-default scale (e.g. obstacles).
+
+## Adding a `TypeFilter` value
+
+`enum class TypeFilter` lives in `src/Waypoint/WaypointFilter.hpp`. The `static_assert(ARRAY_SIZE(type_filter_choices) == unsigned(TypeFilter::COUNT) - 1, ...)` in `src/Dialogs/Waypoint/WaypointList.cpp` will catch a missing dropdown entry at compile time, but the other sites are silent:
+
+1. Add the value **before `COUNT`** and keep it `< _DYNAMIC_FILE_ID_START` (= 100).
+2. `WaypointFilter::CompareType()` in `src/Waypoint/WaypointFilter.cpp` — add the `case`.
+3. `type_filter_choices[]` in `src/Dialogs/Waypoint/WaypointList.cpp` — add the label (reuse the `GetWaypointTypeName` string verbatim so translations apply).
+
+## Translations
+
+Any new English label introduced in the steps above is exposed via `_()` / `N_()`. Run `make update-po` to refresh `po/*.po` and `po/xcsoar.pot` before committing translation changes (commit them at the **end** of the branch, separately, per `xcsoar-project-rules.mdc`).
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -5,6 +5,18 @@ Version 7.45 - not yet released
     ignores the sentence
 * ui
   - infoboxen: new Altitude IGC InfoBox (logger pressure or ISA pressure only)
+  - waypoint search: Type filter no longer lists the .xcm map archive as a
+    selectable filter when its embedded waypoints are not loaded (e.g. when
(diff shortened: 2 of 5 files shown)


lordfolken:
@coderabbitai full review

Issue #2469, opened 2026-05-04
Curated flying profiles: repository + offline defaults (glider/PG/paramotor × XC/local/competition)


lordfolken:
## Summary

XCSoar’s **file repository** (`https://download.xcsoar.org/repository`, see `src/Repository/Glue.cpp`) lists downloadable **maps, airspace, waypoints, tasks, XCI, …** but **not `.prf` profiles**. `FileType` (`src/Repository/FileType.hpp`) and the repository line parser (`src/Repository/Parser.cpp`) have **no `profile` type**, so curated “flying presets” cannot be offered the same way as other data.

Users should be able to pick **sensible default profiles** either **offline** (shipped with the app / data image) or **from the repository** after download.

## Proposed dimensions

Profiles could be described along two axes (names indicative—final taxonomy up for discussion):

### Aircraft / craft category

- **Glider** (sailplane—current defaults biased here)
- **Paraglider**
- **Paramotor** (and possibly **hang glider** if distinct enough)

Each category implies different defaults (units, map scale behaviour, InfoBox sets, task rules relevance, wind display, safety heights, etc.).

### Flying pattern / mission

- **Cross-country** — task-centric InfoBoxes, navigation aids (related discussion: #2467)
- **Local flight** — thermalling / site-focused layout, simpler task chrome
- **Competition** — contest-oriented defaults (rules, start/finish helpers, visibility of tactical fields)

A **matrix** of combinations (e.g. glider × XC, paraglider × local, paramotor × local) avoids an explosion of arbitrary filenames while staying discoverable.

### Beginner / simple profile

In addition to the craft × pattern matrix, offer a **beginner** (or **simple**) profile per craft category as a recommended **first-run** choice:

- **Glider — Beginner**
- **Paraglider — Beginner**
- **Paramotor — Beginner**

This is not a third mission axis; it is a **reduced-complexity entry point** for new users or pilots who want conservative, easy-to-understand defaults before graduating to XC / local / competition templates.

Suggested contents for beginner/simple templates:

- **Conservative safety defaults** — higher arrival height and terrain clearance, safety MC enabled, STF risk factor considered, route mode **Both** (terrain + airspace), reach polar **Safety MC**, airspace warnings enabled
- **Simpler UI** — fewer InfoBoxes, essential panels only, less map clutter; expert settings hidden or de-emphasised in config
- **Clear startup communication** — surface active safety values at profile selection (see #2548)
[...]

PR #2544, opened 2026-05-28
Engine/Task: Increase default safety_mc value from 0.5 to 2

DanD222:
The original commit in https://github.com/XCSoar/XCSoar/pull/778 got dropped in a rebase, but I still think the issue is relevant, so I'm opening a new PR.

- When flying at optimal glide speed, a MC setting of 2 should give a pilot a bigger safety margin in terms of the reach area than 0.5 in case of unexpected sink

- MC 2 is a rather arbitrary value, but will give beginners not familiar with XCSoar enough to increase the value themselves a better chance of reaching their intended target

- Pilots flying in mountains should probably set this much higher; the value of 2 is intended as a starting point for flatlands

## Summary by CodeRabbit

* **Chores**
  * Updated default task behavior safety configuration value.

[Review Change Stack](https://app.coderabbit.ai/change-stack/XCSoar/XCSoar/pull/2544?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/Engine/Task/TaskBehaviour.cpp
+++ b/src/Engine/Task/TaskBehaviour.cpp
@@ -36,7 +36,7 @@ TaskBehaviour::SetDefaults()
   calc_glide_required = true;
   goto_nonlandable = true;
   risk_gamma = 0;
-  safety_mc = 0.5;
+  safety_mc = 2;
   safety_height_arrival = 300;
   task_type_default = TaskFactoryType::RACING;
   start_margins.SetDefaults();


lordfolken:
Raising default Safety MC to 2 also raises startup/task MC (Startup.cpp). With default Auto MC off, that value persists on cruise. Even with Auto MC on, it remains until a valid thermal is finished (≥45 s). On performance polars, MC 2 STF is ~150–180+ km/h — near maneuvering speed — which is a poor default for pilots who never touch settings. If we want conservative reach, that should be decoupled from the MC ring (or startup MC should not track Safety MC).

DanD222:
Decoupling Safety MC from STF values sounds like a good idea.

So ,
- in Manual MC, STF should be coupled to whatever value's in the MC Infobox
- in Auto MC, STF could be coupled to MC0, until the first thermal is finished 

?

My understanding is that Reach is separate from STF (or should be). We have Reach Polar:
- Task (whatever's in the MC Infobox?)
- Safety MC

In case Reach Polar is set to Safety MC, then I see no issue, as that should be governed by whatever value's set as the Safety MC

In case Reach Polar is set to Task, then, in case of Auto MC, the issue is what to have the initial MC value for the Reach Polar, until the first thermal is finished.

I would think Safety MC would be the more conservative option in this case.

But in general I see no big issue with the Reach being governed by a different MC than the STF.

Scumi:
I don't think MacCready should be used to increase safety margins. We already have dedicated settings for arrival altitude and terrain clearance, which make the intended reserve explicit.

Mixing performance assumptions and safety reserves makes the resulting behavior harder to understand. Introducing separate "MC" and "Safety MC" values would add complexity, both in the code and for the user, while largely duplicating functionality that already exists.

It also introduces an arbitrary tuning parameter. Why should the safety MC be 2? Why not 1.5, 3, or 5? Unlike arrival altitude or terrain clearance, it has no clear operational meaning and effectively becomes a magic number that users must tune.

The current defaults seem reasonable to me, and pilots who want larger margins can already achieve that through the existing configuration options.

DanD222:
> I don't think MacCready should be used to increase safety margins. We already have dedicated settings for arrival altitude and terrain clearance, which make the intended reserve explicit.

I personally set the arrival height to zero so that I know how exactly at what altitude I arrive (i.e. if it's at 50m or 299m). I would like to keep using it that way.

> Mixing performance assumptions and safety reserves makes the resulting behavior harder to understand. Introducing separate "MC" and "Safety MC" values would add complexity, both in the code and for the user, while largely duplicating functionality that already exists.

There already is a Safety MC, I'm not introducing anything. My issue is that the Safety MC is 0.5, which in my book is no Safety MC at all, but quite the opposite. Someone not knowing this could get caught in a nasty trap.


> It also introduces an arbitrary tuning parameter. Why should the safety MC be 2? Why not 1.5, 3, or 5? Unlike arrival altitude or terrain clearance, it has no clear operational meaning and effectively becomes a magic number that users must tune.

The fact that there are several values to choose from shouldn't prevent us from choosing one, if it makes the situation better. I don't really care what particular value it is, I just don't want it to be 0.5.
It does in fact have a clear operational meaning, see above: 
 
>We have Reach Polar
>- Task (whatever's in the MC Infobox)
>- Safety MC

Scumi:
You are right that we would not introduce anything new, my misinterpretation.

I am still not convinced that 2 is a better default than 0.5. The choice appears largely arbitrary, and I do not see a compelling reason to prefer 2 over any other value.

More importantly, increasing Safety MC does not simply add a safety margin. It changes the assumptions used by the calculation. For pilots who normally fly with lower MC settings, a safety MC of 2 can make the system unnecessarily pessimistic and indicate that a target is unreachable when it may in fact be reachable with their normal flying style. 

To me, this comes down more to individual preference than to a generally applicable improvement. Since the value is already user-configurable, I would prefer to keep the current default.

DanD222:
I am absolutely certain that 2 is a better value than 0.5 as a Safety value. Consider the following scenario:
1. Set Reach Polar to Safety MC (with 0.5)
2. Try to fly to a landing site behind a forest, just on the edge of the reach area
3. ...
What are the odds of you making it? 

If someone who doesn't know the inner workings of XCSoar and relies solely on "but XCSoar told me 0.5 is safe", then they potentially won't be very happy.

I'd also like to re-state my disagreement with choice paralysis :)

lordfolken:
I don't think we will find a "onesizefitsall" solution. 
At some point in the future I hope that we will be able to serve several profiles at startup/or for download: 
https://github.com/xcsoar/xcsoar/issues/2469 There i'd like to have a beginner profile where we can offer more conservative values. 

  The second thing from i take from this discussion, is that all the safety relevant values need to be communicated up front. 
  LXNAV for example puts the Safety Altitudes at startup on the profile selection page. 
  We could also put a message at startup and list all the safety relevant values, such as terrain clearance height, safety height, and safety mc/risk factor. Created an issue here: https://github.com/XCSoar/XCSoar/issues/2548

DanD222:
I'm not trying to find an optimal solution - I'm merely trying to give a newbie trying to reach a landing site on the far edge of the reach area a bigger chance of success. To re-state, I don't care what value it is, just not 0.5. 

Anything extra is fine, but keeping the value at 0.5 and having messages, disclaimers, etc. as an excuse is not particularly responsible IMO.

DanD222:
I'd also insinuate that there's quite a high risk of information overload and people not actually reading what's being presented to them on XCSoar startup, but that doesn't mean XCSoar shouldn't do it. 

I would however not rely on a user making a deliberate action to manually change a non-safe default value of 0.5 to a safer one.

Scumi:
I think this is where we fundamentally disagree.

Calling the current default unsafe is, in my view, opinion-based (reasons outlined above).

If the concern is that pilots should not arrive at an airfield at 0 m, then I would argue that arrival reserve or safety altitude are the more appropriate settings. They express that intent directly, whereas Safety MC changes the underlying assumptions of the calculation.

Given that, I don't see a compelling reason to change the default based solely on the argument that "anything but 0.5 would be better".

@lordfolken has seen our arguments and can decide ;-)

DanD222:
> Calling the current default unsafe is, in my view, opinion-based (reasons outlined above).

I object to my view being labelled as opinion-based.  If you try to reach the edge of the reach envelope at MC 0.5, what's your safety margin? 

> If the concern is that pilots should not arrive at an airfield at 0 m, then I would argue that arrival reserve or safety altitude are the more appropriate settings. They express that intent directly, whereas Safety MC changes the underlying assumptions of the calculation.

I also object to the approach of suggesting other approaches instead of addressing the issue of whether XCSoar should recommend 0.5 as a Safe value.

WolfgangDrescher:
> I don't think MacCready should be used to increase safety margins. We already have dedicated settings for arrival altitude and terrain clearance, which make the intended reserve explicit.
> 
> Mixing performance assumptions and safety reserves makes the resulting behavior harder to understand. Introducing separate "MC" and "Safety MC" values would add complexity, both in the code and for the user, while largely duplicating functionality that already exists.

But I have to say that setting up all the safety factors properly is not really straightforward. I'm not sure I myself understand all the options correctly; I would consider myself as having above-entry-level knowledge of XCSoar, but limited XC experience in a glider.

In my opinion the problem is mainly suboptimal naming (translations) and unclear documentation. Sometimes there is an understandable description of the settings directly in the app, sometimes there isn't. Sometimes you need to read the documentation to get more information. And sometimes you stumble upon explanations by accident, as I did a few weeks ago when I discovered this in the German `XCSoar Blitzeinstieg` manual:

> Es gibt mitunter Konfusion während des Endanfluges mit MC-Werten größer 0. XCSoar halt sich streng an die reine MC-Lehre. Das bedeutet, daß ein Endanflug ein reines Gleiten ist, auf dem nicht mehr gekurbelt wird. Das ist gleichbedeutend mit MC=0. Die "übliche Vorgehensweise" und dies speziell bei Gegenwind, den MC-Wert als Verschiebung der Polare zu mißbrauchen kann daher zu außerordentlich verwirrenden Ergebnissen führen.

At the same time, the "final glide bar" does actually factor in the current MC value. But if you cancel your task, Safety MC should apply. However, when you then use "Goto" to an airfield and change the MC value, it still gets factored into the arrival height, whereas I would expect Safety MC to apply. This is confusing, especially for a novice.

It's a fact that people use higher MC values to deliberately worsen the glide computer's calculations *on the fly*. Even in my club, a very experienced XC pilot and flight instructor advised me to use higher MC values for more conservative calculations (he is mainly using LXNAV, maybe things are different there, I don't know it yet).

> The second thing from i take from this discussion, is that all the safety relevant values need to be communicated up front.

I agree with @lordfolken's statement and would argue that the documentation should be majorly improved for this aspect. We should think about documenting all the relevant factors: `Arrival height`, `Polar degradation`, `Safety MC`, `STF risk factor` (but also `Auto MC mode` and `Predict wind drift`), and explain in detail what impact each of them has. Am I missing a factor?

Or maybe we need an InfoBox that allows changing the `Polar degradation` or `STF risk factor` on the fly?

One more reason for https://github.com/XCSoar/XCSoar/discussions/2517 :D

DanD222:
- The Final Glide Bar is governed by the Current MC (and can display both the Current MC and MC0)
- The Reach area on the map, if set to Task MC, changes together with the Current MC
- If the Reach Polar is set to Safety MC, the Reach area on the map stays the same and is independent of the Current MC / Final Glide Bar 
- It is only the issue of "_What does the reach area show when set to Safety MC?_" that I'm addressing - I want it to be more pessimistic.

From my point of view, the issue is very simple. Yes, there are many factors affecting things, but all I'm addressing is:
1. XCSoar recommends 0.5 as a Safety value to tell me "_Will I get there?_" on the reach area of the map
2. Will I actually get there with 0.5? I would say that maybe yes, maybe not
3. If the answer is "Maybe yes, maybe not", then why is it labelled "Safety"?
4. What I'm recommending is changing the value that XCSoar is recommending, to change the answer from "_Maybe yes, maybe not_" to "_Most probably yes_"

It really doesn't need overthinking, in my opinion.

(OK, there's the coupling to STF in MC Auto, and I think that should be changed, but that's a separate issue)

lordfolken:
See that there is the fundamental misunderstanding: 

MC isn't about "can i get there?". 
MC is about "how **fast** can i get there?" 

This abuses the the margin between MC speed and Vopt. 

The whole safety mc construct is based on this misunderstanding. It shouldn't exist.  
It should be a glide ratio compensated for wind/terrain. 

For example consider sink... if you fly 105 instead of 150 through an extended sink, it will affect you much more and despite your margin you might not reach the field at 105, whereas you would have at 150.

I'm against this change, because this is hiding options from view. Sure it will show you an option at 50KM with +50m. Is it realistic in a pure glide? probably not.  With potential climb, or convergence: absolutely.

With MC 2 you don't see this option, therefore you never get to decide on it.

Its a tactical glide computer. Decisions need to be made. 

If you want absolute safety for a newbie? Then the alternate GR infobox is your friend. If that is below 1:25 you are pretty safe in most scenarios. Its easily understood and already taught in basic. f

DanD222:
> See that there is the fundamental misunderstanding:
> 
> MC isn't about "can i get there?". MC is about "how **fast** can i get there?"

I beg to differ.

Config > System > Configuration > Glide Computer > Route > Reach Polar = **Safety MC** (0.5)
RP_safety

Config > System > Configuration > Glide Computer > Route > Reach Polar = **Task** (in this case MC 2)
RP_task

I think the misunderstanding is that the Reach area on the map **when Safety MC is selected** is coupled to the MC in the infobox. 

It is not. You can have the MC in the infobox selected to whatever value you like, and it will tell you to fly at the speed for that MC setting. The Reach area on the map won't change. I urge you to try this in the simulator.

mlep:
The alternate GR infobox is nice, since it displays the required GR.  And a safety GR is what I was taught to care for! And this is what I am teaching now...

But this infobox only displays the closest airfield (= 'landable waypoint' in XCSoar's terminology). The map displays in green the reachable airfields, which is convenient. However, when I started to use XCSoar (many years ago) --- and read its manual :smile: ---, I was surprised I could not set a safety GR to determine the reach displayed on the map, i.e. I could not use the method I was taught. 

I would expect that XCSoar would allow for the use of a predefined GR. 
That will be easier to understand for most pilots and will be coherent with what newbies are taught. 

By the way, is there any country where a safety MC is taught to pilots?

Maybe the current mode could be kept as an option for more advanced pilots?

DanD222:
I am honestly baffled at the amount of energy everyone is investing into suggesting solutions /other/ than the proposed one, especially when it's so easy to try out and observe the behaviour.

For me, the issue still remains the same - XCSoar shouldn't have a parameter named "Safety" with a value that can potentially lead someone into trouble.

groundhogxc:
I stick with my opinion from that same discussion three years ago:

> I agree with Dan. Something with "safety" in the name should typically provide significant additional safety over the typical use setting. Which in the case of an MC value means significantly higher. There are not many circumstances where MC 0.5 provides significant "additional(!)" safety over even the lowest value (MC 0).
> 
> I would also argue to set the default higher. If not MC 2, maybe 1.5 (in the mountains I would not even count that as "safety MC"). Even 1 is better as a default than 0.5. Personally I never set a value smaller than 0.5 for the "regular" MC value other than in exceptional circumstances (late night glide home in case of tailwind being one of the few exceptions).

The arguments in the new discussion do not change my mind.

I see the points for some, e.g. in this one:

> The whole safety mc construct is based on this misunderstanding. It shouldn't exist.
> It should be a glide ratio compensated for wind/terrain.

But if I think about it then that is pretty much exactly what the concept of safety MC is: Using the regular arrival calculation that includes wind compensation, but applying a steeper glide path to get a distance-proportional additional safety margin in addition to the arrival height setting.

Of course it would also work to use a "Safety bug factor" instead, that would also be fine with me. But the "industry standard" so far has been to use an increased MC value for this.

One thing (but out of the scope of this very narrow PR!) that could make the G/R proponents happier could be to display the calculated G/R next to the MC option in the settings dialog for safety MC?
So the selection drop-down list would say "2 kts (1/30)", "2.2 kts (1/29)" etc., calculated based on the current plane polar and mass settings? The functions are all there already.

Using the MC instead of a pure glide angle as default behavior also has the significant advantage that it scales reasonably with the aircraft type: An aircraft that has a faster "typical glide speed" needs less additional slope to achieve the same safety improvement than a slower one (because the latter is affected more by the same sinking airmass). You do not need a safety glide ratio of below 1:20 with flat terrain in most weather conditions with a Ventus 2cx, but it is realistic for a Schweizer 1-26. Setting a safety MC to effectively choose the safety glide ratio accounts for that in a fairly realistic way.

IMO you just need to think about MC here simply as a "make glide steeper"-Parameter. Similar to, when using risk-based altitude-dependent MC, the MC is a "make my flight safer/riskier"-Factor to manage outlanding risk (although there it is of course also related to speed).

I am still in favor of increasing the default to 2. The fact that XCSoar starts up with that value as MC setting is annoying, indeed - but to get rid of that I would change that startup behavior instead, for example to one of these three options:

- Use the current safety default 0.5 as the startup default always, it is a good starting value
- Just remember the last STF MC setting from before (my favourite, this is also what most glide computers do, I think)
- Introduce a second default setting as the "startup default MC"

Conclusion and reminder: @DanD222 's PR is **very narrow scope**, **about the default value of the current system only**. Not about replacing the system. I am not saying safety MC is the best concept in the world, but it works alright when properly understood. But most definitely the current state of a 0.5 default is not in any way related to any "additional safety over normal operation", and I agree with DanD that it should be higher by default.

DanD222:
Thanks @groundhogxc. That's the first comment that actually understands exactly what I'm aiming for. 

Here's some related reading for @Scumi  , @lordfolken  

https://static1.squarespace.com/static/5e6033a4ea02d801f37e15bb/t/5eeaa63c2b0d574e76b9571a/1592436284733/safety_glides.pdf

(from https://www.johnhcochrane.com/about/soaring)

It starts with:

> “I don’t get it,” said the pilot as his crew was picking him out of the field. “The glide computer said I could make it home with 200 feet to spare! This stupid thing must be busted.”

> “No,” his crew answered, “you set the MacCready to zero. You were guaranteed not to make it.” “Huh?” said the pilot, “I set the MacCready to zero to maximize my glide.”

> Crew explained: “If you had found lift, you wouldn’t have landed. If you’re not in lift, you’re in sink. So the only way you will land is if you find sink and no lift. By using a zero Mc setting you are guaranteeing that if you land, it won’t be at the airport.”




The first three chapters:

> Lesson 1: Learn to speak MacCready
> Lesson 2: Decouple the glide computer and the speed director
> Lesson 3: You may have to use much higher values than you think

But again, like @groundhogxc  correctly mentions, that is not the scope of this pull request, rather a background as to why MC 0.5 is not a good idea as a Safety value, and why having a MC2 doesn't actually mean that we should necessarily be flying at that speed.

lordfolken:
This PR does not address any of the numerous points i have raised. 

While I agree that the calculations are optimistic with the current value, just arbitrarily raising that value is not the solution.

DanD222:
Well, doing nothing doesn't improve the situation at all. The fact is:
1. The reach are on the map is affected by the Safety MC if Reach Polar is set to Safety
2. This setting is too low

Ignoring these facts and wishing that people would use something else doesn't in any way change this.

PR #2557, opened 2026-05-31
Weather/XCTherm: Wave forecast overlay integration (WIP)

lordfolken:
## Summary

Draft PR for XCTherm vertical-wind forecast integration on the weather page and map overlay.

- Add XCTherm API client (index discovery, GeoJSON download, caching, auth)
- Add XCTherm weather dialog with layer/time/altitude controls
- Render forecast GeoJSON on the map with XCTherm colour palette
- Page layout option with bottom controls widget and auto-switch by GPS altitude + UTC time
- Weather settings panel: credentials, region, enable/auto-switch toggles, overlay position

## Status

Work in progress — commit history and UI still need cleanup before review.

## Test plan

- [ ] Configure XCTherm credentials in Weather settings
- [ ] Open Weather → XCTherm tab; verify index load and layer list
- [ ] Download/select a forecast layer and confirm map overlay renders
- [ ] Enable auto-switch; verify altitude/time selection follows flight state
- [ ] Build `TARGET=UNIX` and `TARGET=ANDROID` without warnings

## Summary by CodeRabbit

- New Features
  - Added XC Therm weather tab with authenticated forecast downloads, caching, and GeoJSON overlay rendering.
  - New optional bottom-panel controls for XC Therm with manual steppers and auto switching by altitude/time.
  - Weather settings include XC Therm credentials, region, overlay location, and auto-switch.
  - Tap overlays to view detailed values at the current location.
- Improvements
  - Overlay list displays contextual details; page titles reflect XC Therm; startup enables cached forecasts for offline use.
- Tests
  - Added geometry lookup tests for band selection.


src/Dialogs/Settings/Panels/PagesConfigPanel.cpp
```diff
@@ -388,6 +389,10 @@ PageListWidget::OnPaintItem(Canvas &canvas, const PixelRect rc,
     buffer.AppendFormat(", %s", _("Cross section"));
     break;
 
+  case PageLayout::Bottom::XCTHERM:
+    buffer.AppendFormat(", %s", "XCTherm");
+    break;
```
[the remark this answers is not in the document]
lordfolken: There is an exception and that is brand names XCTherm is a brand name and needs to be added to the cursor rules.

src/Dialogs/Weather/XCThermDialog.cpp
```diff
@@ -0,0 +1,1226 @@
...
+  { "Vertical wind 1500 m AMSL", "1500amsl", 1500, false },
+  { "Vertical wind 2000 m AMSL", "2000amsl", 2000, false },
+  { "Vertical wind 2500 m AMSL", "2500amsl", 2500, false },
+  { "Vertical wind 3000 m AMSL", "3000amsl", 3000, false },
+  { "Vertical wind 4200 m AMSL", "4200amsl", 4200, false },
+  { "Vertical wind 100 m AGL",   "100agl",   100,  true },
+  { "Vertical wind 200 m AGL",   "200agl",   200,  true },
+  { "Vertical wind 400 m AGL",   "400agl",   400,  true },
+  { "Vertical wind 800 m AGL",   "800agl",   800,  true },
+};
```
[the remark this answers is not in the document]
lordfolken: This here could be avoided, by building a string, and only varring the height. Also i'm a bit unsure what to do with non-metric units.

src/Weather/xctherm/XCThermGeoJSONOverlay.cpp
```diff
@@ -0,0 +1,159 @@
...
+  if (mid <= -2.0)  return Color(0x00, 0xD0, 0xFF);  // bright cyan
+  if (mid <= -1.0)  return Color(0x60, 0xE0, 0xFF);  // sky blue
+  if (mid <= -0.5)  return Color(0xA0, 0xE8, 0xFF);  // light blue
+  if (mid <= -0.2)  return Color(0xC8, 0xF0, 0xFF);  // pale blue
+  if (mid <= +0.2)  return Color(0xF0, 0xF0, 0xC0);  // cream/beige
+  if (mid <= +0.5)  return Color(0xFF, 0xFF, 0x00);  // yellow
+  if (mid <= +1.0)  return Color(0xFF, 0xD0, 0x00);  // gold/amber
+  if (mid <= +2.0)  return Color(0xFF, 0xA0, 0x00);  // orange
+  if (mid <= +3.0)  return Color(0xFF, 0x40, 0x00);  // red-orange
+  if (mid <= +4.0)  return Color(0xFF, 0x00, 0x00);  // red
```
lordfolken, line 173: Colors should be in look/*

src/Weather/xctherm/XCThermGeoJSONOverlay.cpp
```diff
@@ -0,0 +1,159 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The XCSoar Project
+
+#include "XCThermGeoJSONOverlay.hpp"
+
+#include "Projection/WindowProjection.hpp"
+#include "ui/canvas/Canvas.hpp"
+#include "ui/canvas/Color.hpp"
+#include "ui/canvas/opengl/Scope.hpp"
```
lordfolken, line 10: **Build break on Kobo / memory-canvas targets**

This file unconditionally includes `ui/canvas/opengl/Scope.hpp` and uses `ScopeAlphaBlend` in `Draw()`. Kobo sets `OPENGL=n` via `build/fb.mk` (`USE_MEMORY_CANVAS=y`), so this will not compile on e-paper builds even though XCTherm sources are gated behind `HAVE_HTTP` (which Kobo still has).

Other renderers guard the OpenGL path, e.g. `WindArrowRenderer.cpp`:

```cpp
#ifdef ENABLE_OPENGL
#include "ui/canvas/opengl/Scope.hpp"
#endif
```

Please wrap the include and `ScopeAlphaBlend` usage in `#ifdef ENABLE_OPENGL`, and provide a fallback draw path for memory canvas (draw without alpha blending, or skip alpha on non-OpenGL backends). Worth verifying with `TARGET=KOBO` or `TARGET=UNIX OPENGL=n VFB=y`.

src/Dialogs/Weather/XCThermControlsWidget.cpp
```diff
@@ -0,0 +1,856 @@
...
+
+    /* Fetch index.json only if not yet loaded.
+
+       FetchIndex now throws on network / HTTP failure — we MUST catch
+       here because this method is noexcept (the widget is constructed
+       during page setup and an escape would terminate the whole app).
+       On any failure we silently fall into "offline mode" — the user
+       just sees the layer list as unavailable, can still open the
+       dialog where errors are surfaced properly via ShowError. */
+    if (!api.IsIndexLoaded()) {
```
lordfolken, line 520: **UI-thread `FetchIndex()` — NetComponents is probably not the right home**

This synchronous `api.FetchIndex()` during widget creation can block page load on slow/no network. That should be fixed, but moving it to `NetComponents` would likely be the wrong layer unless XCTherm is redesigned as a continuous background service.

`NetComponents` today holds long-lived, timer-driven glue (`TrackingGlue`, `TIM::Glue`) that `ProcessTimer` polls when enabled. TIM fetches periodically in the background via `Co::InjectTask`; the map reads cached data later.

XCTherm's index fetch is different: a small on-demand catalog load tied to `XCThermAPI::Instance()` (auth, RAM/disk cache already live there), not ongoing polling. The download worker is already separate.

Suggested direction for this PR:

1. **Simplest:** drop `FetchIndex()` from the controls widget entirely — treat it as cache/index consumer only. The Weather dialog already fetches the index before download; if `!IsIndexLoaded()`, show offline/unavailable state until then.
2. **If the bottom widget must work without opening the Weather tab:** add async fetch on `XCThermAPI` (`Co::InjectTask` + UI notification via `DeferEvent`), not a new NetComponents member.
3. **NetComponents only if** you want TIM-like behaviour: periodic index refresh from `ProcessTimer`, an explicit enable flag, and app-wide catalog freshness without user opening Weather. That's a larger design choice than needed to fix the UI freeze here.

So: fix the blocking call, but keep orchestration in `XCThermAPI`/dialog rather than NetComponents unless the feature scope expands.

lordfolken:
https://github.com/XCSoar/XCSoar/pull/2607

PR #2562, opened 2026-06-01
Device/Driver/LX: High-rate vario, filter time sync, and filtered display outputs

lordfolken:
## Summary

- Fix LX `$LXWP0` parsing: read all six TE vario samples through the shared 5th-order FIR filter (same as LX EOS), fixing misaligned heading/wind fields.
- Prefer `$PLXVF` for high-rate total-energy vario on sVario devices; request it at 10 Hz via NMEARATE so 1 Hz `$LXWP0` cannot overwrite the fast path.
- Parse `$LXWP3` variofil into device settings and apply a wall-clock output filter so gauge, map vario bar, thermal rose, audio vario, and snail-trail colouring match the vario’s configured damping.

## Test plan

- [ ] Connect an LX sVario (or replay LX NMEA): confirm `$PLXVF` at 10 Hz and `$LXWP3` variofil parsed (check debug log / device settings).
- [ ] Verify vario gauge and audio respond smoothly without 10 Hz UI flicker; map vario bar on e-paper updates ~1 Hz when filter active.
- [ ] Confirm thermal detection/averages still use raw sensor vario, not filtered output.
- [ ] Run `make -j$(nproc) TARGET=UNIX USE_CCACHE=y check` (TestDriver LX cases updated).
- [ ] Test with non-LX vario: filtered fields passthrough when variofil unavailable.

## Summary by CodeRabbit

* **New Features**
  * LXNAV devices now apply their vario filter time to gauge output, vario bar, thermal rose, audio vario, and snail-trail coloring.
  * Added support for configuring the vario filter period via the LXWP3 variofil parameter.
  * Vario displays and audio now use filtered vario values for smoother, less noisy presentation.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -11,6 +11,8 @@ Version 7.45 - not yet released
   - FLARM $PGRMZ (when FLARM is detected) feeds igc_pressure_altitude for the
     ISA logger datum; cleared when strong pressure replaces weak pressure or
     ignores the sentence
+  - LXNAV: apply device vario filter time ($LXWP3 variofil) to gauge, map
+    vario bar, thermal rose, audio vario and snail-trail colouring
 * ui
   - macOS: thick solid lines now render at their correct pixel width #2545
   - status dialog, Times: flight time matches takeoff and landing after
--- a/build/libairspace.mk
+++ b/build/libairspace.mk
@@ -2,6 +2,7 @@ AIRSPACE_SRC_DIR = $(SRC)/Engine/Airspace
 
 AIRSPACE_SOURCES = \
 	$(ENGINE_SRC_DIR)/Util/AircraftStateFilter.cpp \
+	$(ENGINE_SRC_DIR)/Util/VarioOutputFilter.cpp \
 	$(AIRSPACE_SRC_DIR)/AirspacesTerrain.cpp \
 	$(AIRSPACE_SRC_DIR)/Airspace.cpp \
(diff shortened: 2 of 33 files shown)


lordfolken:
Thanks for the review. Addressed the two nitpicks in the branch (line wrap in `VarioOutputFilter.cpp`, drop redundant `const` on `provide_vario`).

The five **Major** availability-flag comments are not applicable as written:

`MoreData::FilteredBruttoVario()` and `FilteredNettoVario()` already guard internally — when `filtered_*_vario_available` is false they passthrough to raw `brutto_vario` / `netto_vario`:

```cpp
constexpr double FilteredBruttoVario() const noexcept {
  return filtered_brutto_vario_available
    ? filtered_brutto_vario
    : brutto_vario;
}
```

Adding explicit checks at every call site that fall back to `0.` would break the intended passthrough when the LX filter is inactive (`variofil` ≤ 0 or unavailable).

`GaugeVario.cpp` intentionally checks `brutto_vario_available` (not `filtered_brutto_vario_available`) so the needle shows zero when the sensor is unavailable; using only the filtered flag would still call the accessor and return a stale `brutto_vario` field.

CI failure was unrelated: `DEBUG_REPLAY_SOURCES` compiled `BasicComputer.cpp` without `FilteredVarioComputer.cpp` / `VarioOutputFilter.cpp`, causing linker errors in `everything` builds (PC/WIN64/sanitizer). Fixed in `build/test.mk`.

PR #2570, opened 2026-06-02
Form: Add toggle switch button for boolean values

WolfgangDrescher:
See https://github.com/XCSoar/XCSoar/discussions/2553#discussioncomment-17147508

Bildschirmfoto 2026-06-02 um 16 21 31

I had to come up with a solution to keep the help text for those form fields. For now I added an `i` icon; when you click it, it opens a `HelpDialog`. Should we add a key binding to enable accessing the help text with the keyboard or control stick?

We could consider using `RichTextWidget` in `HelpDialog` instead of `LargeTextWidget`.

### DIFF
--- a/build/form.mk
+++ b/build/form.mk
@@ -31,6 +31,6 @@ FORM_SOURCES = \
 	$(FORM_SRC_DIR)/GridView.cpp \
 	$(FORM_SRC_DIR)/HLine.cpp
 
-FORM_DEPENDS = SCREEN TIME GEO
+FORM_DEPENDS = SCREEN TIME GEO DATA_FIELD
 
 $(eval $(call link-library,form,FORM))
--- a/src/Form/Edit.cpp
+++ b/src/Form/Edit.cpp
@@ -4,6 +4,7 @@
 #include "Form/Edit.hpp"
 #include "Look/DialogLook.hpp"
 #include "DataField/Base.hpp"
+#include "DataField/Boolean.hpp"
 #include "ui/canvas/Canvas.hpp"
 #include "ui/canvas/Features.hpp"
 #include "Screen/Layout.hpp"
@@ -147,13 +148,20 @@ WndProperty::BeginEditing() noexcept
     
     OnHelp();
     return false;
-  } else {
-    if (!edit_callback(GetCaption(), *data_field, GetHelpText()))
-      return false;
+  }
 
+  if (data_field->GetType() == DataField::Type::BOOLEAN) {
+    auto &df = static_cast<DataFieldBoolean &>(*data_field);
+    df.ModifyValue(!df.GetValue());
     RefreshDisplay();
     return true;
   }
+
+  if (!edit_callback(GetCaption(), *data_field, GetHelpText()))
+    return false;
+
+  RefreshDisplay();
+  return true;
 }
 
 void
@@ -214,9 +222,29 @@ WndProperty::OnResize(PixelSize new_size) noexcept
   UpdateLayout();
 }
 
+static PixelRect
+GetHelpIconRect(const PixelRect &edit_rc) noexcept
+{
+  const int h = edit_rc.GetHeight() * 7 / 10;
+  const int left = edit_rc.left + Layout::GetTextPadding();
+  const int top = edit_rc.top + (edit_rc.GetHeight() - h) / 2;
+  return {left, top, left + h, top + h};
+}
+
 bool
 WndProperty::OnMouseDown([[maybe_unused]] PixelPoint p) noexcept
 {
+  const bool is_boolean = data_field != nullptr &&
+    data_field->GetType() == DataField::Type::BOOLEAN;
+
+  if (is_boolean && HasHelp() && GetHelpIconRect(edit_rc).Contains(p)) {
+    help_icon_pressed = true;
+    pressed = true;
+    Invalidate();
+    SetCapture();
+    return true;
+  }
+
   if (!IsReadOnly() || HasHelp()) {
     dragging = true;
     pressed = true;
@@ -231,6 +259,18 @@ WndProperty::OnMouseDown([[maybe_unused]] PixelPoint p) noexcept
 bool
 WndProperty::OnMouseUp([[maybe_unused]] PixelPoint p) noexcept
 {
+  if (help_icon_pressed) {
+    help_icon_pressed = false;
+    ReleaseCapture();
+    if (pressed) {
+      pressed = false;
+      Invalidate();
+      if (GetHelpIconRect(edit_rc).Contains(p))
+        OnHelp();
+    }
+    return true;
+  }
+
   if (dragging) {
     dragging = false;
     ReleaseCapture();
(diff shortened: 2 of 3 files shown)


Scumi:
How well readable is this on monochrome devices (Kobo)? Shall I test? 
Also (as you said) the helptext also needs to be accessible via remote stick / keyboard.

WolfgangDrescher:
> How well readable is this on monochrome devices (Kobo)? Shall I test?

That would be great as I do not have access to such a device. I added the On/Off labels for this reason.

>Also (as you said) the helptext also needs to be accessible via remote stick / keyboard.

I'll have a look at it. What key binding should I use for this?

lordfolken:
Thank you for attempting to implement. :)

* full row should be selectable and also it should be clear that the full row can be selected
* contrast is paramount: dark/white or even black/white

Always think of the Paragliding pilot with gloves/frozen fingers in a rough thermal and the phone/device is lying on the harness in front of him. 

For the help text: 
Technically a joystick has 4 way controls enter,esc and two keys. These are by default assigned to Zoom in/out (F2/F3) in for example the pan window. Technically one could select the help with that.. but that is hard to communicate. 

Another option would be long vs short press. With long press open the list and show help text, short press toggle. On joysticks the long press is however a different key than the short press. 

The helptext is also the main reason i haven't changed the ui element yet.

Scumi:
I would say not sufficient yet on monochrome. Remove the colors maybe? Squared toggles would fit more with the other styles. I like it in general.

IMG_20260602_193842

WolfgangDrescher:
I updated the colors. Here are screenshots for normal, greyscale, dithered (I overwrote `HasColors` and `IsDithered` in `src/Asset.hpp`, untested on a real Kobo or similar).

Bildschirmfoto 2026-06-02 um 22 01 46
Bildschirmfoto 2026-06-02 um 22 00 32
Bildschirmfoto 2026-06-02 um 21 57 34

What do you think of the `COLOR_XCSOAR` blue for the active state vs. the green I had before?

Scumi:
Will try tomorrow on Kobo. I like the "XCSoar Blue".

fraktal-sb:
tested on UNIX/Android/KOBO; sometimes "I" button is hidden

s1

Screenshot_20260603-111921

WolfgangDrescher:
Thank you for testing this. What resolution and DPI are you using so I can better simulate it on my machine and don't have to involve you in every change I make?

fraktal-sb:
KOBO GLO HD 1448x1072 pixels, or 300ppi; XCSoar Setup Display Resolution "automatic"
Google Pixel 2 1920×1080 pixels, or 441 ppi; XCSoar Setup Display Resolution "automatic"

WolfgangDrescher:
I cannot really reproduce these resolutions and DPI. Does this have to do with my retina screen? I have a lot more pixels on the width with these settings than in your screenshot.

fraktal-sb:
The display resolution  of the Pixel-2 (1080 x1920) is from the technical data; however the screenshot has only 864 x1536 pixels.
The KOBO screenshot has exactly the same pixel size as from the technical data (1072 × 1448 pixels). When I play around with the display resolution in XCSoar and enter a lower display resolution value everything is OK (Symbols and Buttons are much smaller) But then I have to increase the text sizes. I think this is a general problem.

fraktal-sb:
In addition I have set the "Textsize" to 130% and the "Size of landables" to 130%. As I read in the helptext "automatc" adapts to those sizes.

lordfolken:
You need to VptScale() DrawInfoIcon().
The help icon and toggle are both placed in edit_rc, but nothing reserves space between them.

xcsoar can be launched with --dpi=100x100 to simulate dpi. 
also -heightxwidth will do custom resolutions.

WolfgangDrescher:
I'm a bit stuck here. Neither can I reproduce the screen resolution and dimensions of @fraktal-sb, nor can I improve the styles with `VptScale`.

How should the info icon and the toggle switch behave when they overlap? Should they both shrink? Or should the info icon move to the left?

> Technically a joystick has 4 way controls enter,esc and two keys. These are by default assigned to Zoom in/out (F2/F3) in for example the pan window. Technically one could select the help with that.. but that is hard to communicate.
> 
> Another option would be long vs short press. With long press open the list and show help text, short press toggle. On joysticks the long press is however a different key than the short press.
> 
> The helptext is also the main reason i haven't changed the ui element yet.

This is also still unresolved, and I'm genuinely unsure what the best approach is. The cleanest solution would be to make the info icon focusable, but that would increase the number of button presses needed to navigate the menu with a control stick. Does anyone have a preference here? Maybe it's not that important after all. Once you know what a setting does, you'll probably never go back to the help text. And if we had better documentation, you could just look it up there.

fraktal-sb:
Don't worry Wolfgang, I repeated today my tests and now I could no more reproduce it at least at the KOBO's. On Android I think the problem is to find on an other place. Here the comparision between UNIX and Android. The reserved space for one line with text and graphics changes (why?)

lordfolken:
hastouchscreen() sizes the ui elements up to make it easier to use with gloves. 


```c++
 minimum_control_height = std::min(FontScale(23),
                                    min_screen_pixels / 12);
  if (HasTouchScreen()) {
    /* larger rows for touch screens */
    maximum_control_height = PtScale(30);
    if (maximum_control_height < minimum_control_height)
      maximum_control_height = minimum_control_height;
  } else {
    maximum_control_height = minimum_control_height;
  }
  hit_radius = PtScale(HasTouchScreen() ? 28 : 6);
```

fraktal-sb:
The usability does not become better when we size up only a part of the masks (those with less information) the others still remain less usable with the smaller buttons. The "Automatic resolution" setup in look/language causes those effects. So which is the recommendation for the user? Or is there an other solution by software?

WolfgangDrescher:
https://github.com/user-attachments/assets/b919764a-b4ff-4eee-9bd2-678e286f3bd3

Have a look at this video. There is something weird going on when you rescale the window.

1. Initially the toggle switch and info icon look okay (knob is a bit small)
2. After I shrink the width of the window I can reproduce a similar situation to your previous screenshots @fraktal-sb
3. Refreshing the view with a gesture did not change the situation after the flicker
4. Go to the next page and back to `Gauges > Vario`: same small window width but the toggle switch and the info icon are displayed and spaced correctly now (knob is way too small)
5. Make the window width bigger again: the toggle switch disappears
6. After refreshing the view with a gesture the toggle switch is displayed again
7. No changes when switching to the next page and back to `Gauges > Vario`

So what I think is going on is that on the `Gauges > Vario` page there are no other elements that reserve some width for the form column of the form views. Can we set a minimum width that should be reserved for the whole toggle switch including the info icon?

Although this hypothesis may not hold, since in one of the previous screenshots from @fraktal-sb the overlap between the toggle switch and the info icon also occurs on other pages where other form elements are present.



But I cannot understand why the new form element is rendered differently when you navigate to it by accessing the next page than when you resize the window.

PR #2574, opened 2026-06-03
build/libdbus: Fix non-OpenGL UNIX builds

groundhogxc:
For the Wifi control that was recently introduced, libdbus is used. libdbus was gated on USE POLLING, which in turn depends on OpenGL=y, making the non-OpenGL UNIX builds fail (except KOBO, which has special handling).

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration to properly support Android devices with libdbus library configuration.


lordfolken (approved):

groundhogxc:
Actually, one test fails, but on first glance this does not seem to be related to the PR itself, master itself seems to fail on this, too.

lordfolken:
The test is another problem. Lets get this in.

PR #2528, opened 2026-05-24
Sort FileList Conditionally

cjwoolley:
Thank you for contributing to XCSoar!

We truly appreciate your time and effort in making XCSoar better. We know
that contributing to an open-source project can be challenging, and we're
grateful that you've chosen to help.

This checklist is here to help guide you through the submission process.
Don't worry if you're not sure about something—feel free to ask questions
or submit your PR, and we'll work together to get it ready.

For coding, style guide, architecture information please see our
[development guide](https://xcsoar.readthedocs.io/en/latest/index.html).

For Git tips and tricks, including interactive rebase, fixup commits, and
common workflows, see the
[Git Tips](https://xcsoar.readthedocs.io/en/latest/git_tips.html)
documentation.

For testing and debugging utilities, see the
[Test & Debug
Utilities](https://xcsoar.readthedocs.io/en/latest/test_debug_utilities.html)
documentation.

## Summary

There are many menu selections that allow file selection based on ascending order.

This works well except for IGC / NMEA log files where descending order is preferred i.e. latest log first and oldest last.

This change conditionally sorts files to cater for:
- Config->Weglide Upload->IGC-FilePicker — log files displayed in descending order
- Config->Replay->File — blank first (to run the demo) followed by log files displayed in descending order
- All other files are displayed in ascending order

## Pre-Submission Checklist

Please verify the following before submitting your PR:

#### Branch & Rebase
- [ ] PR is rebased on current `master`
- [ ] Use `git rebase -i` to clean up commit history before PR
  submission

#### Commit Format & Messages
- [ ] All commits follow the format: `<Component>: <Summary>` (no
  `src/` prefix)
- [ ] Use present tense in commit messages ("Fix" not "Fixed", "Add"
  not "Added")
- [ ] Commit messages explain *why* the change was made, not just
  *what* changed
- [ ] Commit message body (if needed) provides detailed reasoning and
  context

#### Commit Structure
- [ ] Each commit is atomic and builds successfully (every commit
  must compile)
- [ ] One commit per logical change (don't mix refactoring with
  feature changes)
- [ ] Self-contained commits (each commit changes one thing)
- [ ] No fixup commits (squashed into parent commits using
  `git rebase -i`)
- [ ] No duplicate commits (check with `git log --oneline`)
- [ ] No "WIP" or "testing" commits (clean up before PR)

---

- [ ] I'm ready to merge

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed file list sorting behavior across different file types for consistent organization
[...]

PR #2578, opened 2026-06-03
Form/DataField/File: Sort IGC and NMEA lists in descending order

cjwoolley:
Follow-up to closed #2528.

Log file pickers (replay, WeGlide IGC upload) should list newest
filenames first. Other file lists stay in ascending order.

Sorting uses **filename collation**, not file modification time, so it
works best for date-prefixed log names (typical IGC/NMEA layout).

## Changes

- `FileDataField::Sort(order, preserve_first)` with deferred sort support
- Replay: blank demo row preserved, remaining logs descending
- WeGlide IGC picker: descending order
- `RowFormWidget::GetFileDataField()` helper (used by replay dialog)

## Test plan

- [ ] Config → Replay: blank first, then logs newest-first
- [ ] Config → WeGlide upload → IGC file picker: newest-first
- [ ] Other file pickers (waypoints, airspaces, …): still ascending

### DIFF
--- a/src/Form/DataField/File.cpp
+++ b/src/Form/DataField/File.cpp
@@ -77,7 +77,7 @@ FileDataField::FileDataField(DataFieldListener *listener) noexcept
    // Set selection to zero
    current_index(0),
    loaded(false), postponed_sort(SortOrder::NO_ORDER),
-   postponed_value(nullptr)
+   postponed_preserve_first(false), postponed_value(nullptr)
 {
   file_types.append() = FileType::UNKNOWN;
 }
@@ -182,7 +182,8 @@ FileDataField::ScanDirectoryTop(const char *filter) noexcept
     VisitDataFiles(filter, fv);
   }
 
-  Sort();
+  if (postponed_sort == SortOrder::NO_ORDER)
+    Sort();
 }
 
 void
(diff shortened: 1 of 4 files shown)


src/Form/DataField/File.cpp
```diff
@@ -324,20 +324,26 @@ FileDataField::Sort(SortOrder order) noexcept
     postponed_sort = order;
     return;
   }
-
-  if (order == SortOrder::ASCENDING) {
-    // Sort the filelist in ascending order
-    std::sort(files.begin(), files.end(), [](const Item &a,
+  if (files.size() > 1) {
+    if (GetFileType() == FileType::REPLAY) {
+      // Replay if first record blank then remainder sorted descending
```
Scumi: This pushes UI policy into this generic data field layer, while the handling should happen in the caller.

src/Repository/FileType.hpp
```diff
@@ -22,6 +22,7 @@ enum class FileType : uint8_t {
   PLANE,
   XCI,
   LUA,
+  REPLAY,
```
Scumi: Replay is not a special file type. Replays are NMEA or IGC files. I suppose this just exists to target the sorting behaviour.

Scumi (changes requested):
Looking at this, I think the real gap is API-level: we currently have no clean way to keep one entry pinned at the top while sorting the rest of the list (or adding the null-item after creating and sorting the list).

A small and generic quick fix would be to extend Sort() with an option to preserve the first item:

```
/**
  * Sorts the filelist by filenames, optionally leaving the first
  * entry untouched.
 */
void Sort(SortOrder order = SortOrder::ASCENDING,
          bool preserve_first = false) noexcept;
```

Then the replay caller can do:
`file_df.Sort(FileDataField::SortOrder::DESCENDING, true);`

This keeps the change small, avoids replay-specific logic in `File.cpp`, and avoids adding targeted kludges in individual callers.

cjwoolley:
> Looking at this, I think the real gap is API-level: we currently have no clean way to keep one entry pinned at the top while sorting the rest of the list (or adding the null-item after creating and sorting the list).
> 
> A small and generic quick fix would be to extend Sort() with an option to preserve the first item:
> 
> ```
> /**
>   * Sorts the filelist by filenames, optionally leaving the first
>   * entry untouched.
>  */
> void Sort(SortOrder order = SortOrder::ASCENDING,
>           bool preserve_first = false) noexcept;
> ```
> 
> Then the replay caller can do: `file_df.Sort(FileDataField::SortOrder::DESCENDING, true);`
> 
> This keeps the change small, avoids replay-specific logic in `File.cpp`, and avoids adding targeted kludges in individual callers.

I've made the coding changes you suggested however I need advice on the best approach for the following issue.

The correct parameters are being passed to the sort on the first pass however the data for the sort was flagged as not loaded at the time.

On the subsequent pass the data was flagged as loaded however the sort parameters were set back to the default setting.

lordfolken:
The sort parameters aren't really "reset" on the second pass — they're **cleared during loading**.

What happens:

1. You call `Sort(DESCENDING, true)` before the list is loaded → only `postponed_sort` is stored.
2. Loading runs in `EnsureLoaded()` → `ScanDirectoryTop()` scans files, then calls `Sort()` with **defaults** (ascending, no `preserve_first`).
3. That default sort sets `postponed_sort = NO_ORDER`.
4. When `EnsureLoaded()` tries to replay the deferred sort, there is nothing left to apply.

Also, `preserve_first` is never stored for deferred sorts — only the `SortOrder` is.

**Best approach:**

1. **Store both deferred values:** add `postponed_preserve_first` and pass it when replaying in `EnsureLoaded()`.
2. **Stop clobbering deferred sort:** at the end of `ScanDirectoryTop()`, only call `Sort()` if `postponed_sort == NO_ORDER`.
3. **In ReplayDialog:** call `Sort(DESCENDING, true)` **after** `LoadValue()` (so the list is loaded), or rely on the deferred-sort fix above if you call it earlier.

**Quick workaround without changing `File.cpp` yet:** call `Sort(DESCENDING, true)` after `LoadValue()` in `ReplayDialog::Prepare()` — by then `loaded` is true and the sort applies directly, bypassing the deferred-sort bug.

cjwoolley:
@coderabbitai resume

cjwoolley:
@coderabbitai review

cjwoolley:
@coderabbitai full review

src/Form/DataField/File.cpp
cjwoolley: Updated as requested. Please review.

cjwoolley:
Rebased against master, resolved conflicts, and force-pushed. The PR is up to date and ready for review.

lordfolken:
Review follow-up (lordfolken): title/body updated; commit rebased on current `master` with:

- `Form/DataField/File: Sort IGC and NMEA lists in descending order` (reworded)
- `RowFormWidget::GetFileDataField()` helper; replay dialog uses it
- PR body notes sorting is by **filename**, not mtime

Updated branch: [`XCSoar:fix/pr-2578-sort-filelist`](https://github.com/XCSoar/XCSoar/tree/fix/pr-2578-sort-filelist) (could not push to `cjwoolley` fork). @cjwoolley you can reset `Sort_FileList_by_FileType` from that branch, or we can merge from the upstream branch directly.

lordfolken:
Superseded by #2675 (rebased on master with review fixes from #2578).

PR #2585, opened 2026-06-06
GlidePolar: apply air density ratio for altitude-dependent speed-to-fly

### LINKED ISSUE #1569: Make speed to fly altitude dependent
XCSoar version
--------------
7.43
After reading the XCS handbook I believe that the speed to fly is calculated altitude independent. At least the handbook does not mention, that altitude is taken into account.
However, in real life the polar ist slightly shifted due to altitude. See chapter 3 of following document.
https://journals.sfu.ca/ts/index.php/op/article/download/1796/1717

At moderate altitude, this makes little to no difference, however, if flying in wave it starts to become noticable.

Possibly altitude dependency is already in the code. I did not find where XCS does the polar calculations.
If this is already the case than this issue shall be closed.

As XCS has already code to calculate the air density ratio in [AirDensity.cpp](https://github.com/XCSoar/XCSoar/blob/6696635126fc457cf03fe4f6c5eb57f09d7c598f/src/Atmosphere/AirDensity.cpp#L20) it should be possible to take the altitude dependency of the polar into account.

Unfortunately, I did not find where in the code the polar is adjusted for weight of glider, etc. Hints appreciated.

dspreitz:
## Summary

Fixes #1569 -- speed-to-fly and task calculations were computed at sea-level polar values regardless of altitude.

The glide polar is defined in indicated-airspeed (IAS) terms. At altitude, the air density ratio DR = sqrt(rho0/rho) > 1, so optimal TAS speeds are higher and sink rates in m/s also increase -- while best L/D is preserved. XCSoar already had AirDensityRatio() used for IAS<->TAS conversion, but it was never applied to the polar.

**Changes:**

- GlidePolar gains a density_ratio field (default 1.0). SinkRate(V_TAS) now evaluates at the IAS-equivalent V_TAS/DR and scales the result by DR. Vmin, VbestLD, and GetBestGlideRatioSpeed() are all scaled by DR, so MacCready receives correctly elevated TAS speeds while best L/D is preserved.
- GlideComputer::ProcessGPS() calls SetDensityRatio(AirDensityRatio(altitude)) at the start of every GPS cycle, so all sub-computers (task, airdata, warnings) see a corrected polar.
- Update() guards against zero density_ratio that arises when the struct is zero-initialised via DerivedInfo::Reset().
- MacCready.hpp comment updated to remove the stale \\ir density is NOT taken into account\\ note.

**Tests:**

8 new assertions in TestGlidePolar::TestDensityRatio() verify that at DR=1.1: speeds scale by DR, sink rates scale by DR, best L/D is unchanged, and reset restores sea-level values. All 9805 existing tests continue to pass.

## Test plan

- [ ] Build and run \make TARGET=UNIX check\ -- all tests pass
- [ ] Fly a cross-country task at high altitude (>2000m MSL): verify STF increases noticeably compared to sea-level polar
- [ ] Check that best L/D infobox value is unchanged with altitude
- [ ] Verify task ETE/ETA adjusts correctly when altitude changes

## Summary by CodeRabbit

* **New Features**
  * Glide computer now applies air‑density correction to polar calculations, making speed‑to‑fly and task guidance altitude‑aware while preserving best glide ratio.

* **Tests**
  * New unit tests validating the air‑density correction behavior and its effect on computed speeds and sink rates.

* **Documentation**
  * Added a repository development guide and updated release notes describing the altitude‑corrected glide polar.

* **Chores**
  * Added a new contributor entry to the AUTHORS file.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -84,6 +84,7 @@ The following people and organizations have contributed code to XCSoar:
  Yorick Reum <xcsoar@yorickreum.de>
  kobedegeest <kobedegeest@gmail.com>
  B-4pt <bapt80@proton.me>
+ Dominic Spreitz <dominic@spreitz.de>
 
 Documentation:
 
@@ -306,4 +307,3 @@ Libraries and code segments are included from the following projects:
 
  Icons from:
    * oNline Web Fonts http://www.onlinewebfonts.com
- Dominic Spreitz <dominic@spreitz.de>
--- a/AUTHORS
+++ b/AUTHORS
@@ -306,3 +306,4 @@ Libraries and code segments are included from the following projects:
 
  Icons from:
    * oNline Web Fonts http://www.onlinewebfonts.com
+ Dominic Spreitz <dominic@spreitz.de>
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,116 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## What is XCSoar
+
+XCSoar is a tactical glide computer for glider pilots, written in C++20. It runs on Android, iOS, Linux, macOS, Windows, Kobo e-readers, and Raspberry Pi. The core logic is shared; platform-specific code handles rendering, I/O, and sensors.
+
+## Build System
+
+XCSoar uses GNU Make. Build outputs go into `output/<TARGET>/`. The `TARGET` variable selects the platform:
+
+```sh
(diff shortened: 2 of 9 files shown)


lordfolken:
Closing in favor of #2586, which incorporates this glide-polar density-ratio work and extends it (task manager / safety polars, netto vario STF, Condor3 UDP, airspace warning fix).

dspreitz:
@lordfolken Thanks for your work on this and thanks for picking #2585 up in your other PR.

Issue #1396, opened 2024-04-06
Vario Gauge - Graphical Issues


lordfolken:
Original
![image](https://github.com/XCSoar/XCSoar/assets/407371/6adcacd4-c885-4346-ba5a-8f6e8f221c46)

WIP:
![image](https://github.com/XCSoar/XCSoar/assets/407371/4a4d7d4a-45e8-46cd-8238-987affbf3766)

Issues:
 - [x] Labels NetAvg / Gross / MC are maligned e5b2ffc95403bb19d2ba264fcb5e10ce22b69dbf
 - [x] The arc labels overflow the box 964d45e532def1368d67ac31532479c433d47db6
 - [x] The values under NetAvg/Gross/MC do not resize, the labels themselves do. 21fe2d0caa3a9d5e087f3e66d96a6304b08fbbd8
 - [x] The values should be large and the labels small 21fe2d0caa3a9d5e087f3e66d96a6304b08fbbd8
 - [ ] arc is overlapping rc

I think a design along the butterfly vario, would be dooable and look good.

![image](https://github.com/XCSoar/XCSoar/assets/407371/d6721b23-69fc-414c-b5c9-f532057801cc)

DanD222:
I remember seeing the vario scale replaced by vector lines a while ago on the forum, might be worth looking into as well? The vario scale looks a lot better this way IMO.

https://forum.xcsoar.org/viewtopic.php?p=17530#p17530

[0001-Replace-bitmap-vario-scale-by-drawn-lines.patch.txt](https://github.com/XCSoar/XCSoar/files/14896109/0001-Replace-bitmap-vario-scale-by-drawn-lines.patch.txt)

lordfolken:
None of this is a  bitmap. I think that patch was merged long ago.

DanD222:
Here’s the photo from the forum:
![IMG_6491](https://github.com/XCSoar/XCSoar/assets/48011765/ca7960f5-0f7d-4e9e-bf37-96aa2b3f122b)

It doesn’t look very similar to the current implementation to me.

> None of this is a bitmap

You mean the two vario scales in your first post?

lordfolken:
Yes. Neither scala is a bitmap. Its just maths and then rendered. The bitmap scale was removed some time ago.
Ill have a look at the patch? Maybe there is some inspiration there

DanD222:
OK, thanks. Maybe see how it looks without the arc? (that doesn’t look all that great IMO)

[...]

PR #2623, opened 2026-06-16
Gauge/GaugeVario: Modernize scale and vario bar rendering

WolfgangDrescher:
This PR targets the mockup by @elgandoz in https://github.com/XCSoar/XCSoar/issues/1396#issuecomment-2045215892.

This is still very much at an early stage and a work in progress. There is also still quite a bit of LLM-generated code that I have not yet sufficiently reworked. What do you think, is this heading in the right direction? It is not entirely straightforward to implement the mockup as designed, as the Vario Gauge InfoBox can have different sizes and is not guaranteed to have a 1:2 aspect ratio.

Here are some screenshots:

Bildschirmfoto 2026-06-16 um 20 53 26
Bildschirmfoto 2026-06-16 um 20 54 46
Bildschirmfoto 2026-06-16 um 20 55 38
Bildschirmfoto 2026-06-16 um 21 20 02
Bildschirmfoto 2026-06-16 um 20 54 33

As you can see with different screen sizes, it is not so easy to set the perfect circle size for the gauge. How should the scaling be handled?

1. Width is always 100% → height can be larger (or smaller?) than the available space and overflow the InfoBox
2. Height is always 100% → width can be smaller (or larger?) than the available space
3. Always fit the circle so that both width and height fit within the available space
4. Could we use an ellipse to make the gauge always fit within the available space regardless of the InfoBox's aspect ratio?

I still have to check and improve all Vario options available in the settings: speed arrows, avg., MacCready, bugs, ballast, gross, needle, thermal avg. needle.

@elgandoz do you have an idea how we could lay out the avg. needle and the other options from above?

### DIFF
--- a/src/Gauge/GaugeVario.cpp
+++ b/src/Gauge/GaugeVario.cpp
@@ -12,6 +12,8 @@
 #include "lib/fmt/ToBuffer.hxx"
 
 #include <algorithm> // for std::clamp()
+#include <array>
+#include <cstdlib> // for std::abs()
 
 static constexpr double DELTA_V_STEP = 4.0;
 static constexpr double DELTA_V_LIMIT = 16.0;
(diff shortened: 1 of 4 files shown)


lordfolken:
I kind of had 
image
in https://github.com/XCSoar/XCSoar/pull/2559 maybe you can improve it? dark mode looks better.. light mode i didn't like so much. 

image

lordfolken:
And i admit, the geometry is challenging with an arc. 
XCVario: 
image
Freevario: 
image
Butterfly: 
image

I tried to get the arc to level out, but with the current drawing functions this was tricky.

PR #2630, opened 2026-06-18
Display/PageActions: Keep RASP overlay visible in pan mode (fixes #2629)

### LINKED ISSUE #2629: RASP layer disappears in pan mode
Version: Current Master (3fec8bb6d114f064b98112fc5682f64baaf8e3c4)
Tested on: UNIX

XCSoar versions not having the problem:
Bug came in with the EDL branch merging.

Use the same pattern that was implemented for EDL with a special flag that preserves the RASP state

(Feels like there should be a simpler way though, but I am assuming there is a reason it was done this way for EDL).

I have a PR for this.

groundhogxc:
With the introduction of EDL, overlay handling in pan mode was changed. Overlays are removed by default, with EDL being excepted. This provides the same exception for the RASP weather layer, so it stays visible while in panning mode.

Fixes #2629 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RASP weather overlay behavior during map panning: the overlay is now correctly suspended when panning operations begin and automatically restored when panning ends or is cancelled. This prevents the weather layer from interfering with navigation interactions, providing a smoother and more responsive panning experience.

### DIFF
--- a/src/PageActions.cpp
+++ b/src/PageActions.cpp
@@ -69,7 +69,8 @@ void
 PageActions::ClearPageOverlays() noexcept
 {
   WeatherUIState &weather = CommonInterface::SetUIState().weather;
-  weather.map = -1;
+  if (!weather.IsRaspSuspendedForPan())
+    weather.map = -1;
 
 #ifdef HAVE_EDL
   if (!EDL::IsDedicatedPageSuspendedForPan())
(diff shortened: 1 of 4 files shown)


lordfolken (approved):
I had a fix for this in my xctherm branch. But I'm fine with your approach.

PR #2607, opened 2026-06-11
feature/issue 2557 xctherm integration pheinrich

lordfolken:
- **Weather/xctherm: Add JWT authentication and API client**
- **feat: XCTherm wave forecast integration (WIP)**
- **working on overall functionality**
- **cleaned the code - integrated better download logic**
- **better UI**
- **better UI-working api**
- **fixed shown time intervalls AUTO**
- **cleaned up code**
- **touch on polygon to see climb value now works**
- **Dialogs/Weather/XCThermControlsWidget: Fix cursor bar layout on narrow map**
- **Dialogs/Weather/XCThermControlsWidget: Relayout cursor bar on resize**
- **Dialogs/Weather/XCThermControlsWidget: Use dialog look in dark mode**
- **net/client: Add JWT Bearer session, SyncHttp, and XCTherm HTTP**
- **Weather/xctherm: Run span downloads on Net::AsyncTask coroutines**
- **Weather/xctherm/XCThermGeoJSONOverlay: Use FmtBuffer for map popup text**
- **Dialogs/Weather/XCThermControlsWidget: Use SymbolButtonRenderer for steppers**
- **Weather/xctherm: Add shared layer and region catalog**
- **Weather/xctherm/XCThermAPI: Map profile region id to tiles API slug**
- **Weather/xctherm: Add shared map overlay apply helpers**
- **NetComponents: Own XCTherm span download glue lifecycle**
- **Weather/xctherm: Refactor controls and share overlay helpers**
- **Weather/xctherm/XCThermForecastTime: Share cursor-bar time label formatting**
- **PageSettings: Unify XCTherm bottom widget with weather controls**

Thank you for contributing to XCSoar!

We truly appreciate your time and effort in making XCSoar better. We know
that contributing to an open-source project can be challenging, and we're
grateful that you've chosen to help.

This checklist is here to help guide you through the submission process.
Don't worry if you're not sure about something—feel free to ask questions
or submit your PR, and we'll work together to get it ready.

For coding, style guide, architecture information please see our
[development guide](https://xcsoar.readthedocs.io/en/latest/index.html).

For Git tips and tricks, including interactive rebase, fixup commits, and
common workflows, see the
[Git Tips](https://xcsoar.readthedocs.io/en/latest/git_tips.html)
documentation.

For testing and debugging utilities, see the
[Test & Debug
Utilities](https://xcsoar.readthedocs.io/en/latest/test_debug_utilities.html)
documentation. 

## Pre-Submission Checklist

[...]

PR #2643, opened 2026-06-22
Dialogs: Improve empty file lists with download hint

lordfolken:
## Summary

- When no local files exist but repository download is supported, file pickers show "None" / "Press here for more" and tapping the row opens download
- Adds shared `EmptyDownloadList` helper used by single-file, multi-file, and list picker paths
- Updates `FileManager`, `NOAAList`, and related pickers consistently

Split out from #2607 so the XCTherm integration PR stays focused.

## Test plan

- [ ] Open a file picker for a type with no local files but an available repository (e.g. RASP, NOAA)
- [ ] Verify empty list shows "None" / "Press here for more"
- [ ] Tap the row and confirm download/add flow opens
- [ ] Verify non-empty lists and multi-file pickers still behave as before
- [ ] Build: `TARGET=UNIX`, `TARGET=WIN64`, `TARGET=KOBO`

## Summary by CodeRabbit

* **New Features**
  * Added empty-state UI rows to several file/list pickers, showing clear prompts to add or download when no items are available.
  * Improved download option behavior so download is offered only for supported file types.

* **Bug Fixes**
  * Fixed empty-list interactions so the empty-state row remains selectable and activates the correct action (instead of doing nothing).
  * Standardized “extra button” dialog results so downstream flows refresh/handle actions reliably.

* **Chores**
  * Updated the build to include the new empty-state dialog support.

### DIFF
--- a/build/main.mk
+++ b/build/main.mk
@@ -83,6 +83,7 @@ DIALOG_SOURCES = \
 	$(SRC)/Dialogs/DataField.cpp \
 	$(SRC)/Dialogs/ComboPicker.cpp \
 	$(SRC)/Dialogs/FilePicker.cpp \
+	$(SRC)/Dialogs/EmptyDownloadList.cpp \
 	$(SRC)/Dialogs/MultiFilePicker.cpp \
 	$(SRC)/Dialogs/HelpDialog.cpp \
 	$(SRC)/Dialogs/WifiDialog.cpp \
--- a/src/Dialogs/DownloadFilePicker.cpp
+++ b/src/Dialogs/DownloadFilePicker.cpp
@@ -2,6 +2,8 @@
 // Copyright The XCSoar Project
 
 #include "DownloadFilePicker.hpp"
+#include "EmptyDownloadList.hpp"
+#include "Renderer/TextRowRenderer.hpp"
 #include "Error.hpp"
 #include "WidgetDialog.hpp"
 #include "DownloadFileModal.hpp"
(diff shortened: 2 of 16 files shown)


groundhogxc:
Just one thing: I think it should be more explicit that clicking leads to a download option. "Press here for more" is ambiguous (could also mean: press here for more information), and you might not have internet anyway, so the promise of more might be unfulfilled...

And maybe a comment:
This is a very heavy PR for such a relatively small feature. 350+ lines of C++ code added in 22 changed files, just for this list entry plus the click call-back seems enormous. I suspect before AI this would have been much more compact... :) 
Not sure if there is a way to make the tools be more conservative on this. Or whether this is a concern at all.

lordfolken:
There are 4 features in this PR, not just the empty list issue. 

The empty list feature was 13 files changed, 335 insertions(+), 47 deletions(-)
With those ~280 lines we gain:
  * A unifed solution for displaying empty lists in pickers 
  * We need to touch all the pickers ListPicker, MultiFilePicker, FileManager,, DownloadFilePicker
  * Each picker needs a hook.
  * Correct when a refresh of the repository is needed
  * we get a feature for file types whether they are downloadable or not. 
  * keyboard/tab activation

A human wouldn't have thought of all these aspects, resulting in the fragmented UX/UI behavior logic that XCSoar currently is.
  
I trimmed it a bit, but that drops the noaa dialog unification, and the defence against blank list items.
 
I changed the wording as you suggested.

PR #2647, opened 2026-06-22
Weather: Add EDL/RASP map-bottom cursor bars

lordfolken:
## Summary

- Add reusable `CursorBarWidget` map-bottom stepper strip
- Add dedicated EDL and RASP cursor-bar controls (`EdlControlsWidget`, `RaspControlsWidget`) with shared labels and input-event parsing
- Replace monolithic `MapOverlayControlsWidget` with `EdlSettingsWidget` for EDL cache management
- Add background download progress for weather fetches and RASP auto-update glue
- Wire cursor bars to map pages via `PageActions`, including pan suspension and session persistence
- Add `WeatherOverlay` input event for in-flight time/level stepping (EDL/RASP)

Split out from #2607 so XCTherm integration can rebase on this foundation.

## Test plan

- [ ] Configure map page with EDL overlay + Weather controls bottom widget; step forecast time and pressure level
- [ ] Configure map page with RASP overlay + Weather controls; step quarter-hour time
- [ ] Enter pan from EDL/RASP page; verify overlay preserved and cursor bar restores after leave pan
- [ ] Weather dialog → EDL tab shows cache management (EdlSettingsWidget)
- [ ] RASP Info tab Update button triggers configured download
- [ ] `WeatherOverlay` input events (default.xci bindings) adjust active overlay
- [ ] Build: `TARGET=UNIX`, `TARGET=WIN64`, `TARGET=KOBO` (non-OpenGL overlay helpers)

## Summary by CodeRabbit

* **New Features**
  * Added **weather overlay cursor bar** “Forecast Controls” to step **forecast time** and **altitude/level**, including **auto-advance** modes.
  * Added **Weather** mode entry and quick-menu access (**Forecast Controls**).
  * Added **EDL weather settings** tab to manage **downloaded/cached forecast days** (precache/cleanup).
  * Added a **background download progress** indicator for weather downloads.
* **Bug Fixes**
  * Improved behavior when leaving map panning so weather overlays don’t leave the UI stuck.
* **Documentation**
  * Updated input event documentation for the new **weather overlay** controls, modes, and bindings.


groundhogxc:
@lordfolken , some comments (I might have written two of these before, some time ago - sorry, do not remember - currently I have very little free time):
* This obviously is a clear improvement over the controls that are in master!
* A bug: XCSoar freezes completely for malformed RASP archives (e.g. a jp2 file inside the archive does not contain the geo boundaries or is otherwise corrupted). This is an issue that already exists in master, but it is triggered more directly in this branch
* The time cursor is always in 15 minute increments, irrespective of what data is available. It should jump forward/backward to the next available time instead - the next 15 minute dataset might be missing, or the data might have a different interval
* There should be hardening against the case where a timeslot is available for one field, but not for another (with the current data sources I believe this might be impossible, but depending on how data will be downloaded in the future, this is a possibility. So: moving time cursor always just jumps to next available point; field cursor should probably always go to the next field as long as ANY time slot exists, and, if that combination of time/field is not available, adjust the time to the closest available one
* There should be a field "none" to switch off the weather layer when desired
* There is now the confusing situation that there is a quick menu entry "weather" (for moving the weather cursor), as well as the entry info->weather (for the weather config dialog). At least one of them needs renaming...
* RASP still does not show properly without patching, but that is fixed on master I think (I tested the non-rebased original of the branch)

lordfolken:
Thanks a lot for the detailed feedback — it was very helpful. I’ve addressed these points in #2686.

lordfolken:
Closing this in favor of #2686, which contains the updated implementation and follow-up fixes.

PR #2664, opened 2026-06-26
Reduce TrailRenderer Catmull-Rom rendering cost

### LINKED ISSUE #2661: Excessive increase of CPU-load

Builds after May 3, 2026

Builds prior to April 21, 2026

System information
------------------

The problem exists on both the Cubie2 and the Pi 4. It might also occur on a Linux PC, but that doesn't run into CPU bottlenecks so quickly. I don’t know how to track CPU load on other platforms.
On the Pi 4, Trixie is running, and xcsoar is compiled natively. (No OV image!) It makes no difference whether X11 or Wayland is configured (via raspi-config). I’m running xcsoar from the CLI.

I have 2 NMEA logs attached:
short_2026-06-20_11-47.nmea to illustrate the CPU-load while Circling
2026-06-20_11-47.nmea to illustrate CPU-load towards the end of the flight, after 3-4 hours.

Expected behavior
-----------------

Even on a Cubie2 based system the CPU-load hardly ever exceeded 10%. And the dependency on the trail length was not significant.

Actual behavior
---------------

I've noticed that my Cubie2 based OpenVario starts to slow down after 2–3 hours of flight time, responding to mouse and keyboard with a annoying delay.
When circling, CPU usage rises from an initial 9% to 23% after just 12 minutes and to 35% after 22 minutes. While cruising, the increase in CPU usage can also be observed, but this occurs later, after 2–3 hours of flight duration. It reaches 50% and more.

The increased CPU usage immediately returns to normal levels (< 10%) as soon as "Trail length" is set to OFF.
It makes no measurable difference whether the trail length is set to Full, Short, or Long. Only when the trail length is set to Off does the CPU usage drop to < 10%. 
The remaining trail settings: 
Trail drift: Off 
Trail type: Vario 1 
Trail scaled: On

Zoom level during circling: 1.5 km
Zoom level during cruising: 38 km

@Scumi and I are looking at the Commit history. One candidate we are analyzing: 9110a2ea909
But I can't figure out how to resolve the conflicts when reverting that commit.

Scumi:
This reduces the rendering cost by
- requesting less rendering work in total (no sub-division for sharp turns)
- don't build the temporary vector by drawing the spline pieces directly
- use of buffered merge vario knots

In my tests, it still looks nicer than before the introduction of the Catmull-Rom algorithm, but is less expensive. 

Closes #2661

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trail rendering smoothness and consistency, especially at different map scales.
  * Reduced unnecessary smoothing for dot-only trail modes, making those displays more accurate.
  * Updated color blending along trails so interpolation is handled more reliably during rendering.

### DIFF
--- a/src/Renderer/TrailRenderer.cpp
+++ b/src/Renderer/TrailRenderer.cpp
@@ -218,38 +218,6 @@ CatmullRomInterpolate(const PixelPoint &p0, const PixelPoint &p1,
   );
 }
 
-/**
- * Generate interpolated points between two trace points using Catmull-Rom spline.
- * @param p0 Control point before p1
- * @param p1 Start point
- * @param p2 End point
- * @param p3 Control point after p2
- * @param num_segments Number of segments to generate
- * @return Vector of interpolated points (including p1 and p2)
- */
-static std::vector<PixelPoint>
-InterpolateSegment(const PixelPoint &p0, const PixelPoint &p1,
-                   const PixelPoint &p2, const PixelPoint &p3,
-                   unsigned num_segments) noexcept
-{
-  std::vector<PixelPoint> result;
-  result.reserve(num_segments + 1);
-
-  // Always include the start point
-  result.push_back(p1);
-
-  // Generate intermediate points
-  for (unsigned i = 1; i < num_segments; ++i) {
-    const double t = static_cast<double>(i) / num_segments;
-    result.push_back(CatmullRomInterpolate(p0, p1, p2, p3, t));
-  }
-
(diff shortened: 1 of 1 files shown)


lordfolken:
Closing in favour of #2665, which rebases and combines this bounded Catmull-Rom approach with the zoom-adaptive thinning, merge-vario archive, and segment-cache work for #2661.

The hybrid commit credits Stefan's #2664 changes explicitly. Review should continue on #2665.

PR #2665, opened 2026-06-26
Renderer/TrailRenderer: Reduce trail CPU load and improve merge-vario colouring

### LINKED ISSUE #2661: Excessive increase of CPU-load

lordfolken:
Fixes #2661

## Summary

- Reuse draw buffers each frame to avoid per-segment heap allocations during trail rendering
- Thin the GPS trace by map zoom (canvas pixels via `DistancePixelsToMeters`) and deduplicate vertices that map to the same screen pixel
- Disable Catmull-Rom splines above 6 km map scale; cap spline density from on-screen segment length (lower on embedded targets)
- Batch consecutive same-colour polyline pieces into single draw calls
- Replace per-segment binary search over merge-vario samples with a monotonic sliding window
- Fix merge sample collection to half-open `[t0, t1)` so high-rate vario ticks at the leg start GPS second are included
- Colour trail segments from merge-vario samples via segment parameter u (spreading equal timestamps), on both splined and direct paths

## Test plan

- [ ] Replay `short_2026-06-20_11-47.nmea` in simulator: circling at ~1.5 km zoom, trail ON — CPU stays reasonable vs OFF
- [ ] Replay `2026-06-20_11-47.nmea`: cruise at ~38 km zoom — coloured trail visible, no disappearance at wide zoom
- [ ] Verify vario trail shows colour changes within a single trace leg (merge-rate samples between GPS fixes)
- [ ] Confirm altitude trail and dots modes unchanged
- [ ] Build: `TARGET=UNIX WERROR=y`, `TARGET=WIN64 WERROR=y`

## Summary by CodeRabbit

* **New Features**
  * Added a Lua API (`replay.process_all`) to process all pending replay fixes and refresh the map.
  * Added region-based trace point filtering to support rendering/thinning within selected geographic bounds.
* **Bug Fixes**
  * Improved trail coloring by interpolating vario/merged-vario using sampled breakpoints (including the open trail segment).
  * Updated trail thinning/smoothing to be map-scale aware and refined merged-vario handling with better deduplication and time-filtered loading.
* **Tests**
  * Added a callgrind datapath Lua driver to exercise rendering across multiple zoom scales.

### DIFF
--- a/src/CalculationThread.cpp
+++ b/src/CalculationThread.cpp
@@ -112,6 +112,32 @@ CalculationThread::Tick() noexcept
 }
 
 void
+CalculationThread::ProcessReplayFix() noexcept
+{
+#ifdef HAVE_CPU_FREQUENCY
+  const ScopeLockCPU cpu;
+#endif
+
+  {
+    const std::lock_guard lock{device_blackboard.mutex};
+    glide_computer.ReadBlackboard(device_blackboard.Basic());
+  }
+
+  {
+    const std::lock_guard lock{mutex};
+    glide_computer.ReadComputerSettings(settings_computer);
+  }
+
+  glide_computer.Expire();
+  glide_computer.ProcessGPS(true);
+
+  {
+    const std::lock_guard lock{device_blackboard.mutex};
+    device_blackboard.ReadBlackboard(glide_computer.Calculated());
+  }
+}
+
+void
 CalculationThread::ForceTrigger() noexcept
 {
   {
--- a/src/CalculationThread.hpp
+++ b/src/CalculationThread.hpp
@@ -56,6 +56,12 @@ public:
 
   void ForceTrigger() noexcept;
 
+  /**
+   * Run one calculation cycle for bulk replay (no UI triggers).
+   * Call only while the worker thread is suspended.
+   */
+  void ProcessReplayFix() noexcept;
+
 protected:
   void Tick() noexcept override;
 };
--- a/src/Computer/TraceComputer.cpp
+++ b/src/Computer/TraceComputer.cpp
@@ -5,15 +5,52 @@
 #include "Settings.hpp"
 #include "NMEA/MoreData.hpp"
 #include "NMEA/Derived.hpp"
+#include "Geo/GeoBounds.hpp"
 
-static constexpr unsigned full_trace_size = 1024;
+#include <cmath>
+
 static constexpr unsigned contest_trace_size = 256;
 static constexpr unsigned sprint_trace_size = 128;
 
 static constexpr auto full_trace_no_thin_time = std::chrono::minutes{2};
 
+/** Minimum |Δvario| to store another merge-vario sample (m/s). */
+static constexpr float MERGE_VARIO_DEDUPE_EPS = 0.05f;
+
+/** Near-zero band: always keep samples for Nullschieber colouring. */
+static constexpr float MERGE_VARIO_NULL_BAND = 0.05f;
+
+[[gnu::const]]
+static bool
+MergeVarioSampleWorthKeeping(float prev, float next) noexcept
+{
+  if (std::abs(next - prev) >= MERGE_VARIO_DEDUPE_EPS)
+    return true;
+
+  if ((prev > 0.f) != (next > 0.f))
+    return true;
+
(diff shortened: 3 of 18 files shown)


lordfolken:
Updated with a hybrid of this PR and #2664 (Scumi):

- **From #2664:** bounded smoothed tail (last 180 visible points), zoom-scaled spline segment budget, inline `DrawLinePiece` per Catmull-Rom piece (no interpolated vector on the hot path). Authorship preserved via `Co-authored-by: Stefan Schumann` and code comments referencing #2664.
- **From this PR:** `[t0, t1)` merge-vario lookup, u-parameter breakpoints, merge colouring on direct segments, LoadTrace zoom thinning, splines disabled above 6 km, colour-run batching on non-smoothed segments.

Commit: 84e71d0385

lordfolken:
Pushed history cleanup (authorship preserved) plus CI fixes:

- **KOBO compile:** explicit `{x, y}` for ribbon join circles (`BulkPixelPoint` → `PixelPoint` is ambiguous on memory canvas).
- **Dots-only batching:** flush pending colour batch before skipping sink runs so same-colour segments do not connect across suppressed pieces.
- **Earlier fixups** already on branch: open-leg merge-vario lookup, null-band dedup, cache trim helpers.

**CodeRabbit triage (deferred):**
- `Trace::GetPoints` / `PushMergeVarioDeduped` `noexcept` vs allocating paths — matches existing Trace/TraceComputer conventions; no behaviour change intended here.
- Bounds gap re-fill in `Trace.cpp` — valid optimisation idea for a follow-up; current behaviour preserves trail continuity across viewport exits.

Sanitizer + unit tests pass locally (`TARGET=UNIX VFB=y SANITIZE=y everything check`).

lordfolken:
Addresses #2661

lordfolken:
Addressed remaining CodeRabbit findings (amended into `Computer/TraceComputer: Archive merge vario and replay fast path`):

1. Dropped `noexcept` on allocating merge-vario archive helpers (`PushMergeVarioDeduped`, `ArchiveMergeVarioForLegUnlocked`)
2. Bounds `GetPoints` no longer reinserts off-screen gaps between visible runs
3. Dropped `noexcept` on allocating `Trace::GetPoints` thinning / bounds overloads (so existing `bad_alloc` handlers in `TrailRenderer` can recover)
4. Dropped `noexcept` on `Replay::ProcessAllFixes` (may throw via `Update()`)

PR #2680, opened 2026-07-05
io/Sqlite.cpp needs sqlite3

bomilkar:
It may not be obvious because most systems (used to build xcsoar) have sqlite anyway, but some systems don't.
Therefor sqlite3 should be in install-debian-packages.sh .
Perhaps also in install-android-tools.sh and/or install-darwin-packages.sh .

## Summary by CodeRabbit

* **Chores**
  * Added `sqlite3` to the set of packages installed during Debian setup, ensuring it is available in the environment.


lordfolken (approved):
I concur that this is needed. We also saw this on the openvario side.

PR #2697, opened 2026-07-11
Fix reading SeeYou .cup files with Angled ObservationZones

chrgro:
Previously these kinds of tasks where you use a custom angle for the turnpoints where not correctly read by XCSoar. You can create such tasks in SeeYou and export to .cup, but when reading into XCSoar you got plain FAI quadrants instead of the intended circular sector.

---

This is my first contribution to XCSoar, thank you for an excellent piece of software!

There are probably other ways to structure this, happy to hear opinions.

I plan to also make a PR for a UI update to make this adjustable in-app, but wanted to hear thoughts or merge this initial PR first. There is already partial support for circular sectors in the UI, but the angle adjustment only appears if you load a task with a pre-existing angle set on a task point.

Here is an example task information from SeeYou which shows the intended task:
image

This is the .cup file (file format definition in https://github.com/naviter/seeyou_file_formats/blob/main/CUP_file_format.md ) : 
```
name,code,country,lat,lon,elev,style,rwdir,rwlen,rwwidth,freq,desc,userdata,pics
"Start D",STARTD,,6237.714N,00953.831E,812.0m,1,,,,,,,""
"TP1-D",TP1-D,,6234.110N,00946.407E,991.0m,1,,,,,,,""
"TP2-D",TP2-D,,6240.845N,00946.497E,1076.0m,1,,,,,,,""
"TP3-D",TP3-D,,6234.194N,00945.801E,921.0m,1,,,,,,,""
"Finish D",FINISHD,,6237.982N,00953.952E,739.0m,1,,,,,,,""
-----Related Tasks-----
"Task Delta","???","Start D","TP1-D","TP2-D","TP3-D","Finish D","???"
Options,Short=false,MultiStart=false
ObsZone=0,Style=2,R1=2000m,A1=45,Line=1,SpeedStyle=0
ObsZone=1,Style=1,R1=2000m,A1=90
ObsZone=2,Style=1,R1=2000m,A1=90
ObsZone=3,Style=1,R1=2000m,A1=90
ObsZone=4,Style=3,R1=1500m,A1=180,SpeedStyle=2
```

When loading this into current XCSoar, you get this incorrect task:
image

After my PR, you get the indended task:
image
image

---

[...]

## Pre-Submission Checklist

#### Branch & Rebase
- [x] Use `git rebase -i` to clean up commit history before PR
  submission

#### Commit Format & Messages
- [ almost] All commits follow the format: `<Component>: <Summary>` (no
  `src/` prefix)
- [x] Use present tense in commit messages ("Fix" not "Fixed", "Add"
  not "Added")
- [x] Commit messages explain *why* the change was made, not just
  *what* changed
- [x] Commit message body (if needed) provides detailed reasoning and
  context

#### Commit Structure
- [x] Each commit is atomic and builds successfully (every commit
  must compile)
- [x] One commit per logical change (don't mix refactoring with
  feature changes)
- [x] No fixup commits (squashed into parent commits using
  `git rebase -i`)

---

- [x] I'm ready to merge

## Summary by CodeRabbit

* **New Features**
  * Added a factory for creating symmetric circular observation sectors with caller-defined radius and angle.
* **Bug Fixes**
  * Fixed `.cup` task loading so non-standard observation-zone angles are parsed correctly, preventing incorrect FAI quadrant behavior.
  * Improved symmetric racing-task observation-zone generation to honor explicitly provided radii/angles when available.
* **Tests**
  * Expanded `.cup` parsing tests to cover additional angle/radius fields and verify observation-zone shapes and parameters more thoroughly.

### DIFF
--- a/src/Engine/Task/ObservationZones/SymmetricSectorZone.hpp
+++ b/src/Engine/Task/ObservationZones/SymmetricSectorZone.hpp
@@ -58,6 +58,25 @@ public:
   }
 
   /**
+   * An arbitrary angle circular sector centered at the bisector of incoming/outgoing legs
+   * Similar to an FAI quadrant but with configurable radius or other angles than 90 degrees
+   *
+   * @param radius1 The radius of the sector
+   * @param angle1 The angle of the sector, where 180 is a half-circle
+
+   */
+  static auto CreateSymmetricCircularSectorZone(const GeoPoint loc,
+                                  const double radius,
+                                  const Angle angle
+                                  ) noexcept {
+    std::unique_ptr<SymmetricSectorZone> oz(new SymmetricSectorZone(Shape::SYMMETRIC_QUADRANT, true, false, loc,
+                                                                    radius,
+                                                                    angle));
+    oz->UpdateSector();
+    return oz;
+  }
+
+  /**
    * Create a 180 degree sector centered at the inverse of the
    * outgoing leg.
    *
@@ -71,9 +90,9 @@ public:
     return oz;
   }
 
-  /** 
+  /**
    * Accessor for angle of sector (angle between start/end radials)
-   * 
+   *
    * @return Angle (deg) of sector
    */
   constexpr Angle GetSectorAngle() const noexcept {
--- a/src/Task/TaskFileSeeYou.cpp
+++ b/src/Task/TaskFileSeeYou.cpp
@@ -44,7 +44,7 @@ struct SeeYouTaskInformation {
 
 struct SeeYouTurnpointInformation {
   /** CUP file contained info for this OZ */
-  bool valid = false;
+  bool has_oz = false;
 
   enum Style {
     FIXED,
@@ -57,7 +57,13 @@ struct SeeYouTurnpointInformation {
   bool is_line = false;
   bool reduce = false;
 
+  bool radius1_set = false;
+  bool radius2_set = false;
   double radius1 = 500, radius2 = 200, max_altitude = 0;
+
+  bool angle1_set = false;
+  bool angle2_set = false;
+  bool angle12_set = false;
   Angle angle1{}, angle2{}, angle12{};
 };
 
@@ -175,27 +181,35 @@ ParseOptions(SeeYouTaskInformation &task_info, std::string_view src) noexcept
 static void
 ParseOZs(SeeYouTurnpointInformation &tp_info, std::string_view src) noexcept
 {
-  tp_info.valid = true;
+  tp_info.has_oz = true;
 
   for (std::string_view i : IterableSplitString(src, ',')) {
     if (SkipPrefix(i, "Style="sv))
       tp_info.style = ParseStyle(i);
     else if (SkipPrefix(i, "R1="sv)) {
       const double radius = ParseRadius(i);
-      if (radius > 0)
+      if (radius > 0) {
         tp_info.radius1 = radius;
-    } else if (SkipPrefix(i, "A1="sv))
+        tp_info.radius1_set = true;
+      }
+    } else if (SkipPrefix(i, "A1="sv)) {
+      tp_info.angle1_set = true;
       tp_info.angle1 = ParseAngle(i);
-    else if (SkipPrefix(i, "R2="sv)) {
+    } else if (SkipPrefix(i, "R2="sv)) {
       const double radius = ParseRadius(i);
-      if (radius > 0)
+      if (radius > 0) {
         tp_info.radius2 = radius;
+        tp_info.radius2_set = true;
+      }
     }
-    else if (SkipPrefix(i, "A2="sv))
+    else if (SkipPrefix(i, "A2="sv)) {
+      tp_info.angle2_set = true;
       tp_info.angle2 = ParseAngle(i);
-    else if (SkipPrefix(i, "A12="sv))
+    }
+    else if (SkipPrefix(i, "A12="sv)) {
+      tp_info.angle12_set = true;
       tp_info.angle12 = ParseAngle(i);
-    else if (SkipPrefix(i, "MaxAlt="sv))
+    } else if (SkipPrefix(i, "MaxAlt="sv))
       tp_info.max_altitude = ParseMaxAlt(i);
     else if (SkipPrefix(i, "Line="sv))
       tp_info.is_line = i.starts_with('1');
@@ -317,8 +331,9 @@ CreateOZ(const SeeYouTurnpointInformation &turnpoint_infos,
   const bool is_intermediate = (pos > 0) && (pos < (size - 1));
   const Waypoint *wp = &*wps[pos];
 
-  if (!turnpoint_infos.valid)
-    return nullptr;
+  // For turnpoints that don't have an explicit ObsZone line, default to an FAI Quadrant
+  if (!turnpoint_infos.has_oz)
+    return SymmetricSectorZone::CreateFAISectorZone(wp->location, is_intermediate);
 
   if (factType == TaskFactoryType::RACING &&
       is_intermediate && isKeyhole(turnpoint_infos)) {
@@ -350,13 +365,51 @@ CreateOZ(const SeeYouTurnpointInformation &turnpoint_infos,
 
   else if (factType == TaskFactoryType::RACING) {
 
-    // XCSoar does not support fixed sectors for RT
-    if (turnpoint_infos.style == SeeYouTurnpointInformation::FIXED)
+    // XCSoar does not support FIXED sectors for RT, fall back to a basic cylinder
+    if (turnpoint_infos.style == SeeYouTurnpointInformation::FIXED) {
       return std::make_unique<CylinderZone>(wp->location,
                                             turnpoint_infos.radius1);
-    else
+
+    // Handle SYMMETRICAL OZs for Racing tasks
+    } else if (turnpoint_infos.style == SeeYouTurnpointInformation::Style::SYMMETRICAL) {
+
+      /*
+      Compute "effective" radius and angle used for OZ type selection, depending on
+      whether the .cup file explicitly sets these explicitly or not.
+      */
+      const double eff_radius = turnpoint_infos.radius1_set ? turnpoint_infos.radius1 : 3000.0;
+      const Angle eff_angle  = turnpoint_infos.angle1_set ? turnpoint_infos.angle1 * 2: Angle::QuarterCircle();
+
+      /*
+      For FAI observation zones, SeeYou creates .cup files with a radius of 3000m
+      and an angle of 45 degrees. Detect these (with some margin of error, allowing for
+      any radius larger than 3000m as long as the angle is 45 degrees) and create
+      FAI quadrant OZs for them.
+      */
+      if (eff_radius >= 3000.0 &&
+          eff_angle.CompareRoughly(Angle::QuarterCircle(), Angle::Degrees(1))) {
+
+        return SymmetricSectorZone::CreateFAISectorZone(wp->location,
+                                                         is_intermediate);
+      } else {
+        /*
+        In case of a OZ definition with a non-FAI radius or other angle than a quadrant,
+        then specify a generic circular sector zone here instead of an FAI quadrant.
+        Even if the above heuristic for detecting FAI quadrants is not perfect, this remains
+        functionally correct and in line with what is specified in the .cup file.
+        */
+        return SymmetricSectorZone::CreateSymmetricCircularSectorZone(
+          wp->location, eff_radius, eff_angle);
+      }
+    } else {
+      /*
+      Default to FAI Quadrant OZ if we don't have a better option.
+      This will often not be what the task designer intended.
+      */
       return SymmetricSectorZone::CreateFAISectorZone(wp->location,
-                                                      is_intermediate);
+                                                         is_intermediate);
+
+    }
 
   } else if (is_intermediate) { //AAT intermediate point
     assert(wps[pos + 1]);
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -278,6 +278,9 @@ Version 7.45 - not yet released
   - add NOTAM settings and filters (IFR, effective time, max radius, hidden Q-codes)
   - add NOTAM details integration in airspace dialogs
 * data files
+  - .cup tasks: More correctly parse SeeYou .cup task files with observation zones 
+    with uncommon angles, i.e. angles other than 90 degree quadrants or cylinder. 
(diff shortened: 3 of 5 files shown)


src/Engine/Task/ObservationZones/SymmetricSectorZone.hpp
```diff
@@ -57,6 +57,25 @@ class SymmetricSectorZone: public SectorZone
...
+   * @param radius1 The angle of the sector, where 180 is a half-circle
+
+   */
+  static auto CreateSymmetricCircularSectorZone(const GeoPoint loc,
+                                  const double radius1,
+                                  const Angle angle1
+                                  ) noexcept {
+    std::unique_ptr<SymmetricSectorZone> oz(new SymmetricSectorZone(Shape::SECTOR, true, false, loc,
+                                                                    radius1,
+                                                                    angle1));
```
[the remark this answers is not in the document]
chrgro: I think I disagree. This is intentionaly not a quadrant (a quadrant being a 90degree zone), it can be arbitrary angles. Shape::SECTOR *seems* correct, although I am by no means an expert.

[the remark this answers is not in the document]
chrgro: Aha, thanks. Fair, I buy that argument. WIll update.

src/Engine/Task/ObservationZones/SymmetricSectorZone.hpp
```diff
@@ -57,6 +57,25 @@ class SymmetricSectorZone: public SectorZone
...
+   *
+   * @param radius1 The radius of the sector
+   * @param angle1 The angle of the sector, where 180 is a half-circle
+
+   */
+  static auto CreateSymmetricCircularSectorZone(const GeoPoint loc,
+                                  const double radius1,
+                                  const Angle angle1
+                                  ) noexcept {
+    std::unique_ptr<SymmetricSectorZone> oz(new SymmetricSectorZone(Shape::SYMMETRIC_QUADRANT, true, false, loc,
```
chrgro, line 72: Somewhat annoying that this Shape type is misnamed (its not actually neccesarily a quadrant, see the AI discussion). If someone says they're interested in a refactor of the name of this shape type (and corresponding UI updates), then I can tackle that. If not, I'll leave it be

lordfolken (commented):
Thanks for the contribution — and welcome! This is a real bug (custom symmetrical sectors from SeeYou collapsing to FAI quadrants), and the overall approach of routing those through a configurable `SymmetricSectorZone` is the right fix. Nice first PR.

A few notes from looking at the selection logic and the CUP format:

### `A1 * 2` — correct

The published CUP spec ([Naviter CUP format](https://github.com/naviter/seeyou_file_formats/blob/main/CUP_file_format.md)) only says “Angle 1 in degrees” and does not spell out half vs full angle. In practice Naviter treats `A1` as the angle from the orientation axis to each radial (max 180° → full cylinder), which matches your SeeYou screenshot (`A1=90` → half-circle) and how XCSoar already uses `angle1` for AAT sectors and the `A1≈180` cylinder special case. So multiplying by 2 for `SymmetricSectorZone` (which stores the full sector angle) is right.

### Selection heuristic — please tighten a bit

```cpp
(turnpoint_infos.angle1.Degrees() > 0 && turnpoint_infos.angle1.Degrees() != 45) ||
turnpoint_infos.radius1 != DEFAULT_TP_RADIUS1
```

A few edge cases with this condition:

1. **Zero-angle sector if `R1` is set but `A1` is omitted**  
   CUP allows missing fields (“default settings will be used instead”). If only `R1` differs from 500, we still enter the new factory with `angle1 == 0` → a degenerate 0° zone. Please require `angle1 > 0` (and ideally `radius1 > 0`) before creating the custom sector.

2. **Classic FAI CUP files may change shape type**  
   A typical FAI ObsZone like `Style=1,R1=10000m,A1=45` hits the `radius1 != 500` arm and becomes `SYMMETRIC_QUADRANT` instead of `FAI_SECTOR`. Geometry is roughly the same (90° / 10 km), but UI label, rendering, and FAI task validation (`NON_FAI_OZS`) differ. Worth keeping `CreateFAISectorZone` when the values look like a standard FAI quadrant (≈45° half-angle / ≈10 km, or start/finish 1 km).

3. **`500` is the parser default, not an FAI radius**  
   Using `DEFAULT_TP_RADIUS1` as the “custom vs FAI” sentinel cannot distinguish “field omitted” from “user intentionally chose 500 m”. That is a limitation of the CUP format (defaults are unspecified), so a heuristic is fine — just maybe document it, and prefer matching on “looks like FAI” rather than “differs from parser default”.

4. **Exact `!= 45`**  
   Nearby keyhole detection uses `fabs(... - 45) < 2`. A small tolerance here would be more consistent.

Something along these lines would address the above while still fixing your example (`A1=90`, `R1=2000m`):

```cpp
if (style == SYMMETRICAL &&
    turnpoint_infos.angle1.AbsoluteDegrees() > 0 &&
    turnpoint_infos.radius1 > 0) {
  const Angle full = turnpoint_infos.angle1 * 2;
  // Keep FAI_SECTOR when CUP values match a classic FAI quadrant
  if (/* ~90° full angle and ~FAI radius */)
    return SymmetricSectorZone::CreateFAISectorZone(...);
  return SymmetricSectorZone::CreateSymmetricCircularSectorZone(
    wp->location, turnpoint_infos.radius1, full);
}
return SymmetricSectorZone::CreateFAISectorZone(...);
```

### Small nits (non-blocking)

- In the new factory, prefer parameter names `radius` / `angle` over `radius1` / `angle1` (CUP field names do not belong in the Engine layer).
- Commit message typo: “where” → “were”.
- A short `NEWS.txt` entry would help, since this is user-visible on `.cup` import.
- A unit test around `CreateOZ` selection would lock the FAI vs custom cases in.

Again — thanks for digging into this and for the clear before/after screenshots. Happy to re-review once the heuristic is tightened.

src/Task/TaskFileSeeYou.cpp
```diff
@@ -350,13 +352,27 @@ CreateOZ(const SeeYouTurnpointInformation &turnpoint_infos,
...
+    if (turnpoint_infos.style == SeeYouTurnpointInformation::FIXED) {
       return std::make_unique<CylinderZone>(wp->location,
                                             turnpoint_infos.radius1);
-    else
+    } else if (
+      turnpoint_infos.style == SeeYouTurnpointInformation::Style::SYMMETRICAL &&
+      (
+        (turnpoint_infos.angle1.Degrees() > 0 && turnpoint_infos.angle1.Degrees() != 45) ||
+        turnpoint_infos.radius1 != DEFAULT_TP_RADIUS1
+      )
```
lordfolken: Thanks for digging into this — a few edge cases with this condition, if you do not mind tightening it:

1. **Zero-angle sector if `R1` is set but `A1` is omitted** — CUP allows missing fields ("default settings will be used instead"). If only `R1` differs from 500, we still enter the new factory with `angle1 == 0` → a degenerate 0° zone. Please require `angle1 > 0` (and ideally `radius1 > 0`) before creating the custom sector.

2. **Classic FAI CUP files may change shape type** — e.g. `Style=1,R1=10000m,A1=45` hits `radius1 != 500` and becomes `SYMMETRIC_QUADRANT` instead of `FAI_SECTOR`. Geometry is roughly the same, but UI label, rendering, and FAI validation (`NON_FAI_OZS`) differ. Worth keeping `CreateFAISectorZone` when values look like a standard FAI quadrant (≈45° half-angle / ≈10 km).

3. **`500` is the parser default, not an FAI radius** — it cannot distinguish "field omitted" from "user chose 500 m". Prefer matching on "looks like FAI" rather than "differs from parser default".

4. **Exact `!= 45`** — nearby keyhole detection uses `fabs(... - 45) < 2`; a small tolerance here would be more consistent.

Also: `A1 * 2` is correct. The CUP spec only says "Angle 1 in degrees", but Naviter treats `A1` as half-angle from the axis (max 180° → cylinder), which matches your SeeYou screenshot and XCSoar's existing AAT / `A1≈180` cylinder logic.

Suggested shape:

```cpp
if (style == SYMMETRICAL &&
    turnpoint_infos.angle1.AbsoluteDegrees() > 0 &&
    turnpoint_infos.radius1 > 0) {
  const Angle full = turnpoint_infos.angle1 * 2;
  // Keep FAI_SECTOR when CUP values match a classic FAI quadrant
  if (/* ~90° full angle and ~FAI radius */)
    return SymmetricSectorZone::CreateFAISectorZone(...);
  return SymmetricSectorZone::CreateSymmetricCircularSectorZone(
    wp->location, turnpoint_infos.radius1, full);
}
return SymmetricSectorZone::CreateFAISectorZone(...);
```

Non-blocking nits: prefer `radius`/`angle` over `radius1`/`angle1` in the Engine factory; commit message "where" → "were"; a short `NEWS.txt` entry would help.

lordfolken (changes requested):
Thanks again for the contribution — and welcome! The bug is real and routing custom symmetrical sectors through a configurable `SymmetricSectorZone` is the right approach.

Please tighten the selection heuristic before merge (details on the inline comment). Happy to re-review once that is in.

chrgro:
Thanks for the constructive feedback! I adressed hopefully most/all of your comments, and did some assorted refactors after thinking more about the problem:

* I added explicit fields denoting whether an explicit value for R1, A2 etc is present in the .cup file (in the absence of a null value). Let me know if you prefer a different method for this.
* I swapped the logic in the zone type selection a bit: Now I'm instead actively looking to see if I do have something that looks like an FAI Quadrant. If I do, pick that type. Otherwise, pick a more generic SYMMETRIC_QUADRANT (misnamed still, can be non-90 degree). Even though this feels marginally more invasive of a change, it should still be fine to do, as you *do* always get the radius and angle you asked for in the .cup file. And then the detection of whether that constitutes an FAI Quadrant or not is a bit more best-effort. 
* Added unit tests for the functionality (or more specifically, extended an existing test to test a bit deeper).

I intend to do a bit more manual testing to check that I can convert back and forth between the different task formats as you mentioned without losing any detail, but not today :) Theres enough here for now for another review at least.

chrgro:
(Excuse the monologing...) I've found additional issues with the parsing I also would like to address. I've tried to excercise more of the various turnpoint types in SeeYou, and created this task as a result:

image

The exported .cup file for this looks as follow: 
```
name,code,country,lat,lon,elev,style,rwdir,rwlen,rwwidth,freq,desc,userdata,pics
"AF-START",START,,6237.383N,00950.620E,685.0m,1,,,,,,,""
"AF1",CUW_2,,6236.720N,00942.794E,676.0m,1,,,,,,,""
"AF2",AF2,,6239.050N,00941.726E,1472.0m,1,,,,,,,""
"AF3",AF3,,6239.122N,00945.251E,1082.0m,1,,,,,,,""
"AF4",AF4,,6240.587N,00948.022E,867.0m,1,,,,,,,""
"AF-FINISH",AF-FINIS,,6240.649N,00954.476E,570.0m,1,,,,,,,""
-----Related Tasks-----
"Task All Features","???","AF-START","AF1","AF2","AF3","AF4","AF-FINISH","???"
Options,Short=false,MultiStart=false
ObsZone=0,Style=2,R1=3000m,A1=45,Line=1
ObsZone=2,Style=1,R1=2000m,A1=120
ObsZone=3,Style=1,R1=400m,A1=180,SpeedStyle=1
ObsZone=4,Style=1,R1=3000m,A1=20,SpeedStyle=1
ObsZone=5,Style=3,R1=3000m,A1=180
````

Note in particular that it is skipping ObsZone=1, going from 0 to 2. That means it's relying on "default" values for the first turnpoint, which in SeeYou is drawn as an FAI quadrant with 3km radius. In XCSoar on the other hand, the parsing logic puts a cylinder turnpoint here.

File: src/Task/TaskFileSeeYou.cpp [hunk 4]
```diff
@@ -44,7 +44,7 @@ struct SeeYouTaskInformation {
 
 struct SeeYouTurnpointInformation {
   /** CUP file contained info for this OZ */
-  bool valid = false;
```
chrgro, line 47: The terminology "valid" was confusing, actually what this bool indicates is whether there is a related ObsZone line in the .cup file. The turnpoint itself is still valid without this, just that it uses a "default observation zone".

There is no clear definition of what the default should be, but since this is SeeYou's format, and seeyou will default to an FAI Quadrant here, then I've switched to that.

chrgro:
@lordfolken , with [dc4770e] I think this patch is ready for review again. I have fixed additional issues, including .tsk deserialization, so now tasks can translate back and forth correctly. Also fixed my previously mentioned issue when no observation zone type is specified for a turnpoint (default to FAI Quadrant like SeeYou does, instead of defaulting to cylinder).

src/Engine/Task/ObservationZones/SymmetricSectorZone.hpp
lordfolken: I'm all for renaming. Misnomers have been sources of bugs in XCSoar, as they where used differently than intended via abstraction layers. Separate commit please if you do.

src/Task/TaskFileSeeYou.cpp
```diff
@@ -350,13 +365,43 @@ CreateOZ(const SeeYouTurnpointInformation &turnpoint_infos,
...
+
+      // Compute "effective" radius and angle used for OZ type selection, depending on
+      // whether the .cup file explicitly sets these explicitly or not.
+      const double eff_radius = turnpoint_infos.radius1_set ? turnpoint_infos.radius1 : 3000.0;
+      const Angle eff_angle  = turnpoint_infos.angle1_set ? turnpoint_infos.angle1 * 2: Angle::QuarterCircle();
+
+      // For FAI observation zones, SeeYou creates .cup files with a radius of 3000m
+      // and an angle of 45 degrees. Detect these (with some margin of error, allowing for
+      // any radius larger than 3000m as long as the angle is 45 degrees) and create
+      // FAI quadrant OZs for them.
```
lordfolken: Multiline comments with /* */ please.

lordfolken (changes requested):
commit messages should be of the format 
```
<file>: <changes>
<reasoning>

-or-

<system,topic>: <short changes>
<reasoning>
```
please avoid words like fix, or update. Also in the news.txt. 

contentwise it looks good.  Thank you. :-)

chrgro:
Commit message and NEWS.txt and block comments addressed with [46d7aaa], ready for review again.

lordfolken (approved):
Thank you.

PR #2719, opened 2026-07-14
Add STF to audio vario and expose events

Scumi:
- expose vario volume + mute to events
- guard STF by checking for valid polar
- add STF to audio vario and expose settings
- expose STF switch to events

## Summary by CodeRabbit

* **New Features**
  * Added input commands for audio vario volume (mute/unmute, up/down, show) and vario/STF mode switching (auto/manual, vario/stf, toggle, show).
  * Added an audio vario “Mode switching” setting with profile persistence.
* **Bug Fixes**
  * Speed-to-fly and STF-based rendering/displays now properly handle STF availability/missing glide polar.
  * Audio vario output now correctly mutes/suppresses when STF conditions aren’t met.
* **Documentation**
  * Updated the input event list with the new commands.
* **Tests**
  * Added new audio vario and vario synthesiser test coverage and test harness wiring.

### DIFF
--- a/src/Audio/VarioAudioValue.cpp
+++ b/src/Audio/VarioAudioValue.cpp
@@ -1,0 +1,51 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The XCSoar Project
+
+#include "VarioAudioValue.hpp"
+
+#include <algorithm>
+#include <cmath>
+
+namespace AudioVarioGlue {
+
+static constexpr double STF_FULL_SCALE = 16.0;
+static constexpr double VARIO_FULL_SCALE = 5.0;
+static constexpr double STF_SPEED_DEAD_BAND = 2.0;
+
+[[gnu::const]]
+static double
+STFSpeedErrorToPseudoVario(double speed_error) noexcept
+{
+  /* Reuse the existing tone model: sink-like tones command speeding up,
+     climb-like tones command slowing down. */
+  return std::clamp(-speed_error * (VARIO_FULL_SCALE / STF_FULL_SCALE),
+                    -VARIO_FULL_SCALE, VARIO_FULL_SCALE);
+}
+
+std::optional<double>
+ComputeAudioValue(const VarioAudioInput &input,
+                  const VarioSoundSwitchingMode switching_mode,
+                  const VarioSoundManualMode manual_mode) noexcept
+{
+  const bool use_stf = switching_mode == VarioSoundSwitchingMode::AUTO
+    ? !input.circling
+    : manual_mode == VarioSoundManualMode::STF;
+
+  if (!use_stf)
+    return input.vario;
+
+  if (input.stf_speed_error) {
+    if (std::abs(*input.stf_speed_error) <= STF_SPEED_DEAD_BAND)
+      return std::nullopt;
+
+    return STFSpeedErrorToPseudoVario(*input.stf_speed_error);
+  }
+
+  /* Automatic switching keeps the vario useful until STF inputs become
+     available.  Explicit manual STF instead stays silent. */
+  return switching_mode == VarioSoundSwitchingMode::AUTO
+    ? input.vario
+    : std::nullopt;
+}
+
+} // namespace AudioVarioGlue
--- a/src/Audio/VarioAudioValue.cpp
+++ b/src/Audio/VarioAudioValue.cpp
@@ -8,7 +8,6 @@
 
 namespace AudioVarioGlue {
 
-static constexpr double STF_FULL_SCALE = 16.0;
 static constexpr double VARIO_FULL_SCALE = 5.0;
 static constexpr double STF_SPEED_DEAD_BAND = 2.0;
 
@@ -16,9 +15,12 @@ static constexpr double STF_SPEED_DEAD_BAND = 2.0;
 static double
 STFSpeedErrorToPseudoVario(double speed_error) noexcept
 {
-  /* Reuse the existing tone model: sink-like tones command speeding up,
-     climb-like tones command slowing down. */
-  return std::clamp(-speed_error * (VARIO_FULL_SCALE / STF_FULL_SCALE),
+  /* Square-root scaling compresses large speed errors while preserving
+     useful cue separation just outside the quiet band.  Sink-like tones
+     command speeding up; climb-like tones command slowing down. */
+  const double cue = -std::copysign(std::sqrt(std::abs(speed_error)),
+                                    speed_error);
+  return std::clamp(cue,
                     -VARIO_FULL_SCALE, VARIO_FULL_SCALE);
 }
 
--- a/src/Audio/VarioGlue.cpp
+++ b/src/Audio/VarioGlue.cpp
@@ -11,7 +11,6 @@
 #include "SLES/Init.hpp"
 #endif
 
-#include <algorithm>
 #include <atomic>
 #include <cassert>
 
@@ -29,19 +28,6 @@ static std::atomic<VarioSoundSwitchingMode> switching_mode{
 static std::atomic<VarioSoundManualMode> manual_mode{
   VarioSoundManualMode::VARIO
 };
-
-static constexpr double stf_full_scale = 16.0;
-static constexpr double vario_full_scale = 5.0;
-
-[[gnu::const]]
-static double
-STFSpeedErrorToPseudoVario(double speed_error) noexcept
-{
-  /* Reuse the existing tone model: sink-like tones command speeding up,
-     climb-like tones command slowing down. */
-  return std::clamp(-speed_error * (vario_full_scale / stf_full_scale),
-                    -vario_full_scale, vario_full_scale);
-}
 
 bool
 AudioVarioGlue::HaveAudioVario()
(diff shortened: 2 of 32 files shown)


src/Audio/VarioGlue.cpp
```diff
@@ -21,6 +23,23 @@ static bool have_sles;
...
+
+static constexpr double stf_limit = 5.0;
+
+[[gnu::const]]
+static double
+STFSpeedErrorToPseudoVario(double speed_error) noexcept
+{
+  /* Reuse the existing tone model: sink-like tones command speeding up,
+     climb-like tones command slowing down. */
+  return std::clamp(-speed_error, -stf_limit, stf_limit);
```
lordfolken: **STF tone scale vs gauge / V_stf**

This clamps `-(V_stf − TAS)` into ±5 m/s (the synthesiser full scale) **1:1**. The STF arrows in `GaugeVario` use `DELTA_V_LIMIT = 16` m/s and step 4 m/s.

So once speed error is already ≥ ~5 m/s (~18 km/h), raising/lowering MacCready (and thus V_stf) still moves the arrows / Vopt InfoBox, but the tone stays pegged — it feels like the audio is not following V_stf.

Suggestion: scale into the synthesiser range, e.g.

```cpp
static constexpr double stf_full_scale = 16.0; // match GaugeVario DELTA_V_LIMIT
static constexpr double vario_full_scale = 5.0;

return std::clamp(-speed_error * (vario_full_scale / stf_full_scale),
                  -vario_full_scale, vario_full_scale);
```

Polarity (sink tones = speed up, climb tones = slow down) matches the gauge; it is only the scale that diverges.

src/MergeThread.cpp
```diff
@@ -120,12 +119,19 @@ MergeThread::Tick() noexcept
...
-      vario_available = basic.filtered_brutto_vario_available;
-      vario = vario_available ? basic.FilteredBruttoVario() : 0;
+      if (basic.filtered_brutto_vario_available)
+        vario_audio_input.vario = basic.FilteredBruttoVario();
     }
+
+    if (calculated.V_stf_available && basic.airspeed_available &&
+        basic.total_energy_vario_available && calculated.flight.flying)
+      vario_audio_input.stf_speed_error =
+        calculated.V_stf - basic.true_airspeed;
```
lordfolken: **Simulator testing caveat**

`stf_speed_error` is only set when airspeed, TE vario, and `flight.flying` are all valid. The built-in simulator provides none of the first two by default (GPS/GS only), so STF audio never engages in SIM unless a device feed or synthetic ASI/TE is added.

In that case Manual STF → silence, AUTO (cruise) → vario fallback — MacCready will not change the tone. Worth a short note in the PR description or docs for anyone trying to validate this without hardware.

lordfolken (commented):
Tested the STF audio path in the built-in simulator. A few notes from that exercise (local sim/harness hacks aside):

### STF tone scale vs V_stf / gauge arrows
`STFSpeedErrorToPseudoVario()` maps speed error 1:1 into the vario synthesiser (±5 m/s). The STF needles use `DELTA_V_LIMIT = 16` m/s (`GaugeVario`). Once `|V_stf − TAS| ≥ 5` m/s the tone pegs while Vopt / push-pull arrows can still move (e.g. with MacCready). Worth scaling so ±16 m/s speed error → full tone range, matching the gauge.

### Simulator has no ASI / TE by default
Built-in SIM only feeds GPS (location, track, GS, altitude). Without `airspeed_available` and `total_energy_vario_available`, MergeThread never fills `stf_speed_error` → Manual STF is silent, AUTO falls back to vario, so MacCready changes appear to do nothing. Useful to document, or optionally synthesise TAS≈GS (+ level TE/netto) in the simulator for testing.

### Manual mode + non-persisted `manual_mode`
Profile stores `VarioSoundSwitchingMode` (Manual/Auto) but not `manual_mode`. After restart, Manual always starts as Vario — needs `VarioAudioMode stf` / toggle before STF audio is active. Documented in the event help; easy to miss when flight-testing.

src/Audio/VarioGlue.cpp
Scumi: Good point, will have a look at it (or most probably just take your suggestion).

src/MergeThread.cpp
Scumi: True. Can be tested with an nmea-file that has a high vario input frequency. I thought about a testing mode, but will probably just document this for now.

Scumi:
> ### Manual mode + non-persisted `manual_mode`
> 
> Profile stores `VarioSoundSwitchingMode` (Manual/Auto) but not `manual_mode`. After restart, Manual always starts as Vario — needs `VarioAudioMode stf` / toggle before STF audio is active. Documented in the event help; easy to miss when flight-testing.

This is also by intend to not "accidentally" start in STF mode and confuse the pilot that forgets to switch as current status is for now not shown easily. But the documentation should be added, agreed.

src/Audio/VarioGlue.cpp
lordfolken: Thanks — the square-root cue in `VarioAudioValue.cpp` addresses this well. Pegging is no longer at ±5 m/s speed error; |Δv| ≈ 16 m/s (gauge clamp) maps to ~4 m/s on the synthesiser, and full tone scale needs |Δv| = 25. Much more usable with MacCready changes.

The separate fixed `STF_SPEED_DEAD_BAND = 4` m/s (~14 km/h / one gauge step) is quite wide compared with the vario deadband — intentional?

src/MergeThread.cpp
lordfolken: Nice — the Mode switching help text now covers the SIM ASI/TE requirement. Agree documenting (or testing via NMEA replay) is enough for now.

src/Input/InputEventsSettings.cpp
```diff
@@ -150,6 +163,80 @@ InputEvents::eventAudioDeadband(const char *misc)
...
+    StringFormatUnsafe(Temp, "%d", (int)settings.volume);
+    Message::AddMessage(_("Volume"), Temp);
+    return;
+  }
+
+  if (settings.volume > 0)
+    last_unmuted_vario_volume = settings.volume;
+
+  AudioVarioGlue::Configure(settings);
+  Profile::Set(ProfileKeys::SoundVolume, settings.volume);
```
lordfolken, line 206: **Unknown `VarioVolume` args still mutate settings**

Unlike `eventVarioAudioMode` (which returns early in the `else`), an unrecognised/`null` misc still falls through to `Configure()` and `Profile::Set(SoundVolume, …)`.

Suggest mirroring the audio-mode handler:

```cpp
  } else if (StringIsEqual(misc, "show")) {
    …
    return;
  } else {
    return;
  }
```

src/UIReceiveBlackboard.cpp
```diff
@@ -60,13 +60,14 @@ UIReceiveSensorData(OperationEnvironment &env)
...
-   * (if the location is available the CalculationThread will send the
-   * Command::CALCULATED_UPDATE message which will update them)
-   */
-  if (modified || !CommonInterface::Basic().location_available) {
+  /* Most infoboxes are refreshed by calculated updates.  STF may also be
+     refreshed by the merge-rate sensor calculation. */
+  static bool last_stf_available = false;
+  const auto &basic = CommonInterface::Basic();
+  const bool stf_changed = (bool)basic.V_stf_available != last_stf_available;
+  last_stf_available = (bool)basic.V_stf_available;
```
lordfolken: **Availability-only dirty vs merge-rate values**

Casting to `bool` correctly avoids Validity-timestamp thrash (good). But `stf_changed` only fires when availability flips, so while `V_stf_available` stays true the dolphin InfoBox will not refresh at merge rate — only on the next calculated update (or location loss / external settings).

The gauge path reads `GetSTFSpeed()` on paint and is fine. If InfoBoxes are intentionally calc-rate only, maybe soften the comment above ("may also be refreshed") so it doesn’t imply boxes track each merge sample.

src/Audio/VarioAudioValue.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include "VarioAudioValue.hpp"
+
+#include <algorithm>
+#include <cmath>
+
+namespace AudioVarioGlue {
+
+static constexpr double VARIO_FULL_SCALE = 5.0;
+static constexpr double STF_SPEED_DEAD_BAND = 4.0;
```
lordfolken: **Fixed 4 m/s STF quiet band**

`STF_SPEED_DEAD_BAND = 4` matches one `DELTA_V_STEP` on the gauge arrows, which is neat — but ±4 m/s (~±14 km/h) is a wide quiet zone before any tone. Pilots used to a narrow vario deadband may think STF audio is broken when slightly off target.

Not a blocker if deliberate; a short note in NEWS / the Mode switching help (or making it configurable later) would help set expectations.

lordfolken (commented):
## Re-review (after merge-rate STF + √ cue commits)

Re-checked the updated branch in a worktree. CI is green; `TestAudioVario` 19/19 and `TestVarioSynthesiser` 5/5 pass locally.

### Addressed since the earlier review
- **Tone scale vs V_stf / arrows** — square-root cue scaling is a good fix (see reply on the old `VarioGlue` thread).
- **SIM caveat** — documented in the Mode switching setting help.
- **Mode labels** — wrapped with `_()`.
- **Docs order** — `VarioAudioMode` / `VarioVolume` now sit with the `V*` events.
- **Tests** — real coverage for merge-rate STF, mode selection, fixed deadband, polarity, and silence.
- **Merge-rate STF** — `ComputeSTFSpeedError()` + `MoreData::V_stf` + `GetSTFSpeed()` is a clean split; deriving netto from the same TE/ASI sample is the right call.

### Remaining nits

1. **`eventVarioVolume` still applies changes on unknown args** — unlike `eventVarioAudioMode`, a typo/`null` still runs `Configure()` and writes `SoundVolume` to the profile. Easy early-return guard.
2. **InfoBox dirty only tracks availability, not value** — see inline note on `UIReceiveBlackboard`.
3. **Hard-coded STF audio deadband (4 m/s)** — see inline note on `VarioAudioValue.cpp`.

Otherwise this looks ready — happy to approve once (1) is fixed (2–3 can be follow-ups).

Issue #2596, opened 2026-06-10
Settings: InfoBox geometry changes not propagated to Look > InfoBox Sets without exit from Settings


DanD222:
XCSoar versions having the problem
-------------------------------------------------
7.44

Steps to reproduce the behavior
-------------------------------

1. Navigate to Config > System > Look > Screen Layout
2. In InfoBox geometry, change the layout to a different one (e.g. from "4 Top or Left" to "12 Top or Left")
3. Navigate to Look > InfoBox Sets and select one of the layouts, e.g. "Circling", "Cruise", etc.

Expected behavior
-----------------
The infobox layout should be the one I changed to in the previous steps, i.e. "12 Top or Left"

Actual behavior
---------------
The layout is the original one, i.e. "4 Top or Left" 

Leaving the setting to the map and navigating to  Look > InfoBox Sets > Circling / Cruise / etc. now shows the expected layout, i.e. "12 Top or Left"

System information
------------------
moto g05, Android 15

PR #2724, opened 2026-07-15
Infoboxes: Custom layout per set (page) of Infobox

B-apt:
- Introduce the possibility to define the Infobox layout (number & positionning) per layout set instead of a global, unique layout.
- Keep the existing setting as the "default" layout when a new page is added by the user

**Goals:**
- This will give users more flexibility to have some light pages for better readability (for thermal eg) and some other with every bit of informations/stats they need

feature request: https://github.com/XCSoar/XCSoar/issues/1863

image

image

image

## Summary by CodeRabbit

- **New Features**
  - Added per-panel InfoBox geometry configuration (split, side, top/vario variants, and fixed-size layouts).
  - Added a “Default InfoBox geometry” option for the standard panel layout.
  - InfoBox panel geometry is now saved and restored with profiles.

- **Bug Fixes**
  - Layouts now automatically reinitialize when the active panel’s geometry changes.
  - Switching panels reliably applies each panel’s configured geometry, with a default fallback when needed.

- **Tests**
  - Added coverage for profile save/load of per-panel geometry.

### DIFF
--- a/build/test.mk
+++ b/build/test.mk
@@ -336,8 +336,13 @@ $(eval $(call link-program,TestWrapClock,TEST_WRAP_CLOCK))
 
 TEST_PROFILE_SOURCES = \
 	$(SRC)/LocalPath.cpp \
+	$(SRC)/InfoBoxes/InfoBoxSettings.cpp \
+	$(SRC)/PageSettings.cpp \
 	$(SRC)/Profile/Profile.cpp \
+	$(SRC)/Profile/InfoBoxConfig.cpp \
+	$(SRC)/Profile/PageProfile.cpp \
 	$(TEST_SRC_DIR)/tap.c \
+	$(TEST_SRC_DIR)/FakeLanguage.cpp \
 	$(TEST_SRC_DIR)/FakeLogFile.cpp \
 	$(TEST_SRC_DIR)/TestProfile.cpp
 TEST_PROFILE_DEPENDS = PROFILE MATH IO OS UTIL
--- a/src/ActionInterface.cpp
+++ b/src/ActionInterface.cpp
@@ -288,6 +288,16 @@ GetPanelIndex(const UIState &ui_state)
     return InfoBoxSettings::PANEL_CRUISE;
 }
 
+[[gnu::pure]]
+static InfoBoxSettings::Geometry
+GetPanelGeometry(const InfoBoxSettings &settings, unsigned panel_index) noexcept
+{
+  if (panel_index < InfoBoxSettings::MAX_PANELS)
+    return settings.panels[panel_index].geometry;
+
+  return settings.geometry;
+}
+
 static void
 UpdateMapScalePageInfo(UIState &state,
                        const UISettings &settings) noexcept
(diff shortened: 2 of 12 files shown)


B-apt:
@coderabbitai review

WolfgangDrescher:
Have a look at https://github.com/XCSoar/XCSoar/pull/1864.

B-apt:
> Have a look at #1864.

I missed this one... No problem to close my PR if you prefer to rebase & have yours being merged.

DanD222:
I think this is a really useful feature that would be good to have in 7.45 - do you maybe want to co-ordinate as to who moves forward with it? Having tried #1864, it worked well-enough, with two nit-picks:

- Allowing the user to select Portrait layouts in Landscape mode and vice versa (if that looks weird, then as a user I can just select another layout I suppose)

- Profile settings being saved at settings exit, and not on setting a specific setting (#2596), but that's not specific to the feature

I'll try to have a look at this as well.

@WolfgangDrescher  What do you think?

DanD222:
So I've tried this on MacOS, and it looks good (I was able to select different layouts for Circling and Cruise).

The "_Default InfoBox geometry_" parameter in "_Look > Screen Layout_" then becomes somewhat redundant, as changing the default geometry doesn't change the layouts in the remaining InfoBox Sets that don't have non-default layouts (e.g. AUX-1, etc.) and these then need to be changed manually, but it's something I can certainly live with.

I also wouldn't necessarily worry about the Portrait/Landscape thing for the moment and allow the user to select an incorrect InfoBox geometry for his/her orientation, in favour of getting the feature merged into master.

The only other issue was with the path to IOIOLibCore when compiling for Android.

B-apt:
> So I've tried this on MacOS, and it looks good (I was able to select different layouts for Circling and Cruise).
> 
> The "_Default InfoBox geometry_" parameter in "_Look > Screen Layout_" then becomes somewhat redundant, as changing the default geometry doesn't change the layouts in the remaining InfoBox Sets that don't have non-default layouts (e.g. AUX-1, etc.) and these then need to be changed manually, but it's something I can certainly live with.
> 
> I also wouldn't necessarily worry about the Portrait/Landscape thing for the moment and allow the user to select an incorrect InfoBox geometry for his/her orientation, in favour of getting the feature merged into master.
> 
> The only other issue was with the path to IOIOLibCore when compiling for Android.

Thank you for testing this PR and your comments!
I would like to know if @WolfgangDrescher is planning to update/rebase its PR before closing mine. 
Otherwise I can rebase mine on latest HEAD at some point. 

For the 7.45 goal, I think @lordfolken is working on releasing it in a near future, so it's maybe not the perfect time to add a new functionnality. Hopefully we will have a shorter release cycle for 7.46 😉 

By the way, what was the issue with IOIOLibCore, was it preventing the compilation ? How did you resolve it ?

B-apt:
Closing this one in favor of #1864

DanD222:
> By the way, what was the issue with IOIOLibCore, was it preventing the compilation ? How did you resolve it ?

Forgot to reply, sorry - the issue was the path, e.g.
`android/ioio/IOIOLibCore` vs `android/ioio/software/IOIOLibCore`, but I didn't pursue it further.

Issue #2735, opened 2026-07-17
MapWindow: Publish coherent frame snapshot for DrawThread (OPENGL=n)


lordfolken:
## Summary

On `OPENGL=n`, the DrawThread renders from live UI-owned MapWindow state after intentionally releasing `DoubleBufferWindow::mutex`. That avoids blocking the UI during expensive paints, but it leaves projection and related frame-UI fields unprotected.

Issue #2734 covers the concrete startup hang/crash from an uninitialized / stale `screen_bounds`. The quick fix (initialize bounds to `GeoBounds::Invalid()` and skip render when bounds are invalid) addresses that symptom. This issue tracks the underlying design gap: DrawThread still consumes live UI state without a coherent publish/consume boundary.

Related: #2734, PR #2632 (contains the bounds init/guard commit).

### Current non-OpenGL paint path

1. `DrawThread` triggers `GlueMapWindow::Repaint()` → `DoubleBufferWindow::Repaint()`.
2. `Repaint()` holds `mutex`, then `OnPaintBuffer()` **unlocks** it for the duration of `Render()`.
3. At the start of `Render()`, the DrawThread does `render_projection = visible_projection` and continues painting.
4. Meanwhile the UI thread may mutate `visible_projection` and other frame fields (resize, pan, scale, orientation, follow mode, overlays).

`Projection::IsValid()` only checks `geo_location`. Cached `screen_bounds` is updated separately via `UpdateScreenBounds()`. Those are not atomic with respect to the DrawThread copy.

Blackboard settings already use a safer pattern (`next_*` under `next_mutex`, exchanged via `ExchangeBlackboard()`). Projection / frame-UI state do not.

### Shared state of concern

| State | UI writes | DrawThread reads | Impact |
|---|---|---|---|
| `visible_projection` | resize, pan, scale, angle, location | copied into `render_projection` | torn / incoherent geometry |
| `follow_mode` | pan mode | `IsNearSelf()` / `IsPanning()` during render | inconsistent map/pan decisions |
| `compass_visible` | orientation update | compass / scale overlay | visual glitches / UB |
| `bottom_margin` | overlay layout | HUD / map-scale placement | layout glitches / UB |
| `ui_generation` / `buffer_generation` | quick redraw / resize vs after paint | buffer validity checks | stale-buffer bookkeeping races |

[...]

Issue #2734, opened 2026-07-16
Crash / Hang on startup stochastically, particularly for OPENGL=n


groundhogxc:
Version: Current master (but was more likely to occur in #2632 for some reason to do with timing during startup)
System: UNIX non-OpenGL

Sometimes XCSoar would crash / hang on startup with a white map and empty infoboxes, and 100% CPU load in the DrawThread.

I traced it down to corrupted render_projection due to an uninitialized object. It randomly occurs whenever that gets accessed from the drawthread during the startup before the projection has been calculated and set.

There is a quick fix that fixes this specific issue that is part of PR #2632 : Commit 4e57518b728c504f789a14115eb554af01d9410c - "[MapWindow/Renderer: Fix uninitialized projection, and guard against invalid bounds](https://github.com/XCSoar/XCSoar/pull/2632/changes/4e57518b728c504f789a14115eb554af01d9410c)

However, it seems to me that there is a lot more scope for other issues of this kind, for other objects, since the underlying race conditions still exist with other accesses. "visible_projection" has no protection and is accessed from various places in more than one thread.

So the fundamental underlying issue is not fixed I think, just the occurrence is far less likely with my quick fix (and the unitialized bounds is anyway not good style, particularly since a proper invalid bounds value exists).

There might also be other objects that have this issue with the DrawThread, not sure.

groundhogxc:
@coderabbitai Please investigate thoroughly  the underlying issue: Is this indeed a problem beyond the initialization? And are there other objects in similarly endangered states?
If so, how best to fix?

lordfolken:
We dug into this further on master.

**Worst offender for the hang:** `WindowProjection::screen_bounds` was uninitialized, so a DrawThread paint could run with a “valid” location but garbage bounds and wedge the map (white map / empty InfoBoxes / DrawThread pegged). That matches the quick fix already identified in #2632 (`4e57518`).

**For the release:** land that guard (initialize bounds as invalid and skip render until bounds are ready). See #2736.

**After the release:** the broader DrawThread vs UI `visible_projection` race and the remaining publish/consume / related snapshot work tracked in #2735. Those are real follow-ups, but they are not what was needed to stop this particular startup hang.

lordfolken:
Closing: the startup hang is addressed by #2736 (invalid `screen_bounds` default + render guard + coherent projection snapshot for DrawThread).

Remaining DrawThread publish/consume work (follow/compass/margin/generations, etc.) continues in #2735.

PR #2736, opened 2026-07-17
MapWindow: Fix uninitialized projection and publish DrawThread snapshot

lordfolken:
## Summary
- Initialize `WindowProjection::screen_bounds` to invalid and skip map render when bounds are not ready, fixing the intermittent non-OpenGL startup hang/white map (#2734).
- On `OPENGL=n`, publish a coherent `visible_projection` snapshot after `UpdateScreenBounds()` / create / resize; the DrawThread copies it under a short lock at the start of `Render()` instead of reading the live UI projection (#2735, minimal scope).
- OpenGL path unchanged (no DrawThread).

## Out of scope
Follow-mode / compass / bottom-margin / buffer-generation snapshotting and RASP renderer handoff remain for a follow-up (#2735).

## Test plan
- [ ] `make TARGET=UNIX OPENGL=n USE_CCACHE=y WERROR=y`
- [ ] `make TARGET=UNIX USE_CCACHE=y WERROR=y`
- [ ] Start with a configured home location on `OPENGL=n`; confirm no hang and map appears after first full redraw
- [ ] Resize, pan, and zoom while redrawing; confirm map stays coherent
- [ ] Smoke-test default OpenGL UNIX build

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an intermittent startup hang that could display a blank map in non-OpenGL builds.
  - Improved map rendering stability during startup, resizing, and window creation.
  - Prevented invalid screen-bound data from being used before map bounds are initialized.
  - Improved rendering consistency by using a synchronized projection snapshot in non-OpenGL builds.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -193,6 +193,8 @@ Version 7.45 - not yet released
   - snail trail: Catmull-Rom smoothing between fixes; vario colouring samples
     high-rate netto from the merge thread between GPS points #2447
   - Add renderer to display distance rings around the current location #2522
+  - Fix intermittent white-map hang on startup for non-OpenGL builds when
+    drawing began before map bounds were ready #2734
 * ui
   - infoboxen: refresh titles after changing the interface language #2314
   - infoboxen: add "Home" InfoBox (waypoint name, arrival height at home,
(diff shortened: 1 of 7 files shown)


lordfolken:
For the release, the important part of this PR is the **uninitialized `screen_bounds` fix** (invalid init + skip render until bounds are ready). That is the worst offender behind the #2734 startup hang on `OPENGL=n`.

The coherent projection snapshot for DrawThread (and the rest of the #2735 surface — follow-mode / compass / buffer-generation snapshotting, RASP handoff, etc.) we can take **after the release**. Happy to trim this PR to only the hang fix for the release cut if that helps landing.

PR #2751, opened 2026-07-20
ObservationZones/SymmetricSectorZone: Correct incorrect bounding box

chrgro:
For symmetric sector observation zones with angles over 180 degrees, the bounding box would be incorrectly computed, leading to the task advance/turnpoint trigger not firing. Setting a switch enables the arc to be included in the bounding box computation.

The below image shows the kind of OZ that previously would've missed detection. Bug was introduces as part of #2697 

Screenshot 2026-07-20 at 22 39 09

----

#### Branch & Rebase
- [x] Use `git rebase -i` to clean up commit history before PR
  submission

#### Commit Format & Messages
- x ] All commits follow the format: `<Component>: <Summary>` (no
  `src/` prefix)
- [x] Use present tense in commit messages ("Fix" not "Fixed", "Add"
  not "Added")
- [ x Commit messages explain *why* the change was made, not just
  *what* changed
- [x] Commit message body (if needed) provides detailed reasoning and
  context

#### Commit Structure
- [x] Each commit is atomic and builds successfully (every commit
  must compile)
- [x] One commit per logical change (don't mix refactoring with
  feature changes)
- [x] No fixup commits (squashed into parent commits using
  `git rebase -i`)

---

- [x] I'm ready to merge

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the setup for symmetric circular sector zones to use the intended configuration.
  * Improves the accuracy of sector behavior while preserving existing update behavior and outcomes.

### DIFF
--- a/src/Engine/Task/ObservationZones/SymmetricSectorZone.hpp
+++ b/src/Engine/Task/ObservationZones/SymmetricSectorZone.hpp
@@ -78,7 +78,7 @@ public:
                                   const double radius,
                                   const Angle angle
                                   ) noexcept {
-    std::unique_ptr<SymmetricSectorZone> oz(new SymmetricSectorZone(Shape::SYMMETRIC_SECTOR, true, false, loc,
+    std::unique_ptr<SymmetricSectorZone> oz(new SymmetricSectorZone(Shape::SYMMETRIC_SECTOR, true, true, loc,
                                                                     radius,
                                                                     angle));
     oz->UpdateSector();


chrgro:
Essentially its this bit of code that needs arc_boundary to be true:
https://github.com/XCSoar/XCSoar/blob/0765c25a4361b0b304e9f6d4fbe1fd46a0dfce10/src/Engine/Task/ObservationZones/SectorZone.cpp#L9

And thats set by the above argument :)

PR #2758, opened 2026-07-23
Engine/Trace: Bounds-first trail viewport query (long-flight CPU)

lordfolken:
## Summary
- Change `Trace::GetPoints(..., bounds, ...)` to **bounds/leg-filter first** (flat AABB), then spacing-thin, so Full-trail map paints no longer thin+`GeoClip` the whole flight history every frame at circling zoom (#2661).
- Limit `LockedTrailQuery` merge-vario copies to the kept point time span.
- Add `TestTraceBounds` and a NEWS note.

## Motivation
At circling zoom, spacing is only a few metres, so thin-then-filter barely reduces a Full (~25k) store and still pays expensive Angle/`GeoClip` work over nearly the whole history. Prep cost grew with flight length even when the on-screen trail looked steady.

## Measurement (Bruno Capelle IGC, desktop, `--trail-type=vario1 --circle-radius=750`)
| | before `circle_ms` | after |
|---|---:|---:|
| end-only (30 draws) | 89.8 | 51.8 |
| timeline @ 6 h (10 draws) | 51.7 | 18.7 |

Circling cost vs store size is ~10× flatter after the change.

## Test plan
- [x] `make TARGET=UNIX WERROR=y output/UNIX/bin/TestTraceBounds` — 16/16 ok
- [ ] Smoke: long IGC replay, Trail Full / Vario #1, circling zoom — map stays responsive; trail still draws crossing legs into view
- [ ] Confirm Trail Short (10 min) / Long (1 h) still look correct vs Full

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -179,6 +179,9 @@ Version 7.45 - not yet released
     pre-filtered to recently-used waypoints; listed in the Gesture Help
     dialog with its own icon #336
 * map
+  - snail trail: reduce long-flight map CPU by clipping the trail to the
+    visible area before spacing-thinning (Full trail no longer walks the
+    whole history every redraw) #2661
   - dynamically adjust contour line density to zoom factor #2233
   - provide different contour density profile settings for mountains,
     low lands, etc. to have sensible contours in a given landscape type
--- a/build/test.mk
+++ b/build/test.mk
@@ -109,6 +109,7 @@ TEST_NAMES = \
 	TestOGNAprsParser \
 	TestMETARParser \
 	TestIGCParser \
+	TestTraceBounds \
 	TestStrings TestUnescapeCString TestUTF8 TestWrapText \
 	TestInputConfig \
 	TestCRC16 TestCRC8 \
@@ -1021,6 +1022,14 @@ TEST_TRACE_SOURCES = \
 	$(TEST_SRC_DIR)/TestTrace.cpp
 TEST_TRACE_DEPENDS = IO OS GEO MATH UTIL
 $(eval $(call link-program,TestTrace,TEST_TRACE))
+
+TEST_TRACE_BOUNDS_SOURCES = \
+	$(TEST_SRC_DIR)/tap.c \
+	$(SRC)/Engine/Trace/Point.cpp \
+	$(SRC)/Engine/Trace/Trace.cpp \
+	$(TEST_SRC_DIR)/TestTraceBounds.cpp
+TEST_TRACE_BOUNDS_DEPENDS = GEO MATH UTIL
+$(eval $(call link-program,TestTraceBounds,TEST_TRACE_BOUNDS))
 
 FLIGHT_TABLE_SOURCES = \
 	$(SRC)/IGC/IGCParser.cpp \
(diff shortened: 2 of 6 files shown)


lordfolken:
@bomilkar Could you test this on your Cubie2 / OpenVario when you get a chance?

It changes the Full snail-trail path so map paints clip to the visible area **before** spacing-thinning (and trims the merge-vario copy to that window). On a long IGC at circling zoom, desktop stress showed a clear drop in draw prep cost that previously grew with flight length (#2661).

Suggested check (same idea as your Pi trail A/B):
1. Build this PR branch
2. Replay a long flight (or fly) with **Trail type = Vario #1**, **Trail length = Full**, Contours as you normally use
3. After ~1–1.5 h, see whether zoom / UI stay usable on the Cubie2
4. Optional A/B: Full vs Short vs Off, and Vario #1 vs Vario #1 (dots)

No rush — whenever you next have the box on the desk or in the glider. Thanks!

lordfolken:
Pushed CI/review fixes:
- Correct `--help` NMEA example to \`OpenVario <file>\` (CodeRabbit)
- GDI: use \`VirtualCanvas\` for the offscreen bench surface; paint \`--visual\` in \`OnPaint\` (fixes WIN64/PC \`BufferCanvas::Create(PixelSize)\`)
- \`TestTraceBounds\`: remove unused \`start\` (fixes iOS \`-Wunused-variable\` / \`-Werror\`)

bomilkar:
> Could you test this on your Cubie2

Will do, perhaps next week. I have the OV-box on my desk and new testcases.

bomilkar:
It doesn't look good. Unless I made a mistake. Here is what I built:
```
SRC_URI = "git://github.com/lordfolken/XCSoar.git;protocol=https;branch=perf/trail-bounds-first \
        file://ovmenu-x.service \
"
```
I'm running on a OV-box with an Cubie2 at 720 MHz max. variod and sensord are disabled.

It captures process stats using /proc/PID/stat every 10 seconds.
**The blue samples are from this PR.** "%CPU_SmoothTrail-Full_bounds-filtered"
Red "%CPU_SmoothTrail-Full" is the current XCSoar Master, Trail length = Full
Yellow "%CPU_Trail-OFF" the current XCSoar Master, Trail length = OFF
Green "%CPU_Trail-Full" the current XCSoar Master with a patch to turn off TrailSmoothing, Trail length = Full

The "patch to turn off TrailSmoothing":
`UseTrailSmoothing()` in src/Renderer/TrailRenderer.cpp:232 always returns false.

Trail_performance

Below **without the blue trace**, to better show, the reference samples.

Trail_performance_withoutPR2758

It clearly shows: **this PR doesn't do the trick**.
The big impact seems to comes from Trail Smoothing.

As soon as the CPU load exceeds 60% it doesn't react on key strokes as fast as you would like. It reports to xcsoar.log:
`libinput: client bug: timer event2 debounce short: scheduled expiry is in the past (-50ms), your system is too slow`
When this happens it misses key strokes! Very annoying.

lordfolken:
Superseded by #2777 and #2783.

PR #2797, opened 2026-08-02
Dialogs/Plane/WeGlideTypePicker: Fix crash when offline

lordfolken:
## Summary
- Catch download/network exceptions from `ShowCoFunctionDialog` when loading the WeGlide aircraft type list, so Plane Details no longer crashes offline (reproduced on Android 7.44 with no network and an empty cache).
- Show the error via `ShowError` instead of aborting; cancel remains silent.
- Recreate the list renderer after Refresh to avoid a dangling reference to a moved-from vector.

## Test plan
- [ ] Android (or UNIX): clear/remove `weglide-aircraft-v1.json` cache, disable network, open Plane Details → WeGlide Type — expect an error dialog, no crash
- [ ] With network: same path downloads and shows the picker
- [ ] With a populated cache offline: picker opens from cache; Refresh without network shows an error and keeps the existing list
- [ ] Cancel during download returns to Plane Details without a spurious “could not load” message

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed a crash that could occur when opening the aircraft type picker without an internet connection.
  - Download errors are now shown clearly, while cancelled downloads remain silent.
  - Improved aircraft list loading and refreshing when no list is available.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -24,6 +24,9 @@ Version 7.45 - not yet released
   - document ``AirspaceLabels`` input event
   - document map pinch-to-zoom on touchscreens that report both
     finger positions #2790
+* WeGlide
+  - fix crash when opening the aircraft type picker without a network
+    connection
 * ui
   - Add pinch-to-zoom on the map for touchscreens that report both
     finger positions (Android, iOS): two fingers scale about the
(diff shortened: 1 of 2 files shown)


lordfolken:
@coderabbitai full review

Scumi:
Inflight bug report 😊

lordfolken:
I even had a solution before you landed. ;)

Issue #1019, opened 2022-08-26
Water ballast and crew weight not preserved when XCSoar restarts


iltis42:
XCSoar version
--------------
epic, 7.25 and before

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------
Preserve water ballast and crew weight, like bug settings are preserved when XCSoar restarts.
Today water ballast is set to 0 liters and crew weight is set to 90 kg. If user restarts XCSoar in flight,
from any reason maybe also a the result of a crash, corresponding settings are gone.

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------
Expectation is that XCSoar will come up with the same settings as before during flight.
Water ballast shall be dumped explicitly or may even be the same on next flight and also 
crew may have same weight at next flight, esp. in a single seater (except after a successful diet ;), 
so there is no reason to reset this value to a default value any time XCSoar starts up.

lordfolken:
Was implemented in #840 #807 by @hjr 
There is an explenation for the crew weight:
crew mass is volatile and part of the flight configuration, whereas a default can now be entered to the pilot identity found for the logger. (open for suggestions here)
Related issues and discussions

lordfolken:
## Proposed solution (design)

### Problem recap
Water ballast and **crew mass** used for the task polar are effectively **runtime** values. **Crew mass** from normal Flight Setup is **not** written to the profile; only **`CrewWeightTemplate`** (under **Logger** settings, `LoggerSettings::crew_mass_template`) is persisted as a **default/template**. On startup, `Startup.cpp` seeds the glide polar with `gp.SetCrewMass(computer_settings.logger.crew_mass_template)`, so if the user only changed crew in Flight Setup, restarts still look like **0 L ballast** and **90 kg** (or whatever the template is) unless that template was set in Logger config.

### Direction: session state file (not OS temp)
Store **operational** polar state in a **durable file under the XCSoar data directory** (e.g. via `LocalPath`), e.g. crew mass, water ballast (litres), optionally bugs, plus a **`flying`** flag taken from `DerivedInfo::flight.flying` when saving.

**Restore rule on cold start:** if the session file says **`flying == true`**, load crew/ballast (etc.) from the file and **apply to the glide polar** during startup **before** relying on live GPS/movement. Use **only the persisted `flying` flag** for this decision — at boot, `Calculated().flight.flying` is not meaningful until sensors and `FlyingComputer` have run.

[...]

PR #2819, opened 2026-08-05
Logger: Resume a flight after a crash or power cut

OBrown92:
Resumes a Flight after XCSoar dies mid-air — a crash, an OS kill, or the
power being cut. Today the restart begins from zero: the task shows no
progress, flight time resets, the trail is empty and the statistics are
gone, while the pilot is still airborne on the same task. The flight record
is also split across two IGC files, the first unsigned.

**Draft** — one open review finding and two things unverified, both listed
at the bottom.

## What it does

https://github.com/user-attachments/assets/a8e653d8-f794-4503-9fe2-d6a0ab5823e0

At takeoff after a restart, in this order:

1. Scan the IGC directory for a **Cut Session** — a file with no G-record.
   Signing only happens from `StopLogger`, so an unsigned file means the
   session ended without XCSoar's knowledge.
2. Replay it through the glide computer behind a modal progress dialog with
   a Cancel button, rebuilding task progress, takeoff time, trail,
   statistics and contest.
3. Check the replayed session ended **airborne**.
4. Only then tell the logger to continue that file rather than open a new
   one.

Resume is automatic. The pilot is never asked for permission, only given a
way out: Cancel, plus a 120 s backstop. Either keeps the partial result.

Time proximity is not the discriminator — it would wrongly resume an
ordinary relaunch half an hour after landing. It survives only as a 15 min
staleness bound, compared GPS time to GPS time, because the system clock is
never set from GPS on Linux or Android. A simulated session is never signed
either, so those are rejected by the absence of `HFFXA`.

## Nothing is appended before the reconstruction is validated

Thanks to @lordfolken for pushing on this — the original ordering had
`StartLogger` commit to the append before the sweep could judge it.

Appending cannot be undone: once two flights share one IGC file, nothing
separates them again. Computer state is cheap to discard. So the
irreversible step goes last. The logger no longer decides anything; it is
told, via `SetResumeTarget()`, after the check has passed. On reject there
is nothing to undo in the file, because the logger has not opened anything
yet.

Sweeping first costs nothing: the calculation thread is suspended for the
sweep either way, so no live fix is recorded during it whether the logger
is open or not.

Two things fell out of the change. The sweep no longer replays a file that
is concurrently being appended to, which had been producing an off-by-a-few
fix count and, in one case, replaying records the sweep had itself caused
to be written. And the rollback shrinks to computer state only.

## On gap continuity and scoring

Raised on the thread: reboot plus sweep can push the Gap past FAI
continuity (~1 min) and WeGlide (~120 s), so is one-file-via-append still
worth it?

Nothing here can shrink the Gap — that data is gone the moment the device
dies. What the sweep recovers is what the *restart* destroyed: hours of
task progress, statistics and trail that were never lost from the file,
only from memory. A pilot flying an unscoreable file still wants the
computer to know where they are on task.

Most cross-country pilots carry a separate logger for the scoreable trace
anyway, which is also why this deliberately does not guess at a turnpoint
rounded entirely inside the Gap: the backup device already knows.

## Design notes

`GlideComputer::ProcessGPS` is deliberately **unmodified**. It runs for
every fix of every live flight and is the one place a conditional could
break flights that already work. Everything a sweep must not do is
suppressed in the component that does it — the logger, the condition
monitors, the task events — rather than by branching in the hot path.

Replay uses a separate idle entry point rather than a flag on
`ProcessIdle`, so `LogComputer::Run` is unreachable rather than merely
skipped. That is not the only route to the file, though: a replayed task
start reaches `OnStartTask()` and so `LogComputer::StartTask()`, which
writes an E record, and `IGCWriter::LogEvent` appends a B record after it
as the IGC spec requires. Those records land in the file the sweep is
reading, hours out of order, and replaying them warps time backwards and
resets the flying state. Hence the suppression flag there too.

Terrain is detached for the sweep. It is attached during startup, so
without this the route planner fires on nearly every replayed fix.

## Verified

- All 16 commits build standalone (`TARGET=MACOS`), checked one by one
- `make check` passes: 106 files, 10 668 assertions
- `test/TestFlightResume` drives the real `GlideComputer` over an IGC twice,
  live-shaped and sweep-shaped, and asserts the results are identical — and
  that a sweep logs no fix at all
- End-to-end on an Android emulator against two IGC files from a real
  interrupted flight:

```
Cut Session found: 2026-08-01-XCS-AAA-01.igc (10916 fixes)
Resume swept 10916/10916 fixes; replayed Session ends airborne
Continuing Flight in 2026-08-01-XCS-AAA-01.igc
```

  Resulting file: 10 930 B-records, one header, timestamps monotonic
  through the seam.

- **Cancel exercised** against a 49 478-fix candidate, cancelled at 7 %:

```
Resume swept 3456/49478 fixes; replayed Session ends airborne
Continuing Flight in 2026-08-01-XCS-AAA-01.igc
```

  The sweep stopped within a second, the partial result was kept rather
  than discarded, and the file was still continued correctly — cancel is
  not rejection.

## Review findings addressed

- append could not work on **Windows**: `FileOutputStream` opens with share
  mode `0`, so `GRecord::LoadFileToBuffer()` could never open the file. The
  repair and hash load now happen before the stream is opened, and
  `grecord` is declared before `file` so initialiser order enforces it
- a prior **simulator** IGC is never signed and so qualified for ever; now
  rejected by the absence of `HFFXA`
- the suppression flags were written while the calculation thread was live;
  thread suspension now comes first
- `[[gnu::pure]]` removed from `IsExpired()`, which reads the clock
- a sweep that throws is now rolled back, unlike a cancelled one
- `File::Truncate` no longer grows a file
- tests assert the exact `H`-record count and express `plan_tests` as an
  addition

## Still open

- **`Print()` in `TestFlightResume` takes a raw `const char *`.** Left as
  is: every caller passes a literal. Happy to change it if you would rather
  the guideline were applied uniformly.
- **No real-hardware timing.** 120 s was chosen against a projection of
  41–103 s for 20 513 fixes on a Kobo. If measurement lands near the top of
  that range the answer is a coarser sweep, not a longer freeze.

## Checklist

- [x] Rebased on current `master` (`2584f77fe4`)
- [x] Commits follow `<Component>: <Summary>`, no `src/` prefix
- [x] Present tense; messages explain *why*
- [x] Every commit compiles — verified individually
- [x] One logical change per commit; no fixup, WIP or duplicate commits
- [x] `NEWS.txt` updated
- [x] I'm ready to merge — draft, pending the items above

## Summary by CodeRabbit

* **New Features**
  * Added flight recovery after crashes or power loss.
  * Resumes logging in the previous IGC file when possible.
  * Reconstructs flight progress, timing, trail, and statistics through a cancellable progress dialog.
  * Detects recent interrupted flight logs and repairs incomplete records before continuing.
* **Bug Fixes**
  * Prevented duplicate flight-fix records during replay.
  * Preserved live flight state when recovery cannot be completed.
  * Suppressed replay notifications and logging side effects during reconstruction.


lordfolken:
## Question: append commitment before resume validation

Thanks for the careful ADRs — the CutSession detection and replay-safe reconstruction path look solid. One design point I’d like to hear more on:

**`StartLogger` already opens/continues the cut IGC (append) before `ResumeFlight` runs the sweep and decides accept/reject** (ADR-0006 / ADR-0007). On reject, computer state is rolled back but the file is left as-is, so a false positive can permanently put two flights in one IGC.

Could you walk through why that ordering is preferable to something like:

1. Detect the cut session and rebuild state (sweep) first, still logging to a **new** file (or buffering), then
2. Only switch to append / splice once the airborne (and any other) checks pass —

or, if append must stay, what makes undoing or avoiding the append on reject impractical?

Also relevant for scoring: reboot + sweep can already push the Gap past ~1 min (FAI continuity) / ~120 s (e.g. WeGlide). Does mandatory one-file-via-append still win if the cockpit restore worked equally well against a new sequence file?

Happy to be convinced — just want the tradeoff spelled out on the thread before this moves out of draft / gets split.

lordfolken:
So this is good timing, as i'd like to address the persistence issues anyway. 

Please drop all the docs. that just makes this pr large. Decisions should be in the git commit messages.

OBrown92:
@lordfolken — both answered, details in the updated description.

**1. Append before validation.** You were right; changed. `ResumeFlight` now
runs *before* `AutoLogger start`: the handler scans, sweeps, checks the
session ended airborne, and only then calls `Logger::SetResumeTarget()`. The
logger no longer judges anything — it is told, after the check passes. On
reject there is nothing to undo, because it has not opened a file yet.

Sweeping first costs nothing, which is what I had wrong: the calculation
thread is suspended for the sweep either way, so no live fix is recorded
during it whether the logger is open or not. Two bugs fell out of the fix —
the sweep no longer reads a file that is being appended to underneath it,
and the rollback shrinks to computer state only.

**2. Continuity.** Agreed there is nothing to do about it, and it does not
change the case for appending. The sweep cannot shrink the Gap; that data is
gone when the device dies. What it recovers is what the *restart* destroyed —
hours of task progress, statistics and trail still present in the file but
lost from memory. A pilot on an unscoreable file still wants the computer to
know where they are on task.

So the argument for one file is one file to sign and upload, not a scoring
one. Most cross-country pilots carry a separate logger for the scoreable
trace anyway — which is also why this deliberately does not guess at a
turnpoint rounded entirely inside the Gap.

Cancel is now exercised too (3456/49478 fixes, partial result kept). Still
unverified: real-hardware timing, and one unreproduced
`GCE_LANDING`/`GCE_TAKEOFF` pair.

OBrown92:
@lordfolken yes this is bothering me also for a long time as the OpenVarios we use like to crash from time to time. Loosing all the tracked data is not nice. I'm not sure if this is the best way but it looked kind of good, I have two example igc files of a flight where exactly this happened and the approach works great.

> Also relevant for scoring: reboot + sweep can already push the Gap past ~1 min (FAI continuity) / ~120 s (e.g. WeGlide). Does mandatory one-file-via-append still win if the cockpit restore worked equally well against a new sequence file?

The data is lost anyways and most pilots have another logger for a valid and accepted igc anyways, so I would say this is not a issue.

lordfolken:
@WolfgangDrescher if you are working on replay improvements, could you base your work on top of this branch? Id love to have a third pair of eyes here.

OBrown92:
One thing which came into my mind are the "Flight Settings". Like QNH, Weight, ... I'm not sure if we can recover this from somewhere, I think it is not written to a file. We can show a message but the messages disappear quickly and maybe overseen.

lordfolken:
> One thing which came into my mind are the "Flight Settings". Like QNH, Weight, ... I'm not sure if we can recover this from somewhere, I think it is not written to a file. We can show a message but the messages disappear quickly and maybe overseen.

We have quite a few topics here with the persistence: 
https://github.com/XCSoar/XCSoar/issues?q=is%3Aissue%20state%3Aopen%20label%3Atopic-profile

Particular https://github.com/XCSoar/XCSoar/issues/1019 where we have to consider all the settings in Flight Setup

lordfolken:
Just for a working instruction: In general can you please [rebase](https://github.com/XCSoar/XCSoar/discussions/2766) your code on upstream master? I would like a linear history in git.

lordfolken:
g++ @output/UNIX/bin/RunFlyingComputer.rsp
/usr/bin/ld: output/UNIX/dbg/src/IGC/IGCWriter.o: in function `PrepareGRecord(Path, IGCWriter::Mode)':
/__w/XCSoar/XCSoar/src/IGC/IGCWriter.cpp:31:(.text+0x22): undefined reference to `RepairIgcForAppend(Path)'
g++ @output/UNIX/bin/RunCirclingWind.rsp

I think the test is missing an include.

PR #2846, opened 2026-08-10
Dialogs/Plane/PlaneListDialog: Rebuild plane list with active checkbox

lordfolken:
Mark the active plane with a list-row checkbox, show type and
competition ID in the standard two-row layout, prefer Activate as the
default action (New when empty or sole-active), auto-activate the first
created plane, and announce polar changes or an invalid polar.
Share GetListRowCheckBoxRect with multi-select file lists and add list
hotspot activation for the checkbox.

Thank you for contributing to XCSoar!

We truly appreciate your time and effort in making XCSoar better. We know
that contributing to an open-source project can be challenging, and we're
grateful that you've chosen to help.

This checklist is here to help guide you through the submission process.
Don't worry if you're not sure about something—feel free to ask questions
or submit your PR, and we'll work together to get it ready.

For coding, style guide, architecture information please see our
[development guide](https://xcsoar.readthedocs.io/en/latest/index.html).

For Git tips and tricks, including interactive rebase, fixup commits, and
common workflows, see the
[Git Tips](https://xcsoar.readthedocs.io/en/latest/git_tips.html)
documentation.

For testing and debugging utilities, see the
[Test & Debug
Utilities](https://xcsoar.readthedocs.io/en/latest/test_debug_utilities.html)
documentation. 

## Pre-Submission Checklist

#### Branch & Rebase
- [ ] Use `git rebase -i` to clean up commit history before PR
  submission

#### Commit Format & Messages
- [ ] All commits follow the format: `<Component>: <Summary>` (no
  `src/` prefix)
- [ ] Use present tense in commit messages ("Fix" not "Fixed", "Add"
  not "Added")
- [ ] Commit messages explain *why* the change was made, not just
  *what* changed
- [ ] Commit message body (if needed) provides detailed reasoning and
  context

#### Commit Structure
- [ ] Each commit is atomic and builds successfully (every commit
  must compile)
- [ ] One commit per logical change (don't mix refactoring with
  feature changes)
- [ ] Self-contained commits (each commit changes one thing)
- [ ] No fixup commits (squashed into parent commits using
  `git rebase -i`)
- [ ] No duplicate commits (check with `git log --oneline`)
- [ ] No "WIP" or "testing" commits (clean up before PR)

---

- [ ] I'm ready to merge

### DIFF
--- a/src/Dialogs/Plane/PlaneListDialog.cpp
+++ b/src/Dialogs/Plane/PlaneListDialog.cpp
@@ -7,7 +7,9 @@
 #include "Dialogs/WidgetDialog.hpp"
 #include "Widget/ListWidget.hpp"
 #include "Form/Button.hpp"
+#include "Form/CheckBox.hpp"
 #include "Renderer/TwoTextRowsRenderer.hpp"
+#include "Screen/Layout.hpp"
 #include "Plane/Plane.hpp"
 #include "Plane/PlaneGlue.hpp"
 #include "Plane/PlaneFileGlue.hpp"
@@ -21,6 +23,7 @@
 #include "Look/DialogLook.hpp"
 #include "Interface.hpp"
 #include "Language/Language.hpp"
+#include "Message.hpp"
 #include "util/StringAPI.hxx"
 #include "util/StringCompare.hxx"
 #include "Components.hpp"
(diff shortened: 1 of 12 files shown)


WolfgangDrescher:
The message text should be `Polar changed to [Name]` or `Polar changed: [Name]`.

Technically a checkbox is the wrong UI element when the option cannot be multi-selected. A radio button would be the correct choice here, but I think it would look silly. So maybe it would be better to just highlight the active plane with `COLOR_YELLOW`, or to introduce a `COLOR_SELECTED` subject.

Other observations and inconsistencies:

* Ticking the checkbox changes the active plane but does not close the window (this is fine, I guess).
* A single tap or double tap on a plane list row opens a confirmation dialog. Pressing Yes or No does not close the window (I mean the plane dialog, not the confirmation dailog). A single tap should only select the row, or it should activate the plane, but then we don't need the "Activate" button (at least on touch devices). There should not be a confirmation dialog at all in my opinion.
* Clicking the "Activate" button closes the window. I don't think that should happen.
* If you activate several planes one after another, you get a message for each activation over the map, and it may become confusing which one is active now. Ideally the message should be sent only once when you close the dialog, so that only the polar of the currently active plane is shown.

---

Here is some inspiration I quickly sketched in Swift Playgrounds:

Bildschirmfoto 2026-08-31 um 17 54 37
Bildschirmfoto 2026-08-31 um 17 46 58
Bildschirmfoto 2026-08-31 um 17 52 08

If we want to use a checkmark here, we should not display it as a checkbox. I like the active badge.

PR #2855, opened 2026-08-12
Text entry: use the system keyboard on iOS, add a Paste button

### LINKED ISSUE #2894: Cant enter special characters on iOS

7.45

System information
------------------

Iphone 16 Pro Max iOS 26.5.2

Try to enter a skysight password with special characters

Maybe it would be better to use the system keyboard of iOS instead of the XCSoar one. This would also allow support for copy-pasting.

WolfgangDrescher:
XCSoar's built-in on-screen keyboard cannot enter `_`, `/`, `:` and other special characters. Where the operating system provides an on-screen keyboard (iOS), the text entry dialog now uses that one instead. This is the new behaviour of the `Default` text input style, while `Keyboard` keeps XCSoar's own one and platforms without a system keyboard are unaffected. Also adds a `Paste` button for the system clipboard, and `_` on the built-in keyboard via shift.

Can anyone check if this works on Android as well?

Bildschirmfoto 2026-08-13 um 01 57 21

## Summary by CodeRabbit

* **New Features**
  * Added operating-system keyboard support for text-entry dialogs where available, with automatic fallback to the built-in keyboard.
  * Added clipboard paste support and character validation in text-entry fields.
  * Added clearer backspace labeling with “⌫” when supported.
  * Improved keyboard layouts for switching between “-” and “_”.
  * Added system-font usage on macOS and iOS, improving platform-native text rendering.

* **Documentation**
  * Updated development notes and text-entry keyboard configuration guidance.

### DIFF
--- a/src/Dialogs/DialogSettings.hpp
+++ b/src/Dialogs/DialogSettings.hpp
@@ -8,12 +8,25 @@
 struct DialogSettings {
   enum class TextInputStyle : uint8_t {
     /**
-     * Use the platform default - i.e. keyboard if the device has a
-     * pointing device.
+     * Use the platform default - i.e. the keyboard of the operating
+     * system where there is one (#SystemKeyboard), else our own
+     * keyboard if the device has a pointing device.
      */
     Default,
+
+    /**
+     * Always use XCSoar's own on-screen keyboard.
+     */
     Keyboard,
+
     HighScore,
+
+    /**
+     * Use the on-screen keyboard provided by the operating system
+     * (e.g. on iOS), which gives access to all special characters.
+     * Falls back to #Keyboard if the platform has none.
+     */
+    SystemKeyboard,
   };
 
   enum class TabStyle : uint8_t {
--- a/src/Dialogs/DialogSettings.hpp
+++ b/src/Dialogs/DialogSettings.hpp
@@ -20,13 +20,6 @@ struct DialogSettings {
     Keyboard,
 
     HighScore,
-
-    /**
-     * Use the on-screen keyboard provided by the operating system
-     * (e.g. on iOS), which gives access to all special characters.
-     * Falls back to #Keyboard if the platform has none.
-     */
-    SystemKeyboard,
   };
 
   enum class TabStyle : uint8_t {
--- a/src/Dialogs/TouchTextEntry.cpp
+++ b/src/Dialogs/TouchTextEntry.cpp
@@ -10,8 +10,11 @@
 #include "Asset.hpp"
 #include "Screen/Layout.hpp"
 #include "ui/event/KeyCode.hpp"
+#include "ui/event/TextInput.hpp"
 #include "UIGlobals.hpp"
 #include "Language/Language.hpp"
+#include "util/CharUtil.hxx"
+#include "util/StringAPI.hxx"
 #include "util/StringCompare.hxx"
 #include "util/TruncateString.hpp"
 #include "ui/window/Window.hpp"
@@ -21,18 +24,26 @@
 namespace {
 struct TextEntryLayout {
   PixelRect editor;
+  PixelRect paste;
   PixelRect backspace;
   PixelRect keyboard;
   PixelRect ok, cancel, clear;
 };
 
+/**
+ * @param with_paste reserve room for the @em Paste button
+ */
 static void
-ComputeTextEntryLayout(const PixelRect &rc, TextEntryLayout &o) noexcept
+ComputeTextEntryLayout(const PixelRect &rc, bool with_paste,
+                       TextEntryLayout &o) noexcept
 {
   const int client_height = rc.GetHeight();
   const int padding = Layout::Scale(2);
   const int backspace_width = Layout::Scale(36);
   const int backspace_left = rc.right - padding - backspace_width;
+  const int paste_width = with_paste ? Layout::Scale(60) : 0;
+  const int paste_left = backspace_left - (with_paste ? padding : 0) -
+    paste_width;
   const int editor_height = Layout::Scale(22);
   const int editor_bottom = padding + editor_height;
   const int button_height = Layout::Scale(40);
@@ -69,7 +80,8 @@ ComputeTextEntryLayout(const PixelRect &rc, TextEntryLayout &o) noexcept
     ? rc.right
     : clear_left + Layout::Scale(50);
 
-  o.editor = {0, padding, backspace_left - padding, editor_bottom};
+  o.editor = {0, padding, paste_left - padding, editor_bottom};
+  o.paste = {paste_left, padding, paste_left + paste_width, editor_bottom};
   o.backspace = {backspace_left, padding, rc.right - padding, editor_bottom};
   o.keyboard = {padding, keyboard_top, rc.right - padding, keyboard_bottom};
   o.ok = {ok_left, button_top, ok_right, button_bottom};
@@ -80,7 +92,8 @@ ComputeTextEntryLayout(const PixelRect &rc, TextEntryLayout &o) noexcept
 static void
 ApplyTextEntryLayout(const TextEntryLayout &L, WndProperty &editor, Button &ok,
                      Button &cancel, Button &clear, KeyboardWidget &keyboard,
-                     Button &backspace, ContainerWindow &client_area) noexcept
+                     Button &backspace, Button *paste,
+                     ContainerWindow &client_area) noexcept
 {
   editor.Move(L.editor);
   ok.Move(L.ok);
@@ -88,6 +101,8 @@ ApplyTextEntryLayout(const TextEntryLayout &L, WndProperty &editor, Button &ok,
   clear.Move(L.clear);
   keyboard.Move(L.keyboard);
   backspace.Move(L.backspace);
+  if (paste != nullptr)
+    paste->Move(L.paste);
   client_area.Invalidate();
 }
 } // namespace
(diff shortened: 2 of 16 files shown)


WolfgangDrescher:
Never mind about the system keyboard on Android. I missed that it does not render with SDL. Is a native keyboard available on Android at all?

src/Dialogs/TouchTextEntry.cpp
```diff
@@ -110,10 +139,35 @@ static char edittext[MAX_TEXTENTRY];
...
+  const char *allowed = AllowedCharactersCallback(edittext);
+  if (allowed == nullptr || StringFind(allowed, ch) != nullptr)
+    return ch;
+
+  /* the allowed set is usually upper case only */
+  if (const char upper = ToUpperASCII(ch);
+      upper != ch && StringFind(allowed, upper) != nullptr)
+    return upper;
+
+  return 0;
```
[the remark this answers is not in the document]
WolfgangDrescher: FilterCharacter can never see a non-ASCII value, because both callers gate first (FormCharacter on `ch >= 0x80`, OnPaste on IsPrintableASCII). The clipboard path drops whole characters rather than splitting them, since every byte of a multi-byte sequence has the high bit set. And the system keyboard path receives decoded code points, not bytes: ShowModal iterates LengthUTF8 and Event::GetCharacter() decodes via NextUTF8, so a typed "ü" arrives as one call and is dropped as a whole. What the finding really describes is that this dialog is ASCII-only by design, marked by the "TODO: ASCII only for now" comment in FormCharacter that predates this branch.

That is worth doing as a separate change, and it is much more tractable now that TCHAR is gone: text is UTF-8 char end to end, and util/UTF8.hpp already provides UnicodeToUTF8, SequenceLengthUTF8 and CopyTruncateStringUTF8, so the work is confined to TouchTextEntry.cpp. I would rather land that first than fold it into this series.

@lordfolken What do you think? In my opinion it's nothing we desperately need right now, but we could add it as an issue for the future.

DanD222:
> Is a native keyboard available on Android at all?

It is, for example when entering LiveTrack24 credentials in settings, or a TCP client IP address in devices.

yorickreum:
I am certainly in favour of using the iOS default display.

Making it configurable would be a bonus, imo we can just fully switch to the system default keyboard.

src/Dialogs/DialogSettings.hpp
```diff
@@ -8,12 +8,25 @@
...
     Keyboard,
+
     HighScore,
+
+    /**
+     * Use the on-screen keyboard provided by the operating system
+     * (e.g. on iOS), which gives access to all special characters.
+     * Falls back to #Keyboard if the platform has none.
+     */
+    SystemKeyboard,
```
lordfolken: `SystemKeyboard` is redundant with `Default` on iOS. `TextEntry.cpp` already handles both in the same case, so the extra enum value does the same thing as Default wherever `HasScreenKeyboard()` is true.

We only need Default vs Keyboard (vs HighScore): Default uses the OS keyboard when there is one, Keyboard keeps XCSoar’s grid. A third clone of Default is extra settings noise (see the duplicate entries in `text_input_list_system`).

Please drop `SystemKeyboard` and keep Default as “OS keyboard if present, else our grid”.

lordfolken:
SDL `StartTextInput()` only brings up the default iOS keyboard for every string. That is enough for `_` `/` `:` on Default, but it cannot set email / password / URI traits (`secureTextEntry`, `UIKeyboardTypeEmailAddress` / `URL`, Android `InputType`).

Those need a real field (`UITextField` / `EditText`), not more SDL hints. Fine to leave out of this PR; Default = stock OS keyboard is the right first step.

Please keep **Keyboard** as XCSoar’s grid so we still have a cursor-key path (Up/Down/Left/Right, Enter, Esc) when the IME is the wrong flavour or the platform has none.

WolfgangDrescher:
I don't think the traits are necessary right now. Having the system keyboard is already a huge improvement over the previous keyboard: every special character that matters can be entered. A few of them still don't make it through, because of the UTF-8 charset limitation mentioned above.

Bildschirmfoto 2026-08-19 um 23 17 44
Bildschirmfoto 2026-08-19 um 23 17 56


Apart from that, I think we should revise the labels.

English:

- `Default`
- `Keyboard`
- `HighScore Style`

`Default` is fine. For clarity, `Keyboard` should be renamed to `XCSoar keyboard`, `On-screen keyboard` or something along those lines. `HighScore Style` should probably be renamed too, to something like `Rotary knob entry`. Can we agree on sensible labels here? I would then add them in this PR right away.

German:

- `Voreinstellung`
- `Tastatur`
- `Ranglisten Stil`

`Voreinstellung` is really bad and misleading, it should be `Standard`. For `Tastatur` and `Ranglisten Stil` the same criticism applies as in English.

We should also revise the help text.

src/Dialogs/DialogSettings.hpp
WolfgangDrescher: Solved in 0cd4442.

WolfgangDrescher:
Something like this:

```diff
diff --git a/po/de.po b/po/de.po
index 55cb0ca47b..50153f3384 100644
--- a/po/de.po
+++ b/po/de.po
@@ -2823,7 +2823,7 @@ msgstr "Alle Lufträume unter dem Segler werden gezeigt."
 
 #, no-c-format
 msgid "Default"
-msgstr "Voreinstellung"
+msgstr "Standard"
 
 #, no-c-format
 msgid ""
@@ -3413,18 +3413,30 @@ msgid "This determines how long menus will appear on screen if the user does not
 msgstr "Die Zeitkonstante bestimmt wie lange ein Menü angezeigt wird, wenn der Benutzer keine Eingabe tätigt und nicht mit dem Rechner interagiert."
 
 #, no-c-format
-msgid "Keyboard"
-msgstr "Tastatur"
+msgid "The keyboard of the operating system where there is one (e.g. iOS), otherwise XCSoar's own keyboard."
+msgstr "Die Tastatur des Betriebssystems, sofern vorhanden (z. B. iOS), sonst XCSoars eigene Tastatur."
 
 #, no-c-format
-msgid "HighScore Style"
-msgstr "Ranglisten Stil"
+msgid "XCSoar keyboard"
+msgstr "XCSoar-Tastatur"
+
+#, no-c-format
+msgid "XCSoar's own on-screen keyboard; can also be operated with cursor keys."
+msgstr "XCSoars eigene Bildschirmtastatur; auch mit den Pfeiltasten bedienbar."
+
+#, no-c-format
+msgid "Rotary knob entry"
+msgstr "Drehknopf-Eingabe"
+
+#, no-c-format
+msgid "Pick one character at a time with up/down, move with left/right; for devices without a touch screen."
+msgstr "Zeichen einzeln mit Hoch/Runter auswählen, mit Links/Rechts die Position wechseln; für Geräte ohne Touchscreen."
 
 msgid "Text input style"
 msgstr "Texteingabestil"
 
-msgid "Determines how the user is prompted for text input (filename, teamcode etc.)"
-msgstr "Bestimmt, wie der Benutzer zur Texteingabe aufgefordert wird (Dateiname, Teamcode, usw.)"
+msgid "How text is entered (file names, team code, waypoint names). \"Default\" uses the keyboard of the operating system where there is one, and XCSoar's own keyboard otherwise."
+msgstr "Legt fest, wie Text eingegeben wird (Dateinamen, Teamcode, Wegpunktnamen). „Standard“ nutzt die Tastatur des Betriebssystems, sofern vorhanden, sonst XCSoars eigene Tastatur."
 
 #, no-c-format
 msgid "OS settings"
diff --git a/src/Dialogs/Settings/Panels/InterfaceConfigPanel.cpp b/src/Dialogs/Settings/Panels/InterfaceConfigPanel.cpp
index 46b1fc9435..f633991fce 100644
--- a/src/Dialogs/Settings/Panels/InterfaceConfigPanel.cpp
+++ b/src/Dialogs/Settings/Panels/InterfaceConfigPanel.cpp
@@ -160,15 +160,18 @@ InterfaceConfigPanel::Prepare(ContainerWindow &parent,
   SetExpertRow(MenuTimeout);
 
   static constexpr StaticEnumChoice text_input_list[] = {
-    { DialogSettings::TextInputStyle::Default, N_("Default") },
-    { DialogSettings::TextInputStyle::Keyboard, N_("Keyboard") },
+    { DialogSettings::TextInputStyle::Default, N_("Default"),
+      N_("The keyboard of the operating system where there is one (e.g. iOS), otherwise XCSoar's own keyboard.") },
+    { DialogSettings::TextInputStyle::Keyboard, N_("XCSoar keyboard"),
+      N_("XCSoar's own on-screen keyboard; can also be operated with cursor keys.") },
     { DialogSettings::TextInputStyle::HighScore,
-      N_("HighScore Style") },
+      N_("Rotary knob entry"),
+      N_("Pick one character at a time with up/down, move with left/right; for devices without a touch screen.") },
     nullptr
   };
 
   AddEnum(_("Text input style"),
-          _("Determines how the user is prompted for text input (filename, teamcode etc.)"),
+          _("How text is entered (file names, team code, waypoint names). \"Default\" uses the keyboard of the operating system where there is one, and XCSoar's own keyboard otherwise."),
           text_input_list, (unsigned)settings.dialog.text_input_style);
   SetExpertRow(TextInput);
 
```

lordfolken:
Having the traits would allow the Keyboard to behave properly in each case. 
For example when you enter a password the password is currently printed on screen. (Android)
An email or url would have the allowed characters per their definitions. 

But as stated its not necessary especially if we treat this as a bugfix for #2894. I'd also leave out retranlsation for that usecase.

lordfolken:
Superseded by #2920, which is the same work cherry-picked onto current master with NEWS under 7.46.

Issue #1270, opened 2023-07-23
QR code for importing tasks


Persiantraveler:
XCSoar version
--------------

 7

What should XCSoar do differently, what functionality should be added?
Adding QR code import feature that allows tasks to be ported using the camera on the device. XCtrack uses this feature and it is a very time saving,error reducing and more accurate way to imprt task into their program. I hope Xcsoar would consider adding this feature too. 
----------------------------------------------------------------------

Described it in above

What are you trying to do, what is the use case for the suggestion? As explained above using QR to import tasks is a very good feature to have an it is very popular feature of XCtrack which could be used in other apps such as xcsoar. 
-------------------------------------------------------------------

MaxKellermann:
How does that work, technically? A QR code encodes a string of data. What is inside this string of data? Please post a link to a specification and upload a few samples with data how these samples are supposed to look like in XCSoar.

Scumi:
Related: #1080

Persiantravelerr:
Basically it works by encoding the task info such as coordinates for waypoints, start and finish and other variables such as SSS and ESS as time stamp into a QR code which then can be shared and imported into relevant apps that pilots use such as XcTrack, Naviter Oudie N, FlySkyHi...etc.


For example this is a task we had recently during the US Open.

As you can see the organizers had to only send that QR code to all participants as a text message and when we scanned that QR through whichever apps each pilot may use, in my case XcTrack, then the app would convert that QR to the Task map and Task way point list including the time start, finish ...etc in a matter of a second. 
Please take a look at these pictures for reference. 

 
![image](https://github.com/XCSoar/XCSoar/assets/140318247/00e82ab3-26cd-4d41-98b8-8eef3d283742)

![image](https://github.com/XCSoar/XCSoar/assets/140318247/4c54913d-a492-4879-8309-850e7afd2140)

![image](https://github.com/XCSoar/XCSoar/assets/140318247/ec7fbb52-fcec-4d3d-b604-242fe4728f1e)

Hope the following link and example codes I found online can help to illustrate how XCtrack seem to generate and also use QR code. 

https://pkg.go.dev/github.com/twpayne/go-xctrack

Here is a video in german of how you can create a task on one of the XcPlanner kind of Websites and then generate a QR code and import it into your flight app of choice. 
Please fast forward to minute 8:10 if you prefer to see how the presenter sets his task and you can see by minute 13:10 he has created the QR code. 

[...]

PR #2861, opened 2026-08-13
Task: Scan task QR codes with the camera (portable zxing-cpp decoder)

OBrown92:
Adds a **Scan QR Code** button to the task manager, so an XCTrack task can be
imported with the built-in camera instead of being routed through the phone's
camera app and an app chooser.

https://github.com/user-attachments/assets/0f4000da-e3a9-4c70-8908-1484788a1f40

Requested in #1270.

## Where the code lives

The decoder is portable C++, not Android code. `src/Task/QRDecoder.cpp` takes
nothing but an 8-bit luminance buffer and its row/pixel strides — no platform
types, no JNI — and is built on **zxing-cpp**, added as a `CmakeProject`
alongside the other third-party libraries. So the iOS and desktop ports can
reuse it as soon as they have a camera to point at it, with no second decoder
and no duplicated task parsing.

Only the capture layer is platform-specific, and only that part is in Java:
`QRScannerActivity` drives Camera2, and hands each frame's Y plane straight
across JNI. Because `ZXing::ImageView` accepts strides directly, the plane is
never repacked, and since Camera2 gives out a direct `ByteBuffer`, native code
reads the camera's memory in place rather than copying a frame at a time.

Nothing about the *decode* path is new: it already existed for the `xctsk:` URI
scheme (7.38). `ReceiveTaskQRCode()` strips the scheme prefix and reuses
`ReceiveXCTrackTask()`, so a scanned code and a scanned-elsewhere link end up
in exactly the same place. The prefix match is case-insensitive because XCTrack
writes `XCTSK:` while an intent URI arrives lower-cased.

The button lives on the task manager's *Manage* page, next to Save. A scan
started there leaves the dialog open, and the scanned task is handed to that
dialog via `TaskManagerReceiveTask()` rather than stacking a second task
manager on top — so cancelling a scan leaves the pilot's task untouched.

## zxing-cpp

Built with the reader only and only for QR: the encoders and the other five
symbologies are switched off, leaving 42 objects and none for 1D, Aztec,
DataMatrix, MaxiCode or PDF417, and no bundled zint.
`ZXING_DEPENDENCIES=LOCAL` keeps the build off the network. Apache-2.0, noted
in `THIRD_PARTY_NOTICES.txt`.

One build caveat worth knowing: zxing-cpp does not compile for Android with the
NDK we pin (r26d) out of the box. Below API 24 libc++ falls back to
`__bsd_locale_fallbacks.h`, and *precompiling* that header fails with a
`va_list` mismatch on aarch64. The affected translation unit is fine without
the PCH, so the recipe passes `-DZXING_DISABLE_PCH=ON`, with the reason
recorded there.

## Notes for review

- `android:hardwareAccelerated="true"` is spelled out on the activity. A
  `TextureView` in a software-rendered window draws nothing and never hands out
  its `SurfaceTexture`, which left the preview black and the camera unopened.
  Nothing else in the app depended on it, since the main activity drives its
  own GLES `SurfaceView`.
- The camera thread lives as long as the activity and is quit *from* that
  thread, so the teardown posted by `onPause()` has run first; otherwise the
  camera framework finds the looper gone while still delivering close
  callbacks.
- `onPause()` takes the session, device, reader and preview surface as locals
  and clears the fields before posting the teardown, and `onOpened()` /
  `startPreview()` re-check the paused flag, so a camera or surface arriving
  late is handed straight back instead of leaking.
- The camera permission is requested directly rather than through
  `PermissionHelper`, which reports every permission as granted in simulator
  mode without asking for it.
- Cancelling returns through `XCSoar` rather than just finishing, because the
  main activity is `singleInstance` and a bare `finish()` drops the pilot on
  the home screen.

## Testing

On a Pixel 7a (Android 17, arm64) and in the emulator:

- Scanning a real XCTrack task QR code imports it into the task manager, with
  turnpoints, cylinder radii and task distance as expected.
- The existing `xctsk:` intent path still opens a populated task manager.
- Cancel, hardware Back and camera errors all return to the map, with the task
  manager left as it was.
- `DecodeQRCode()` exercised off-device against the same zxing-cpp release,
  decoding the payload through a packed plane and through `rowStride +1`,
  `rowStride +64`, `pixelStride 2` and `pixelStride 2 + rowStride +32` — which
  also demonstrates it building and running outside Android. Its input guards
  are covered too: null pointer, zero size/width/height/stride, and a buffer
  too small for the geometry all return empty rather than throwing.
- Builds clean for `ANDROIDAARCH64`; the non-Android build is unaffected
  (`HaveQRScanner()` is a `constexpr false` stub off Android).
- All 33 catalogues pass `msgfmt --check`, and the button renders translated on
  a German device.

The translations are a separate commit so native speakers can correct them
without disturbing the feature.

## Known limitations

- Only Android has the capture side so far. iOS needs `AVFoundation` capture
  plus lifting `ReceiveTask.cpp` and `MainWindow::OnTaskReceived()` out of
  `#ifdef ANDROID`; the decoder it would call is already in place.
- The preview transform derives its orientation from the display rotation, so
  it may be stretched on a device whose natural orientation is landscape. The
  scanner is locked to portrait for now. Decoding is unaffected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -190,6 +190,7 @@ include $(topdir)/build/libjpeg.mk
 include $(topdir)/build/libsqlite.mk
 include $(topdir)/build/libtiff.mk
 include $(topdir)/build/netcdf.mk
+include $(topdir)/build/zxing.mk
 include $(topdir)/build/coregraphics.mk
 include $(topdir)/build/appkit.mk
 include $(topdir)/build/uikit.mk
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -1,4 +1,7 @@
 Version 7.46 - not yet released
+* android
+  - add "Scan QR Code" to the task manager to import an XCTrack task
+    with the camera, instead of opening the code in another app #1270
 * development
   - documentation: document the release and post-release version workflow
 
--- a/THIRD_PARTY_NOTICES.txt
+++ b/THIRD_PARTY_NOTICES.txt
@@ -1,5 +1,217 @@
 # Third-Party Notices
 
+## zxing-cpp
+
+XCSoar uses **zxing-cpp** to decode the QR codes read by the task QR
+scanner.  It is built with the QR reader only; the encoders and the
+other barcode formats are disabled.
+
+**License:** [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0)
+
+Copyright (c) zxing-cpp contributors.
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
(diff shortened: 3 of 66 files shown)


lordfolken:
Could we use `zxing-cpp` instead of the Java ZXing implementation?

It fits XCSoar’s C++ architecture better, is available as Debian’s `libzxing-dev`, and would allow the decoder to be shared across Android, Linux, and potentially iOS. On Android, Camera2 could remain in Java while passing the Y plane through JNI to the native decoder. The Android build could compile only the QR reader and disable writers and other formats.

yorickreum:
Would love to see this implemented cross-platform, not in in Java/Android-only part of the codebase…

OBrown92:
> Would love to see this implemented cross-platform, not in in Java/Android-only part of the codebase…

@yorickreum @lordfolken  I just changed it to `zxing-cpp` I can also do the iOS part. But will do it in a separate pr. I think it's ready. I'm not sure if it is correct to put the translations here as well.

OBrown92:
I also have the ios one ready, based on this pr. Need to confirm it on a real device as the simulator does not support virtual camera. Will do it next week. Would love to see this feature in the next version.

PR #2863, opened 2026-08-14
Quick Menu: internal grid separators with top and bottom borders

### LINKED ISSUE #2853: Grid separators for Quick Menu
I think these would improve the Quick Menu legibility quite a bit, especially in cases where there are multi-line items above each other.

DanD222:
Draw only the internal cell separators, while keeping the top and bottom edges visible to avoid clipping at the panel boundary.

Full disclosure: this is Copilot, and I don't really know what I'm doing. I did build it for Android and MacOS and it looked OK. 
I don't know if this kind of input is welcome/beneficial, or rather if it detracts from more serious code reviews.

Closes https://github.com/XCSoar/XCSoar/issues/2853

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved grid rendering when navigating between pages.
  * Added clearer separators and borders around visible grid content.
  * Ensured grid panels are drawn consistently across light and dark display modes.

### DIFF
--- a/src/Form/GridView.cpp
+++ b/src/Form/GridView.cpp
@@ -2,16 +2,22 @@
 // Copyright The XCSoar Project
 
 #include "Form/GridView.hpp"
+#include "Look/DialogLook.hpp"
+#include "Screen/Layout.hpp"
+#include "ui/canvas/Color.hpp"
+#include "ui/canvas/Pen.hpp"
+#include "ui/canvas/Canvas.hpp"
 #include "Math/Util.hpp"
 
 void
-GridView::Create(ContainerWindow &parent, const DialogLook &look,
+GridView::Create(ContainerWindow &parent, const DialogLook &_look,
                  const PixelRect &rc, const WindowStyle style,
                  unsigned _column_width, unsigned _row_height)
 {
   column_width = _column_width;
   row_height = _row_height;
-  PanelControl::Create(parent, look, rc, style);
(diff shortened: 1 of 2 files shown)


lordfolken:
I don't mind ai written code, as it tends to follow the .cursor rules i specified relatively well, therefore taking care of a lot of aspects that would require a deeper knowledge of the codebase and conventions.
What i do see however, is that its often bloated for what it is, and requires several iterations to boil down to the essentials. Also use the planning stage and the self review stage extensively before you let it write code.

However you must be able to read and understand and explain it. Its the litmus test whether you have control of the thing or not. The text responder will do things you instruct, it doesn't think in a larger context.

WolfgangDrescher:
Should we add a `.cursor/rules/documentation.mdc` file?

DanD222:
> I don't mind ai written code, as it tends to follow the .cursor rules i specified relatively well, therefore taking care of a lot of aspects that would require a deeper knowledge of the codebase and conventions. What i do see however, is that its often bloated for what it is, and requires several iterations to boil down to the essentials. Also use the planning stage and the self review stage extensively before you let it write code.
> 
> However you must be able to read and understand and explain it. Its the litmus test whether you have control of the thing or not. The text responder will do things you instruct, it doesn't think in a larger context.

OK, thanks. I'll need to do do a bit more code-staring, so I'll convert to draft for now

PR #2865, opened 2026-08-15
Kobo clara colour gcc14

anj1:
Adds Kobo Clara BW/Colour support when launched from Nickel, using the GCC 14 NickelTC toolchain.

See: https://github.com/XCSoar/XCSoar/discussions/2827

Includes:

    KOBO_NICKEL build target with static libstdc++/libgcc
    FBInk framebuffer integration
    Clara BW/Colour model detection and display defaults
    Color framebuffer support with grayscale fallback
    Nickel-specific input, display, and text compatibility fixes
    Focused model and canvas export tests

Validated by building with GCC 14 and WERROR=y, deploying on-device, and testing both color rendering and controls.

## Summary by CodeRabbit

- **New Features**
  - Added support for building and running on Kobo Nickel devices.
  - Added Kobo Clara BW and Clara Colour detection, display defaults, 300-DPI support, battery reporting, Wi‑Fi selection, and model-specific hardware handling.
  - Added FBInk-based framebuffer rendering and improved grayscale/color display conversion.
  - Added a higher-priority custom font location for Kobo devices.

- **Bug Fixes**
  - Improved input device handling and prevented invalid audio resource playback.
  - Improved device detection and compatibility across Kobo hardware.

- **Documentation**
  - Added Kobo Nickel build requirements, commands, output details, and graphics support documentation.


anj1:
I'll have a look at the build errors.

lordfolken:
Undefined symbols for architecture arm64:
  "DetectKoboModel()", referenced from:
      DisplaySettings::SetDefaults() in DisplaySettings.o

I think this needs to be gated.

anj1:
Addressed review feedback, including the OpenSSL cross-prefix handling, grayscale export safety, input-grab failure handling, and `VersionNumber` cleanup. I also gated `TestKoboModel` to Kobo targets, resolving the cross-platform `DetectKoboModel()` linker failures.

anj1:
Fixed the major/minor version clash, and the failing code paths for the older KOBO build.

anj1:
Fix for failing KoboRoot build: fix  `ExpandDitheredGreyscale()`

src/SystemSettings.cpp
```diff
@@ -17,10 +17,15 @@ SystemSettings::SetDefaults()
...
     devices[0].path = "COM1:";
+#else
+#ifdef TARGET_IS_KOBO_NICKEL
+    devices[0].path = "/dev/ttyS0";
+    devices[0].baud_rate = 9600;
 #else
     devices[0].path = "/dev/tty0";
-#endif
     devices[0].baud_rate = 4800;
+#endif
```
lordfolken: `/dev/ttyS0` at 9600 is the MediaTek Clara UART, not a Nickel-launch property. `KOBO_NICKEL` already defines `KOBO`, so this should follow battery, DPI, and display type and key off `IsKoboMediaTek()` / `DetectKoboModel()` at runtime.

A Nickel binary on a non-MediaTek Kobo (or a future KoboRoot Clara build) would otherwise get the wrong default port.

src/ui/canvas/memory/Canvas.cpp
```diff
@@ -224,6 +258,7 @@ RenderText(const Font *font, std::string_view text) noexcept
 
   return TextCache::Get(*font, text);
 }
+#endif
```
lordfolken: This disables `TextCache` for the whole Nickel binary and re-renders on every `DrawText` / `DrawTransparentText` / `DrawClippedText`. Nothing here is Nickel-specific.

If the problem is color memory-canvas, gate on `#ifndef GREYSCALE` (this target already sets `DITHER=n` / `GREYSCALE=n`). If `TextCache` is buggy, fix `TextCache`. As written this is a performance hit on every text draw, with no comment explaining why the cache cannot be used.

src/Audio/Sound.cpp
```diff
@@ -27,7 +27,7 @@
 #endif
 
 bool
-PlayResource(const char *resource_name)
+PlayResource([[maybe_unused]] const char *resource_name)
```
lordfolken: `[[maybe_unused]]` is fine for the Nickel build (`ENABLE_ALSA=n`, so this hits the `#else return false` path). While touching this, `PlayResource` should also reject a null or empty `resource_name` at function entry.

`eventPlaySound()` can pass `misc` through unchecked, and the Android/Win32 paths then call `strstr(resource_name, ...)`. A cheap `if (resource_name == nullptr || *resource_name == '\0') return false;` at the top would close that.

src/ui/canvas/memory/Canvas.cpp
anj1: I actually went further and decided to test if with the new changes and new FBInk backend, is this workaround even needed. After some testing it turns out it isn't. So I restored the `TextCache`. I did this as a fixup commit, you can squash if you like.
https://github.com/XCSoar/XCSoar/pull/2865/commits/d2411df93b180c943e60b38825c96b9a89972f7c

anj1:
Let me know if there's anything I can do to help progress this along.

PR #2878, opened 2026-08-16
Support for Polish Start, cleanup of Task types

### LINKED ISSUE #2389: Feature: Polish Start — declare task start inside start cylinder (PEV)
### Summary

Support competition starts where the start is a **start cylinder** and the pilot **starts the task from any position inside** that cylinder by pressing **PEV**, with **distance and speed** counted from that point (not only from exiting the cylinder after an outside start line crossing).

This matches the **“Polish Start”** style of task start used to reduce start gaggles and mid-air risk; it is also referenced for upcoming major contests (e.g. EGC/WGC trials).

### Background

- **Discussion:** https://github.com/XCSoar/XCSoar/discussions/2381  
- **Article (Nordic Gliding):** [“Polish Start”: Hur minskar vi risken för ”midairs”?](https://nordicgliding.com/polish-start-hur-minskar-vi-risken-for-midairs/)

Typical properties described in those sources:

- Start area is a **circle/cylinder** (e.g. ~10 km radius; task-dependent).
- Pilot presses **PEV** when ready **while still inside** the cylinder; that fixes the **start point** and starts the task clock.
- **Task distance** (and speed / scoring as defined by rules) is measured from that **in-cylinder** point, so the flown distance can differ from a nominal task depending on where inside the cylinder the pilot starts.
- This is distinct from behaviour where the task only starts when **leaving** the start cylinder after having entered it.

### Request

Add or complete XCSoar support so pilots can **start the task inside the start cylinder** via **PEV** (and related displays/navigation logic stay consistent with that start method), aligned with the rules and task formats where this applies.

### Notes

- Exact parameters (cylinder dimensions, PEV wait/start windows, height rules, RT vs AAT) come from the **task brief / rules**; the software should follow the same definitions as other start types.
- If parts of this already exist, this issue can be narrowed to the **missing** behaviour (e.g. explicit in-cylinder PEV start vs exit-only).

### FAI Sporting Code — Annex A (official rules text)

Authoritative document: **Annex A to Section 3 – Gliding** (FAI). Current edition on the FAI site includes e.g.  
https://www.fai.org/sites/default/files/sc3a_2025.pdf  
(Relevant sections: **§7.4** Starting.)

Annex A defines **two** start *types* (briefing + task sheet say which is used each day):

1. **Line Start (§7.4.3)**  
   - Cross a **Start Line** toward the first turn/area.  
   - Optional **PEV Start**: PEV **before** crossing the line; **PEV Wait Time** and **PEV Start Window** (each **5–10** minutes) published on the task sheet; penalty-free crossing in the resulting interval.

2. **Cylinder Start (§7.4.4)** — matches the in-cylinder PEV / “Polish start” *pattern*  
   - **Start** by recording **PEV on the primary flight recorder while inside** the **Start Cylinder**; time/place of start are normally **when/where PEV was pressed**.  
   - **Start cylinder:** centered on Start Point; radius on task sheet; **should not be less than 10 km**.  
   - **Start time:** latest **PEV inside** the cylinder, **or** if **no** PEV inside, latest **exit** from the cylinder.  
   - **Second/subsequent PEV inside** invalidates the **previous** start, resets the **10 minute** minimum interval between PEVs, cancels penalties on the previous start; **unlike** line start, the **latest** start is used for scoring.  
   - **Validity:** fix **inside** cylinder **after** start opens. If **no** PEV inside, valid start on **exiting** cylinder **with a penalty**; if **any** PEV inside was recorded, **exit alone is not** the start.  
   - **PEVs outside** the cylinder: **ignored**.  
   - **Maximum groundspeed at start**, **maximum loss of height** (start altitude − finish altitude): on task sheet; penalties if exceeded. **Marking distance** from **credited start point** (§7.4.4.6).

**IGC waiver:** §7.4.3.6 states that **Cylinder Start (§7.4.4) will not be used unless approved by a specific waiver from the IGC Bureau.** Implementation in XCSoar remains useful where this is authorized and for preparation/training.

### Relation to current XCSoar (for implementers)

- **PEV wait/window** in settings matches **Line Start / PEV Start (§7.4.3.4)** more closely than **Cylinder Start (§7.4.4)** (cylinder rules use **in-cylinder PEV** as primary, **10 min** between PEVs, exit-with-penalty fallback if no PEV, etc.).  
- Default **start cylinder** behaviour in docs is **exit** to count the start; **§7.4.4** adds **in-cylinder PEV** as the primary scoring case when that start type is in use.

chrgro:
Fixes #2389.

DRAFT.

Initial preview of work to enable clearer separation of start types, including support for Polish start.

* Splits out the "start" type selection into a separate dropdown in the task options. Selecting the task start type disables or enables various options surrounding the tasks. E.g. selecting FAI start/finish means only FAI waypoints are supported, Polish or PEV start shows PEV button but not Arm task button and so on.

* Adds support for Polish start: A start method where the start location can be anywhere within an observation zone, and the task is scored from that location.

I'm currently not entirely happy with the UX surrounding Polish start, it needs clearer markings on where the task was started from, update of navigation lines drawn and so on.

I also need to add substantial amount of help text.

image

image

This is currently not ready for merging.

---

---

- [ ] I'm ready to merge

## Summary by CodeRabbit

* **New Features**
  * Added a unified task start-mode selector for Normal, FAI, Pilot Event (PEV), and Polish starts.
  * Added Polish task-start support based on current flight conditions.
  * PEV start settings are now saved and restored with task configuration.
  * Updated navigation shortcuts for Calculator/Task Manager and Pilot Event actions.

* **Improvements**
  * PEV timing descriptions and Start Open information now use clearer terminology.
  * PEV controls appear only when the PEV start mode is selected.

### DIFF
--- a/Data/Input/default.xci
+++ b/Data/Input/default.xci
@@ -662,10 +662,19 @@ location=1
 
 mode=Nav1
 type=key
-data=6
+data=2
 event=Calculator
 event=Mode default
 label=Task Manager
+location=4
+
+mode=Nav1
+type=key
+data=6
+event=StatusMessage Pilot event announced
+event=Logger note PEV
+event=PilotEvent
+label=Pilot Event Announce
 location=5
 
 mode=Nav1
@@ -730,16 +739,6 @@ location=6
 mode=Nav2
 type=key
 data=8
-event=Mode default
-event=StatusMessage Pilot event announced
-event=Logger note PEV
-event=PilotEvent
-label=Pilot Event Announce
-location=7
-
-mode=Nav2
-type=key
-data=9
 event=Setup Target
 event=Mode default
 label=Target$(CheckTask)$(CheckTaskResumed)
--- a/Data/Input/default.xci
+++ b/Data/Input/default.xci
@@ -674,8 +674,8 @@ data=6
 event=StatusMessage Pilot event announced
 event=Logger note PEV
 event=PilotEvent
-label=Pilot Event Announce
-location=5
+label=$(PilotEventAnnounce)
+location=7
 
 mode=Nav1
 type=key
@@ -689,7 +689,7 @@ type=key
 data=8
 event=AdjustWaypoint nextarm
 label=$(WaypointNextArm)
-location=7
+location=5
 
 mode=Nav1
 type=key
--- a/src/Dialogs/Settings/Panels/TaskRulesConfigPanel.cpp
+++ b/src/Dialogs/Settings/Panels/TaskRulesConfigPanel.cpp
@@ -106,14 +106,14 @@ TaskRulesConfigPanel::Prepare(ContainerWindow &parent,
   SetExpertRow(spacer_3);
 
   AddDuration(_("PEV start wait time"),
-              _("Wait time in minutes after Pilot Event and before start gate opens. "
+              _("Wait time in minutes after Pilot Event (PEV) and before start gate opens. "
                 "0 means start opens immediately."),
               {}, minutes{30}, minutes{1},
               task_behaviour.ordered_defaults.start_constraints.pev_start_wait_time);
   SetExpertRow(PEVStartWaitTime);
 
   AddDuration(_("PEV start window"),
-              _("Number of minutes start remains open after Pilot Event and PEV wait time."
+              _("Number of minutes start remains open after Pilot Event (PEV) and PEV wait time."
                 "0 means start will never close after it opens."),
               {}, minutes{30}, minutes{1},
               task_behaviour.ordered_defaults.start_constraints.pev_start_window);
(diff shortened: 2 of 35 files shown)


marek1313:
With the new start rules introduce also "Maximum Loss of Height" between start and finish which set to different values depending on conditions - it is not only 1000m as in FAI task. So the setting should be made available together with the start inside sector function. In Poland we also use sometimes this loss of height for other task types - so the setting should be visible and working for other types of tasks. 
I think introducing start types is cleaner architecture then mine: https://github.com/marek1313/XCSoar-fork-for-PEV-start). But you skip going through "SmartTaskAdvance" - not sure if it matters at all. When I prepared the implementation for Polish competitions, there was possibility to press PEV outside cylinder and start when first fix inside was registered. Now FAI regulation do not allow this so the whole logic is not needed and task may be started directly on PEV only when in cylinder.
Additionally, LX software, for example kept the "PEV window" start options within "Normal start" option. Only added start inside sector for this type new regulation. I would rather keep only 3 start types: NORMAL, FAI, POLISH. Then the "windowed" start but with normal procedure of exiting sector/crossing line apply - then XCSoar would work in similar way.

PR #2884, opened 2026-08-17
iOS: Remove white 1px frame around full-screen dialogs

WolfgangDrescher:
On iOS, full-screen dialogs like the Fly/Simulator prompt showed a thin bright frame at the display edges. The usable area of the main window was recomputed from the screen dimensions and truncated, which can end up one pixel off from the actual framebuffer size, so a dialog sized from it was no longer recognised as filling its parent and got a raised border drawn around it. The usable area now comes from the window's real framebuffer size, the window size rounds instead of truncating, and the check whether a window fills its parent uses the parent's usable area rather than its raw size, which changes nothing on any other platform.

Two questions: is `std::lround()` right for the window size, or `std::ceil()`? And should the safe-area insets in `GetClientRect()`, still truncated with `static_cast<int>()`, use the same?

| Before | After |
|-|-|
| Bildschirmfoto 2026-08-17 um 10 27 46 | Bildschirmfoto 2026-08-17 um 10 40 29 |

(it's better visible on a real device)

## Bug Fixes
- Fixed Android crashes.
- Full-screen iOS dialogs now extend cleanly to display edges without thin white frames.
- Improved iOS window sizing, pixel alignment, safe-area handling, and maximized-window detection.
- Corrected USB HID device re-enumeration.
- Fixed SkySight metadata downloads.
- Improved high-DPI map-label rendering.
- Corrected ADS-B speed handling.
- Fixed Checklist keyboard navigation.

## Documentation
- Updated release workflow documentation.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -2,6 +2,10 @@ Version 7.46 - not yet released
 * android
   - fix crash when exiting the app #2873
   - fix crash when restarting the app immediately after exit
+* iOS
+  - remove the thin white frame around full-screen dialogs (such as
+    the Fly/Simulator screen), which no longer stop one pixel short of
+    the display edge
 * input
   - fix USB HID remotes (e.g. SteFly) that stop sending keys on
     OpenVario after the stick re-enumerates #2820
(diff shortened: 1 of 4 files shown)


WolfgangDrescher:
I have addressed rabbit's feedback in 608cdab. No force push this time to avoid making cherry-picking harder for you.

PR #2911, opened 2026-08-21
Replay: seek to a flight minute or flight phase, and make +10' a real skip

### LINKED ISSUE #933: replay +10' not working

WolfgangDrescher:
Closes #933.

Adds seek to a chosen flight minute and seek to the next circling or cruise phase, and turns the +10' button into a real ten minute skip instead of fast forward mode, as asked for in #933. The flight minute seek picks up the idea from the mockup in #2472. Each seek applies every fix through merge and calculation, so trail, statistics and phases stay consistent, and it only restarts the recording when the target lies behind the current position.

Every seek scans inline for up to 500 ms, so short ones show no dialog. Only when that budget runs out does the rest move to a worker thread and the progress dialog appears, with a bar and a cancel button.

Bildschirmfoto 2026-08-21 um 14 32 37
Bildschirmfoto 2026-08-21 um 14 33 05

Two questions for reviewers: 19a637f fixes the cached clock in the SDL event queue, where timers never became due while events kept arriving, which is why the progress bar stayed empty on macOS and iOS. It is an independent fix in shared code, so can it go in without putting the rest of XCSoar at risk? And 4f88e3b still contains a fair amount of LLM written code: is this an acceptable way to implement the seek logic in `src/Replay/Replay.cpp`, especially suspending and resuming the worker threads around the scan? I am not yet confident about thread handling in C++, so a close look there would be welcome.

I have not yet got around to testing this on top of #2819.

## Summary by CodeRabbit

* **New Features**
  * Added replay seeking by elapsed flight time.
  * Added controls to jump forward 10 minutes or move to the next circling or cruise phase.
  * Added progress reporting and cancellation for longer seek operations.
  * Added Lua support for replay seeking.
* **Bug Fixes**
  * Improved replay restart handling by resetting flight state.
  * Improved replay accuracy when seeking through recorded flights.
  * Improved event timing and responsiveness during event processing.
  * Improved handling of replay speed and interpolated positions during seeking.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -82,12 +82,18 @@ Version 7.46 - not yet released
 * ui
   - keep pan-mode menu buttons about 20 mm tall on tall phones,
     large enough for a gloved tap
+* replay
+  - add seek to a chosen flight minute or to the next circling or cruise
+    phase, showing progress and allowing cancellation
+  - skip 10 minutes instead of fast-forwarding (+10' button) #933
 * documentation
   - describe the T Next Leg InfoBox and how to use it at a turn
 * development
   - documentation: document the release and post-release version workflow
   - macOS/iOS: read TESTING=y from darwin/.env, so Xcode builds can
     produce the testing flavour
+  - lua: add replay seek functions seek_to_flight_minutes,
+    seek_to_next_circling and seek_to_next_cruise
 
 Version 7.45 - 2026-08-15
 * highlights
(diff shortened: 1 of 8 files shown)


lordfolken:
Nice!
Also checkout https://github.com/XCSoar/XCSoar/pull/2910/changes/90e3124bbf0240f8ca532646f2b9aa96faf4f211 which will enable to have a replay at 10x speed without skipping any data.

Issue #2905, opened 2026-08-21
Kobo- Wind not being calculated


ulipo:
XCSoar versions having the problem
-------------------------------------------------
v7.44 and v7.45

XCSoar versions not having the problem
-------------------------------------------------
v7.43

System information
------------------
tested on a Kobo GloHD with BluefLy vario and another GloHD with GoFly vario

Steps to reproduce the behavior
-------------------------------
- flash v7.43 on a Kobo
- set up a wind infobox and set wind calculation to be:
  - Circling: ON
  - Zig Zag: ON
  - External: OFF
  - Trail drift: ON
- replay a track at 1x speed (e.g. [this one attached](https://github.com/user-attachments/files/31292933/2026-08-19-XCS-UPE-01.txt), rename the file to .igc)
- wait a few minutes to see the wind arrow appear and the wind value being calculated 
- flash v7.44 or v7.45 on the same device
- repeat the above steps
- see the wind not being calculated

Expected behavior
-----------------
the wind arrow and intensity should appear both near the aircraft symbol and in the infobox

Actual behavior
---------------
No wind arrow appears and the infobox remains at "---".
Tried also with a fresh new profile, made no difference.

Do you have any idea what may have caused this?
-----------------------------------------------

Do you have an idea how to solve the issue?
-------------------------------------------

lordfolken:
The 7.44 rewrite estimates wind from a cosine fit of ground speed, then throws the circle away if GPS track rate is not steady.

That gate is wrong for GPS-only aircraft in a decent wind. Track rate is not constant when wind is a large fraction of airspeed (typical PG). The attached IGC had plenty of full circles; all of them died on that check. 7.43 used min/max ground speed and did not care about a round GPS rate.

What this does in `CirclingWind.cpp`:

- If there is no TAS and no gyro, skip the roundness reject and score quality from the cosine fit.
- If TAS or a gyro is present, the old gate stays.
- Also loosen the 5% / 2 s sample-spacing checks a bit; they were tighter than replay and phone GPS jitter.

The other commits are only so IGC replay still feeds ~1 Hz GPS (especially at 10× on a fast LCD). E-paper / slow CPUs keep the existing calculation-thread limit.

Fix: https://github.com/XCSoar/XCSoar/pull/2910

PR #2919, opened 2026-08-22
test: Add GPS track-bucket wind prototype

lordfolken:
## Summary

Prototype of discussion #2918: a GPS-only wind estimate from ground speed versus ground track. It is **test-only** (`test/src/BucketWind.*`) and is **not** wired into Auto wind or the flight computer.

Circling wind needs a detected thermal and a fairly round GPS circle. This method only needs a spread of tracks — a partial arc or a slow cruise heading change is enough.

## How the prototype works

Ground speed around the compass looks like a cosine: fastest over the ground is downwind, slowest is upwind. The usual light-wind model is

```text
ground_speed(track) ≈ TAS + wind × cos(track − downwind)
```

The code keeps **72 bins of 5°**. Each GPS fix overwrites the bin for that track (latest sample wins). `Fit()` ignores bins older than **180 s**, then least-squares-fits that cosine. Wind **from** is the opposite of the downwind peak.

```mermaid
flowchart TD
  GPS["GPS fix: time, track, ground speed"] --> Update
  Update["Update: write GS into the 5° track bin"] --> Bins["72 bins around the compass"]
  Bins --> Fit["Fit: drop bins older than 180 s"]
  Fit --> LS["Least squares: GS ≈ TAS + b·cosθ + c·sinθ"]
  LS --> Speed["Wind speed = hypot(b, c)"]
  LS --> Down["Downwind = atan2(c, b)"]
  Down --> From["Wind FROM = reciprocal of downwind"]
```

`Fit()` also rejects the result if fewer than 12 bins are fresh, the 3×3 system is singular (tracks too clustered), TAS is below 1 m/s, or wind is outside 0.1–30 m/s.

This is still a toy: latest sample per bin, three-minute expiry, and the small-wind `GS ≈ TAS + W·cos` model (a bit high when W/V is large, as on the PG circle below).

## Compile

UNIX debug build of the comparison test only:

```bash
make -j$(nproc) TARGET=UNIX USE_CCACHE=y WERROR=y output/UNIX/bin/TestBucketWind
```

The binary is `output/UNIX/bin/TestBucketWind`. Object files go under `output/UNIX/dbg/`.

## Run tests

```bash
output/UNIX/bin/TestBucketWind
```

`TestBucketWind` is also part of `make check`. It feeds the same synthetic GPS traces to `CirclingWind` and `BucketWind`.

Typical result on current master:

| Scenario | CirclingWind | Buckets | Truth |
|---|---|---|---|
| PG full circle (9 m/s TAS, 5 m/s from E) | no wind | 5.3 m/s / 091° | 5.0 / 090° |
| Glider circle (25 m/s TAS, 5 m/s from W) | 4.9 m/s / 266° | 5.0 m/s / 270° | 5.0 / 270° |
| PG 220° heading arc | no wind | 4.7 m/s / 089° | 5.0 / 090° |
| Cruise 360° sweep (not circling) | no wind | 6.0 m/s / 358° | 6.0 / 000° |

The PG full-circle case is the #2905 gap: circling wind stays quiet; buckets still recover the wind.

## Test plan

- [ ] `make -j$(nproc) TARGET=UNIX USE_CCACHE=y WERROR=y output/UNIX/bin/TestBucketWind`
- [ ] `output/UNIX/bin/TestBucketWind` (12 TAP tests)
- [ ] Read `test/src/BucketWind.hpp` / `.cpp` against the diagram above

### DIFF
--- a/build/test.mk
+++ b/build/test.mk
@@ -90,7 +90,8 @@ TEST_NAMES = \
 	TestValidity TestUTM \
 	TestAllocatedGrid \
 	TestRadixTree TestGeoBounds TestGeoClip \
-	TestLogger TestGRecord TestClimbAvCalc TestFilteredVarioComputer \
+	TestLogger TestGRecord TestClimbAvCalc TestBucketWind \
+	TestFilteredVarioComputer \
 	TestVarioSynthesiser TestAudioVario \
 	TestWaypointReader TestThermalBase \
 	TestFlarmNet TestFlarmMessaging \
@@ -645,6 +646,16 @@ TEST_CLIMB_AV_CALC_SOURCES = \
 	$(TEST_SRC_DIR)/TestClimbAvCalc.cpp
 TEST_CLIMB_AV_CALC_DEPENDS = MATH
 $(eval $(call link-program,TestClimbAvCalc,TEST_CLIMB_AV_CALC))
+
+TEST_BUCKET_WIND_SOURCES = \
+	$(SRC)/Computer/Wind/CirclingWind.cpp \
+	$(SRC)/Atmosphere/AirDensity.cpp \
+	$(TEST_SRC_DIR)/FakeLogFile.cpp \
+	$(TEST_SRC_DIR)/tap.c \
+	$(TEST_SRC_DIR)/BucketWind.cpp \
+	$(TEST_SRC_DIR)/TestBucketWind.cpp
+TEST_BUCKET_WIND_DEPENDS = LIBNMEA GEO MATH UTIL TIME FMT UNITS
+$(eval $(call link-program,TestBucketWind,TEST_BUCKET_WIND))
 
 TEST_FILTERED_VARIO_COMPUTER_SOURCES = \
 	$(SRC)/Atmosphere/AirDensity.cpp \
--- a/test/src/BucketWind.cpp
+++ b/test/src/BucketWind.cpp
@@ -1,0 +1,147 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright The XCSoar Project
+
+#include "BucketWind.hpp"
+
+#include <cmath>
+
+namespace {
+
+constexpr unsigned MIN_BINS = 12;
+constexpr double MIN_TAS = 1;
+constexpr double MIN_WIND = 0.1;
+constexpr double MAX_WIND = 30;
+
+[[gnu::const]]
+double
+Det3(double a00, double a01, double a02,
+     double a10, double a11, double a12,
+     double a20, double a21, double a22) noexcept
+{
+  return a00 * (a11 * a22 - a12 * a21)
+       - a01 * (a10 * a22 - a12 * a20)
+       + a02 * (a10 * a21 - a11 * a20);
+}
+
+/**
+ * Least-squares fit of y = mean + b·cosθ + c·sinθ.
+ *
+ * The 3×3 system is the normal equations for those three unknowns.
+ * Returns false if the tracks do not span enough angles (singular).
+ */
+bool
+FitMeanCosSin(double n,
+              double sum_cos, double sum_sin,
+              double sum_cos2, double sum_sin2, double sum_cos_sin,
+              double sum_y, double sum_y_cos, double sum_y_sin,
+              double &mean, double &b, double &c) noexcept
+{
+  const double det = Det3(n,       sum_cos,     sum_sin,
+                          sum_cos, sum_cos2,    sum_cos_sin,
+                          sum_sin, sum_cos_sin, sum_sin2);
+  if (std::fabs(det) < 1e-9)
+    return false;
+
+  /* Cramer's rule: replace one column at a time with the right-hand
+     side (sum_y, sum_y_cos, sum_y_sin). */
+  mean = Det3(sum_y,     sum_cos,     sum_sin,
+              sum_y_cos, sum_cos2,    sum_cos_sin,
+              sum_y_sin, sum_cos_sin, sum_sin2) / det;
+
+  b = Det3(n,       sum_y,     sum_sin,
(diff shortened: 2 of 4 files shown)


bomilkar:
@ulipo do you have any testcases for this?

Ideally NMEA protocols, but IGC will do as well. However, please include notes on what results can be expected from the wind calculation.

ulipo:
I have just .igc files, this one should be easy to test:
[2026-06-20-XCT-UPE-11.igc](https://github.com/user-attachments/files/31384326/2026-06-20-XCT-UPE-11.igc.txt)
the wind should be about 1-4 km/h from east at the beginning, then it should rapidly rotate from south and south west, with increased speed and a peak of about 15 km/h during the longest thermal

This other one contains more thermals, light wind in the range of 3-10 km/h rotating from multiple directions during the flight:
[2026-08-23-XCT-UPE-09.igc](https://github.com/user-attachments/files/31384456/2026-08-23-XCT-UPE-09.igc.txt)

bomilkar:
Hi @ulipo 
> I have just .igc files, ...

These 2 IGCs are hopeless. Replay these and watch ground speed and track. It jumps all over the place. I cannot think of any algorithm which would calculate a reasonable wind from that data.

Can you use an Android phone on your next flight? If so, I would like you to install a test version of xcsoar and turn on the NMEA logger (Config -> Setup -> Logger -> NMEA_Logger = On). This "test version" is the 7.45 release including PR https://github.com/XCSoar/XCSoar/pull/2616 .
It produces a NMEA log from the built-in GNSS (if you have this device enabled). The GPRMC/GPGGA records have ground speed and track right from the GNSS receiver. That is much cleaner.
If you want, you can build the APK yourself or I'll send you a link to an APK to install.

ulipo:
If you have already an .apk at hand, I'd be happy to test it, but at this moment I don't have the time to compile it.

In the meanwhile, try this other .igc recorded with a different instrument to see if it is any better:
[2026-05-09-XCS-UPE-01.igc](https://github.com/user-attachments/files/31425054/2026-05-09-XCS-UPE-01.igc.txt)
Should sport wind mostly from SW between 5 and 15 km/h.

bomilkar:
@ulipo 
> If you have already an .apk at hand, I'd be happy to test it, but at this moment I don't have the time to compile it.

https://magentacloud.de/s/9HcJkxF7kxoWT3X

> In the meanwhile, try this other .igc recorded with a different instrument to see if it is any better...

This IGC looks much better. Still not sure whether it's good enough for wind calculation. I'll test it.
The problem with IGCs is that they only contain the coordinates of the position fixes. The GNSS receiver has much better means of determining track and speed than the first derivative of the position.

bomilkar:
> Should sport wind mostly from SW between 5 and 15 km/h.

During the first circling phase of that flight it calculates the wind to 19 km/h, 211°

BTW: not sure if you know: "ZigZag Wind" requires airspeed data from a real instrument (not calculated from ground speed and wind). At least the IGC doesn't have that.

Here is the result of RunCirclingWind for that IGC:
[2026-05-09_CirclingWind.txt](https://github.com/user-attachments/files/31452580/2026-05-09_CirclingWind.txt)
Is that what you would expect?

ulipo:
> BTW: not sure if you know: "ZigZag Wind" requires airspeed data from a real instrument (not calculated from ground speed and wind). At least the IGC doesn't have that.
I don't have any Pitot or specific wind measurement instrument, it's only GPS for me, I'll disable that setting.

> Here is the result of RunCirclingWind for that IGC: [2026-05-09_CirclingWind.txt](https://github.com/user-attachments/files/31452580/2026-05-09_CirclingWind.txt) Is that what you would expect?
I think the direction is OK, intensity is a bit on the high side, specially between 13.29.54 and 13.31.31; with a paraglider ~25 km/h is a lot of wind, I don't remember to have withstanded so much wind in that flight.

I installed the .apk, next flight I'll try to record some NMEA logs.

bomilkar:
Here is what the new "BucketWind" algorithm produces with the IGC dated 2026-05-09:
[2026-05-09_RunBucketWind.txt](https://github.com/user-attachments/files/31507786/2026-05-09_RunBucketWind.txt)

@ulipo please compare this to the result using CirclingWind. Which one is more realistic? When does it work better/worse?
Is the calculated TAS data realistic? If so, the BucketWind algo might not work well.

ulipo:
> @ulipo please compare this to the result using CirclingWind. Which one is more realistic? When does it work better/worse? Is the calculated TAS data realistic? If so, the BucketWind algo might not work well.

Sorry for the delay, I was sure I had already replied to this message...
The wind seems OK, TAS is almost OK except for some high and low peaks (paraglider TAS is almost fixed between about 25-40 km/h, we are sky snails).

Here attached the .nmea and .igc logs of today's flight, recorded with your provided .apk on my android phone:
[2026-08-30.zip](https://github.com/user-attachments/files/31617322/2026-08-30.zip)
The wind should be in the range of 3-10 km/h with varying directions, mostly from north at the beginning and from south at the end.

bomilkar:
@ulipo 
> Here attached the .nmea and .igc logs of today's flight, recorded with your provided .apk on my android phone: [2026-08-30.zip](https://github.com/user-attachments/files/31617322/2026-08-30.zip) The wind should be in the range of 3-10 km/h with varying directions, mostly from north at the beginning and from south at the end.

The nmea data is quite good! But the results don't look good at all.
The assumption you are flying at constant airspeed is not valid at all.
I'd like to discuss with you a bit deeper and don't want to "pollute" this PR here. Please reply to this email if you agree: ronald.niederhagen@gmail.com

It's not hopeless. Does your smartphone have a gyro? You can use an app like "Sensor Test" to find out.
We have to use more than just GPS data to calculate wind.

lordfolken:
Closing as superseded by #2984.

Issue #2912, opened 2026-08-21
Alternates target : Manual vs Auto : inverted?


mlep:
@B-apt : Thanks for contributing this new feature!

XCSoar versions having the problem
-------------------------------------------------

7.45

XCSoar versions not having the problem
-------------------------------------------------

7.44 (feature was not included)

System information
------------------

Android

Steps to reproduce the behavior
-------------------------------

By clicking on the Alternate Infobox, a menu is displayed.
One of the buttons allows to switch between "Auto" to "Manual" modes.
When this button displays "Manual", the "Choose Alternate" button is active.
When this button displays "Auto", the "Choose Alternate" button is inactive (= grayed out).

Expected behavior
-----------------

In XCSoar, it seems to me that the standard convention in the GUI is that the label of a button tells the user what will happen if this button is clicked on. Example: "Config. page 2/3" will get you to page 2, implying you are on page 1.

So, in the Alternate page, I expect:
- "Choose Alternate" to be active when the button setting the mode displays "Auto".
- "Choose Alternate" to be inactive when the button setting the mode displays "Manual".

Actual behavior
---------------

Currently, the behaviour seems inverted:
- "Choose Alternate" is active when the button setting the mode displays "Manual".
- "Choose Alternate" is inactive when the button setting the mode displays "Auto".

Do you have any idea what may have caused this?
-----------------------------------------------

A (common) misunderstanding in the way the GUI is designed.

Do you have an idea how to solve the issue?
-------------------------------------------

Invert the button labels (i.e. displays "Auto" instead of "Manual", and vice-versa)

lordfolken:
The Auto/Manual control was showing the **current** mode (`Altn mode: AUTO` / `Altn mode: MANUAL`), while XCSoar buttons usually name the **next action**. That is why Manual on the button plus an enabled Choose Alternate felt backwards.

Inverting those two captions would have made it worse (`Altn mode: AUTO` while actually in MANUAL). The enable-gate on Choose Alternate was also the source of a two-tap path: switch to Manual, then select — even though Select as Alternate already sets MANUAL.

#2921 changes the model instead of swapping labels:

- **Select as Alternate** is always available (with a row selected). It pins the field and switches that InfoBox to MANUAL.
- **Alternate Mode: Auto** only returns the slot to the computed list. It is enabled in MANUAL and greyed in AUTO.

[...]

PR #2921, opened 2026-08-22
Dialogs/Task/AlternatesList: Pin a field without a mode toggle

lordfolken:
## Summary
- Let **Select as Alternate** pin a landing field from AUTO (it already switches the slot to MANUAL).
- Replace the Auto/Manual toggle with **Alternate Mode: Auto**, enabled only while a field is pinned, greyed when the slot is already following the computed list.
- Avoids the inverted-looking labels in #2912 without a next-action toggle that would bring back the two-tap path.

## Test plan
- [ ] Open Alternate 1 in AUTO: **Alternate Mode: Auto** is greyed; **Select as Alternate** is enabled with a row selected
- [ ] Select as Alternate pins the field, InfoBox title shows MAN, dialog closes
- [ ] Reopen: **Alternate Mode: Auto** is enabled; tapping it returns the slot to AUTO (InfoBox title AUTO)
- [ ] GoTo still navigates and aborts the task
- [ ] Alternate 2 is independent of Alternate 1

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -45,6 +45,9 @@ Version 7.46 - not yet released
     #2903
 * dialogs
   - let Shift type '_' on XCSoar's on-screen keyboard
+  - let Select as Alternate pin a landing field from AUTO, and use
+    Alternate Mode: Auto only to return the Alternate InfoBox to the
+    computed list #2912
   - fix keyboard wrap in the Checklist: Down from Close returns
     to the list and highlights an item so Enter acts on that row
     #2862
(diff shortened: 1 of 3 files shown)


B-apt:
Thanks for this evolution, I will test this in the next day 👍 
Do you think it should have any impact on what we decided in #2730 ?

lordfolken:
> Thanks for this evolution, I will test this in the next day 👍 Do you think it should have any impact on what we decided in #2730 ?

Hey, this is really to address the bug that was raised in the minor release that will go out soonish (7.45.1).  (And also fix the color in the infobox  for good/bad glide ratio.)
Its meant to be minimal and just address the issue, not be a redesign or feature extension unlike your pr 2370.

I don't mind if you replace it all again, or find a better solution.

PR #2961, opened 2026-08-26
Remove Airspace padding flickering

### LINKED ISSUE #2479: Airspace shading / padding is spuriously disappearing
When the map moves during flight, or when panning, there is an effect where airspace shading / padding appears and disappears as the position of the map changes.

This is an old bug that has annoyed me for years. Since I am currently working on the airspace rendering for the clearance business, I decided to also look into this one finally, and did some digging last night to find the cause.

First an illustration of the problem: (Example with London City airport airspace): The difference from left to right picture is that the map has been panned a bit to the left. As a result the filling of the airspace re-appears (The correct state is with filling visible).

Image

Another example is London TMA 4 - Again the difference between both images is just horizontal panning:

Image

I think I have localized the problem to the polygon triangularization for rendering in OpenGL. The triangularization can fail if the polygon becomes degenerate (self-intersecting or self touching).

In particular, the triangularization loop will exit in that case in `AirspaceRendererGL.cpp`:

```
if (heat++ > num_points) {
  // if polygon edges overlap we may loop endlessly
  delete[] next;
  return 0;
}
```

It is relatively easy to see for the first example: The airspace is clipped to 1.1x the screen rectangle before triangularization, and in the first example this leads to two separate areas connected by a "line" of overlapping airspace boundaries, where points from two different sides of the polygon lie on the same vertical line which leads to failure of triangularization.

The second case is less obvious, but I thing it is caused by something similar, just not due to clipping this time, but rather due to rounding. If you look at the London TMA airspace from the OpenAir file in detail (OpenAIP version), you find this:

Image

Here, it seems that rounding can lead to the points 11-14, especially 12 and 13 to generate a back-and-forth with a point on the same line, which will also lead to failure of triangularization. This is supported by the fact that when zooming in more, the effect disappears.

I did not check all the other airspaces that "blink" to see if these two effects explain all these "blinkings", but it seems likely.

I see the following possible solutions:

1. Do not clip to screen (will only help for the first example)
2. Use a different triangularization algorithm
3. Do some preprocessing of the points after clipping but before triangularization to remove degeneracies
4. Use a different filling algorithm that does not require area-conserving triangularization, but instead uses e.g. triangle fans and an odd-even filling strategy

So far there is not a clear path for me yet. Solution 2. would be most clean, have properly researched yet how easy it would be to find one.

- Not sure what the performance impact of 4. would be.
- I have an idea how to do 3. at least to avoid the first example type problem, but will have to think how to make it work for the second example.

Anyone with a truly elegant idea, speak up...

kobedegeest:
Fixes #2479.

Related: #3127 (OpenGL map-renderer architecture; this is the airspace fill child, not reach #3131).

OpenGL airspace fill uses even-odd stencil instead of `PolygonToTriangles`, so clipped or self-touching polygons keep their padding while the map pans.


lordfolken:
@coderabbitai review

groundhogxc:
It looks pretty good. Overall, I like it. It definitely fixes the flickering, and is implemented in the expected way plus some cleaner and simpler vertex handling.
(I only tested UNIX and Windows (OpenGL) so far though. Android will be next.

* The one thing I personally do not liek (visually) is the beveling. I prefer the previous mitered rendering style of concave edges; although there could be a limit is beveling extreme acute angles. But right now it is beveling a bit too aggressively for my taste. But that might be a taste question. The official ICAO maps I looked at do neither of the two, they to a rounded boundary in all cases. But that would require some kind of drawing with a circular large pen, which we do not really have available.

Here the more technical issues I see:

* This approach does not work if no stencil buffer is available, or if not enough bits are there for stencil buffer. In GLES2, stencil buffers with 8bit are not guaranteed. I suspect that for all relevant platforms, there are 8 bits though. Nonetheless, it would be cleaner to keep the old algorithm as a fallback when the stencil buffer properties are insufficient.
* Heap allocation: The new algorithm grows lists on the heap, so it generates a lot of heap turnover. According to coding guidelines, this should be avoided in favor of static allocation. Especially since this happens once on every frame. I think you could just do it with static arrays just like the original. I suppose there are more arrays now, but I think that should be fine still in terms of static memory use. I guess doing that would bring back the previously existing max number of vertices per airspace limit, but that limit was 64000, and the largest airspace I found so far has just over 3000 points. So that appears to be not a problem.
* You change the GL state, but do not restore it at the end.

I think these should be addressed at minimum. Switching to static would also remove the eternal "noexcept" discussions with the rabbit :)

Also:
* The OUTLINE_STENCIL does not actually get properly used, it seems. That apparently was already the case in master, you doid not introduce this (I wondered about when working on the airspace clearance branch); but now that you explicitly name the buffers, maybe that should be fixed now
* The stencil buffer is prepared even if the fill mode is NONE. No problem, just unnecessary operations.

It seems that CPU load is reduced a bit with this algorithm, too, while GPU load increases on OpenGL. But I suspect that, so far still the CPU is going to be typically limiting even for the weaker GPU devices, so that at leas is "the right direction", and probably, if necessary, we could optimize more to reduce GPU usage as well. But that would be outside of this PR's scope, since the increase is not that large.

Nonetheless, just to make sure this should also be tested on Cubieboard (OpenVario) and other weak devices.

lordfolken:
@kobedegeest can you address @groundhogxc concerns?

Issue #1836, opened 2025-08-02
Smart ring for monitoring vital signs


Blaubart:
I've been experimenting a bit with a smart ring from China. It can measure blood oxygen saturation, heart rate, and blood pressure. The measurement results seem reasonably reliable. I can connect it to my Android device via Bluetooth and select it as a BLE sensor in XCSoar. The NavBox heart rate already exists, and my heart rate is displayed there. Unfortunately, blood oxygen saturation isn't available yet, but I think this NavBox would be even more important.

There are two other problems, though. I set the ring to a refresh interval of 10 minutes using an app on my smartphone. Unfortunately, it doesn't allow for any less. For us, however, an update interval of no more than 1 minute would certainly be useful. If no measured values are transmitted, the connection between the ring and XCSoar is lost.

So, a driver would have to be written in which the refresh interval can be set and which maintains the connection permanently, or establishes it once within the interval and queries the value.

I Think that would be a great Feature for XCSoar!!

A ring for monitoring vital signs would have huge advantages for us glider pilots. It's completely unobtrusive and can be worn under a glove when flying in the mountains! Reliable monitoring would therefore be possible!

lordfolken:
I like the idea. A new navbox with a different value would be easy. 
What i haven´t found is a blood saturation bluetooth service id.

Blaubart:
That sounds very good! I’m on holidays. When I’m back I will see if I could find out how to adjust measurement interval

Blaubart:
I found some information at the internet. Measurement interval can only be set to multiples of 5 minutes. But there are some commands for the Android App to connect the ring, start a measurement and close the connection again:

1. Set up a listening
All blood oxygen related data will be called back through CRPBloodOxygenChangeListener.
bleConnection.setBloodOxygenChangeListener(CRPBloodOxygenChangeListener);

2. Start measuring blood oxygen
bleConnection.startMeasureBloodOxygen();

3. Stop measuring blood oxygen
Stop measuring blood oxygen. If the measurement time is too short, there will be no measurement result. The
result is passed through the CRPBloodOxygenChangeListener.onBloodOxygen() callback.
bleConnection.stopMeasureBloodOxygen();

I have no idea if these information are helpful in our case. But if the driver is able to start the connection, measure the value and close connection again, we could monitor Blood Oxygen Saturation every 30, 60, 90s for example.

alenz-68:
[...]

PR #1966, opened 2025-12-05
Renderer/InfoBoxes: Add outline to infobox values and units for non-b…

lordfolken:
…lack/white colors

Add outline rendering support for infobox values and units when their colors are not black or white. This improves readability of colored text on various backgrounds.

Changes:
- Extended RenderShadowedText() to accept custom text and outline colors
- Added outline support to infobox values in PaintValue()
- Added outline support to infobox units via UnitSymbolRenderer::Draw()
- Outline thickness is adaptive: thicker for small fonts (≤16px) for better visibility, thinner for larger fonts
- Outline color is automatically chosen based on background brightness
- All changes are DPI-aware using font height calculations

Your PR should:
  * all description should be in the commit messages (no text in the pr)
  * your pr should be rebased on the current HEAD
  * one commit per atomic feature (every commit should build)
  * no fixup commits
  * pass all the compile and CI targets

For coding, style guide, architecture information please see our development guide:
https://xcsoar.readthedocs.io/en/latest/index.html

## Summary by CodeRabbit

* **Style**
  * Enhanced text rendering with improved outlines for better readability and visual contrast in info boxes and unit symbols.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #2971, opened 2026-08-27
Look/InfoBoxLook: make the caution colour readable on light backgrounds

### LINKED ISSUE #2970: InfoBoxLook: colour index 4 (yellow) is unreadable on the default look
## Problem

`InfoBoxLook` maps colour index 4 to pure yellow:

https://github.com/XCSoar/XCSoar/blob/master/src/Look/InfoBoxLook.cpp#L47

```cpp
colors[4] = inverse ? COLOR_INVERSE_YELLOW : COLOR_YELLOW;
```

with `COLOR_YELLOW = Color(0xff, 0xff, 0x00)` and `COLOR_INVERSE_YELLOW = COLOR_YELLOW` (`src/ui/canvas/Color.hpp`).

The InfoBox background is white unless the inverse look is active (`src/Look/InfoBoxLook.cpp`):

```cpp
background_color = inverse ? COLOR_BLACK : COLOR_WHITE;
```

Pure yellow on white is a contrast ratio of **1.07 : 1**. That is not "hard to read", it is effectively invisible. On the inverse look the same colour is fine.

| Colour | on white (default look) | on black (inverse look) |
| --- | --- | --- |
| normal text | 21.0 : 1 | 21.0 : 1 |
| `colors[1]` red — `#FF0000` / `#FF7070` | 4.0 : 1 | 7.8 : 1 |
| `colors[3]` green — `#00C000` / `#00FF00` | 2.46 : 1 | 15.3 : 1 |
| **`colors[4]` yellow — `#FFFF00`** | **1.07 : 1** | 19.6 : 1 |

(WCAG relative luminance.)

Note that green already sets the precedent for what is proposed below: `colors[3]` is *not* symmetric, it is `inverse ? COLOR_INVERSE_GREEN : COLOR_LIGHT_GREEN`, i.e. `#00FF00` on the inverse look but a darkened `#00C000` on the default look. Yellow is the odd one out in that it uses the same full-brightness value for both, via `COLOR_INVERSE_YELLOW = COLOR_YELLOW`.

## Why nobody has noticed

Colour index 4 is currently **unreachable**. Every call site in `src/InfoBoxes/` passes 0, 1, 2, 3 or 5, and the only computed index, `GetAlternateInfoBoxValueColor()`, returns 0, 1 or 2. The single use of index 5 is `SetCommentColor(5)` in `src/InfoBoxes/Content/Radio.cpp`. So the palette entry exists but has never been used for anything.

That also means **changing it cannot regress any existing InfoBox** — there is nothing to regress.

## How this came up

While adding traffic-light colouring to the SpO2 InfoBox for #1836. A caution band is exactly what index 4 is there for, and it turned out to be unusable on the default look. Rather than work around it locally in one InfoBox, it seems better to fix the palette entry, since the next feature that wants a caution colour will hit the same wall.

There is a safety angle specific to #1836: the SpO2 box is read at altitude, and colour discrimination degrades under hypoxia and in low light. A caution colour that is already near-invisible under ideal conditions is not a good starting point.

## Proposal

Keep `#FFFF00` for the inverse look, where it works well, and use a darker amber for the default look:

```cpp
colors[4] = inverse ? COLOR_INVERSE_YELLOW : COLOR_AMBER;
```

Candidate values, contrast against the respective background:

| Candidate | on white | on black |
| --- | --- | --- |
| `#FFA200` (existing `COLOR_ORANGE`) | 2.02 : 1 | 10.4 : 1 |
| `#D07000` | 3.50 : 1 | 6.00 : 1 |
| **`#B35A00`** | **4.80 : 1** | 4.38 : 1 |
| `#8F6000` | 5.47 : 1 | 3.84 : 1 |

`#B35A00` is the best balance if a single value is wanted for both looks; it clears 4.5 : 1 on white and stays usable on black. The existing `COLOR_ORANGE` is not sufficient on white at 2.02 : 1.

Alternatively, `COLOR_INVERSE_YELLOW` could stop being an alias of `COLOR_YELLOW` and the two looks could diverge properly, which is what the other inverse colours already do (`COLOR_INVERSE_RED`, `COLOR_INVERSE_BLUE`).

## Open questions for maintainers

- Is a new `COLOR_AMBER` constant in `src/ui/canvas/Color.hpp` acceptable, or should this reuse/adjust `COLOR_ORANGE`?
- Green is darkened for the default look but only reaches 2.46 : 1. Is that deliberate and considered good enough in the field, or worth revisiting in the same pass?

Happy to submit a patch once there is a preference on the above.

Related: #1836

dspreitz:
Fixes #2970.

InfoBox colour index 4 is pure yellow in both looks, while the InfoBox background is white unless the inverse look is active. That is a contrast ratio of **1.07 : 1** — not "hard to read", effectively invisible.

| Colour | on white (default look) | on black (inverse look) |
| --- | --- | --- |
| normal text | 21.0 : 1 | 21.0 : 1 |
| `colors[1]` red — `#FF0000` / `#FF7070` | 4.0 : 1 | 7.8 : 1 |
| `colors[3]` green — `#00C000` / `#00FF00` | 2.46 : 1 | 15.3 : 1 |
| `colors[4]` yellow — `#FFFF00` **before** | **1.07 : 1** | 19.6 : 1 |
| `colors[4]` amber — `#B35A00` **after** | **4.80 : 1** | 4.38 : 1 |

The fix follows the pattern index 3 already uses: a muted variant for light backgrounds, the bright variant for dark ones. Yellow was the odd one out only because `COLOR_INVERSE_YELLOW` is an alias of `COLOR_YELLOW`, so both looks got the same full-brightness value.

### No visual regression

Colour index 4 is currently unreachable. Every call site in `src/InfoBoxes/` passes 0, 1, 2, 3 or 5, and the only computed index, `GetAlternateInfoBoxValueColor()`, returns 0, 1 or 2. The single use of index 5 is `SetCommentColor(5)` in `src/InfoBoxes/Content/Radio.cpp`. So no existing InfoBox changes appearance — this only makes the slot usable.

### Why now

It came up while adding traffic-light colouring to the SpO2 InfoBox for #1836, where index 4 is exactly the caution band. There is a safety angle there: that box is read at altitude, and colour discrimination degrades under hypoxia and in low light, so a caution colour that is already invisible under ideal conditions is a poor starting point.

### On the open questions from #2970

- **New constant vs. reusing `COLOR_ORANGE`:** this adds `COLOR_AMBER` to `src/Look/Colors.hpp`, next to `COLOR_LIGHT_GREEN` and with the same style of comment. `COLOR_ORANGE` (`#FFA200`) only reaches 2.02 : 1 on white, so it is not sufficient. Happy to switch to adjusting `COLOR_ORANGE` instead if that is preferred.
- **Green in the same pass:** deliberately left alone here. At 2.46 : 1 it is below where I would want it, but it has existing callers and changing it would alter how shipping InfoBoxes look, which belongs in its own change with its own discussion.

### Testing

Contrast values are computed from the WCAG relative luminance formula against the two `background_color` values in `InfoBoxLook::Initialise()`. I have not been able to build locally (no `pkg-config` on this machine), so this has not been compiled or seen on a device — it is a two-line constant change, but please treat the visual result as unverified.

## Summary by CodeRabbit

* **Style**
  * Improved the readability of informational panels on light backgrounds by updating one display color to a darker amber shade.
  * Preserved the existing inverse-mode color treatment.

### DIFF
--- a/src/Look/Colors.hpp
+++ b/src/Look/Colors.hpp
@@ -58,6 +58,14 @@ static constexpr Color COLOR_ADMONITION_TIP =
 static constexpr Color COLOR_LIGHT_GREEN = Color(0x00, 0xc0, 0x00);
 
 /**
+ * A dark gold readable on light backgrounds.
+ * Standard COLOR_YELLOW (255,255,0) is invisible on white, while a
+ * burnt orange would sit too close to COLOR_RED to serve as the
+ * caution step of a traffic light.
+ */
+static constexpr Color COLOR_AMBER = Color(0x80, 0x70, 0x00);
+
+/**
  * Airspace warning list / map-item status badge colours.
  */
 static constexpr Color COLOR_AIRSPACE_WARNING_INSIDE =
--- a/src/Look/InfoBoxLook.cpp
+++ b/src/Look/InfoBoxLook.cpp
@@ -44,7 +44,7 @@ InfoBoxLook::Initialise(bool _inverse, bool use_colors,
     colors[1] = inverse ? COLOR_INVERSE_RED : COLOR_RED;
     colors[2] = inverse ? COLOR_INVERSE_BLUE : COLOR_BLUE;
     colors[3] = inverse ? COLOR_INVERSE_GREEN : COLOR_LIGHT_GREEN;
-    colors[4] = inverse ? COLOR_INVERSE_YELLOW : COLOR_YELLOW;
+    colors[4] = inverse ? COLOR_INVERSE_YELLOW : COLOR_AMBER;
     colors[5] = inverse ? COLOR_INVERSE_MAGENTA : COLOR_MAGENTA;
   } else
     std::fill(colors + 1, colors + 6, inverse ? COLOR_WHITE : COLOR_BLACK);


src/Look/Colors.hpp
```diff
@@ -57,6 +57,12 @@ static constexpr Color COLOR_ADMONITION_TIP =
  */
 static constexpr Color COLOR_LIGHT_GREEN = Color(0x00, 0xc0, 0x00);
 
+/**
+ * A dark amber readable on light backgrounds.
+ * Standard COLOR_YELLOW (255,255,0) is invisible on white.
+ */
+static constexpr Color COLOR_AMBER = Color(0xb3, 0x5a, 0x00);
```
lordfolken: `#B35A00` is too near red. Darken a gold/mustard instead of a burnt orange so caution stays distinct from `COLOR_RED` — e.g. `#807000`.

lordfolken (changes requested):
The readability fix is the right idea (yellow on white is unusable), but `#B35A00` is too close to warning red for a caution colour.

Hue is about 30° with G only half of R, so on a white InfoBox it reads as brick/rust next to `colors[1]` (`COLOR_RED`). That is a problem for a traffic-light palette: caution has to stay visually distinct from warning, including under hypoxia and for red-green colour vision deficiency. Project convention is red = warning, orange/yellow = caution.

`COLOR_ORANGE` (`#FFA200`) is too light on white, as you said — but the answer is to darken a yellower hue, not to pull toward red. Please keep ~4.5:1 on white and shift toward gold/mustard (around hue 50°), e.g. `#807000` (`Color(0x80, 0x70, 0x00)`).

lordfolken:
Moving this to **v7.46**. I wasn't able to find a good colour: high-contrast ambers read too close to warning red, and yellower shades either look olive/muddy or fail contrast on the white InfoBox background.

dspreitz:
You are right, and I measured it rather than argue: `#B35A00` is too close to red, and worst exactly where it matters.

| colour | contrast on white | L\* | C\* | hue | ΔE to red | …deuteranopia | …protanopia |
| --- | --- | --- | --- | --- | --- | --- | --- |
| `#B35A00` (mine) | 4.80 | 48.1 | 65.3 | 60.9° | 49.7 | **11.2** | **7.5** |
| `#807000` (yours) | 4.96 | 47.2 | 53.5 | 94.4° | 85.5 | 15.1 | 12.8 |
| `#FF0000` red | 4.00 | 53.2 | 104.6 | 40.0° | — | — | — |

ΔE is CIE76 in Lab; the colour-deficient columns are the two colours run through a Viénot simulation first. A ΔE of 7.5 is about the width of a "just noticeable" step — for a protanope my burnt orange and the warning red were nearly the same colour. That is exactly the failure mode a traffic light must not have, and it lines up with 14 CFR 25.1322 keeping red for warnings and amber for cautions.

**Changed to `#807000`.**

I also went looking for something more chromatic that would still hold up, since your note about moving this to v7.46 says you could not find one. I searched the sRGB gamut under: contrast ≥ 4.5 on white, hue 65–110°, ΔE to red ≥ 13 under deuteranopia and ≥ 12 under protanopia, and ΔE ≥ 40 from `COLOR_LIGHT_GREEN`.

The most chromatic survivors are around `#6E7C00` — C\* 57.3 against your 53.5, so **four units of chroma**, bought by moving the hue to 109°. That is yellow-green, which is both the "olive" you described and a collision with the green class. Below 4.5:1 nothing opens up either: at amber hues, chroma is bounded by lightness, and lightness is bounded by the white background.

So your instinct was right and there is nothing better to find — `#807000` is effectively the optimum for a flat colour on white.

**The constraint is the white background, though, not the palette.** Your #1966 puts an outline behind InfoBox values; with one, contrast is against the outline rather than the background, and a real aviation amber becomes usable: `#FFBF00` is C\* 83.7 at hue 82.9° — 12.7:1 against a black outline, and ΔE 30.9 / 53.9 from red under the two deficiencies, far clearer than anything reachable today. If #1966 lands, this constant is worth revisiting; I would rather have the vivid amber than the gold.

Contrast of the new value: **4.96:1 on white**, 4.24:1 on black. The inverse look keeps `COLOR_INVERSE_YELLOW` as discussed, since yellow on black is already 19.6:1.

dspreitz:
Closing this. Solution will be addressed in #2999

PR #2935, opened 2026-08-23
Draw the distance rings with a fine pen

WolfgangDrescher:
The current distance rings are a bit to bold. Should I experiment with opacity as well?

Before | After |
-|-
Bildschirmfoto 2026-08-24 um 00 46 47 | Bildschirmfoto 2026-08-24 um 00 46 30

Here is another screenshot on darker terrain. 

Bildschirmfoto 2026-08-24 um 01 00 23

## Summary by CodeRabbit

* **New Features**
  * Updated map distance rings to appear thinner and less prominent on high-DPI displays.

* **Documentation**
  * Added a release note documenting the updated distance-ring appearance in Version 7.46.

Issue #2917, opened 2026-08-21
Modernize the XCSoar UI


WolfgangDrescher:
The XCSoar UI has grown over almost two decades and shows its age: styling is inconsistent between dialogs, widgets and the map overlay, and many controls were designed for small PDA screens operated with a stylus.

This is not only about a nicer look. It is mainly about handling: fewer steps to reach a function, predictable navigation, readable text in sunlight, and an input model that works properly on every device class we support (touch phones and tablets, e-ink, panel mount with hardware buttons, remote stick or rotary encoder, keyboard and mouse).

Below is a list of todos that I would tackle in the sense of a fundamental but iterative modernization. Where issues for them already exist, they are linked as sub issues below.

## Todo

* [ ] Rework the dialog drop shadow #2916
* [ ] Rework the dialog titles
* [ ] Standardize the appearance of widgets in XCSoar #3014
* [ ] Rework buttons #2871
* [ ] Rework main menu #2852
* [ ] Improve scrollbars
* [ ] Rework the sound system #2593
* [ ] Add one or more modern InfoBox border themes #2553 and #2876
* [x] Sharper icons and graphics on retina displays #2985
* [ ] Rework gesture input path (color and shape)
* [ ] Rework "tab dialog" such as `Weather` #2926
* [ ] Modern map layout (MapLibre?) #2833
* [ ] Improve Docs #2517

gratefulfrog:
I would also add, rework directory structure handling to allow for migration from pre 7.45.  

Perhaps consider that creation of a user editable text config file entries (in default.prf) , that would be updated by the System config menus as well, but would allow for simple user editing with a text editor.   

Currently, it seems that some, or most of the path configuration is not user editable.

Hope this helps!
Cheers,
Bob

gabriel-briffe:
Don't know if this is of any help here?
[https://github.com/XCSoar/XCSoar/discussions/3057](url)

PR #2999, opened 2026-08-29
InfoBoxes: outline coloured values and units (rebase of #1966)

dspreitz:
Rebase of #1966 onto current master.

The InfoBox value colours are picked to convey a meaning, not for contrast against
the background, so an amber or yellow value can be hard to read — on a light
background in particular. This draws those values (and their unit symbol) with the
same halo that `RenderShadowedText()` already provides for map labels. Values that
are plain black or white keep the cheaper unshadowed path, so nothing changes for
the vast majority of InfoBoxes.

Credit for the idea and the original implementation goes to @lordfolken in #1966.
That branch no longer applies to master (the `TCHAR` → `char` migration, plus a fair
amount of reformatting), so this is a hand-port of its substance. Three deliberate
differences:

1. **The halo offset formula is master's, unchanged.** #1966 added a second formula,
   `h <= 16 ? h/20 : h/32`. Master's `DrawTextHalo()` already uses
   `std::max(1u, GetFontHeight() / 12u)` with `std::clamp(4 * offset, 8u, 16u)`
   copies distributed around a circle, which is both thicker at large font sizes and
   already DPI-aware. Adopting the PR's formula would make outlines *thinner* from
   24 px up, so there is nothing to port here.

2. **No `Red()` on a `Color` in portable code.** #1966's `GetOutlineColor()` calls
   `background_color.Red() + .Green() + .Blue()`, which does not compile under
   `GREYSCALE` — those accessors only exist in the non-`GREYSCALE` branch of
   `memory/Color.hpp`, so the Kobo target would have failed to build. The brightness
   test here is behind `#ifdef GREYSCALE` and uses `GetLuminosity()` there.

3. **No duplicated symbol-drawing code.** Rather than copying the body of
   `UnitSymbolRenderer::Draw()`, it is factored into a static
   `DrawSymbol(..., const Color *halo)` that both overloads share; `halo == nullptr`
   means "as before". This also picks up a line #1966 missed — the second symbol
   line in the `size1.width > size2.width` branch was still drawn without a halo.

Likewise, `RenderShadowedText(canvas, text, p, bool inverted)` now delegates to the
new colour-taking overload instead of carrying its own copy of the drawing code.

Checked with `-fsyntax-only` for the three touched translation units, with and
without `GREYSCALE`.

Supersedes #1966.

## Summary by CodeRabbit

* **Improved Readability**
  * Added contrasting outlines around colored values and unit symbols in information boxes.
  * Text and units now remain clear against both light and dark backgrounds.

### DIFF
--- a/src/InfoBoxes/InfoBoxWindow.cpp
+++ b/src/InfoBoxes/InfoBoxWindow.cpp
@@ -9,6 +9,7 @@
 #include "Input/InputEvents.hpp"
 #include "Renderer/GlassRenderer.hpp"
 #include "Renderer/UnitSymbolRenderer.hpp"
+#include "Renderer/TextInBox.hpp"
 #include "Screen/Layout.hpp"
 #include "ui/canvas/Canvas.hpp"
 #include "ui/event/KeyCode.hpp"
@@ -20,6 +21,25 @@
 
 /** timeout of infobox focus */
 static constexpr std::chrono::steady_clock::duration FOCUS_TIMEOUT_MAX = std::chrono::seconds(20);
+
+/**
+ * Pick an outline colour that separates coloured text from the given
+ * InfoBox background: a light outline on a dark background and the
+ * other way round.
+ */
+static Color
+GetOutlineColor(Color background_color) noexcept
+{
+#ifdef GREYSCALE
+  const unsigned luminosity = background_color.GetLuminosity();
+#else
+  const unsigned luminosity = (background_color.Red()
+                               + background_color.Green()
+                               + background_color.Blue()) / 3;
+#endif
+
+  return luminosity < 0x80 ? COLOR_WHITE : COLOR_BLACK;
+}
 
 InfoBoxWindow::InfoBoxWindow(ContainerWindow &parent, PixelRect rc,
                              unsigned border_flags,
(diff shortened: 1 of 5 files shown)


lordfolken:
Thank you for taking over and cleaning it up. 😀

lordfolken:
Would you mind removing the claude attribution in the commits? Its a setting in the ide.

dspreitz:
> Would you mind removing the claude attribution in the commits? Its a setting in the ide.

Sure. Should be disabled for future commits.

dspreitz:
So, after limited testing of the MacOS Branch on my desk, this PR seems to do what is expected. See black surrounding on McReady numbers.
Bildschirmfoto 2026-08-29 um 12 25 13
Further testing would be appreciated, but for the time being, I think I am done with this PR @lordfolken

WolfgangDrescher:
Sorry to say it, but visually I really don't find this very appealing, and especially with regard to my vision in #2917 we should be careful with changes like this.

I understand the problem and I won't stand in the way of such a change, but I would prefer to see it behind a new setting as an opt-in option. I would suggest a generic `ProfileKeys::ContrastMode`, which could also change the UI in other places:

```cpp
enum class ContrastMode : uint8_t {
  DEFAULT = 0,
  LOW = 1,
  HIGH = 2,
  MAX = 3,
};
```

For the InfoBoxes this could map to: `DEFAULT` keeps the current rendering without any outline, `LOW` adds a very subtle light grey outline, `HIGH` a stronger grey one, and `MAX` uses the black outline proposed in this PR. That way users can pick how much contrast they want instead of everyone getting the strongest variant.

This would also be useful for greyscale builds and for devices using the new display type setting (`ProfileKeys::DisplayType` with `E_INK` and `COLOR_E_INK`), where colored values carry much less information and a stronger contrast is genuinely helpful. The default for those could even be derived from the display type.

Another example is my distance rings PR (#2935), which could be rendered with an additional, different style when a higher contrast mode is active. That is just one of many places where such an option would make sense.

lordfolken:
@WolfgangDrescher i'm holding off on this one, until you tell me the grand scheme of things.

WolfgangDrescher:
Okay, but it will take some time before I get to details like these. First I want to fix the major issues such as dialog styles, list styles, buttons, safe area, etc.

Issue #2992, opened 2026-08-28
Repository ALPS.xcm points at old 38 MB map, not current 247 MB build


gabriel-briffe:
In-app download of ALPS.xcm uses the repository URL:

http://download.xcsoar.org/source/map/region/ALPS.xcm (38 MB)

That map still has water_area,100 in topology.tpl, so large lake labels stay visible when zoomed far out.

A newer ALPS build (247 MB, water_area_large,5, etc.) exists on the same host at a different path (.../source/source/map/region/ALPS.xcm). It is not what the repository serves.

Is this on purpose? Is it widespread?

I investigated this trying to find out why at low zoom levels we see some random lakes only, but not the main cities.. when viewing skysight satellite, it might be useful to see only the main cities instead..

lordfolken:
The maps currently in the repo are very old and use outdated and sparse data. 
The other maps (247MB)  where an attempt to improve the situation. I however discovered that a lot more optimization was needed in order for those to perform properly, especially at low-end hardware. 

I have since a few days created some heavy maps to improve the renderer. 
https://download.xcsoar.org/beta/map/ on the basis of https://github.com/XCSoar/xcsoar-mapgen-topology
With the help of the LLMs i was able to build a harness with which i could optimize the layers and polygon renderers. 
Its not perfect yet, sometimes there are artifacts.
Topology is now renderered in a separate thread, and therefore no longer impacts UI or interactivity. It also already displayes while still building (defeered rendering)
Besides that, there is an OpenGL shader for Hillshading, which is at present done in software. 


The renderer improvments are in pr: https://github.com/XCSoar/XCSoar/pull/2895
Please help test that.

gabriel-briffe:
trying to get XCtherm polygons to display well and without artifacts or flicker, i replaced the PolygonToTriangles ear-clip by Mapbox Earcut algorithm.

So on Stefly now i have polygons with holes, no flicker, all good.

On top of that i added this PR and it is all perfect on the phone so far.
And the shader is great.

I am not sure if code from Mapbox would be a big No No for xcsoar, but it definitely solves the artifacts and concave shapes. I restored polygons with holes, but that is not a necessity..

tell me if you want a PR or a showcase branch, and with what..

edit: found the polygon split on lakes to end up not touching each other, investigating

gabriel-briffe:
in here: https://github.com/gabriel-briffe/XCSoar/tree/show/topology-dense-demo
[...]

Issue #666, opened 2021-07-16
New dependency to libcares


hjr:
XCSoar version
--------------
#HEAD

What should XCSoar do differently, what functionality should be added?
----------------------------------------------------------------------
Our dev manual does not mention the dependency to libcares. I am not sure which we it should be resolved e.g. for Linux, but a simple "make" does end on link level missing this library.

What are you trying to do, what is the use case for the suggestion?
-------------------------------------------------------------------
Please add the necessary hint to get this library dependency resolved.

lordfolken:
We have all the dependencies needed in the provisioning script: https://github.com/XCSoar/XCSoar/blob/e7db3c374c7f3abd4f3d425052719f6cbab35324/ide/provisioning/install-debian-packages.sh#L42

PR #3008, opened 2026-08-30
Openvario/Cubiebaord: Work around shader compiler bug with Mali 400 GPUs using the mainline Lima DRM driver

hor63:
### Current XCSoar (as of V7.44) has two major issues on Cubieboards:

1. No text is being rendered. Only the blank background shows up. No matter if text in text fields, in buttons or on the map.
2. Cross compilation for Cubies (and others) does not work. Reason is that libgeotiff does not have a pkgconfig file. Therefore for Linux no one is expected, but only for local builds, not for cross compilation builds.

The development and test was executed on Ubuntu 24.04 LTS with current updates.
Kernel on the Cubieboard is a vanilla home-baked V6.18.35 LTS.
Correct function was verified on X86_64 machines with

- Intel integrated graphics on a 2012 laptop
- 2013 desktop with Nvidia card
- as VMWare guest with the SVGA driver

Both with MESA_KMS on a text console and as X11 application

And functionality was of course verified on a Cubieboard2.

## Pre-Submission Checklist

#### Branch & Rebase
- [X] PR is rebased on current `master`
- [X] Use `git rebase -i` to clean up commit history before PR
  submission

#### Commit Format & Messages
- [X] All commits follow the format: `<Component>: <Summary>` (no
  `src/` prefix)
- [X] Use present tense in commit messages ("Fix" not "Fixed", "Add"
  not "Added")
- [X] Commit messages explain *why* the change was made, not just
  *what* changed
- [X] Commit message body (if needed) provides detailed reasoning and
  context

#### Commit Structure
- [X] Each commit is atomic and builds successfully (every commit
  must compile)
- [X] One commit per logical change (don't mix refactoring with
  feature changes)
- [X] Self-contained commits (each commit changes one thing)
- [X] No fixup commits (squashed into parent commits using
  `git rebase -i`)
- [X] No duplicate commits (check with `git log --oneline`)
- [X] No "WIP" or "testing" commits (clean up before PR)

---

- [X] I'm ready to merge

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved text and monochrome rendering so transparency and color are applied more consistently.
  - Corrected inverted rendering to preserve intended colors alongside texture transparency.
  - Improved visual consistency for colored, transparent, and inverted canvas content, including on supported graphics hardware.

- **Build Improvements**
  - Improved libgeotiff configuration for Linux cross-compilation, providing more consistent behavior across Linux host and target environments.


lordfolken:
@scumi can i have your take here, should we merge this?
@hor63 there are several opengl shader efforts at the moment, would be great for testing. can you test pr #2895 ?
and also see https://github.com/XCSoar/XCSoar/issues/2992#issuecomment-5470854524
@gabriel-briffe this might be interesting for you aswell.

Scumi:
Thanks, I looked again at the history and remembered why 932cb8a mattered on meta-openvario: waiting synchronously for page-flip completion could block input processing, and the non-blocking implementation fixed that.

I tested current master and the complete PR on an OpenVario/Cubieboard2 with Mesa 24.0.7 and Linux 6.12.79. I cannot reproduce either the sluggish mouse behavior or the reported shader issues, and both versions appear responsive and render correctly. Could you check whether these problems are reproducible with Mesa 24.x and possibly Linux 6.12? This may be a version-dependent interaction.

I am cautious about calling ModeSetCrtc() every frame because, with the sun4i driver in Linux 6.12, it appears to become a blocking atomic commit and may therefore reintroduce the input stalls that 932cb8a fixed. Its return value is also ignored before releasing the old buffer.

Since this affects every MESA_KMS build, it would be useful to identify why the non-blocking page-flip behavior and shader rendering differ on the newer stack. I am happy to test targeted variants on OpenVario.

lordfolken:
@hor63 can you provide feedback to @scumi, so the problem can be located?

hor63:
@Scumi, I can track the shader issue, not rendering text, down to a regression in Mesa.
I compiled mesa 24.0.7 on the Cubieboard, and ran XCSoar without my fix against it. The alpha blending works just fine.
image
image

With Mesa 25.2.8 that comes with Ubuntu 24.04 LTS and latest updates, the alpha blending bug shows up:
image
image

I also compiled the latest Mesa version 26.2.1 and tested XCSoar. The error is still appearing. So I guess it is going to stay for a while.

All done on the same user session on the same machine without re-booting. Just setting LD_LIBRARY_PATH to point to the local builds of MESA in version 24.0.7 and 26.2.1. So I guess we can root out the version of the Linux kernel.

In two weeks I have a week off. Then I have time enough to git bisect Mesa to nail down the root cause and raise an error report to the maintainers. However, it will take quite some time to materialize in distributions. So we may want to apply at least that work-around [18b1716](https://github.com/XCSoar/XCSoar/pull/3008/commits/18b1716f822bf36ff65af78d5c0c43bb3ed1369c), the the optimization using a uniform for the color [53ce827](https://github.com/XCSoar/XCSoar/pull/3008/commits/53ce827ab2ab0360b1363a5ff6cde25dfa58e1fb) and  enabling cross-compiling at all [53ce827](https://github.com/XCSoar/XCSoar/pull/3008/commits/53ce827ab2ab0360b1363a5ff6cde25dfa58e1fb).

hor63:
@lordfolken, [pr2895](https://github.com/XCSoar/XCSoar/pull/2895) is working fine together with the shader related fixes from this PR.
The shading is a lot more precise in high zoom levels (circling mode).

hor63:
@lordfolken, @Scumi, the issue with the lagging mouse movements is clearly related to the resolution of the screen, i.e. the workload of the GPU.
With  [pr2895](https://github.com/XCSoar/XCSoar/pull/2895) my commit [18b1716](https://github.com/XCSoar/XCSoar/pull/3008/commits/18b1716f822bf36ff65af78d5c0c43bb3ed1369c) does not have any positive effect when I run my Cubie with an HDMI monitor at 1920x1200 resolution. With or without the commit the mouse hardly moves at all. 
When I run it at 720x480 resolution the responsiveness is just fine.

So I guess that requires more in-depth investigation. Please disregard commit [18b1716](https://github.com/XCSoar/XCSoar/pull/3008/commits/18b1716f822bf36ff65af78d5c0c43bb3ed1369c) for now. No one will fly with his or her office monitor in the cockpit :)

lordfolken:
> @lordfolken, [pr2895](https://github.com/XCSoar/XCSoar/pull/2895) is working fine together with the shader related fixes from this PR. The shading is a lot more precise in high zoom levels (circling mode).

Thanks for the report. :)

lordfolken:
> @lordfolken, @Scumi, the issue with the lagging mouse movements is clearly related to the resolution of the screen, i.e. the workload of the GPU. With [pr2895](https://github.com/XCSoar/XCSoar/pull/2895) my commit [18b1716](https://github.com/XCSoar/XCSoar/pull/3008/commits/18b1716f822bf36ff65af78d5c0c43bb3ed1369c) does not have any positive effect when I run my Cubie with an HDMI monitor at 1920x1200 resolution. With or without the commit the mouse hardly moves at all. When I run it at 720x480 resolution the responsiveness is just fine.

So XCSoar never implemented a hardware cursor in the KMS/MESA case. If your cubie is fill rate strapped, (very likely at full hd), moving the cursor will cause a full repaint of the window with an updated cursor position. That is probably why its sluggish.

hor63:
> So XCSoar never implemented a hardware cursor in the KMS/MESA case. If your cubie is fill rate strapped, (very likely at full hd), moving the cursor will cause a full repaint of the window with an updated cursor position. That is probably why its sluggish.

It is not only sluggish, but does not even react when you keep moving the mouse. Only when you push the mouse quickly in a burst, then rest, it the cursor jumps accordingly in the next update cycle.

Quick question: Are the libinput file handles polled in the main loop that also performs the drawing? If so, moving polling libinput handles into an own thread could get us out of the weeds.
The other more elegant approach would be implementing the cursor on a second DRI plane with an own small frame buffer that is panned over the primary plane, showing the scene. The old All winners do not have a dedicated cursor plane. One of the four regular plane must suffice. I will explore that in the next weeks closer.

lordfolken:
No adding a thread for libinput wouldn't solve the issue. You would still wait on ui and map. 
What you need is a hardware cursor that doesn't invalidate(), so for example drmModeMoveCursor.

hor63:
OK, I stripped down the pull request to the essential: Working around the alpha blending regression in Mesa 3D, and fixing the cross-compilation issue.
I will investigate the sluggish mouse issue further more in-depth in the next week of vacation. I will raise another pull request, and may be an issue in due time.

hor63:
With the re-hash of my branch I also re-based it to the current head of branch master.

hor63:
@lordfolken, @Scumi, is there anything more requires to move this PR forward?

Scumi:
I'll have a look, thanks.

src/ui/canvas/opengl/Shaders.cpp
```diff
@@ -110,10 +114,13 @@ static constexpr char invert_fragment_shader[] =
...
     uniform sampler2D texture;
+    varying vec4 colorvar;
     varying vec2 texcoordvar;
     void main() {
-      vec4 color = texture2D(texture, texcoordvar);
-      gl_FragColor = vec4(vec3(1) - color.rgb, color.a);
+      vec4 color;
+      vec4 alpha = vec4(vec3(0),texture2D(texture, texcoordvar).a);
+	  color = vec4(vec3(1) - colorvar.rgb, 0);
+      gl_FragColor = color + alpha;
```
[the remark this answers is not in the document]
hor63: You are mis-understanding the intent and connection between texture and fragment shaders :
The vertex shader "texture_vertex_shader" is re-used for multiple fragment shaders: "texture_fragment_shader", "invert_fragment_shader", "alpha_fragment_shader" and combine_texture_fragment_shader.
The the texture shader texture_vertex_shader has an attribute "color" which is assigned to a varying "colorvar".
However, the fragment shaders "texture_fragment_shader" and "invert_fragment_shader" are _not_ using the varying "colorvar". The color value is coming in these fragment shaders from the texture only. This is intendend this way.
Thus the varying "colorvar" together with the attribute "color" are likely to be optimized away by the linker in the final program.
Anyway setting the attribute "color" for these programs is not required, and will probably even rejected.

src/ui/canvas/opengl/Shaders.cpp
```diff
@@ -126,10 +140,44 @@ static constexpr char alpha_fragment_shader[] =
...
+ )glsl";
+static constexpr char alpha_fix_color_fragment_shader[] =
+  GLSL_VERSION
+  GLSL_PRECISION
+  R"glsl(
+    uniform sampler2D texture;
+    varying vec2 texcoordvar;
+    uniform vec4 color;
+    void main() {
+      gl_FragColor = vec4(color.rgb, texture2D(texture, texcoordvar).a);
```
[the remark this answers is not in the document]
hor63: The work-around for alpha_fragment_shader is not required in the case of the fixed-color shader.
I verified the correct behavior on a Cubieboard2 with Allwinner A7 SoC using a Mali400 GPU.
Seemingly it matters if the color RGB components are pulled from local variables or varyings or if they are pulled from a uniform.

hor63:
I copy-pasted too much from alpha_fragment_shader into invert_fragment_shader. Ups.
Thank you CodeRabbitAI for pointing it out.
It is corrected now.

src/ui/canvas/opengl/Shaders.cpp
```diff
@@ -112,8 +116,18 @@ static constexpr char invert_fragment_shader[] =
...
+      vec4 alpha;
+      texture_sample = texture2D(texture, texcoordvar);
+      alpha = vec4(vec3(0),texture_sample.a);
+      color = vec4(vec3(1) - texture_sample.rgb, 0);
+      /* Workaround for the LIMA GPU driver for OpenVario:
+       * Assigning the alpha component directly leads to
+       * render errors in blending mode
+       * Full-vector additions or multiplications however work.
+       */
+      gl_FragColor = color + alpha;
```
lordfolken, line 127: `invert_shader` has to invert the sampled texel, which this version does. An earlier revision inverted `colorvar` instead.

`invert_shader` only binds `position` and `texcoord`. `GLTexture::Draw()` never writes a color, so an unbound `colorvar` stays at the GLES default `(0,0,0,1)` and every opaque texel becomes white.

`Data/icons/map_thermal_hotspot.svg` is the icon that shows it. Both fills are grey (`#666` and `#e6e6e6`), so `has_colors` is false. `WaypointIconRenderer` calls `MaskedIcon::Draw(canvas, point, height)` when the icon is scaled (the waypoint list, or a map icon scale other than 100%), and a dark background selects this shader. A real invert keeps the cloud and the inner arrow as different shades. Inverting `colorvar` collapses them into one white shape.

`colorvar` belongs to the alpha-tint shader. Please keep `texture_sample` here.

lordfolken (commented):
The thermal waypoint icon is the case that shows why this shader has to invert the sampled texel.

src/ui/canvas/opengl/Shaders.cpp
```diff
@@ -126,10 +140,44 @@ static constexpr char alpha_fragment_shader[] =
...
+static constexpr char alpha_fix_color_fragment_shader[] =
+  GLSL_VERSION
+  GLSL_PRECISION
+  R"glsl(
+    uniform sampler2D texture;
+    varying vec2 texcoordvar;
+    uniform vec4 color;
+    void main() {
+      gl_FragColor = vec4(color.rgb, texture2D(texture, texcoordvar).a);
+	}
```
lordfolken: This closing brace is a tab. `.clang-format` sets `UseTab: Never` and the surrounding shader text is indented with spaces. Please use two spaces, same as the other braces in this string.

.clang-format
```diff
@@ -10,7 +10,7 @@ NamespaceIndentation: None
 # CamelCase for class/function names
 # Lower case with underscores for attributes/variables
 # All upper case for constants/enums
-Standard: Cpp20
+Standard: c++20
```
lordfolken, line 13: `c++20` is the spelling clang-format accepts (`Cpp20` is not a documented `LanguageStandard` value), but this edit does not belong in the libgeotiff commit. Please move it to its own commit, or drop it from this PR. The libgeotiff change is the cross-build `.pc` lookup.

lordfolken (commented):
Two style nits: a tab in the new shader, and the `.clang-format` edit sitting in the libgeotiff commit.

src/ui/canvas/opengl/Shaders.cpp
```diff
@@ -126,10 +140,44 @@ static constexpr char alpha_fragment_shader[] =
...
     void main() {
-      gl_FragColor = vec4(colorvar.rgb, texture2D(texture, texcoordvar).a);
+      vec4 color = vec4(colorvar.rgb, 0);
+      vec4 alpha = vec4(vec3(0),texture2D(texture, texcoordvar).a);
+      /* Workaround for the LIMA GPU driver for OpenVario:
+       * Assigning the alpha component directly leads to
+       * render errors in blending mode
+       * Full-vector additions or multiplications however work.
+       */
+      gl_FragColor = color + alpha;
```
lordfolken: `alpha_shader` is unused after the switch to `alpha_fix_color_shader`. Text and `StretchMono` both go through `PrepareColoredAlphaTexture()`, and nothing else calls `alpha_shader->Use()` to draw.

This program is still compiled, and `UpdateShaderProjectionMatrix()` / `UpdateShaderTranslate()` still switch to it. The Lima workaround in this shader never runs. Please delete `alpha_shader` and those updates.

src/ui/canvas/opengl/Canvas.cpp
```diff
@@ -618,8 +618,8 @@ Canvas::CalcTextSize(std::string_view text) const noexcept
 static void
 PrepareColoredAlphaTexture(Color color) noexcept
 {
-  OpenGL::alpha_shader->Use();
-  color.Bind();
+  OpenGL::alpha_fix_color_shader->Use();
```
lordfolken, line 649: This is the user-visible fix: text (and mono stretch) comes back on a Cubieboard / OpenVario with Mesa 25 and the Lima driver. Please add a `NEWS.txt` bullet under version 7.46.

lordfolken (commented):
Two leftovers: the old alpha shader is unused, and the OpenVario text fix needs a NEWS.txt line.

src/ui/canvas/opengl/Shaders.cpp
hor63: Will do.
I contemplated about removing it. However, without the original shader you cannot use the shader with the color attribute as part of the vertex list. We are limited to constant color, no changing colors like fading effects.

src/ui/canvas/opengl/Canvas.cpp
hor63: I just amended NEWS.txt. I only mentioned the disappeared text.
mono stretch is nowhere being used. The method is stale.

hor63:
Re-based to current master.
Update of NEWS.txt moved to an own commit.

PR #3042, opened 2026-09-05
THIRD_PARTY_NOTICES: document the embedded third-party code

dspreitz:
`THIRD_PARTY_NOTICES.txt` covers ANGLE and CM4all libcommon. Nine further
components are vendored into the source tree or shipped with the binaries
and are not mentioned at all, even though their licences require the
copyright and permission notices to be reproduced:

| Component | Location | Licence |
| --- | --- | --- |
| GLM | `lib/glm/` | MIT |
| IOIO | `android/ioio/` | BSD-2-Clause |
| JasPer | `src/Terrain/jasper/` | JasPer License 2.0 |
| MapServer | `src/Topography/shapelib/` | MIT |
| UsbSerial | `android/UsbSerial/` | MIT |
| zziplib | `src/zzip/` | LGPL / MPL |
| XML parser | `src/XML/` | LGPL-2.1 |
| Freescale mxcfb | `src/ui/canvas/fb/mxcfb.h` | LGPL-2.1-or-later |
| libtap | `test/src/tap.{c,h}` | BSD-2-Clause |

GLM, IOIO and UsbSerial are git submodules, which is presumably why they
were easy to overlook: they are absent from a shallow checkout, but their
code is compiled into the binaries we hand out all the same.

Also credits the Volkslogger protocol code, which is under XCSoar's own
licence but carries a separate copyright holder, and records the five
additional copyright holders that `mapstring.cpp` and `mapsearch.c` carry
beyond the University of Minnesota.

Every section names where the code lives and, where one exists, its
upstream. Four carry a caveat worth a reviewer's attention:

- the XML parser upstream has since relicensed to the Aladdin Free Public
  License, which is not a free software licence. The copy in this tree is
  version 1.08 and states LGPL-2.1 in its own header, which is what the
  notice reproduces.
- libtap's original site (jc.ngo.org.uk) is gone, so the link points at a
  mirror carrying the same notice.
- GLM is dual licensed under the Happy Bunny License and the MIT License;
  the notice takes the MIT option.
- `android/ioio/LICENSE` puts that repository under Apache-2.0, but 54 of
  the 59 files `build/android.mk` actually compiles carry a BSD-2-Clause
  header instead, so that is the notice reproduced here. The remaining
  five, all under `IOIOLibCore`, bear no header and fall back on the
  repository licence.

The licence texts are quoted verbatim from the files in the tree, not
from memory or from upstream websites. Every path in the file was checked
against the current tree, and the submodule licences were read from
checked-out submodules rather than assumed.

Rebased onto current master, so the ANGLE section carries 74000b5bf
("ANGLE is not a macOS-only dependency") with the location and upstream
lines added on top.

Documentation only; no code is affected.

## Summary by CodeRabbit

- **Documentation**
  - Expanded third-party notices with detailed licensing information for IOIO components.
  - Added the complete Apache License 2.0 text for applicable IOIO files.
  - Updated authorship information to direct license-required notices to the third-party notices document.
  - Removed duplicate third-party component entries from the authorship information.
  - No exported or public interfaces were changed.

### DIFF
--- a/THIRD_PARTY_NOTICES.txt
+++ b/THIRD_PARTY_NOTICES.txt
@@ -19,6 +19,12 @@ binaries that use it.
 
 **Upstream:** [chromium.googlesource.com/angle/angle](https://chromium.googlesource.com/angle/angle)
 
+Windows builds do not compile ANGLE themselves: they download pre-built
+libraries from the packaging repository
+[github.com/XCSoar/angle-libs](https://github.com/XCSoar/angle-libs), pinned by
+`windows/fetch-angle-from-github.sh` to release tag `a96fca8`, commit
+`a96fca8d5ee2ca61e8de419e38cd577579281c9e`.
+
 **License:** [BSD 3-Clause](https://opensource.org/licenses/BSD-3-Clause)
 
 Copyright (c) 2002-2024 The ANGLE Project Authors.
@@ -53,10 +59,14 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 XCSoar incorporates selected headers and sources derived from **CM4all libcommon**.
 Location in the tree: spread over `src/util/`, `src/net/`, `src/event/`, `src/io/`,
 `src/java/`, `src/lib/`, `src/time/`, `src/lua/`, `src/thread/`, `src/system/`,
-`src/json/`, `src/co/` and a few others; the files carry
-`SPDX-License-Identifier: BSD-2-Clause` and can be listed with
+`src/json/`, `src/co/` and a few others.  The files carry
+`SPDX-License-Identifier: BSD-2-Clause`, but that marker on its own does not
+establish CM4all provenance: XCSoar has BSD-2-Clause sources of its own under
+`src/UIUtil/`, `src/Android/`, `src/Audio/SLES/` and `src/Tracking/SkyLines/`.
+The derived files are the ones whose header also names the copyright holder:
 
-    grep -rl "SPDX-License-Identifier: BSD-2-Clause" src/
+    grep -rl "Copyright CM4all GmbH" src/
+    grep -rl "Copyright Max Kellermann" src/
 
 **Upstream:** [github.com/CM4all/libcommon](https://github.com/CM4all/libcommon)
 
@@ -333,11 +343,11 @@ Several files in this directory are XCSoar modifications and carry
 
 Location in the tree: `src/XML/`
 
-**Upstream:** [applied-mathematics.net/tools/xmlParser.html](http://www.applied-mathematics.net/tools/xmlParser.html)
+**Upstream:** [applied-mathematics.net/tools/xmlParser.html](https://www.applied-mathematics.net/tools/xmlParser.html)
 
-Note that the upstream author has since relicensed: the version offered there
-today (2.44) is under the Aladdin Free Public License, which is not a free
-software licence.  The copy in this tree is version 1.08 and states the GNU
+Note that the upstream author has since relicensed: version 2.44, the one
+offered upstream, is under the Aladdin Free Public License, which is not a
+free software licence.  The copy in this tree is version 1.08 and states the GNU
 Lesser General Public License 2.1 in its own header, quoted below.
 
 **License:** GNU Lesser General Public License, version 2.1
--- a/THIRD_PARTY_NOTICES.txt
+++ b/THIRD_PARTY_NOTICES.txt
@@ -1,10 +1,23 @@
 # Third-Party Notices
 
+XCSoar itself is distributed under the GNU General Public License, version 2 or
+later; see COPYING.  This file collects the notices of third-party code that is
+embedded in the source tree or shipped with the binaries, and whose licences
+require their copyright and permission notices to be reproduced.
+
+Components are listed with the directory they occupy, so that the notice can be
+checked against the tree.
+
+
 ## ANGLE (Almost Native Graphics Layer Engine)
 
 XCSoar uses **ANGLE** to provide OpenGL ES where the platform has no
 native implementation: over Metal on macOS, over Direct3D 11 on
 Windows builds with ANGLE enabled.
+Not part of the source tree: it is fetched at build time and shipped with the
+binaries that use it.
+
+**Upstream:** [chromium.googlesource.com/angle/angle](https://chromium.googlesource.com/angle/angle)
 
 **License:** [BSD 3-Clause](https://opensource.org/licenses/BSD-3-Clause)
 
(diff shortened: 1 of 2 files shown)


lordfolken:
I guess we should remove the 3rd party components  from AUTHORS and move them to this file.

dspreitz:
On the missing LGPL/MPL texts: the premise is right, and it seems worth a
maintainer decision rather than a quick patch.

Verified in the tree: `build/resource.mk:257` lists `THIRD_PARTY_NOTICES.txt`
in `TEXT_FILES`, so it is gzipped into the binary; `src/Dialogs/dlgCredits.cpp:76`
inflates and shows it in the Credits dialog; `build/osx.mk:70` copies it into the
macOS bundle. The file is therefore a shipped licence document, not just
repository documentation. And there is no LGPL or MPL full text anywhere in the
tree — `COPYING` holds the GPL-2 text and nothing else.

So three components ship without the licence text their licence asks us to
carry: zziplib (`src/zzip/`, dual LGPL/MPL), the XML parser (`src/XML/`,
LGPL-2.1) and the Freescale mxcfb header (`src/ui/canvas/fb/mxcfb.h`,
LGPL-2.1-or-later).

This predates the pull request — the file was already embedded before it — but
the change makes the gap easier to see, so it seemed fair to raise it here
rather than leave it.

What I have not done is fix it, because the obvious fix is expensive in a place
users actually read. LGPL-2.1 is 501 lines and MPL 1.1 is 469; appending both
would roughly triple a document pilots scroll through in a dialog. Note also
that the zziplib headers say only "use under copyleft (LGPL,MPL)" without
naming a version of either, so which texts apply needs settling first.

Reasonable ways forward:

1. Append the texts to `THIRD_PARTY_NOTICES.txt` and accept the length.
2. Ship `COPYING.LGPL` and `COPYING.MPL` beside `COPYING`, embed them the same
   way, and have the notices file point at them.
3. Leave it, if the project takes the view that shipping the corresponding
   source satisfies the obligation.

I have no stake in which; happy to implement whichever you prefer, in a separate
commit so it stays cherry-pickable.

The other four review comments are addressed in the commits just pushed. On the
CM4all one I did not follow the suggestion literally: grepping `src/` for the
BSD-2-Clause SPDX marker returns 304 files, of which only 81 name CM4all GmbH or
Max Kellermann, and seven are XCSoar's own code. The notice now greps for the
copyright holder instead, which selects those 81 and nothing outside the
directories it lists.

lordfolken:
Please go with option 1: append the full licence texts to `THIRD_PARTY_NOTICES.txt` and accept the length.

Keep them in this file so Credits stays a single third-party page rather than adding extra `COPYING.*` tabs.

For the versions: append GNU LGPL 2.1 once (that covers the XML parser's 2.1 and the Freescale header's 2.1-or-later) and MPL 1.1 for zziplib. Upstream zziplib is dual "LGPL version 2 or later" / MPL; LGPL 2.1 is the "or later" text we already need for the other two components.

A separate commit is fine, as you suggested.

dspreitz:
Done in `e8494f64f`, as a separate commit.

Followed your instruction exactly: LGPL 2.1 once, MPL 1.1 for zziplib, both inside
`THIRD_PARTY_NOTICES.txt` so Credits stays one page.

* **Appendix A — LGPL 2.1**, introduced as covering all three: the XML
  parser (names 2.1), the Freescale mxcfb header (2.1 or later), and
  the LGPL half of zziplib's dual offer.
* **Appendix B — MPL 1.1**, for the other half of zziplib's offer.

Each of the three component sections now points at the appendix that
applies to it. The zziplib section says outright that its headers name
no version and records your reasoning — upstream offers "LGPL version 2
or later", so the 2.1 text we need anyway satisfies it.

Texts taken verbatim from gnu.org and mozilla.org, 501 and 469 lines,
and diffed against the downloads after insertion to be sure nothing was
reflowed.

Size, since you said to accept it: 19k to 72k raw, **5.7k to 22k
compressed**, so about 16k more in the binary. For scale, `COPYING` is
already 6.9k compressed through the same `TEXT_FILES` list. The Credits
dialog inflates the whole file into memory when opened, which on a Kobo
means roughly 70k instead of 19k, and only while the dialog is up.

Build and full test suite green on macOS; the gzipped resource
round-trips to the original byte for byte.

THIRD_PARTY_NOTICES.txt
```diff
@@ -68,3 +98,1544 @@ HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
...
+
+    android/ioio/IOIOLibCore/src/main/java/ioio/lib/api/Closeable.java
+    android/ioio/IOIOLibCore/src/main/java/ioio/lib/impl/FixedReadBufferedInputStream.java
+    android/ioio/IOIOLibCore/src/main/java/ioio/lib/impl/InterruptibleQueue.java
+    android/ioio/IOIOLibCore/src/main/java/ioio/lib/impl/ResourceLifeCycle.java
+    android/ioio/IOIOLibCore/src/main/java/ioio/lib/impl/Version.java
+
+The Apache License 2.0 asks in section 4(a) for a copy to be given to the
+recipients of the work, so its text is reproduced in Appendix C.  The upstream
+repository carries no `NOTICE` file, so section 4(d) does not apply.
```
dspreitz, line 160: Confirmed and fixed in `f8e1a32c4` — the Apache 2.0 text is now Appendix C.

Verified the premise against the tree rather than taking it on trust.
`build/android.mk` compiles 59 files from the submodule, and exactly
five of them carry no licence header at all:

```
ioio/lib/api/Closeable.java
ioio/lib/impl/FixedReadBufferedInputStream.java
ioio/lib/impl/InterruptibleQueue.java
ioio/lib/impl/ResourceLifeCycle.java
ioio/lib/impl/Version.java
```

`Closeable.java` opens straight with `package ioio.lib.api;`. They fall
under `android/ioio/LICENSE`, which is the unmodified Apache 2.0 text —
diffed against apache.org, the only difference is a missing final
newline.

One correction to the finding, which changes the reasoning but not the
outcome: the licence text is **not** absent from the tree. It sits in
`android/ioio/LICENSE` and always has. The gap is narrower — that file
is inside a git submodule and nothing ships it. `build/resource.mk:257`
embeds `AUTHORS`, `COPYING`, `NEWS.txt` and this file, and nothing
else, so whoever holds the APK never sees it. Section 4(a) asks for a
copy to be given to the recipients of the work, and they are not
getting one.

Two alternatives considered and rejected:

* Adding `android/ioio/LICENSE` to `TEXT_FILES`. `build/resource.mk` is
  included unconditionally from `Makefile:121`, so every target would
  start depending on a submodule only Android needs — a shallow clone
  without it would break the macOS, UNIX, Kobo and Windows builds. It
  would also add a fourth Credits tab, which the maintainer explicitly
  ruled out in this PR.
* Adding a `NOTICE` file to the submodule fork. Section 4(d) is
  conditional on the work *having* a NOTICE, and upstream has none, so
  writing one would not discharge 4(a) — and it would put attribution
  text in the copyright holder's name that he never wrote.

So the text is duplicated: once in the submodule, once here. That is the
price of this file being self-contained, and it is the only version that
travels with the binary.

The section now also names the five files, so the claim can be checked
against the tree the way the rest of this document can, and says
outright that 4(d) does not apply rather than leaving the next reader to
work it out.

dspreitz:
@lordfolken — a third licence text has turned up, beyond the two you
approved. Flagging it rather than slipping it in, since it changes the
size you agreed to.

CodeRabbit's latest review pointed at the IOIO section, and it is right.
That section already recorded that five of the files XCSoar compiles
carry no header and fall under `android/ioio/LICENSE` — the Apache
License 2.0 — and then reproduced only the BSD 2-Clause notice the
*headered* files carry. I checked: `build/android.mk` compiles 59 files
from the submodule and exactly five have no header, down to
`Closeable.java` starting straight with `package ioio.lib.api;`.

`f8e1a32c4` adds it as **Appendix C**, following your ruling on the
other two. Reasoning is in the inline comment; the short version is
that the text does exist in the tree but sits in a git submodule that
nothing ships, and section 4(a) asks for recipients of the work to be
given a copy. I looked at putting `android/ioio/LICENSE` into
`TEXT_FILES` instead — that would make every target depend on a
submodule only Android needs and add the fourth Credits tab you ruled
out.

**Running total**, since you said to accept the length:

| | raw | compressed |
| --- | --- | --- |
| before this PR's licence work | 19k | 5.7k |
| + LGPL 2.1 and MPL 1.1 | 72k | 22k |
| + Apache 2.0 | **85k** | **25k** |

So about 19k more in the binary than when you approved, against 5.7k
originally. For scale, `COPYING` is 6.9k compressed through the same
list. Say the word if that is past what you meant by accepting the
length and I will take Appendix C back out — the BSD notice would then
stand alone as it does today, with the five files still named so the
gap is at least documented.

Build and full test suite green; the gzipped resource round-trips to
the original byte for byte.

dspreitz:
@coderabbit Full review please

Issue #3076, opened 2026-09-09
Task optimisation with BGA enhanced OZ


xtophe80:
XCSoar versions having the problem
-------------------------------------------------
7.45.1 + master

XCSoar versions not having the problem
-------------------------------------------------
none

System information
------------------
Android + Linux

Steps to reproduce the behavior
-------------------------------

Fly/Simulate a task with BGA Enhanced OZ (aka showery sectors). When you have signficantly deviated from the straight line, XCSoar still point you at the small barrel.

<img width="843" height="727" alt="Image" src="https://github.com/user-attachments/assets/613a174c-4703-436d-8ee6-1491ba1c6d45" />

Expected behavior
-----------------
At some point the lateral displacement, make it shorter to go to the straight part of the OZ rather than the barrel. (next to the "7.5 km range circle label)

Actual behavior
---------------

XCSoar keeps pointing at the small barrel or at the edge/corner of the OZ if the lateral displacement is extreme.

Do you have any idea what may have caused this?
-----------------------------------------------

Lack of nodes along the straight edges of the OZ so the Dijkstra algorithm can't take them in consideration.

Do you have an idea how to solve the issue?
-------------------------------------------

Add more nodes in GetBoundary() in src/Engine/Task/ObservationZones/KeyholeZone.cpp
For completness src/Engine/Task/ObservationZones/SectorZone.cpp need to be addressed as well

xtophe80:
The issue has also been noticed for the start zone by @Vib38 in #3068

Vib38:
Hello, yes i'm working on something similar for the start #2388. It's a good idea to do the same for turnpoint (even though this type of turnpoint is not that common) and also for the finish line.

PR #3079, opened 2026-09-09
Task: Add more nodes along edge of circular OZ

### LINKED ISSUE #3077: Big jump of aiming point with large OZ circle/ keyhole with large inner radius

Master

none

System information
------------------

Android and Linux

Fly a task with large circular OZ or keyhole OZ with large inner diameter. That typically happens flying a low handicap glider in a distance handicapped competition (e.g DHT rules in the UK)

Expected behavior
-----------------

Smooth evolution of the aiming point on the edge of the OZ

Actual behavior
---------------

The aiming point jumps by 1km or 2 km (radius of OZ = 5 or 7 km)

Not enough nodes along the OZ boundary (20)

Increase the number of nodes in GetBoundary() in src/Engine/Task/ObservationZones/CylinderZone.cpp and GenerateArcExcluding() in src/Engine/Task/ObservationZones/KeyholeZone.cpp.

xtophe80:
This allow the Dijkstra algorithm to return more precise min distances and aiming points. Fixes #3077.

The maximum number of nodes is set to 360. There doesn't seems to be any performance impact on Linux on a laptop. 

## Summary by CodeRabbit

* **Improvements**
  * Improved the accuracy and smoothness of circular observation-zone boundaries by adapting detail levels to the zone size.
  * Large-radius zones now render with more boundary segments, while smaller zones retain efficient rendering.

### DIFF
--- a/src/Engine/Task/ObservationZones/Boundary.cpp
+++ b/src/Engine/Task/ObservationZones/Boundary.cpp
@@ -4,11 +4,13 @@
 #include "Boundary.hpp"
 #include "Geo/GeoVector.hpp"
 
+#include <algorithm>
+
 void
 OZBoundary::GenerateArcExcluding(const GeoPoint &center, double radius,
                                  Angle start_radial, Angle end_radial) noexcept
 {
-  const unsigned steps = 20;
+  const unsigned steps = std::max(20,std::min(360, (int) (radius / 25)));
   const Angle delta = Angle::FullCircle() / steps;
   const Angle start = start_radial.AsBearing();
   Angle end = end_radial.AsBearing();
--- a/src/Engine/Task/ObservationZones/CylinderZone.cpp
+++ b/src/Engine/Task/ObservationZones/CylinderZone.cpp
@@ -20,7 +20,7 @@ CylinderZone::GetBoundary() const noexcept
 {
   OZBoundary boundary;
 
-  const unsigned steps = 20;
+  const unsigned steps = std::max(20,std::min(360, (int)(GetRadius() / 25)));
   const auto delta = Angle::FullCircle() / steps;
 
   GeoVector vector(GetRadius(), Angle::Zero());


xtophe80:
I noticed yesterday evening that it breaks the test_task test. So I review should be paused and I will investigate

lordfolken (commented):
Thanks for this — the jump is real. With 20 fixed vertices, a 7 km cylinder has ~2 km chords, so Dijkstra (and the aiming point) can only snap between those world-frame nodes. `GetBoundary()` is the documented place for arc approximations, so denser sampling is the smallest change that fits the current design.

I don't think 360-everywhere is the right long-term fix, though.

`TaskDijkstra` builds a complete bipartite graph between consecutive OZs (`M × M` geodesic `Distance()` calls per leg). Points on a circle are already the convex hull, so they are not thinned. Default AAT cylinders are 10 km, so they all hit the 360 cap: one remaining leg goes from 400 edges to ~130k, five legs to ~650k. Racing 500 m OZs are unchanged (`500/25 = 20`). That can look free on a laptop and still hurt on Kobo / old Android.

Please don't gate this on `IsSlowCPU()`: this is Engine code (no `Asset` / CPU dependency), and the slow devices are the ones used in the cockpit for the DHT case this fixes.

Better options, cheapest first:

1. Keep ~20 nodes for the combinatorial search, then slide each circular OZ along its arc to minimise `dist(prev)+dist(next)` (or project the winner onto the true circle). Same graph cost, continuous aiming point.
2. Denser samples only on the **active** OZ; remaining TPs stay at 20. The jump is the aiming point. Cost becomes O(360×20) per leg, not 360×360.
3. Closed-form near-side intersection / tangent for cylinders — right geometrically, more code, and sectors/keyholes still need an arc.

A lower static cap (72–90) is a compromise, not a better idea: jumps shrink, you still pay `M²`, and you still snap.

Happy to discuss which of (1) or (2) you want to try.

xtophe80:
To check I understood (1) correctly:
* At the end of OrderedTask::RunDijsktraMin(), check if the active waypoint OZ is a circle or Keyhole. If no, no change. If yes continue
* Minimise dist(A,P)+dist(B,P) sliding along the circle where A is current position, B the vortex found by Dijkstra on the next waypoint OZ and P the point we search on the current waypoint OZ initialised from the Dijkstra vortex.
* Replace the vortex by the new P with SetPointSearchMin ()

Thinking at the same time about #3076, there is (or will be once #3076 is fixed) the possibility that the Dijkstra vortex is on a straight edge of the OZ or at a vortice which is the corner between straight and arc.
One way to solve that is to track in the various SetBoundary() is a vortex is on an arc, straight or corner.

Issue #2388, opened 2026-04-12
Reggata Start


lordfolken:
## Summary (from [discussion #1891](https://github.com/XCSoar/XCSoar/discussions/1891))

**Discussion title:** Regatta Start Development?

Currently when behind the start line the track goes to the center or the end of the line. For regatta starts it would be great if XCSoar could project the start point to where the plane is pointing and then have the time and distance shown to reach that point (or even have a countdown to a specified time) at the current ground speed.

**Safety note from the thread (JacquesGr):** pilots often approach a long start line **perpendicular** to the line; following a cue that aims at the **centre** (or another “optimal” point for the next leg) can put a glider **across** the flow of traffic.

### Discussion participants

cc @rjfrawley @lordfolken @JacquesGr

---

### Current behaviour

For ordered tasks, the start uses `StartPoint::find_best_start()`, which picks a point on the start observation-zone boundary that **minimizes distance via the start line to the next turn** (`DoubleDistance` from aircraft → boundary point → next leg). The map/bearing line therefore follows that **distance-optimal** point, which often feels like “center or end of the line” rather than “where my nose points.”

There is an existing TODO on the start point about **gate start time**, which aligns with optional countdown-style behaviour.

### Requested behaviour (regatta-style)

1. **Projection along the nose** — On the start line, use the point that lies on the aircraft **heading/track** (e.g. intersection of the line with a ray from the current position along flight direction), not the point that minimizes distance to the next turn.
2. **Time and distance** to that point using **current ground speed**.
3. **Optional:** countdown to a **specified start / gate time**.

### Implementation notes

| Area | Notes |
|------|--------|
| **Scope** | Split into: (A) geometry / target point, (B) ETA at ground speed, (C) optional gate countdown — each can be specified and tested separately. |
| **Geometry** | Clarify: segment vs infinite line; **track vs heading** (GPS track vs compass heading, slip/wind). |
| **UI** | Task option, map overlay, and/or InfoBox; must remain usable on small screens and e-paper. |
| **Architecture** | Prefer a dedicated “regatta start” mode or parallel computed fields so classic XC distance minimization and scoring paths stay unchanged. |
[...]

Issue #1737, opened 2025-04-07
LXNav S8/80/10/100 varios: Task declaration start finish/point types incorrect


groundhogxc:
XCSoar versions having the problem
-------------------------------------------------

7.4x, and 7.3x seem to have the problem, probably earlier ones, too

System information
------------------

Seen with Android version

Steps to reproduce the behavior
-------------------------------

* create task in XCSoar or e.g.  download from weglide 
* use default turnpoint types: A cylinder start point and finish point, and symmetric sectors for the turnpoints
* declare task to LXNav device

Expected behavior
-----------------

Turnpoint types should be the same in LXNav as in XCSoar

Actual behavior
---------------

The task shows up  in the LXNav device with symmetric sector type turnpoints for ALL task points, including start and finish

Do you have any idea what may have caused this?
-----------------------------------------------

It looks like somewhere in the declaration process the turn point type info is lost and something (driver or device) then reverts to sector types as a default, maybe?

PNich:
See attached - Output from XCS RS232 traffic captured and response from LXNAV.

XCS sends turnpoint positions OK but does not send turnpoint types.

LXNAV reports task declared, XCS reports declaration failed.

[XCS decl.txt](https://github.com/user-attachments/files/19725043/XCS.decl.txt)

[LXNAV resp.txt](https://github.com/user-attachments/files/19725044/LXNAV.resp.txt)

groundhogxc:
Looking at the dataport specification file V 1.04 (Linked in issue #1248), it seems to be supported by the protocol at least- their example declaration file is:

`HFPLTPILOT:Uros Krasovic
HFCM2CREW2:
HFGTYGLIDERTYPE:
HFGIDGLIDERID:
HFCIDCOMPETITIONID:
HFCCLCOMPETITIONCLASS:
C050221133458050221020502
C0000000N00000000E
C4846667N01015883EAALEN HEIDENHEI::585.00000
C4711500N00942100ED21CH NENZING::572.00000
C4748667N01346550EEBENSEE EBXTRAUN::420.00000
C4744110N01133501EWACKERSBERG::831.00000
C0000000N00000000E
LLXVOZ=-1,Style=2,R1=3000m,A1=45.0,R2=0m,A2=0.0,A12=193.6,Maxa=0m,Line=1,Autonext=1,Lat=4846.667N,Lon=01015.883E,Near=0
[...]

PR #3080, opened 2026-09-09
Task: navigate to the nearest point of the start and finish

Vib38:
This answers #2388 on regatta start.

XCSoar aims at the point of the start or finish zone which makes the task shortest. On a line the boundary offers only the centre and the two ends, so it picks the centre; under a 20 km start line that is kilometres from where the glider will cross, and the distance, the time and the arrival height all describe a place nobody is flying to.

This adds a task rule which aims at the nearest point of the zone instead, while the start or the finish is the active task point. The point moves as the glider moves.

image

- `LineSectorZone` and `CylinderZone` implement `GetNearestPoint()`. Sectors and keyholes return an invalid location and keep the point the task already uses. So does a cylinder the glider is inside, where the nearest point of the rim is the wrong way out, away from the next task point.
- A line gains the most. On a cylinder the task already refers to a point of the rim, but the one which makes the task shortest rather than the one the glider is nearest to, quantised to one of twenty boundary samples; on the 5 km start cylinder of the test the two are 800 m apart.
- `GetLocationNavigation()` is a new virtual on `TaskPoint`, so the planned distance, the distance minimisation, the scoring and the AAT isolines keep using `GetLocationRemaining()`. The remaining vectors follow the nearer point, and with them the remaining distance, the glide solution and `TaskStats::current_leg`.
- Stored on `OrderedTaskSettings`, `.tsk` attribute `navigate_nearest`, profile key `NavigateNearest`. Off by default. Config → System → Task Defaults → Task Rules seeds new tasks and tasks read from a file which carry no such attribute; the "Rules" page of the task manager edits the task in use.

Image

Following the review: the rule moved off `StartConstraints` onto `OrderedTaskSettings`, since it now covers the finish as well and says where to navigate rather than what makes a start valid. A `.tsk` written by an earlier version of this branch loses the flag; nothing on master is affected.

Not in this PR: the LXNAV declaration still sends `Near=` hardcoded (`LLXVTSK Near=true`, finish `Near=1` / start `Near=0`). Plumbing it through `Declaration` changes what goes out on the port — with the rule off, a finish goes from `Near=1` to `Near=0` — so it deserves its own commit and its own `TestDriver` cases. Happy to follow up, as @lordfolken offered.

Tested with `make TARGET=UNIX check`: 114 files, 11089 tests, all green. `TestOrderedTask` covers the option on and off for a start line, a start cylinder, a start sector, a cylinder the glider is inside, and a finish line and cylinder, plus the planned and minimum distances staying put; `TestTaskSave` covers the `.tsk` round trip. Also flown in the simulator on a 20 km start line.

#### Branch & Rebase

#### Commit Format & Messages

#### Commit Structure

---

- [x] I'm ready to merge

## Summary by CodeRabbit

- **New Features**
  - Added an optional “Navigate to nearest point” task rule for start and finish observation zones.
  - Navigation guidance, displayed distance, estimated time, arrival height, and glide calculations now reflect the observation-zone point closest to the glider when enabled.
  - Added controls in task properties and settings, with the option disabled by default.
  - Start cylinders and lines are supported; sector behavior remains unchanged.

- **Documentation**
  - Added user guidance and limitations for the new task rule.

- **Tests**
  - Added coverage for nearest-point navigation and task-setting persistence.

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -58,6 +58,10 @@ Version 7.46 - not yet released
 * airspace
   - fill airspaces with a transparent colour; hatch patterns and the
     Airspace transparency setting are gone
+* task
+  - add a "Navigate to nearest point" task rule: aim at the start and
+    finish zone where the glider is going to cross it, not where the
+    task is shortest; off by default #2388
 * FLARM
   - fix ADS-B traffic ground speed wrapping above 457 km/h #2887
   - fix no-position targets missing from the radar and map #2847
--- a/doc/manual/en/ch04_xc_tasks.tex
+++ b/doc/manual/en/ch04_xc_tasks.tex
@@ -362,6 +362,9 @@ type allows the pilot to enter the following parameters
   \item [Arm start manually] If set to ``On'', some extra buttons are presented
   in order to let you \bmenug{Arm start} as well as \bmenug{Disarm start} in
   menu \bmenug{Nav 1/2}, controlling  the detection of the start condition.
+  \item [Navigate to nearest point] If set to ``On'', XCSoar aims at
+  the nearest point of the start and finish zone instead of the point
+  which makes the task shortest.  See Section~\ref{sec:task-rules}.
   \item [Start open time] This is the time when the start zone opens.
   \item [Start close time] This is the time when the start zone closes.
   \item [Start max.\ speed] This is the maximum speed allowed in the start observation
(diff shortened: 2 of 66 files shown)


Vib38:
Guard added, thanks. `GetNearestPoint()` now returns `GeoPoint::Invalid()` when the projected line has zero length, and the caller already falls back to the task point's own location. Worth noting the case only exists in release builds: `SetLength(0)` trips `assert(new_radius > 0)` in `CylinderZone::SetRadius()` in debug.

Leaving the two UPPER_CASE nitpicks as they are. `Keys.hpp` and the `TaskRulesConfigPanel` control indexes are CamelCase throughout (`StartMaxSpeed`, `StartHeightRef`, `StartMaxHeightMargin`), so renaming only the new entries would break consistency with the surrounding code.

lordfolken:
Thanks for this — the storage and the nav/scoring split are in the right place. Compared to the plan on #2388, a few gaps if we want XCSoar and the LXNAV declare to match:

**Already good**
- Task-level `StartConstraints` / `.tsk` `start_navigate_to_nearest`, not a task type
- Task Manager → Properties + Config → Task rules
- `GetLocationNavigation()` vs `GetLocationRemaining()` so scoring stays put
- Start line: clamped perpendicular, not centre/ends

**To conform**

1. **Finish.** LXNAV's option is start *and* finish (glide to the rim of a finish cylinder, nearest point on a finish line). This PR only moves the start line. Finish remaining is still the waypoint centre.

2. **Start cylinder/sector.** Only `LineSectorZone::GetNearestPoint()` is implemented. The option is named as if it were line-only; LXNAV applies it to the start *zone* as well.

3. **LXNAV declare.** Still hardcoded: [`NanoDeclare.cpp#L151`](https://github.com/XCSoar/XCSoar/blob/master/src/Device/Driver/LX/NanoDeclare.cpp#L151) `LLXVTSK Near=true`, [`LXNavDeclare.hpp#L151`](https://github.com/XCSoar/XCSoar/blob/master/src/Device/Driver/LX/LXNavDeclare.hpp#L151) finish `Near=1` / start `Near=0`. That came from the dataport example in #1737, not from XCSoar nav. Please plumb `Declaration` and send `Near=` from this setting (start **and** finish when on). FLARM has no field — skip. CUP `NearDis` is a scoring tolerance, not this.

4. **Default.** Code/NEWS are off; the issue is still open on default-on vs opt-out. Either is fine once it is in the `.tsk` and on the declare. Please make the PR body match the code (it still says "on by default" in one place).

Config → Turnpoint types is optional if Task rules already seeds new tasks.

Happy to take finish + declare in a follow-up if you want this start-line piece in first.

PR #3098, opened 2026-09-10
doc: Document the XCSoar .tsk task file format

### LINKED ISSUE #3097: Document the XCSoar .tsk task file format
## Summary

There is no in-tree description of XCSoar's `.tsk` XML format. The user manual only says it is "XCSoar's own XML format". The actual definition is the round-trip in `src/Task/Serialiser.cpp` and `src/Task/Deserialiser.cpp`.

That is awkward for anyone writing a converter, WeGlide export, or a declare path: they have to reverse-engineer the C++. #2388 / #3080 add `start_navigate_to_nearest` on `<Task>` and we told converters they can set it — with nowhere to look it up.

## What to write

A short developer note (same idea as `doc/mapfile.rst`), e.g. `doc/task_file.rst`, linked from `doc/index.rst` and from a comment on the serialiser.

Enough to round-trip a task:

- Root `<Task>`: `type` (`RT`, `AAT`, `MAT`, `FAIGeneral`, …) and the start/finish constraint attributes (`start_requires_arm`, `start_score_exit`, start open/close, max speed/height, PEV, `aat_min_time`, and `start_navigate_to_nearest` once #3080 lands)
- `<Point type="Start|Turn|Area|Finish|OptionalStart">` with `Waypoint` / `Location` and `ObservationZone` (`Line`, `Cylinder`, `Sector`, `FAISector`, `SymmetricQuadrant`, …)
- Load rule: unknown attributes are ignored; missing attributes use defaults
- What this file is **not**: CUP `NearDis`/`NearAlt` (scoring tolerances), LXNAV `LLXVTSK`/`LLXVOZ` `Near=` (declaration)

Do **not** add an XSD. The format is whatever the serialiser writes; a schema would rot.

The user manual can keep a one-liner and point here.

## Why now

#2388 needs converters and the LXNAV declare to agree with the `.tsk`. Without a page that lists the `<Task>` attributes, "don't change the default, external tools will break" is guesswork.

Related: #2388, #3080

dspreitz:
Closes #3097.

## What this does

There is no in-tree description of the `.tsk` XML format — the only definition is the round trip between `src/Task/Serialiser.cpp` and `src/Task/Deserialiser.cpp`, so anyone writing a converter, a WeGlide export or a declare path has to reverse-engineer the C++.

This adds `doc/task_file.rst`, a developer note in the same spirit as `doc/mapfile.rst`:

- The document wrapper: no XML declaration, UTF-8, tab indentation, the five entities plus decimal character references the parser understands, and the 65536-byte limit in `src/XML/Parser.cpp`.
- Every `<Task>` attribute with its unit, its encoding and what `0` means — `aat_min_time` in seconds, `start_max_speed` in m/s, heights in metres, the start gate as `HH:MM` UTC, the PEV pair, `fai_finish`.
- `<Point>` types (`Start`, `Turn`, `Area`, `Finish`, `OptionalStart`), `<Waypoint>`, `<Location>`, and the 10 m waypoint-database matching that can discard the file's `id`/`comment`/`altitude`.
- A table of all eleven `<ObservationZone>` types with their attributes and defaults.
- The load rules a converter actually trips over: unknown attributes are ignored; a **missing** attribute is not "the default" but the pilot's configured task defaults narrowed by the factory; booleans are integers, not `true`/`false`; enum values are case-sensitive even though element and attribute names are not.
- What the format deliberately does **not** carry: CUP `NearDis`/`NearAlt` (scoring tolerances) and the LXNAV `LLXVTSK`/`LLXVOZ` `Near=` field (a declaration property computed in `LXNavDeclare.hpp`).

No XSD, as the issue asks — the format is whatever the serialiser writes, and a schema would rot.

The page is linked from `doc/index.rst`, pointed at from a comment on the serialiser and deserialiser, and the user manual gets a one-liner pointing here.

## Verification

Checked against a built `DumpTaskFile`, not only against the source. That caught a wrong claim in an early draft: the `FAI*` types do not uniformly collapse cylinders into FAI quadrants. `FAIGeneral` and `MAT` rebuild *every* point at the factory's standard dimensions (a 5000 m start line comes back as the FAI 1000 m line, a 1000 m turn cylinder as the 500 m one), while `FAITriangle` leaves legal zones untouched and swaps only the illegal ones. The page describes the observed behaviour.

The complete example in the page is verbatim `DumpTaskFile` output and round-trips byte-identically. The case-sensitivity, unknown-attribute and missing-zone rules were each confirmed with a crafted file. Sphinx builds the new page with no warnings.

## Note on #3080

`start_navigate_to_nearest` is **not** documented here — #3080 has not landed, so there is no such attribute in the tree yet. It needs one entry added to the `<Task>` section when that merges; happy to follow up afterwards, or to fold it in here if #3080 goes first.

Related: #2388, #3080

#### Branch & Rebase

#### Commit Format & Messages

#### Commit Structure

Single commit, `doc: Document the XCSoar .tsk task file format`, on `087f48ea6`.

### Code, tests & docs (where it applies)

- [x] Reuses existing helpers and UI strings (no parallel API, extra abstraction, or near-duplicate `_()` msgid) — n/a, no code changes beyond two comments
- [x] TAP tests for new UI-free logic (parsers, formatters, protocol, math/geo); registered in `build/test.mk` — n/a, no new logic; the existing `DumpTaskFile` was used to verify the page instead
- [x] User-visible behaviour: `NEWS.txt` and English manual (`doc/manual/en/`) — English manual updated (`ch12_data_files.tex` now links to the new page); no `NEWS.txt` entry, as nothing about XCSoar's behaviour changes
- [x] New gettext strings: `make update-po`; `.po` diff only adds new fields — n/a, no new strings

---

- [ ] I'm ready to merge


lordfolken (approved):
Thank you.

PR #3114, opened 2026-09-11
Flight detection improvement

### LINKED ISSUE #3060: 7.45.1 does not detect take-off

7.45.1

not 100% sure but either 7.44 or 7.45

System information
------------------

HiBreak Pro (Android Phone) - will cross check with a Samsung S23 Ultra

Start XCSoar
Check GPS signal is ok, moving map is ok, baro connected
Take-off (paraglider)
Look at XCSoar and check the Flight duration Infobar

Expected behavior
-----------------

 
(1) Flight duration Infobar start to count flight duration

Actual behavior
---------------

(1) Flight duration Infobar displays nothing "---"
(2) no switch from cruise to circling mode: the infobar set remains the default one
(3) GPS position is updated, moving map is updated, ground speed is computed, track is available, altitude via gps (or baro) is available, Height is available
(4) After a while (around 10-15 minutes I would guess) the flight duration start to count (from 0 so the first 10-15 minutes are not taken in account)
(5) The same behavior can be seen by replaying the IGC file via the replay function.

Any change to the threshold speed/logic to automatically detect the takeoff? There is nothing unusual in the take-off sites where I can see this behavior and other applications (e.g. SeeYou Navigator, burnair) can detect the take off without problems.

See above

Scumi:
- Detect actual takeoff, landing and duration in the existing Export Flights list, addressing #2964 and #3060
- Export Flight Dialog improved: scan IGC files asynchronously, newest first, while keeping the dialog responsive and updating results incrementally
- Share IGC-to-NMEA conversion between replay and metadata detection, including missing ground-speed derivation
- Improve slow-launch and low-speed landing detection while retaining polar-dependent behavior for live flights

## Summary by CodeRabbit

* **New Features**
  * Export Flights now detects and displays takeoff time, landing time, and flight duration as flight logs are scanned in the background.
  * Flight-time detection supports slow launches and low-speed wave or ridge flying.

* **Bug Fixes**
  * Reduced false landing detection during sustained low-speed flight.
  * Improved handling of midnight crossings and irregular flight-log timestamps.

* **Documentation**
  * Added guidance for background flight-time detection in the Data Management dialogue.


lordfolken:
@bomilkar could you review this please?

bomilkar:
> @bomilkar could you review this please?

Thanks for inviting me to the review.

After reading through the commits and the one issue report I'm getting an idea of what this is trying to achieve:
1. automatically detect takeoff (and landing), namely for (misconfigured) paragliders
2. generate NMEA from IGC

@Scumi : is this complete? Is there a description of the concepts, other than the commit messages? I'd like to read this before diving into the detail.

Just for background: I'm working with @ulipo on a related issue which is strongly linked to paragliders' characteristics: slow and unknown airspeed, slow and sometimes "negative" ground speed, unusual (in the eyes of a glider pilot) flight pattern.

I know a bit about paragliding. I've been flying paragliders for 20+ years, but that's many years ago. Pragliders have improved a lot I've been told. From that experience I know for sure it is impossible to auto-detect takeoff reliably. (I've been flying near Fort Funston at the Pacific coast. It is not unusual to remain "fixed" at 5m above ground for as long as you want. Similar cases in the Alps.) Even circling is not a reliable indicator because of the "random" pattern they often fly.

Therefor I would recommend to take a more pragmatic approach to this issue; provide a means of "Manual Flight Detection" as opposed to the automatic detection. Perhaps configure a "Takeoff/Landing" Quick-Button which appears over the map as soon as automatic detection is turned off.

And @paolo-ch suggestion to check for obvious misconfiguration is a good idea: There is a message box appearing at the top of the map at startup for a few seconds "Simulator Steer by dragging... ". There could be a message "you have no polar configured!", and similar warnings.

Scumi:
Thanks for the review.

Two clarifications: 
- automatic take-off / landing detection already existed before this PR. It just extends it for slow launches and reuses it for IGC flight times - used by the Export Flights dialog. It does not generate NMEA sentences. The fixes are only mapped to the internal NMEAInfo so the existing detector can process them.
- GPS-only detection will probably never cover every paraglider situation, but I think this PR is a reasonable improvement. Pilots who need to ensure the whole flight is recorded can start the logger manually before take-off or use the existing “Start only” mode. A manual flight-detection button or polar warning could be a separate PR.

paolo-ch:
_"GPS-only detection will probably never cover every paraglider situation, but I think this PR is a reasonable improvement. Pilots who need to ensure the whole flight is recorded can start the logger manually before take-off or use the existing “Start only” mode. A manual flight-detection button or polar warning could be a separate PR."_

Burnair, SeeYou Navigator and XCTrack (and probably many others) are fully capable to detect a paraglider taking off based on the GPS alone.
My vario (Stodeus Ultrabip) is also capable to detect a paraglider taking off.

And actually XCSoar was capable to do reliably in the past.

If the polar is now mandatory  to enable take-off detection it is perfectly fine (the polar is useful for other purposes), just mention this but please do not put a band-aid by adding a button to force logging: If other applications can detect a take-off, then XCSoar should be capable to do that.

Scumi:
Well, this PR adds a polar-independent slow-launch path, so XCSoar should detect paraglider launches more reliably even when no polar is configured.

I’m not saying that GPS-only detection is impossible, only that some situations make it difficult to do reliably. Without knowing the algorithms and sensor inputs used by the other apps, comparisons are hard to evaluate.

In any case, this PR should improve the reported behaviour as well as improve the Export Flights dialogue.

lordfolken:
@paolo-ch Polar is not a new take-off requirement. Detection has always been `Vmin/2` of the loaded polar, for 10 s. The factory default is LS-8-class, so that gate is ~42 km/h. PolarStore Para polars are 12–19 km/h (3.3–5.2 m/s) — the same band XCTrack / Flyskyhy / UltraBip use (~10 km/h GPS). With a PG polar configured, a normal foot launch already starts flight time and the logger the same way those apps do.

paolo-ch:
@lordfolken 
Thanks for the Polar clarification that it has been always needed. Now I begin to figure out what happened.
I have set a Polar ages ago and then simply done a copy&paste of the XCSoar folder from the old version to the new one. The Polar has always been present in any version with the exception of the latest (7.45.1?) in which despite the copy&paste I discovered I was flying without a polar set.
 
@bomilkar
Paragliders are today a bit faster, with a better Glide Ratio (still a far cry compared to gliders), but most of all capable to penetrate the wind, especially the latest generation of two liners with shark nose, less camber, more cells, higher A/R flown by pilots inside an aerodynamic pod. Not like the old paragliders that a friend of mine uses to call "matratzen" due to the giant intakes they had...

lordfolken:
The question here is how to handle the polar. 
I'm personally unfan of having a preconfigured polar because ultimately that is what lead to this whole discussion. The advantage is the "it just works" factor and with the general population's attention span running thin, popping up a lot of dialogs isn't really what is expected. 

An option would be to configure a PG Polar as soon as launch detection is detecting a PG style takeoff?

bomilkar:
> With a PG polar configured, a normal foot launch already starts flight time and the logger the same way those apps do.

I can confirm this from analyzing paraglider NMEA logs.

I think taking HAGL into account is a good idea! That should help gliders on wave flights and PGs in the mountains.

Reading through the code the only change I would suggest: use the slow-launch parameters only when there is no polar (and map) configured. I'm a bit concerned that when you pull a glider to the start point it will have detected takeoff already. @Scumi what do you think?

Scumi:
HAGL is already used after take-off to prevent false landing detection. The slow-launch detection also needs at least 3 m/s, ten seconds of climbing and 20 m height gain, so simply moving a glider should not trigger it. This is also covered by a test.

An uphill tow exceeding those conditions could still look like a launch (but I question someone doing this as its pretty extreme for a non-launch tow). I would rather see a representative NMEA log before restricting this path to missing polars, because that would also remove the fallback for configured paragliders and for imported IGC files (which was the main reason to look at this for me).

I would also not configure any polar based on automated detection. Polar configuration should be a deliberate user choice. A potential false detection could change an unconfigured glider into a paraglider, which would cause very weird effects for that glider pilot.

Edit: I was not aware that if nothing is configured, an LS8-15 polar will be used internally for calculation. I'll have a look at that because this would not be ideal for PG.

bomilkar:
> I would also not configure any polar based on automated detection. Polar configuration should be a deliberate user choice.

I strongly agree with this: changing any deliberate configuration (while flying) is an **absolute no-no** in my opinion.

bomilkar:
> I was not aware that if nothing is configured, an LS8-15 polar will be used internally for calculation. I'll have a look at that because this would not be ideal for PG.

I was hoping that your slow-launch would address it so that for instance `output/UNIX/bin/RunCirclingWind` will work for PG logs, too.

I'll test this later. @Scumi : which branch shall I clone to test it?

Edit: I cloned the branch and ran RunCirclingWind. **And it works!** Obviously the slow-launch does the trick.

Scumi:
Rebased.

I think the issue with the polar defaults need to be solved, just not in this PR. This should be a follow-up as I know there is work on possibly disallowing not setting any aircraft config.

The branch is mentioned at the top of this PR, its `improvement/flight-detection`.

lordfolken:
The IGC flight-list part of this pull request is now in #3196.

Your commits there are the original objects, unchanged. The slow-launch and polar landing commits stay on this branch and are not part of #3196.

#3196 carries that list on into WeGlide Upload and Replay, and adds the hours:minutes total for IGC files with a detected launch and landing.

lordfolken:
Concerns on the live takeoff and landing changes. 

Above 50 m AGL, a flight that has already started no longer ends. `clearly_airborne` forces the moving state at any ground speed, including zero, whenever terrain height is more than 50 m. Previously only 300 m AGL did that. GPS altitude and the terrain grid are often wrong by more than 50 m, so a real landing in a valley or on a ridge can leave the flight running. The logger, statistics, task, and takeoff/landing events all follow `flight.flying`. `CheckLandingEitherSideOfAGLBoundary` locks in “51 m stays flying.” That matches the code, and 50 m is still a bad threshold for live use.

The slow-launch detector is live as well. Below the polar takeoff speed, 3 m/s plus about 10 s of climb and 20 m of net gain calls `Takeoff()`. Taxiing up a slope, or GPS altitude wandering 20 m while rolling, can start a flight. The detected time is the first rising sample, which can be before the aircraft leaves the ground. The header already says so.

Inside that 50 m band, climb no longer blocks a landing quickly. The old climb clock counted as climbing after about one second. `ClimbEvidence` now needs 10 seconds before `landing_climbing` is true. For a slow polar the landing-speed cap does not move (`min(4, takeoff_speed / 2)` stays at half the takeoff speed). Low-level ridge flight at low ground speed, with terrain height at or under 50 m, is more likely to be declared a landing than before.

The headwind adjustment now scales only `landing_speed` and leaves the takeoff threshold alone. While already flying, a speed between those two thresholds freezes both clocks, so it still refuses a landing. That part is sound.

The new tests follow this state machine closely, including the 50 m boundary. They will not catch a landed aircraft whose terrain height stays above 50 m.

Scumi:
- removed the unconditional 50m AGL safeguard
- the "uphill taxi" false-trigger requires a fairly specific combination of circumstances, which I find unlikely to occur. A representative log would be useful. If it does happen, stopping the taxi will normally end the detected flight again. A wrong flight entry can be deleted, but a missed paraglider take-off cannot be recovered.

lordfolken:
I present you my completley nuts airfield with a 50m inclined runway:

Screenshot_20260923-132029_1

towing uphill would trigger this. 
can we still consider the polar vmin?

bomilkar:
> I present you my completley nuts airfield with a 50m inclined runway:

Not so unusual. Another good one is Wasserkuppe. Or la Motte du Caire .

Scumi:
Don't worry guys. I fly on similar fields as well and the takeoff detection should not trigger there for mere taxiing your aircraft. My assumption on the feedback was that you have to pull your aircraft by more than 40 km/h uphill to trigger it, which would be excessive. The fact that it would still trigger right now on lower speeds is because I implemented the 10 second timer incorrectly. 

Anyway, I'll provide an update (and adjust the approach to this a bit) and then we can have a look again :-)

PR #3116, opened 2026-09-11
Device/Driver/LX: Restore signed IGC altitudes from LXN

### LINKED ISSUE #3075: IGC file download with LXNav driver under Android

From version 7.40 upwards under Android 

under Windows the problem does not occur. I cannot test other operation systems.
 Version 7.39

System information
------------------

Samsung Galaxy A71 with Android version 13 and Samsung Tablet A11 with Android version 16

Download an IGC file from LX7007IGC pro or from a Colibri F.

Expected behavior
-----------------
The downloaded IGC file should contain correct data.

Actual behavior
---------------
The downloaded IGC file contains wrong pressure altitude values in the first B-records, which mark the Zero Altitude line.
This leads to a complete wrong evaluation of the flight although the recorded pressure altitudes during the flight are correct.

I think there must be a bug in the SW routine that decodes the values from the original flight data in the LX Navigation  loggers and creates the IGC file.

No.

lordfolken:
## Summary
- Restore the `int16_t` casts when converting LXN logger altitudes to IGC B records. The 7.40 fmt switch dropped them, so negative pressure altitudes (typical on the ground / zero line) wrapped to ~65000 m.
- Add a TAP case that builds a minimal LXN blob with −57 m / −1 m and checks the IGC fields are `-0057` / `-0001`, not `65479` / `65535`.
- Fixes #3075 (Colibri F and LX7007 IGC pro; same conversion on Android and Windows).

## Test plan
- [x] `make -j$(nproc) TARGET=UNIX USE_CCACHE=y output/UNIX/bin/TestLXNToIGC && ./output/UNIX/bin/TestLXNToIGC`
- [ ] Download a Colibri / LX7007 IGC flight that starts below ISA 0 m and confirm the first B records use signed altitudes (e.g. `-0057`), matching 7.39

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed IGC downloads from LX Navigation loggers so negative pressure altitudes are recorded correctly instead of appearing as values around 65,000 m.
  * Affected logger formats include Colibri, LX20, and LX7007 IGC.


nifidel:
Hi,
I tested WINDOWS and Android XCSoar 7.45.2 with the same flight log.
The downloaded igc file is okay.

PR #3187, opened 2026-09-20
build/android.mk: Build the arch-independent outputs only once

### LINKED ISSUE #3186: ANDROID CI: intermittent classes.dex failure — concurrent submakes rebuild the shared noarch/classes.zip
## What happens

The `build (ubuntu-24.04, ANDROID, XCSoar, bin, XCSoar, .apk, true, r26d, debian:trixie-slim)`
job fails intermittently with:

```
make[1]: *** [build/android.mk:383: output/ANDROID/noarch/classes.dex] Error 1
make: *** [build/android.mk:436: output/ANDROID/x86_64/dbg/bin/libxcsoar.so] Error 2
```

D8 did not fail to translate anything. Further up the log it says why it failed:

```
Exception in thread "main" java.lang.RuntimeException: com.android.tools.r8.CompilationFailedException:
    Compilation failed to complete, origin: ./output/ANDROID/noarch/classes.zip
Caused by: com.android.tools.r8.internal.g: java.nio.file.NoSuchFileException: ./output/ANDROID/noarch/classes.zip
```

Its input file was gone by the time it opened it.

## Why the file disappears

`output/ANDROID/noarch/` is shared between all ABIs, and two rules in `build/android.mk`
write it:

```make
# 362
$(NO_ARCH_OUTPUT_DIR)/classes.zip: $(JAVA_SOURCES) $(GEN_DIR)/org/xcsoar/R.java | ...
	... javac ...
	$(Q)rm -f $(NO_ARCH_OUTPUT_DIR)/classes.zip
	$(Q)$(ZIP) -0 -r $(NO_ARCH_OUTPUT_DIR)/classes.zip $(JAVA_CLASSFILES_DIR)

# 382
$(NO_ARCH_OUTPUT_DIR)/classes.dex: $(NO_ARCH_OUTPUT_DIR)/classes.zip
	$(Q)$(D8) ... --output $(NO_ARCH_OUTPUT_DIR) $(NO_ARCH_OUTPUT_DIR)/classes.zip
```

Both are reached from inside the per-ABI **submakes**. `generate-abi` (android.mk:401)
builds each `libxcsoar.so` by recursion:

```make
$$(TARGET_OUTPUT_DIR)/$(2)/$$(XCSOAR_ABI)/bin/lib$(1).so: $(NATIVE_HEADERS) generate boost FORCE
	$$(Q)$$(MAKE) TARGET_OUTPUT_DIR=... TARGET=$(3) ... $$@
```

and in that submake the Android objects depend on the generated JNI headers
(android.mk:453ff), which depend on the dex:

```make
$(NATIVE_HEADERS): $(NO_ARCH_OUTPUT_DIR)/classes.dex
```

`generate-all-abis` instantiates this four times (armeabi-v7a, x86, arm64-v8a,
x86_64) and CI runs `make -j$(nproc)`, so up to four submakes run at once, each
with its own opinion about whether `noarch/classes.zip` is up to date, and with
nothing serialising them. GNU make serialises a target within one make instance;
it cannot serialise it across instances.

The log of https://github.com/XCSoar/XCSoar/actions/runs/35502810349/job/106057520867
shows exactly that. Four submakes start together:

```
09:46:22.8092975Z make[1]: Entering directory '/__w/XCSoar/XCSoar'
09:46:22.8102791Z make[1]: Entering directory '/__w/XCSoar/XCSoar'
09:46:22.8103463Z make[1]: Entering directory '/__w/XCSoar/XCSoar'
09:46:22.8106858Z make[1]: Entering directory '/__w/XCSoar/XCSoar'
```

and one submake's `zip` is still adding entries while another's D8 is already
unwinding:

```
09:46:40.9770980Z   adding: output/ANDROID/noarch/classes/ioio/lib/impl/SequencerImpl$ChannelPwmSpeed.class (stored 0%)
09:46:40.9772031Z Compilation failed
09:46:40.9778019Z   adding: output/ANDROID/noarch/classes/ioio/lib/impl/QueueInputStream$State.class (stored 0%)
09:46:40.9781488Z Exception in thread "main" java.lang.RuntimeException: ...
09:46:40.9802979Z   adding: output/ANDROID/noarch/classes/ioio/lib/impl/InterruptibleQueue$Nudged.class (stored 0%)
09:46:41.0487113Z make[1]: *** [build/android.mk:383: output/ANDROID/noarch/classes.dex] Error 1
```

The `rm -f classes.zip` at the head of the second submake's recipe is what removed
the file the first submake's D8 was about to open.

## It is timing, not branch content

| PR | head | ANDROID job |
| --- | --- | --- |
| #2998 | 02bb558b1 | [failure](https://github.com/XCSoar/XCSoar/actions/runs/35502810349/job/106057520867) |
| #2966 | bde1fe6d9 | [failure](https://github.com/XCSoar/XCSoar/actions/runs/35501584707/job/106054655899) |
| #3000 | e9e5f3ddb | success |
| #3042 | a3d054db9 | success |
| master | c96d4ee2f | success |

#3000 contains every commit #2998 has plus 21 more and passes. None of the four
branches touches `build/android.mk`, `android/`, or any `.java` file. The failures
are reproducible only in the statistical sense: re-running the job is the current
workaround.

## Possible fixes

1. **Write `classes.zip` atomically.** Build it under a temporary name and `mv` it
   into place instead of `rm -f` + `zip` on the live path. `mv` within one
   filesystem is atomic, so a concurrent D8 sees either the old file or the new
   one, never none. The same applies to `classes.dex` and `--output`.
2. **Keep the submakes out of `noarch` altogether.** The parent already lists
   `$(NATIVE_HEADERS)` as a prerequisite of the per-ABI rule (android.mk:415), so
   the headers and the dex exist before any submake starts; the submakes only
   re-derive them because the rules are visible there too. A variable passed down
   by the recursion that turns the `noarch` rules into no-ops would remove the
   second writer instead of making it safe.

The second is the real fix; the first is a small, local change that would stop the
CI noise on its own.

dspreitz:
Fixes #3186.

## The problem

`output/ANDROID/noarch/` holds everything that does not depend on the ABI:
the compiled Java classes, `classes.zip`, `classes.dex` and the generated
JNI headers. A fat-binary build reaches it from each of its four per-ABI
submakes, because the Android objects compiled there depend on the headers
(`build/android.mk:453ff` → `$(NATIVE_HEADERS)` → `classes.dex`).

make serialises a target within one make instance. It cannot serialise it
across instances, so under `-j` all four submakes ran `javac`, `zip` and
`d8` over the same files at the same time. The zip recipe removes
`classes.zip` before writing it, so a D8 that opened it inside that window
found nothing at all:

```
Caused by: com.android.tools.r8.internal.g: java.nio.file.NoSuchFileException:
    ./output/ANDROID/noarch/classes.zip
make[1]: *** [build/android.mk:383: output/ANDROID/noarch/classes.dex] Error 1
make: *** [build/android.mk:436: output/ANDROID/x86_64/dbg/bin/libxcsoar.so] Error 2
```

One line of the CI log shows the collision directly — one submake's `zip`
writing while another's `rm` deletes:

```
09:46:39.7648138Z   adding: output/ANDROID/noarch/classes/org/xcsoar/PermissionHelper$4.classrm -f ./output/ANDROID/noarch/classes.zip
```

In the failing job the second build round (`PLAY=y`, after `aapt2 link`
rewrote `R.java`) started four `javac`, four `rm`+`zip` and four `d8` runs
within two seconds of each other. The first round, where the parent had
already brought `noarch/` up to date, ran exactly one of each.

## The change

The parent has built all of it by the time a submake starts: the generated
headers are prerequisites of the very rule whose recipe invokes the submake
(`build/android.mk:415`). So the submake has nothing to add — it only needs
to keep its hands off.

It is now passed `NO_ARCH_READY=y`, which leaves it without the recipes for
those files. It uses what it was given and writes nothing under `noarch/`.
Single-ABI builds do not set the variable and define the recipes as before.

As a side effect the fat build now runs `javac` once instead of five times.

### Why the header rule stays outside the guard

`$(NATIVE_HEADERS): $(NO_ARCH_OUTPUT_DIR)/classes.dex` carries no recipe of
its own — `javac -h` writes the headers in passing — and it is deliberately
left outside. A rule without a recipe is what lets make accept a target file
that was never created, and `NATIVE_CLASSES` names one such file:
`org_xcsoar_DownloadUtil.h`. Its Java class is not in the repository, `javac`
writes eleven headers and not that one, and its only consumer,
`src/Android/DownloadManager.cpp`, appears in no source list and is never
compiled.

The first revision of this PR guarded that rule away too, and CI showed the
consequence immediately:

```
make[1]: *** No rule to make target 'output/ANDROID/include/org_xcsoar_DownloadUtil.h',
    needed by 'output/ANDROID/armeabi-v7a/dbg/src/Android/EventBridge.o'.  Stop.
```

The submake must be as tolerant as the parent, not stricter. Keeping the rule
outside restores that and costs nothing: a rule with no recipe cannot write to
`noarch/` either. Whether `NATIVE_CLASSES` should still list a class that no
longer exists is a separate question — see below.

## Not addressed here

Two pre-existing oddities this change steps around rather than fixes:

- `NATIVE_CLASSES` is out of sync with the Java sources in both directions: it
  names `DownloadUtil`, which no longer exists, and does not name
  `StorageHotplugReceiver` or `XCSoar`, for which `javac` does write headers.
  `src/Android/DownloadManager.cpp` is dead code that includes the missing
  header.
- `$(GEN_DIR)/org/xcsoar/R.java` has a prerequisite but no recipe — aapt2
  produces it as a side effect of the base package. That is why `classes.zip`
  looks stale again in the second build round and the submakes wanted to
  rebuild it in the first place. With this change the rebuild is harmless,
  because only the parent performs it.

## Testing

The machine used has no Android SDK/NDK, so the ANDROID build was not run
locally — CI is the check that matters here, and it caught the first revision.
For the second, the CI situation was reproduced locally against the rule
database: `noarch/classes.dex` plus exactly the eleven headers `javac` really
writes, with `org_xcsoar_DownloadUtil.h` absent as in CI.

| check | result |
| --- | --- |
| submake asked for the never-generated header | `Nothing to be done` — no error |
| submake asked for a real header | `Nothing to be done` |
| submake asked for `noarch/classes.dex` | `Nothing to be done` — no rebuild, nothing written |
| same, with `classes.dex` removed | `No rule to make target …/classes.dex` — a broken parent contract fails loudly |
| submake without `NO_ARCH_READY` | still walks into the javac rule |
| `gmake TARGET=ANDROIDFAT -p -n` | all four generated submake recipes carry `NO_ARCH_READY=y`; parent keeps its own rules |

#### Branch & Rebase

#### Commit Format & Messages

#### Commit Structure

### Code, tests & docs (where it applies)

- [x] Reuses existing helpers and UI strings (no parallel API, extra abstraction, or near-duplicate `_()` msgid)
- [ ] TAP tests for new UI-free logic — not applicable, this is a build rule
- [x] User-visible behaviour: `NEWS.txt` and English manual (`doc/manual/en/`) — `NEWS.txt` under *development*; no user-visible behaviour, so no manual change
- [x] New gettext strings: `make update-po` — none added

---

- [x] I'm ready to merge

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed intermittent failures during Android fat-binary builds.
  * Improved build reliability by preventing concurrent build steps from conflicting when generating shared Java and DEX outputs.
  * Android builds now complete more consistently when multiple architecture-specific builds run in parallel.

* **Documentation**
  * Added a changelog entry describing the Android build reliability fix.


lordfolken (approved):

lordfolken:
Thanks for fixing this. :)

PR #3195, opened 2026-09-21
Docs: Migrate the manual and developer docs to markdown

WolfgangDrescher:
This replaces the LaTeX manual and the RST developer docs with a markdown site in `docs/`, built with Nuxt, Nuxt Content and Docus. The same content is printed to the PDF manuals. `docs/README.md` explains how to build the site and the PDFs to try this PR out. Draft until the open points below are settled.

Preview available on https://xcsoar-docs.wolfgangdrescher.ch/

## Done

- Nuxt site with Docus in `docs/`, XCSoar colours and logo, link to xcsoar.org, landing page with an overview and the section navigation.
- User manual: all 14 chapters, preface and licence migrated from the LaTeX manual, chapter introductions as pages of their own where needed.
- Quick guide migrated and linked to the user manual.
- InfoBox reference: one page per InfoBox with the metadata from the C++ source, generated by a script that is rerun after code changes; index page; descriptions taken from the LaTeX manual; bodies that only repeat the help text removed.
- Developer docs migrated from RST (diagrams rendered to SVG).
- Hardware and docs pages content migrated from the `XCSoar/website`.
- Syntax highlighting for C++, Lua, XML and make.
- Figures with captions at their natural size, served from the LaTeX directories for now.
- Gestures as drawings in the text, beside paragraphs and in tables; gesture table with IDs and characters; menu overview and default keys as tables.
- Tables sized to their content, wide tables printed with sideways headings.
- PDF export of every section with Paged.js, with the commit on the title page.
- Version badge in the header linking to the release or commit, "Last updated" line below every page, including the history of the migrated LaTeX and RST sources.
- Edit link on every page.
- Text conventions applied throughout: backticks for on-screen text, 80 columns, single spaces, consistent heading spacing, no repeated chapter titles after links.
- Titles, navigation titles and slugs reworked for the sidebar and short URLs.
- Contributor rules rewritten for the new docs in `.cursor/rules/user-manual.mdc`, build and workflow described in `docs/README.md`.
- Workflow builds the pages and the PDFs instead of the LaTeX and Sphinx builds.

## To do

- [ ] Review the developer docs by hand; they have not been checked yet.
- [ ] Migrate the images into `docs/`; they are still served from the old LaTeX directories.
- [ ] Carry over the changes to the old docs on master since the migration base (the commit before da1dee02f).
- [ ] Set the final base commit in `import-history.mjs`, run it once more and commit `content-history.json` before deleting the old sources.
- [ ] Clean up the git history so every markdown file is touched only once; the word wrap and similar fixups still need to be squashed.
- [ ] Point `.cursor/rules/developer-docs.mdc` at the new pages instead of `doc/*.rst`.
- [ ] Delete the LaTeX and RST sources and the old build once everything is migrated.
- [ ] Two figures only exist in the PDF and are missing as PNG (snail trail, waypoint details).
- [ ] Check the default key assignment on the menu buttons page against `default.xci`.

## Future

- New screenshots and migrated figures once https://github.com/XCSoar/XCSoar/issues/2917 has landed.
- Demo values for the InfoBoxes, so every InfoBox page shows what the box looks like in the app. They could live as YAML in the frontmatter of the InfoBox pages.

## Decisions needed

- Rename COPYING to LICENSE? A LICENSE file is the common name today and is what GitHub and most tools look for; the licence page takes its content verbatim either way.
- The PDFs in other languages are not migrated. Proposal: drop them for now and stop building and deploying them.
- The workflow in `.github/workflows/build-docs.yml` still uploads the PDFs to the download server; that step needs rework and testing. Is it still needed if the docs are a project of their own? They could be built and linked directly, and attached as release artifacts.
- I have a script that compares the plain text of the LaTeX docs with the markdown to catch anything the migration dropped or changed. Should it go into the repo?
- Permalinks: the app should be able to link into the docs with short, stable URLs that survive renamed pages. Should we set up link.xcsoar.org or xcsoar.link for that, or keep a redirect table in the docs site itself?


docs/.gitignore
lordfolken, line 1: This is dangerous. We need only stuff in there which is relevant for the proejct. If you have more you can use git to use other include files.

docs/package-lock.json
lordfolken, line 1: Why is this 24000 lines?

docs/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "files": [],
+  "references": [
+    { "path": "./.nuxt/tsconfig.app.json" },
+    { "path": "./.nuxt/tsconfig.server.json" },
+    { "path": "./.nuxt/tsconfig.shared.json" },
+    { "path": "./.nuxt/tsconfig.node.json" }
+  ]
+}
```
lordfolken: newlines please.

docs/package-lock.json
WolfgangDrescher: https://www.reddit.com/r/ProgrammerHumor/s/7Vt838xJwo

`package-lock.json` is the lock file that resolves the dependency graph from `package.json` and pins the exact versions being used. The file is generated automatically, and we should definitely keep it in the repository, since otherwise there is no guarantee that the docs are generated exactly the same way as locally. The `npm ci` command also requires this file.

For us it doesn't really matter that it's so long and contains so many dependencies, since Nuxt uses a bundler and we generate static (but interactive) HTML anyway.

docs/.gitignore
WolfgangDrescher: Why is this dangerous? It only applies to the `docs` subfolder and is the default file that was added when the docs were created. If it bothers you, I can clean it up. As far as I can tell, only the following entries are actually needed:

```gitignore
# Nuxt dev/build outputs
.output
.data
.nuxt
.nitro
.cache
dist

# Node dependencies
node_modules
```

docs/package-lock.json
lordfolken: I dont want to have such insane data in the repo. Can you find another solution please.

WolfgangDrescher: Please read the npm documentation on this first: [package-lock.json](https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json).

> This file is intended to be committed into source repositories, and serves various purposes:
>
> * Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
> * Provide a facility for users to "time-travel" to previous states of `node_modules` without having to commit the directory itself.
> * Facilitate greater visibility of tree changes through readable source control diffs.
> * Optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages.
> * As of npm v7, lockfiles include enough information to gain a complete picture of the package tree, reducing the need to read `package.json` files, and allowing for significant performance improvements.

That is just the way it is in Node.js, and we would be shooting ourselves in the foot by not following the convention. Without the lock file there is no reproducible build and no `npm ci`, and every checkout resolves a slightly different dependency tree.

On the size: my clone of this repository is about 385 MB, of which the lock file is 204 KB compressed, 812 KB in the working tree. That is around 0.05 percent, and it compresses well because every later version is stored as a delta against the previous one.

I still went through the alternatives:

| Option | Gain | Cost |
| --- | --- | --- |
| Switch to pnpm | `pnpm-lock.yaml` has around 15000 instead of 22775 lines | everyone needs pnpm, workflow and README have to change, same order of magnitude in the end |
| No lock file, pinned versions in `package.json` | the file is gone | no reproducible builds, `npm ci` no longer works, transitive dependencies float |
| Tooling in its own repository, only the content stays here | no Node files here | two repositories, content and tooling drift apart |

A separate repository would certainly work, but I would not recommend it, because it grows the dependency situation instead of shrinking it.

A `.gitattributes` entry with `linguist-generated=true` is a pragmatic step: GitHub then collapses the file in diffs, so it stops showing up in reviews. It does not solve @lordfolken's objection though, since the file stays in the tree.

My preference is to keep `package-lock.json`.

